PR #5, opened 2024-07-01
feature: add elastic search

geclos:


apps/web/src/lib/elastic/client.ts
```diff
@@ -0,0 +1,9 @@
+import { Client } from '@elastic/elasticsearch'
+
+export default new Client({
+  node: process.env.ELASTIC_URL,
+  auth: {
+    username: process.env.ELASTIC_USERNAME!,
+    password: process.env.ELASTIC_PASSWORD!,
```
andresgutgon, line 7: Don't use `process` use `env`
https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/drizzle.config.ts

Is a type safe version

docker-compose.yml
```diff
@@ -14,6 +14,12 @@ services:
       - apps/web/.env
     command: pnpm run dev
 
+  elastic:
+    image: elasticsearch:8.14.1
+    environment:
+      - 'discovery.type=single-node'
+      - 'ELASTIC_USERNAME=latitude'
+      - 'ELASTIC_PASSWORD=papapa55'
```
andresgutgon, line 22: Use `secret` to standaraize or change the other to `papapa55`

geclos: done here https://github.com/latitude-dev/latitude-llm/pull/7

PR #6, opened 2024-07-01
Initial data model for prompts

andresgutgon:
# What?
☝️ That

### DIFF
--- a/apps/web/src/db/schema/prompts/promptVersions.ts
+++ b/apps/web/src/db/schema/prompts/promptVersions.ts
@@ -1,0 +1,18 @@
+import { InferSelectModel } from 'drizzle-orm'
+import { bigserial, text, timestamp, uuid, varchar } from 'drizzle-orm/pg-core'
+
+import { latitudeSchema } from '../db-schema'
+import { timestamps } from '../schemaHelpers'
+
+export const promptVersions = latitudeSchema.table('prompt_versions', {
+  id: bigserial('id', { mode: 'bigint' }).notNull().primaryKey(),
+  promptId: uuid('prompt_uuid').notNull(),
+  name: varchar('name').notNull(),
+  path: varchar('path').unique().notNull(),
+  content: text('content').notNull(),
+  hash: varchar('hash').notNull(),
+  deletedAt: timestamp('deleted_at'),
+  ...timestamps(),
+})
+
+export type PromptVersion = InferSelectModel<typeof promptVersions>
--- a/apps/web/src/db/schema/prompts/prompts.ts
+++ b/apps/web/src/db/schema/prompts/prompts.ts
@@ -1,9 +1,0 @@
-import { timestamps } from '$/db/schema/schemaHelpers'
-import { InferSelectModel } from 'drizzle-orm'
-import { pgTable } from 'drizzle-orm/pg-core'
-
-export const prompts = pgTable('prompts', {
-  ...timestamps(),
-})
-
-export type Prompt = InferSelectModel<typeof prompts>
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,59 @@
+name: Linter & Types
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [20.x] # Specify node versions you want to test against
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 2 # Checkout HEAD^
+
+      - name: Setup Node.js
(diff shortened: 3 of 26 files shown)


apps/web/drizzle/0000_graceful_spectrum.sql
```diff
@@ -0,0 +1,46 @@
...
+CREATE TABLE IF NOT EXISTS "latitude"."prompt_versions" (
+	"serial" serial PRIMARY KEY NOT NULL,
+	"prompt_id" integer NOT NULL,
+	"path" varchar NOT NULL,
+	"content" text NOT NULL,
+	"hash" varchar NOT NULL,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"updated_at" timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT "prompt_versions_path_unique" UNIQUE("path")
+);
```
csansoon, line 21: Here `path` is not UNIQUE. Two versions of the same prompt may still have the same path.

apps/web/drizzle/0000_graceful_spectrum.sql
```diff
@@ -0,0 +1,46 @@
...
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."prompts" (
+	"serial" serial PRIMARY KEY NOT NULL,
+	"path" varchar NOT NULL,
+	"content" text NOT NULL,
+	"hash" varchar NOT NULL,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"updated_at" timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT "prompts_path_unique" UNIQUE("path")
+);
```
csansoon, line 31: Missing `prompt_id`, which should be UNIQUE here too

apps/web/src/db/schema/prompts/prompts.ts
```diff
@@ -1,9 +1,20 @@
...
+    id: serial('serial').notNull().primaryKey(),
+    path: varchar('path').unique().notNull(),
+    content: text('content').notNull(),
+    hash: varchar('hash').notNull(),
+    ...timestamps(),
+  },
+  (promp) => ({
+    path: uniqueIndex('prompt_path_idx').on(promp.path),
+  }),
+)
```
csansoon, line 18: Is this supposed to be the "snapshot"?

csansoon: Also, as previously mentioned, it's missing a `promptId`

apps/web/src/db/schema/prompts/promptVersions.ts
```diff
@@ -0,0 +1,31 @@
...
+} from 'drizzle-orm/pg-core'
+
+export const promptVersions = latitudeSchema.table(
+  'prompt_versions',
+  {
+    id: serial('serial').notNull().primaryKey(),
+    promptId: integer('prompt_id')
+      .notNull()
+      .references((): AnyPgColumn => prompts.id),
+    path: varchar('path').unique().notNull(),
```
csansoon, line 21: Mentioned above, `path` cannot be unique here.

apps/web/drizzle/0000_graceful_spectrum.sql
andresgutgon: But also have `prompt_id` pointing to a prompt?

apps/web/src/db/schema/prompts/prompts.ts
andresgutgon: Yes, this is the snapshot

apps/web/drizzle/0000_graceful_spectrum.sql
```diff
@@ -0,0 +1,46 @@
+CREATE SCHEMA "latitude";
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."commits" (
```
geclos, line 3: use uuids instead of auto incremental ids for identification of commits and prompts this way we can safely reference them from our api without revealing information and its easier to reference them from other services like elastic search

apps/web/src/db/schema/prompts/prompts.ts
geclos: maybe we have to call it prompt_snapshot and call prompt_versions prompts instead

OPINIONS EVERYWHERE

apps/web/drizzle/0000_graceful_spectrum.sql
geclos: well maybe for prompts is not needed although for consistency i'd use uuids everywhere? opinions?

apps/web/src/db/schema/prompts/prompts.ts
csansoon: I think we should have `prompt_versions` and `prompt_snapshots`. `prompts` is just something built from those, something theoretical and untangible ✨

PR #7, opened 2024-07-01
Feature/api scaffolding

geclos:


apps/web/src/middleware.test.ts
```diff
@@ -0,0 +1,24 @@
+import { NextRequest } from 'next/server'
+import { describe, expect, it } from 'vitest'
+
+import { middleware } from './middleware'
+
+describe('handler', () => {
+  const req = new NextRequest(new Request('http://acme.com'))
+
+  it('succeeds', async () => {
+    req.headers.set('Authorization', `Bearer ${process.env.API_KEY}`)
```
andresgutgon, line 10: From where we're going to get the token? DB?

andresgutgon: Just saying because this is not Vercel (edge) friendly. And nextjs complains

andresgutgon: Also this auth can be done with a HOF at API level if we see it's not possible in middleware. Or use Express in another app and use there the db connection

apps/web/src/lib/elastic/client.ts
```diff
@@ -1,9 +1,10 @@
+import env from '$/env'
 import { Client } from '@elastic/elasticsearch'
 
 export default new Client({
   node: process.env.ELASTIC_URL,
```
andresgutgon, line 5: What's `ELASTIC_URL`?

andresgutgon: Linter fail. We have to setup CI. I'll do tomorrow
image

apps/web/src/middleware.test.ts
andresgutgon: Thinking about this more maybe it would be better to have an express serving API endpoints. [Express middleware](https://hackernoon.com/effective-use-of-middleware-in-expressjs-practical-approaches) is much nicer. Also, this would for us to adopt from the get-go a more modular approach when building the services that use these end points or the next app.

apps/web/src/lib/elastic/client.ts
andresgutgon: Put defaults values in [env.example](https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/env.example) so we can use it

PR #8, opened 2024-07-01
Import base compiler from latitude-sdk

csansoon:
New language compiler. In this series of PRs, the compiler will be imported from @latitude-sdk and modified to the new conversation-based dynamic prompt language. Example of new syntax:
```md
---
model: gpt-3.5-turbo
temperature: 0.2
tools:
  - code_interpreter
  - run_query
---
<system>
  You are {{name}}, a data engineer AI expert in data science. You must respond to any user request related to data. If the request is not about data, do not respond.
</system>

{{#if use_example}}
  <ref prompt="data_requests_examples" />
{{/if}

<user name={{user_name}}>
  {{request}}
</user>
```

This new compiler will have 3 available methods:

- `compile` will receive the prompt as a string and a list of parameters, and return a conversation object.
- `parse` will receive the prompt as a string and return metadata information, such as:
  - `config`: the prompt's config
  - `hash`: a hash identifying the prompt's content
  - `parameters`: a list of all values used in the query that have not been defined within the query (so they must be given as parameters)
- `serialize` will receive a conversation object, and return the prompt string that defines it.

---

🚂 Trenecito
- [x] Import base package from latitude-sdk (this PR)

**Parser**
  - [x] Mustache logic (already done from latitude-sdk)
  - [x] Conditionals and loops (already done from latitude-sdk)
  - [x] HTML-like tags #9 
  - [x] YAML config #10 

**Compiler**
  - [x] Compile #12 
  - [ ] Parse
  - [ ] Serialize


packages/compiler/package.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "dependencies": {
+    "acorn": "^8.9.0",
+    "code-red": "^1.0.3",
+    "locate-character": "^3.0.0",
+    "yaml": "^2.4.5"
+  },
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "keywords": [],
```
andresgutgon, line 30: 🔥

andresgutgon (approved):

PR #9, opened 2024-07-02
Compiler html like tags

csansoon:
🚂 TRENECITO
Destination: #8 

---
Parse any HTML-like tags. Tags contain a name, and an array of attributes:
```html
<name attr1=value1 attr2=value2> content </name>
```
Attributes can have the following values:
 - If a value is provided and wrapped in quotation marks, it is returned as a `Text` node.
 - If a value is provided and wrapped in custom logic brackets (`{{ }}`), it is returned as a `MustacheTag` with the correspondent expression.
 - If a value is provided as a single word with no wrap, it is returned as a `Text` node.
 - If no value is provided (and there's no = sign either), it is returned as a `true` value.

There is no logic in the Parser about what types of tags are there or what attributes do they expect. That's the compiler's responsibility.

---
Last stop: #8
Next station: #10


geclos:
> If a value is provided as a single word with no wrap, it is returned as a Text node.

I wouldn't do this? I know it's cool but it increases the api surface for no added benefit (not having to write quotes is not a benefit) and it introduces some confusion `name=true` could be expected to parse true as a boolean instead of a string whereas name='true' is crystal clear.

packages/compiler/src/parser/parser.test.ts
```diff
@@ -274,4 +275,138 @@ world */
       ` hello\n  ${CUSTOM_TAG_START}#if condition${CUSTOM_TAG_END}then${CUSTOM_TAG_START}/if${CUSTOM_TAG_END}\nworld `,
     )
   })
+
+  it('Allows tag comments', () => {
+    const fragment = parse('<!-- hello -->')
```
geclos, line 280: very cool!

packages/compiler/src/parser/state/tag.ts
```diff
@@ -0,0 +1,219 @@
...
+				type: 'Text',
+				raw: '',
+				data: ''
+			};
+		} else {
+			currentChunk.raw += parser.template[parser.index++];
+		}
+	}
+	parser.error(PARSER_ERRORS.unexpectedEof);
+}
```
geclos, line 219: is this code copied from somewhere?

andresgutgon: don't worry when he rebase with main prettier will complain

csansoon:
> > If a value is provided as a single word with no wrap, it is returned as a Text node.
> 
> I wouldn't do this? I know it's cool but it increases the api surface for no added benefit (not having to write quotes is not a benefit) and it introduces some confusion `name=true` could be expected to parse true as a boolean instead of a string whereas name='true' is crystal clear.

The correct approach to setting it as `true` would be with `name={{true}}`, as it is a logical value.

andresgutgon (dismissed):

geclos:
> The correct approach to setting it as true would be with name={{true}}, as it is a logical value.

I understand but the point is that it's confusing and there's no point on having this feature since you can already define both booleans and strings with  `foo={{}}` syntax

PR #10, opened 2024-07-02
Parser yaml config

csansoon:
🚂 TRENECITO
Destination: #8 

---
Parse any YAML-like config if it's surrounded by `---`. As a requirement, it MUST be the first thing in the file. Otherwise, it will just be treated as text.

For example:
```md
---
model: gpt-4o
temperature: 0.2
tools:
    - user_search
---
/* Rest of conversation */
```

Contents inside this block will simply be parsed as YAML. This content must be static, so no custom logic is applied via `{{ }}` tags.

---

Previous stop: #9 
Next stop: #12

### DIFF
--- a/packages/compiler/src/parser/state/config.ts
+++ b/packages/compiler/src/parser/state/config.ts
@@ -1,0 +1,32 @@
+import yaml from 'yaml'
+
+import { Parser } from '..'
+import PARSER_ERRORS from '../../error/errors'
+
+export function config(parser: Parser) {
+  const start = parser.index
+  parser.eat('---')
+
+  // Read until there is a line break followed by a triple dash
+  const data = parser.readUntil(/\n\s*\-\-\-\s*/)
+
+  parser.allowWhitespace()
+  parser.eat('---', true)
+
+  let parsedData
+  try {
+    parsedData = yaml.parse(data)
+  } catch (error) {
+    parser.error(PARSER_ERRORS.invalidConfig((error as Error).message), start)
+  }
+
+  const node = {
+    start,
+    end: parser.index,
+    type: 'Config',
+    raw: data,
+    value: parsedData,
+  }
+
+  parser.current().children!.push(node)
+}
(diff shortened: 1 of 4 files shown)


packages/compiler/src/parser/state/config.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Parser } from '..'
+import PARSER_ERRORS from '../../error/errors'
+
+export function config(parser: Parser) {
+  const start = parser.index
+  parser.eat('---')
+
+  // Read until there is a line break followed by a triple dash
+  const data = parser.readUntil(/\n\s*\-\-\-\s*/)
+  console.log('data', data)
```
geclos, line 12: remove

packages/compiler/src/parser/state/config.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  parser.allowWhitespace()
+  parser.eat('---', true)
```
geclos, line 15: again?

packages/compiler/src/parser/state/text.ts
```diff
@@ -12,6 +12,12 @@ export function text(parser: Parser) {
     const isEscaping = ENDS_WITH_ESCAPE_REGEX.test(data)
     if (isEscaping) data = data.slice(0, -1) // Remove the escape character
 
+    if (!isEscaping && parser.match('---')) {
+      // Only parse config if it's the first thing in the file
+      const isFirst = parser.template.slice(0, parser.index).trim() === '' // Ignore any whitespace
```
geclos, line 17: dry a bit

geclos (approved):
missing a test

packages/compiler/src/parser/state/config.ts
```diff
@@ -0,0 +1,32 @@
+import yaml from 'yaml'
+
+import { Parser } from '..'
+import PARSER_ERRORS from '../../error/errors'
```
andresgutgon, line 4: Put alias paths please

andresgutgon (dismissed):

packages/compiler/src/parser/state/config.ts
csansoon: Yes, it must start and finish with `---`
```md
---
foo: bar
---
Actual markdown content
```

packages/compiler/src/parser/state/config.ts
csansoon: will do in a later PR

PR #11, opened 2024-07-02
feature: standalone api

geclos:
We are moving the backend api to a standlone express application. This allows users to independently scale their critical app path (the connection to the LLM provider) independently from the latitude dashboard.


apps/api/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Run the production build (Must be built first).
+
+### `npm start -- --env="name of env file" (default is production).`
+
+Run production build with a different env file.
+
+## Additional Notes
+
+-   If `npm run dev` gives you issues with bcrypt on MacOS you may need to run: `npm rebuild bcrypt --build-from-source`.
```
andresgutgon, line 41: `pnpm` no?

apps/api/src/common/EnvVars.ts
```diff
@@ -0,0 +1,28 @@
...
+      maxAge: Number(process.env.COOKIE_EXP ?? 0),
+      domain: (process.env.COOKIE_DOMAIN ?? ''),
+      secure: (process.env.SECURE_COOKIE === 'true'),
+    },
+  },
+  Jwt: {
+    Secret: (process.env.JWT_SECRET ??  ''),
+    Exp: (process.env.COOKIE_EXP ?? ''), // exp at the same time as the cookie
+  },
+} as const;
```
andresgutgon, line 28: Use the same package t3-env no?

apps/api/spec/support/apiCb.ts
```diff
@@ -0,0 +1,15 @@
+import logger from 'jet-logger'
```
andresgutgon, line 1: I would go with `wistom`
image

apps/api/spec/tests/users.spec.ts
andresgutgon, line 1: We usually use `.test` for tests.

apps/api/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "@types/morgan": "^1.9.9",
+    "@types/node": "^20.14.9",
+    "@types/supertest": "^6.0.2",
+    "@typescript-eslint/eslint-plugin": "^7.15.0",
+    "@typescript-eslint/parser": "^7.15.0",
+    "eslint": "^8.57.0",
+    "eslint-plugin-node": "^11.1.0",
+    "find": "^0.3.0",
+    "fs-extra": "^11.2.0",
+    "jasmine": "^5.1.0",
```
andresgutgon, line 52: Why not vitest?

apps/api/src/server.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * Setup express server.
+ */
+
+import express, { NextFunction, Request, Response } from 'express'
+import helmet from 'helmet'
+import logger from 'jet-logger'
+import morgan from 'morgan'
```
andresgutgon, line 8: Isn't morgan another logger?

apps/api/src/common/EnvVars.ts
geclos: looking at the benefits I don't think it justifies another dependency, went for simple types with zod

apps/api/src/server.ts
geclos: yup the generator i used makes use of two loggers, one for express stuf and another for custom logs. Opted for regular console logs and kept morgan which is a classic out-of-the-box experience for epxress apps logging

apps/api/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "jet-logger": "^1.3.1",
+    "jet-paths": "^1.0.6",
+    "nodemon": "^3.1.4",
+    "supertest": "^7.0.0",
+    "ts-node": "^10.9.2",
```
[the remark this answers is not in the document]
andresgutgon: It handle better newer repos with ESM

apps/api/build.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Start
+ */
+
+;(async () => {
+  try {
+    // Remove current build
+    await remove('./dist/')
+    // Copy front-end files
+    await copy('./src/public', './dist/public')
+    await copy('./src/views', './dist/views')
```
andresgutgon, line 19: Do we have views or public?

geclos: nope! good catch!

apps/api/package.json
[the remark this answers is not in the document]
andresgutgon: Give it a try

apps/api/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+    "esModuleInterop": true /* Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'. */,
+    "skipLibCheck": true /* Skip type checking of declaration files. */,
+    "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */,
+    "paths": {
+      "@src/*": ["src/*"]
+    },
+    "useUnknownInCatchVariables": false
+  },
+  "include": ["src/**/*.ts", "test/**/*.ts", "build.ts"]
+}
```
andresgutgon, line 17: Clean this file of comments and use our base

apps/web/package.json
```diff
@@ -42,7 +39,6 @@
     "@types/pg": "^8.11.6",
     "@types/react": "^18",
     "@types/react-dom": "^18",
-    "dotenv": "^16.4.5",
```
andresgutgon, line 45: Don't remove this one. I need to run CI

geclos: i did nothing it's duplicated from dependencies and i guess pnpm removed the duplicate

apps/api/package.json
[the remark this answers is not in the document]
geclos: coolio, a bit less config

apps/api/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+/** @type {import("eslint").Linter.Config} */
+module.exports = {
+  extends: ['./node_modules/@latitude-data/eslint-config/library.js'],
+  env: {
+    node: true,
+    browser: true,
```
andresgutgon, line 6: No browser here no?

apps/api/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "lint": "eslint --ext .ts src/",
+    "lint:tests": "eslint --ext .ts test/",
+    "start": "node -r module-alias/register ./dist --env=production",
+    "dev": "tsx src/index.ts",
+    "tc": "tsc --noEmit",
+    "test": "vitest --run",
+    "test:watch": "vitest "
+  },
+  "packageManager": "pnpm@8.6.3",
+  "_moduleAliases": {
```
andresgutgon, line 16: What's this?

apps/api/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@latitude-data/api",
+  "version": "0.0.1",
+  "scripts": {
+    "prettier": "prettier --write \"**/*.{ts,tsx,md}\"",
+    "build": "npx ts-node build.ts",
+    "lint": "eslint --ext .ts src/",
+    "lint:tests": "eslint --ext .ts test/",
+    "start": "node -r module-alias/register ./dist --env=production",
+    "dev": "tsx src/index.ts",
```
andresgutgon, line 10: Do `tsx watch src/index.ts` to have watcher

apps/api/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@latitude-data/api",
+  "version": "0.0.1",
+  "scripts": {
+    "prettier": "prettier --write \"**/*.{ts,tsx,md}\"",
+    "build": "npx ts-node build.ts",
+    "lint": "eslint --ext .ts src/",
+    "lint:tests": "eslint --ext .ts test/",
```
andresgutgon, line 8: Can we colocate tests please?

apps/api/package.json
geclos: module-alias package resolves ts's path alias at runtime so that running the production build works appropiately

andresgutgon: I think something that must be as perf as possible should be transpiled to JS and avoid as much as possible TS

geclos: this is all js. To give more details:
- we compile to JS for production build with tsc
- tsc does not change code, it just transpiles, and since we use custom alias paths "@" these end up in the final js build
- since @ is not a valid path for node, this package modifies the node resolution algo at boot time in order to make it support our custom alias paths.

In summary, we trade some milliseconds at boot time in order to not have to deal with vite/rollup builds.

PR #12, opened 2024-07-02
compiler: compile conversation

csansoon:
🚂 TRENECITO
Destination: #8 

---

Compiler's `compile` function now works a bit differently. Differences from last compiler:
- Now compiling must return a `Conversation` object, representing a list of messages and generation config.
- Support for `ContentTag`. There are 3 types of tags:
  - Message tags (`system`, `user` and `assistant`). These cannot have any other `ContentTag` as parent.
  - Content tags (`text`, `image`). Must be inside a message tag as direct parent.
  - References (`ref`). Cannot be used inside any other tag, and cannot contain children.
- Call expressions now just return regular values, there is no "interpolation policy", "static arguments" or complex stuff.
- All parameters are loaded into the scope at start, so there is no need of any function call.

---

Previous stop: #10  
Next stop:


packages/compiler/src/compiler/foo.test.ts
```diff
@@ -0,0 +1,30 @@
...
+      parameters: {
+        name: 'Latitude',
+      },
+    })
+
+    const conversation = await compiler.compile()
+
+    console.log(JSON.stringify(conversation, null, 2))
+  })
+})
```
geclos, line 30: lol the filename, maybe something more meaningful?

geclos, line 30: is this wip? don't see any assertions

packages/compiler/src/compiler/scope.ts
```diff
@@ -15,65 +15,54 @@ export default class Scope {
...
   }
 
   copy(): Scope {
     const scope = new Scope()
-    scope.consts = { ...this.consts }
-    scope.vars = { ...this.vars }
+    scope.pointers = { ...this.pointers }
     return scope
   }
 }
```
geclos, line 68: why do we need to continue to support variables?

geclos (approved):
feels good!

packages/compiler/src/compiler/scope.ts
geclos: also missing tests for this if we indeed need to continue to support it

packages/compiler/src/compiler/scope.ts
```diff
@@ -15,65 +15,54 @@ export default class Scope {
...
    */
-  private static stash: unknown[] = [] // Stash of every variable value in every scope
-  private static readFromStash(index: number): unknown {
-    return Scope.stash[index]
-  }
-  private static addToStash(value: unknown): number {
-    Scope.stash.push(value)
-    return Scope.stash.length - 1
-  }
+  private stash: unknown[] = [] // Stash of every variable value in the global scope
```
geclos, line 22: globalScope would be a better name imo

packages/compiler/src/compiler/foo.test.ts
csansoon: oh I pushed it by accident xd, was just testing some things. Will do a proper test when finished.

packages/compiler/src/compiler/scope.ts
csansoon: creating variables inside of `{{#if }}` and `{{#each}}` statements should be scoped imo. I don't know about tags though.

packages/compiler/src/compiler/scope.ts
csansoon: I'll change the names to `globalStash` and `localPointers`.
\*All\* variable values are in the `globalStash`, but that doesn't mean they're all accessible from any scope. The values accessible to each scope are defined in the `localPointers`, which point to the actual value in the stash.

packages/compiler/src/compiler/scope.ts
geclos: my question was more about whether we need variables in the first place

packages/compiler/src/compiler/foo.test.ts
csansoon: oh, yes sorry. This was still WIP

packages/compiler/src/compiler/scope.ts
csansoon: Well, following the line of latitude-sdk's powerful template engine, here you can not only condition or iterate over variables, but you can also create and modify at will.

packages/compiler/src/compiler/scope.ts
csansoon: The stash is not the scope. The scope is what variables (from the stash) you have access to. The stash is where all variables live.

It's like in javascript. Two variables can point to the same value:
```ts
const var1 = { foo: 'bar' }
const var2 = var1

if (true) {
  var1.foo = 'qux'
}

console.log(var2.foo) // 'qux'
```

Modifying a variable inside the `if` statement, also modifies the value from outside.
However, inside the `if` statement there is a different scope. One copied from the old scope, but with different variable access.

```ts
const var1 = { foo: 'bar' }

if (true) {
  const var2 = var1
  var1.foo = 'qux'
}

console.log(var2.foo) // ReferenceError: var2 is not defined
```

geclos: > The scope is what variables (from the stash) you have access to.

isn't that what local pointer are?

geclos: talked offline. 

this.globalStash is indeed the global scope of variables but we wanna avoid calling globalScope because carlos thinks it can lead devs to thinking the scope has access to the global scope when in fact it only has access to the scope available via local pointers.

geclos: It would have been cool to call them heap and pointers as a memory management joke though (don't do it)

geclos (approved):

packages/compiler/src/compiler/scope.ts
csansoon: The compiler uses variables in several ways, even if the user does not define one themselves. For example:
- Parameters are variables
- When using `{#each}` blocks, both the iterator element and index key are scoped variables
- Any functions we want to add (such as 'parse', for example) are treated as scoped variables for simplicity.

In addition to all of this, users can create their own variables, which will work just in the same way.

geclos: oh ok then 👌🏼

packages/compiler/src/compiler/compile.test.ts
```diff
@@ -0,0 +1,854 @@
...
+  })
+})
+
+describe('messages', async () => {
+  it('allows creating system, user, assistant and tool messages', async () => {
+    const propmt = `
+      <system>system message</system>
+      <user>user message</user>
+      <assistant>assistant message</assistant>
+      <tool>tool message</tool>
```
andresgutgon, line 126: Missing tool id validation

andresgutgon (dismissed):

PR #13, opened 2024-07-02
Database package standalone

andresgutgon:
# What?
This allow to use database connection in web and api apps sharing the same models and same connection config to the database

```javascript
import { listCommits } from '@latitude-data/database'

// Connection is build once in the nextjs app.
// Then is passed to database package where all the Drizzle logic is 
// maintained
import database from '$/db/database'

export default async function Home() {
  const commits = await listCommits({ db: database })
}
```

#### Test using DB from nextJS app
image


apps/web/drizzle.config.ts
```diff
@@ -4,9 +4,10 @@ import env from './src/env'
 
 const testEnv = process.env.NODE_ENV === 'test'
 const connectionString = testEnv ? env.TEST_DATABASE_URL : env.DATABASE_URL
+
 export default {
   dialect: 'postgresql',
-  schema: './src/db/schema/index.ts',
+  schema: './node_modules/@latitude-data/database/src/schema/index.ts',
```
andresgutgon, line 10: The drizzle.config.ts still lives in NextJS because is where the `db:migrate` command and the rest are executed. The difference is that we pick schema from the database package. Not sure if this will be a problem in some way

apps/web/package.json
```diff
@@ -24,12 +24,11 @@
     "node": ">=20.0.0"
   },
   "dependencies": {
+    "@latitude-data/database": "workspace:*",
     "@elastic/elasticsearch": "^8.14.0",
     "@t3-oss/env-nextjs": "^0.10.1",
     "dotenv": "^16.4.5",
-    "drizzle-orm": "^0.31.2",
     "next": "15.0.0-rc.0",
-    "pg": "^8.12.0",
```
andresgutgon, line 32: Drizzle ORM and PG runtime dependencies now lives in database. We call functions from this package. The convention is to put all the SQL calls under database/data-access

apps/web/src/db/database.ts
```diff
@@ -0,0 +1,6 @@
+import { buildDatabaseClient } from '@latitude-data/database'
+import env from '$/env'
+
+const testEnv = process.env.NODE_ENV === 'test'
+const connectionString = testEnv ? env.TEST_DATABASE_URL : env.DATABASE_URL
```
andresgutgon, line 5: We let the consumer of the database package provide Db credentials

apps/web/tsconfig.json
```diff
@@ -9,7 +9,8 @@
       "./node_modules/@types"
     ],
     "paths": {
-      "$/*": ["./src/*"]
+      "$/*": ["./src/*"],
+      "$db/*": ["../../packages/database/src/*"]
```
andresgutgon, line 13: This do the trick and hotreload works and also LSP to go from app to package. I hope in production it works too. Didn't tried. 

Dev experience is perfect

docker-compose.yml
```diff
@@ -8,6 +8,7 @@ services:
       - "4983:4983"
     volumes:
       - ./apps/web:/app/apps/web
+      - ./packages/database:/app/packages/database
```
andresgutgon, line 11: This do the hot reload magic.

apps/web/src/db/database.ts
csansoon: Does it always have to be passed as env variables? Since this is a standalone package, I think this should be passed as a parameter. Then it's the app who decides how to obtain it and, if it's still using an env var, it decides how it's called.

csansoon (approved):

packages/database/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
geclos, line 1: set it to private: true

apps/web/src/db/database.ts
geclos: agree on it being a parameter

apps/web/tsconfig.json
geclos: i don't get it you should be importing everything from the package no? not directly from its source code

packages/database/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { buildDatabaseClient } from '$db/client'
```
geclos, line 1: why is this not imported from @latitude-data/database

packages/database/src/data-access/commits.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export async function listCommits({ db }: { db: Database }) {
+  return db
+    .select({
+      uuid: commits.uuid,
+      title: commits.title,
+      description: commits.description,
+    })
+    .from(commits)
+}
```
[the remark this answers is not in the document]
andresgutgon: > Imo it should only expose the tables and models and let consumers build their own queries

That can be the right approach. But...

We want to encapsulate business logic in the db calls so for example `getCommits` can be used in the same way from API or WEB without these places reimplementing in both places. I would like to have some nice architecture thoughts from the get-go.

That's my idea here

packages/database/package.json
andresgutgon: Again maybe yes, but we might want to publish in npm at some point

apps/web/tsconfig.json
andresgutgon: Did you notice I didn't implement Rollup? 👀 NextJS bundler is doing the work for us and the `packages/**` are plain TS. 

The same as the other questions. Maybe is not the right approach but let's talk offline.

apps/web/src/db/database.ts
andresgutgon: It's a parameter. The confusion here is that this file is called `database` but it lives in `apps/web/src/db/database.ts` It use `@latitude/database` init method and it pass the env configuration.

docker-compose.yml
andresgutgon: I'm not so happy with this. I'm trying to have everything dockerized but if this packages lived in another repo `pnpm link` might not work. I have to test it.

packages/database/package.json
geclos: do we? this is a useless package for anyone else and there's no point with having to deal with package versions/publishing/whatnots, i'd just bundle it with web and api production builds and palantir

packages/database/src/data-access/commits.ts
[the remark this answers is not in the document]
geclos: I agree on DRYing queries but this seems a bit premature, it's very possible i'd might want to query just uuid for example, or maybe select created_at/updated_at also. I'd wait a little bit until we actually see some repeated queries before implementing abstractions for it.

docker-compose.yml
geclos: I think it's wrong to think of database as another npm package, this is really just shared code between api and web, it will never live in another repo and i'd just bundle it with the production builds of web and api.

apps/web/drizzle.config.ts
geclos: spoiler: it is 😅 

Since the database package has its own typescript config doesn't know about, it can't do stuff like resolve custom path aliases in the database package. I'm trying to fix this atm.

PR #15, opened 2024-07-03
feature: convos model

geclos:
This commit implements the convos model and moves logic to migrate database to the database package. It also simplifies a little bit how we use database package in web.


packages/database/env.example
```diff
@@ -0,0 +1,4 @@
+NODE_ENV=development
+
+DATABASE_URL=postgres://latitude:secret@localhost:5432/latitude_development
+TEST_DATABASE_URL=postgres://latitude:secret@localhost:5432/latitude_test
```
geclos, line 4: I will find a way to share .env files between packages in the monorepo in another pr

packages/database/src/schema/models/promptSnaptshots.ts
```diff
@@ -1,4 +1,3 @@
-import { timestamps } from '$db/schema/schemaHelpers'
```
geclos, line 1: since database package is imported as ts files into other projects typescript does not know how to interpret path aliases from other packages, resulting in type errors in those other packages. I've opted for removing path aliases in the database project in order to bypass this issue, but happy to hear any suggestions on how to solve this better.

packages/database/drizzle/0000_military_bruce_banner.sql
```diff
@@ -11,6 +11,14 @@ CREATE TABLE IF NOT EXISTS "latitude"."commits" (
 	CONSTRAINT "commits_uuid_unique" UNIQUE("uuid")
 );
 --> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."convos" (
```
andresgutgon, line 14: can we name it conversations?

packages/database/env.example
andresgutgon: with another package xd

packages/database/src/env.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import 'dotenv/config'
+
+const envvars = z.object({
+  NODE_ENV: z.string(),
+  DATABASE_URL: z.string(),
+  TEST_DATABASE_URL: z.string(),
+})
+
+export default envvars.parse(process.env)
```
andresgutgon, line 11: errors are better handle with the other package, i don't get why you don't use it

packages/database/src/schema/models/convos.ts
```diff
@@ -0,0 +1,35 @@
...
+  {
+    id: bigserial('id', { mode: 'bigint' }).notNull().primaryKey(),
+    uuid: uuid('uuid')
+      .notNull()
+      .unique()
+      .default(sql`gen_random_uuid()`),
+    content: jsonb('content'),
+    promptVersionId: bigint('prompt_version_id', { mode: 'bigint' })
+      .notNull()
+      .references(() => promptVersions.id, { onDelete: 'cascade' }),
```
andresgutgon, line 19: not sure if we should attach conversations to versions or snapshots. Why you choose versions?

packages/database/drizzle/0000_military_bruce_banner.sql
geclos: don't you prefer convos? it's a common shorthand for conversations and we'll write less chars

andresgutgon: We don't write the name of the table ever again I think. This is only on the schema. I would prefer a more formal name

packages/database/src/env.ts
geclos: because I strongly agree with the idea that adding dependencies should be the last option to consider and t3-env brings very [minimal advantages](https://env.t3.gg/docs/introduction#transforms-and-default-values) vs this:
- **defaults and transforms**
  - you should never transform env vars anyways and defaults are nice but hardly a must-have imo
 - **support for multiple environments**
   - can be done with a couple lines of code as done in /api package

packages/database/src/schema/models/convos.ts
geclos: Because you can always trigger a convo by referencing a past commit, and so it makes sense to target the prompt version for that commit.

andresgutgon: But the `prompt_snapshots` is who holds the relation with commits

packages/database/src/env.ts
andresgutgon: I think we're reimplementing a small package that does it super nice

packages/database/src/schema/models/convos.ts
geclos: oh I didn't realize. I think the model is wrong? If I ask for a prompt in a particular commit I'm gonna build the snapshot in memory, use it, and remove it once completed. In our convos we said prompt_snapshot were only the snapshot of the latest commit.

packages/database/drizzle/0000_military_bruce_banner.sql
geclos: I think it would be an unnecessary point of confusion so I'd prefer to keep consistent. conversations or convos everywhere, which one do u prefer? I don't have a strong opinion here.

packages/database/src/schema/models/convos.ts
andresgutgon: You're right. Is missing `commitId`

packages/database/drizzle/0000_military_bruce_banner.sql
andresgutgon: Whatever you prefer. I don't have any logical reason. Just feels more correct conversations

docker/init-db.sh
andresgutgon, line 1: What's this permissions change?

packages/database/docker/Dockerfile
```diff
@@ -0,0 +1,69 @@
+FROM node:20-alpine AS base
+
+ARG DATABASE_URL
+ARG NODE_ENV=development
+
+# Install Pulumi
+RUN apk add --no-cache curl
+
+# TODO: Document the need for libc6-compat
```
andresgutgon, line 9: We have this TODO 3 times. Maybe is time to explain 😂

packages/database/docker/Dockerfile
```diff
@@ -0,0 +1,69 @@
+FROM node:20-alpine AS base
+
+ARG DATABASE_URL
+ARG NODE_ENV=development
+
+# Install Pulumi
```
andresgutgon, line 6: No pulumi here no?

docker/init-db.sh
geclos: sh needs to be executable otherwise it does not work (or it didn't in my case!)

packages/database/docker/Dockerfile
geclos: uh oh good catch

tools/env/env/development.env
```diff
@@ -0,0 +1,9 @@
+NODE_ENV=development
+
+DATABASE_URL=postgres://latitude:secret@localhost:5432/latitude_development
+TEST_DATABASE_URL=postgres://latitude:secret@localhost:5432/latitude_test
+
+ELASTIC_URL=https://elastic:9200
+ELASTIC_USERNAME=latitude
+ELASTIC_PASSWORD=papapa55
```
andresgutgon, line 8: Use secrets

tools/env/index.ts
```diff
@@ -0,0 +1,20 @@
...
+    path: url,
+  })
+
+  console.log('RESULT: ', result)
+  console.log('WAT: ', process.env.DATABASE_URL)
+
+  if (result.error) {
+    throw result.error
+  }
+}
```
andresgutgon, line 20: Promise is last I say this. But take a look at the code. Is super well done
https://github.com/t3-oss/t3-env/blob/main/packages/core/src/index.ts#L215

And small. I would use it in this package as starting point

tools/env/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@latitude-data/env",
+  "version": "0.0.1",
+  "description": "Common envvars for latitude llm",
+  "main": "index.ts",
+  "exports": {
+    ".": "./index.ts"
+  },
```
andresgutgon, line 8: Do we need exports having `main`?

geclos: yeah for typescript tsc

tools/env/index.ts
geclos: what for? this package does not have type validation of env vars (it's implemented in the other packages)

docker-compose.yml
```diff
@@ -36,6 +36,14 @@ services:
       - 'ELASTIC_USERNAME=latitude'
       - 'ELASTIC_PASSWORD=secret'
 
+  database:
```
andresgutgon, line 39: Maybe is a good time to rename now this as talked offline

packages/database/drizzle/meta/0000_snapshot.json
```diff
@@ -82,6 +82,64 @@
...
+          "primaryKey": false,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "content": {
+          "name": "content",
+          "type": "jsonb",
+          "primaryKey": false,
+          "notNull": false
+        },
```
csansoon, line 107: What's the `convo` content? Is it the list of messages?

We can save the convo's content as a latitude prompt, which also includes the convo's initial config.

packages/database/src/schema/models/convos.ts
```diff
@@ -0,0 +1,35 @@
...
+    ),
+  }),
+)
+
+export const convoRelations = relations(convos, ({ one }) => ({
+  promptVersion: one(promptVersions),
+}))
+
+export type Convo = InferSelectModel<typeof convos> & {
+  promptVersion: PromptVersion[]
```
andresgutgon, line 34: This is not an array no?

andresgutgon (dismissed):

docker-compose.yml
geclos: hmm good point

packages/database/src/schema/models/convos.ts
geclos: nein, good catch!

docker/init-db.sh
andresgutgon: Weird it worked in my machine

packages/database/drizzle/meta/0000_snapshot.json
geclos: hmm it could lead to confusion if we use the prompt_snapshots table for both prompt templates and conversation histories

csansoon: Prompt files are representation of `Conversation` objects, which are just a list of messages.

What I mean is that the conversation history can be represented as another .prompt file, as it is just a list of messages too. The only difference is that they are already resolved, so there are no variables.

docker-compose.yml
```diff
@@ -8,7 +8,6 @@ services:
       - '4983:4983'
     volumes:
       - ./apps/web:/app/apps/web
-      - ./packages/database:/app/packages/database
```
andresgutgon, line 11: This container will not have fresh code without this no?

packages/database/drizzle/meta/0000_snapshot.json
andresgutgon: But the compiled version is a json shape. The prompt is a text template no?

PR #16, opened 2024-07-03
Move env to packages and use `db` in environment for connecting to postgres instead of `localhost`

andresgutgon:
# What?
We want to run all the services inside docker containers which means pointing to Postgresql with docker internal network name `db`


docker-compose.yml
```diff
@@ -5,14 +5,13 @@ services:
...
     volumes:
+      # Mount packages for development
+      - ./packages/env:/app/packages/env
+      - ./packages/core:/app/packages/core
       - ./apps/web:/app/apps/web
-      - /app/apps/web/node_modules
     depends_on:
       - db
-    env_file:
-      - apps/web/.env
```
andresgutgon, line 15: No needed no?

packages/core/docker/Dockerfile
```diff
@@ -61,6 +64,19 @@ RUN if [ "$DATABASE_URL" ]; then \
...
+# Ex.:
+#  https://local.drizzle.studio?port=3002&host=0.0.0.0
+
+# CMD pnpm db:migrate
+
+# This entrypoint runs the migration
+ENTRYPOINT ["/entrypoint.sh"]
+
+# This is a hack to keep running the container
+CMD ["tail", "-f", "/dev/null"]
```
andresgutgon, line 81: I need help here. It works but this has to be different for production. 
Also when doing `docker compose stop core` it does not work

packages/core/src/schema/models/promptVersions.ts
```diff
@@ -26,8 +26,11 @@ export const promptVersions = latitudeSchema.table('prompt_versions', {
   ...timestamps(),
 })
 
-export const promptVersionRelations = relations(commits, ({ one }) => ({
```
andresgutgon, line 29: This was a bug. It has to be the table that has the relations. In this case `promptVersions`.

Is very nice to use Drizzle Studio because is super strict and if we have some relation miss configured it does not run

```
Drizzle Studio is up and running on https://local.drizzle.studio?port=3002&host=0.0.0.0
Error: There is not enough information to infer relation "__latitude__.commits.commit"

```

packages/core/docker/Dockerfile
andresgutgon: # EXEC
`docker compose exec core pnpm db:studio`
image

With `compose run` it does not work

packages/core/docker/entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# Run database migrations
+pnpm db:migrate
+
+cleanup() {
+  echo "Core container stopped"
+}
+
+# Check if NODE_ENV is set to development.
+if [ "$NODE_ENV" = "development" ]; then
```
andresgutgon, line 12: This chunk is here to keep alive core container in development to be able to connect to port 3002 to Drizzle studio. AI is awesome

.github/workflows/test.yml
```diff
@@ -69,6 +69,7 @@ jobs:
       - name: Run Server migrations
         env:
           NODE_ENV: test
+          DATABASE_URL: postgres://latitude:secret@localhost:5432/latitude_test
```
andresgutgon, line 72: When I changed to `db` the `packages/env/test.env` the CI started failing. Would be ideal to call `db` the postgresql in CI but I didn't know how. So I override the env here

PR #17, opened 2024-07-04
Jobs component setup

andresgutgon:
# What?
Setup the jobs system as an package that can be used by API and WEB


packages/jobs/docker/Dockerfile
```diff
@@ -0,0 +1,84 @@
+FROM node:20-alpine AS base
```
andresgutgon, line 1: Use base container

packages/jobs/docker/Dockerfile
```diff
@@ -0,0 +1,84 @@
...
+RUN if [ "$REDIS_PORT" ]; then \
+    export REDIS_PORT; \
+  else \
+    unset REDIS_PORT; \
+  fi
+RUN if [ "$REDIS_PASSWORD" ]; then \
+    export REDIS_PASSWORD; \
+  else \
+    unset REDIS_PASSWORD; \
+  fi
```
andresgutgon, line 80: 🔥 Delete all of this variable

apps/web/package.json
```diff
@@ -18,19 +18,20 @@
     "@elastic/elasticsearch": "^8.14.0",
     "@latitude-data/core": "workspace:*",
     "@latitude-data/env": "workspace:^",
+    "@latitude-data/jobs": "workspace:*",
     "@t3-oss/env-nextjs": "^0.10.1",
-    "next": "15.0.0-rc.0",
-    "react": "19.0.0-rc-f994737d14-20240522",
-    "react-dom": "19.0.0-rc-f994737d14-20240522",
```
andresgutgon, line 24: Copy & pasta from a pet project. Too hipster for now

packages/env/src/index.ts
```diff
@@ -1,12 +1,15 @@
 import path from 'path'
+import { fileURLToPath } from 'url'
 
 import dotenv from 'dotenv'
 
+const DIRNAME = path.dirname(fileURLToPath(import.meta.url))
```
andresgutgon, line 6: `__dirname` is not giving the right path when the package is used in `web` or `api`

apps/web/src/actions/example-enqueu.ts
csansoon, line 1: typo on filename

apps/web/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build": "next build",
+    "build": "NODE_ENV='production' next build",
```
geclos, line 7: This should come from the environment instead of being hardcoded here

apps/web/src/app/page.tsx
```diff
@@ -1,14 +1,20 @@
-import { listCommits } from '@latitude-data/core'
+import { Commit, listCommits } from '@latitude-data/core'
+import { exampleEnqueue } from '$/actions/example-enqueu'
 import database from '$/db/database'
 
+export const dynamic = 'force-dynamic'
```
geclos, line 5: what's this

apps/web/package.json
andresgutgon: I have the production docker file in another PR. I'll rebase when merge this one and remove from here.

apps/web/src/env.ts
```diff
@@ -4,14 +4,21 @@ import { createEnv } from '@t3-oss/env-nextjs'
 import { z } from 'zod'
 
 export default createEnv({
+  skipValidation: process.env.BUILDING_CONTAINER == 'true',
```
geclos, line 7: so this did the trick

apps/web/tsconfig.json
```diff
@@ -6,9 +6,11 @@
     "moduleResolution": "Bundler",
     "jsx": "preserve",
     "typeRoots": ["./node_modules/@types"],
+    "baseUrl": ".",
```
geclos, line 9: is this needed?

packages/compiler/package.json
```diff
@@ -21,15 +37,5 @@
...
-  "dependencies": {
-    "acorn": "^8.9.0",
-    "ajv": "^8.16.0",
-    "code-red": "^1.0.3",
-    "locate-character": "^3.0.0",
-    "yaml": "^2.4.5"
-  },
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
-  "author": "Latitude Data"
```
geclos, line 34: what happened here

apps/web/src/app/page.tsx
andresgutgon: Next try to build the pages when doing `pnpm build`. This makes everything fail because we don't have env variables at build time

apps/web/src/env.ts
andresgutgon: I think so. This should be in the next PR but I'll keep in this one

packages/env/src/index.ts
geclos: makes sense

apps/web/tsconfig.json
andresgutgon: Not sure. Did many things on friday. I'll clean up in the next PR when it's working

packages/jobs/rollup.config.mjs
```diff
@@ -0,0 +1,43 @@
...
+    input: 'src/index.ts',
+    output: [{ file: 'dist/index.d.ts', format: 'es' }],
+    plugins: [alias(aliasEntries), dts()],
+  },
+  {
+    input: 'src/server.ts',
+    output: [{ file: 'dist/server.js' }],
+    plugins: COMMON_PLUGINS,
+    external: EXTERNAL,
+  },
```
csansoon, line 42: Exporting as `dsit/index.js` twice?

packages/env/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "@latitude-data/typescript-config/base.json",
+  "compilerOptions": {
+    "module": "ESNext",
+    "moduleResolution": "node",
+    "rootDir": "./src"
+  },
+  "include": ["src/**/*"],
+  "exclude": ["node_modules"]
+}
```
geclos, line 10: it's so bad that we need so much configuration for a pkg that just exports a single function that took 3 minutes to write honestly

packages/jobs/docker/Dockerfile
geclos: use the base latitude container ☝🏼

packages/env/tsconfig.json
andresgutgon: 🤷‍♂️

packages/jobs/rollup.config.mjs
```diff
@@ -0,0 +1,43 @@
...
+const EXTERNAL = ['ioredis', 'bullmq']
+
+/** @type {import('rollup').RollupOptions[]} */
+export default [
+  {
+    input: 'src/index.ts',
+    output: [{ file: 'dist/index.js' }],
+    plugins: COMMON_PLUGINS,
+    external: EXTERNAL,
+  },
```
geclos, line 31: I assume index.ts exports the code that is used in web and api? you don't really need to build it no? It can be exported as typescript to the rest of the projects? 

src/server.ts is the only thing that needs building in truth. Or am I missing something?

packages/jobs/rollup.config.mjs
```diff
@@ -0,0 +1,43 @@
...
+  {
+    input: 'src/index.ts',
+    output: [{ file: 'dist/index.d.ts', format: 'es' }],
+    plugins: [alias(aliasEntries), dts()],
+  },
```
geclos, line 36: why two entries for the same input?

packages/jobs/src/connection/index.ts
```diff
@@ -0,0 +1,15 @@
+import IORedis from 'ioredis'
+
+export type ConnectionParams = {
+  port: number
+  host: string
+  password?: string
+}
+export function buildConnection({ host, port, password }: ConnectionParams) {
+  console.log('🔥 Connecting to Redis', host, port)
+  return new IORedis(port, host, {
```
geclos, line 10: 👾

packages/jobs/src/connection/index.ts
```diff
@@ -0,0 +1,15 @@
...
+    password: password ? password : undefined,
```
geclos, line 11: ternary not needed?

packages/jobs/rollup.config.mjs
andresgutgon: One is for `server.ts`

packages/jobs/src/env.ts
```diff
@@ -0,0 +1,19 @@
...
+  DATABASE_URL: z.string(),
+})
+
+export default envvars.parse(process.env)
+
+declare global {
+  namespace NodeJS {
+    interface ProcessEnv extends z.infer<typeof envvars> {}
+  }
+}
```
geclos, line 19: beautifui :trollface:

packages/jobs/rollup.config.mjs
andresgutgon: Probably you're right. I'll cleanup when fixing docker build

packages/jobs/src/env.ts
andresgutgon: Not convinced at all. I might switch to t3-env.

apps/web/tsconfig.json
geclos: ok don't forget to remove it please

packages/jobs/src/env.ts
geclos: nein

apps/web/src/app/page.tsx
andresgutgon: No, the error is that the variables are not defined and we can't connect to DB or Redis at build time. 

All good. I want to avoid these errors if possible. But the build succeed anyway

packages/jobs/src/env.ts
andresgutgon: [Their rationale is good](https://env.t3.gg/docs/introduction).

1. Transforms in the env setup, what needs to be a number is a number in the final form.
2. Type safe.

With this solution when you use `process.env.THING` `THING` is not typed when used. With t3-env you can see if this is a number or a string or optional.

andresgutgon: >  protection against env transformations, which you should never do anyway

Why?

PR #18, opened 2024-07-04
Compiler: Read metadata

csansoon:
Read metadata from a prompt without actually resolving it. Reading metadata will return the following information:
- `hash`: A unique hash representing the prompt contents. Modifying this prompt, or any that it references, will update this value.
- `config`: The configuration of the config
- `parameters`: A list of all the variables being used in the prompt that must be passed as parameters.
- `referencedPrompts`: A list of all prompts that are being referenced in the query
- `schemaValidation`: Whether the config follows the given schema, or a list of schema errors otherwise.

The function will take the following parameters:
- `prompt`: The prompt to be read
- `referenceFn`: (optional) A function to obtain a prompt given the prompt name/path. It is not needed if no refs are called in the prompt.
- `configSchema`: (optional) The config schema for validation. `schemaValidation` will be `undefined` if not provided.

```ts
const {
  hash,
  config,
  parameters,
  referencedPrompts,
  schemaValidation
} = await readMetadata({ prompt, referenceFn, configSchema })
```

### DIFF
--- a/packages/compiler/src/compiler/config.ts
+++ b/packages/compiler/src/compiler/config.ts
@@ -1,10 +1,28 @@
+import Ajv from 'ajv'
+import type { ErrorObject, JTDDataType } from 'ajv/dist/core'
+
 import { Fragment } from '../parser/interfaces'
 
-export function readConfig(fragment: Fragment): Record<string, unknown> {
+export function readConfig<ConfigSchema>(
+  fragment: Fragment,
+): JTDDataType<ConfigSchema> {
   for (const node of fragment.children) {
     if (node.type === 'Config') {
-      return node.value
+      return node.value as JTDDataType<ConfigSchema>
     }
   }
-  return {}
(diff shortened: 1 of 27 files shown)


packages/compiler/src/compiler/config.ts
```diff
@@ -1,10 +1,28 @@
...
 import { Fragment } from '../parser/interfaces'
 
-export function readConfig(fragment: Fragment): Record<string, unknown> {
+export function readConfig<ConfigSchema>(
+  fragment: Fragment,
+): JTDDataType<ConfigSchema> {
   for (const node of fragment.children) {
     if (node.type === 'Config') {
-      return node.value
+      return node.value as JTDDataType<ConfigSchema>
```
andresgutgon, line 11: Missing typeof here no?

andresgutgon (approved):
Awesome!
Now after this before moving on implement path alias please and you have a 10 : )

PR #19, opened 2024-07-04
feature: docker base image

geclos:
Simplifies our dockerfiles with a common base image

### DIFF
--- a/packages/core/docker/Dockerfile
+++ b/packages/core/docker/Dockerfile
@@ -1,36 +1,14 @@
-FROM node:20-alpine AS base
+FROM latitudedata/base AS installer
 
-ARG DATABASE_URL
 ARG NODE_ENV=development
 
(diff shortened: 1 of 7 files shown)


csansoon (approved):

packages/core/docker/Dockerfile
```diff
@@ -1,67 +1,26 @@
...
-
-COPY . .
 RUN turbo prune @latitude-data/core --docker
 
-FROM base AS installer
-
-ARG NODE_ENV
-
-WORKDIR /app
+FROM installer AS runner
```
csansoon, line 7: Bit weird that the `latitude/base` stage is named `installer`, and the actual `installer` stage is named `runner`

andresgutgon: The base is the nodejs and pnpm installer I guess is for that

geclos: hmm you have to think it as "FROM _this image_ START A NEW IMAGE WITH NAME _installer_"

PR #20, opened 2024-07-04
Compiler: clean package

csansoon:
# WIP
Finish cleaning the compiler package.

Changes:
 - Changed the `compile` method to use a recursive approach instead of an iterative one. This makes the code way more readable and organized. The tests have not been changed to check that everything still works as expected.
 - Improved typing for AST nodes
 - Updated `ToolCall` object to contain both function's `name` and `arguments`
 - Added alias
 - Removed unused error definitions

### DIFF
--- a/packages/compiler/src/compiler/compile.ts
+++ b/packages/compiler/src/compiler/compile.ts
@@ -1,28 +1,28 @@
-import type { Node as LogicalExpression } from 'estree'
-
 import {
   CUSTOM_MESSAGE_ROLE_ATTR,
   CUSTOM_MESSAGE_TAG,
   REFERENCE_PROMPT_ATTR,
   REFERENCE_PROMPT_TAG,
-} from '../constants'
-import CompileError, { error } from '../error/error'
-import errors from '../error/errors'
-import parse from '../parser/index'
-import type {
-  BaseNode,
-  EachBlock,
-  ElementTag,
-  IfBlock,
-  TemplateNode,
-} from '../parser/interfaces'
+  TOOL_CALL_TAG,
+} from '$/constants'
+import CompileError, { error } from '$/error/error'
+import errors from '$/error/errors'
+import parse from '$/parser/index'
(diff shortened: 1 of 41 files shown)


packages/compiler/src/compiler/compile.ts
```diff
@@ -1,28 +1,28 @@
...
-import parse from '../parser/index'
-import type {
-  BaseNode,
-  EachBlock,
-  ElementTag,
-  IfBlock,
-  TemplateNode,
-} from '../parser/interfaces'
+  TOOL_CALL_TAG,
+} from '$/constants'
```
andresgutgon, line 7: namespace alias with `$compiler/*` we need this to run the package from nextjs and api apps

packages/compiler/vitest.config.ts
```diff
@@ -1,6 +1,8 @@
+import tsconfigPaths from 'vite-tsconfig-paths'
 import { defineConfig } from 'vitest/config'
 
 export default defineConfig({
+  plugins: [tsconfigPaths()],
```
andresgutgon, line 5: we might not need rollup. Don't remove yet tho

packages/compiler/src/compiler/compile.ts
csansoon: What do you mean

packages/compiler/vitest.config.ts
csansoon: The tests do not configure the alias without this, so it does not find the imported files and it fails.

packages/compiler/src/compiler/compile.ts
andresgutgon: in tsconfig.ts

packages/compiler/vitest.config.ts
andresgutgon: ahhh sorry this is vitest. ignore

andresgutgon (approved):

PR #23, opened 2024-07-05
Compiler: sync errors

csansoon:
For consistency, both the `compile` and `readMetadata` methods should throw the same errors on the same situations.

> [!NOTE]
> The only exception on this are errors that depend on the resolution of values or logical expressions, which are not evaluated by `readMetadata`

This will make sure that we can evaluate syntax errors in a prompt file in real time without having to compile it, and compiling it afterwards will not throw additional errors that have not been foreseen by the metadata.


packages/compiler/src/compiler/compile.test.ts
```diff
@@ -191,26 +191,16 @@ describe('messages', async () => {
...
-    const action1 = () =>
-      compile({
-        prompt: removeCommonIndent(propmt1),
-        parameters: {},
-      })
-    const action2 = () =>
+    const action = () =>
       compile({
-        prompt: removeCommonIndent(propmt2),
+        prompt: removeCommonIndent(propmt),
```
geclos, line 199: typo

packages/compiler/src/compiler/readMetadata.ts
```diff
@@ -118,134 +169,212 @@ export class ReadMetadata {
...
           isInContentTag,
         })
       }
-      if (node.else) {
-        for await (const childNode of node.else.children ?? []) {
+      return
+    }
+
+    if (node.type === 'ElementTag') {
+      // ToolCall <tool_call id="123" name="foo">{ arguments }</tool_call>
```
geclos, line 242: unnecessary comment imo

packages/compiler/src/compiler/readMetadata.ts
```diff
@@ -118,134 +169,208 @@ export class ReadMetadata {
...
+        }
+        return
       }
-    }
 
-    if (node.type === 'ElementTag') {
       if (
-        node.name === CUSTOM_MESSAGE_TAG ||
-        Object.values(MessageRole).includes(node.name as MessageRole)
+        Object.values(MessageRole).includes(node.name as MessageRole) ||
```
geclos, line 290: you are repeating this Object.values(...).includes(...) in several places, dry a bit. This method is long to read and deals with much complexity, which is ok since you are solving a hard problem! make it easier for devs to understand what is happening by drying code where possible and giving meaningful names to methods.

geclos (approved):

packages/compiler/src/compiler/readMetadata.ts
csansoon: Right, I'll dry a bit and simplify things!

csansoon: should be clearer now!

andresgutgon (approved):

PR #25, opened 2024-07-08
Tenancy schema

csansoon:
Added the tenancy schema.

Auth managed by Next-auth. Added:
- `users`
- `accounts`
- `sessions`

Created a `workspace` table, along with a `members` relation table.

Added both `workspace_id` and `author_id` properties to `commits`

Added `api_token` for API access to a `workspace`


packages/core/drizzle/0004_woozy_sebastian_shaw.sql
```diff
@@ -0,0 +1,87 @@
...
+	"access_token" text,
+	"expires_at" integer,
+	"token_type" text,
+	"scope" text,
+	"id_token" text,
+	"session_state" text,
+	CONSTRAINT "account_provider_providerAccountId_pk" PRIMARY KEY("provider","providerAccountId")
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."authenticator" (
```
andresgutgon, line 16: We don't need this one

packages/core/drizzle/0004_woozy_sebastian_shaw.sql
```diff
@@ -0,0 +1,87 @@
...
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."user" (
+	"id" text PRIMARY KEY NOT NULL,
+	"name" text,
+	"email" text NOT NULL,
+	"emailVerified" timestamp,
+	"image" text
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."verificationToken" (
```
andresgutgon, line 43: No need no?

packages/core/drizzle/0004_woozy_sebastian_shaw.sql
```diff
@@ -0,0 +1,87 @@
...
+	CONSTRAINT "authenticator_credentialID_unique" UNIQUE("credentialID")
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."session" (
+	"sessionToken" text PRIMARY KEY NOT NULL,
+	"userId" text NOT NULL,
+	"expires" timestamp NOT NULL
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."user" (
```
andresgutgon, line 35: Is called user? Tables are named in plural by our convention

packages/core/drizzle/0004_woozy_sebastian_shaw.sql
csansoon: Hmm weird, it's [NextAuth's official docs for Drizzle](https://authjs.dev/getting-started/adapters/drizzle#schemas), but doesn't seem to be even mentioned on the rest of the docs.

packages/core/src/schema/models/commits.ts
```diff
@@ -26,6 +28,12 @@ export const commits = latitudeSchema.table(
     ),
     title: varchar('title', { length: 256 }).notNull(),
     description: text('description'),
+    authorId: text('author_id')
+      .notNull()
+      .references(() => users.id, { onDelete: 'cascade' }),
+    workspaceId: bigserial('workspace_id', { mode: 'bigint' })
+      .notNull()
+      .references(() => workspaces.id, { onDelete: 'cascade' }),
```
csansoon, line 36: Instead of having to specify that this column is a `bigserial`, can't I just infer the type from `workspaces.id` since it's already defined?

andresgutgon: I think you can

packages/core/src/schema/models/users.ts
```diff
@@ -0,0 +1,49 @@
...
+  }),
+)
+
+export const sessions = latitudeSchema.table('sessions', {
+  sessionToken: text('sessionToken').primaryKey(),
+  userId: text('userId')
+    .notNull()
+    .references(() => users.id, { onDelete: 'cascade' }),
+  expires: timestamp('expires', { mode: 'date' }).notNull(),
+})
```
andresgutgon, line 47: Put in different files

packages/core/src/schema/models/apiTokens.ts
```diff
@@ -0,0 +1,22 @@
...
+import { latitudeSchema } from '..'
+import { timestamps } from '../schemaHelpers'
+import { workspaces } from './workspaces'
+
+export const apiTokens = latitudeSchema.table('api_tokens', {
+  id: bigserial('id', { mode: 'bigint' }).notNull().primaryKey(),
+  token: uuid('token')
+    .notNull()
+    .unique()
+    .$defaultFn(() => crypto.randomUUID()),
```
csansoon, line 13: All logs are going to be associated with an `apiToken` (for traceability), so we might want index performance on the `id`.

packages/core/src/schema/models/commits.ts
csansoon: Hm there's no official way of doing this. Weird, I think it makes hella sense

packages/core/src/schema/models/apiTokens.ts
andresgutgon: The token is to authorized the api calls. Also the logs would be on Elastic don't add any index here I think

packages/core/drizzle/0000_youthful_sumo.sql
```diff
@@ -0,0 +1,172 @@
...
+	"token_type" text,
+	"scope" text,
+	"id_token" text,
+	"session_state" text,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"updated_at" timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT "accounts_provider_providerAccountId_pk" PRIMARY KEY("provider","providerAccountId")
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."api_tokens" (
```
geclos, line 20: nitpick but can we call it api_keys? it's the common naming for these

packages/core/drizzle/0000_youthful_sumo.sql
```diff
@@ -0,0 +1,172 @@
+CREATE SCHEMA "latitude";
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."accounts" (
```
geclos, line 3: what's this table for? oauth? I see some mentions to provider

packages/core/package.json
```diff
@@ -22,6 +22,7 @@
   "dependencies": {
     "@latitude-data/env": "workspace:^",
     "drizzle-orm": "^0.31.2",
+    "next-auth": "5.0.0-beta.19",
```
geclos, line 25: living in the edge

geclos (dismissed):

packages/core/drizzle/0000_youthful_sumo.sql
csansoon: This is for OAuth. It is recommended to still add it even if we're not using it, for compatibility with NextAuth providers and easier OAuth integration in the future.

andresgutgon (approved):

PR #27, opened 2024-07-08
small fixes

csansoon:
Some small fixes:

- Fixed `api` Dockerfile to install all required modules correctly
- Fixed typo on `promptSnapshots` model
- Fixed casing on `users`, `sessions` and `acounts` models to follow our convention


apps/api/docker/Dockerfile
```diff
@@ -9,15 +9,15 @@ FROM installer AS runner
...
+WORKDIR /app/apps/api
+
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
     if [ "$NODE_ENV" = "production" ]; then \
         pnpm install --prod --frozen-lockfile; \
     else \
         pnpm install; \
     fi
 
-WORKDIR /app/apps/api
```
geclos, line 19: why this change?

geclos (approved):
good catches on the table names 👌🏼

apps/api/docker/Dockerfile
csansoon: It wasn't installing the correct required packages, resulting in execution errors. Installing them after changing the workdir fixed it.

PR #28, opened 2024-07-08
Web dockerfile production

andresgutgon:
# What?
We need to check if the way we're setting monorepo makes sense. For example we're using TS path alias to link from packages to apps and I want to see if this works when building the apps for production

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,4 @@
+# These 2 config set to `false` avoid having
+# ERR_PNPM_LOCKFILE_CONFIG_MISMATCH errors when installing dependencies
+auto-install-peers=false
+dedupe-peer-dependents=false
--- a/apps/web/docker/Dockerfile
+++ b/apps/web/docker/Dockerfile
@@ -1,25 +1,61 @@
-FROM latitudedata/base AS installer
+FROM latitudedata/base AS base
 
-ARG NODE_ENV=development
+WORKDIR /app
+COPY . .
 
 RUN turbo prune @latitude-data/web --docker
 
-FROM installer AS runner
+# COMMON stage (installer)
+# ------------------------------------------------------
+FROM base AS installer
+COPY --from=base /app/out/json/ .
+COPY --from=base /app/out/pnpm-lock.yaml ./pnpm-lock.yaml
+COPY --from=base /app/out/pnpm-workspace.yaml ./pnpm-workspace.yaml
+COPY --from=base /app/out/full/ .
 
-COPY .gitignore .gitignore
-COPY --from=installer /app/out/full .
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install \
+  --frozen-lockfile \
+  --filter "@latitude-data/web..."
 
-RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
-    if [ "$NODE_ENV" = "production" ]; then \
-        pnpm install --prod --frozen-lockfile; \
-    else \
-        pnpm install; \
-    fi
+# DEVELOPMENT stage
+# ------------------------------------------------------
+FROM installer AS dev-runner
 
 WORKDIR /app/apps/web
+ENV NODE_ENV=development
+CMD pnpm dev
 
-ENV NODE_ENV $NODE_ENV
+# PRODUCTION stage
+# ------------------------------------------------------
+FROM installer AS prod-builder
 
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
+  BUILDING_CONTAINER=true \
+  pnpm turbo build --filter='@latitude-data/web...'
+
+FROM base AS prod-runner
+
+WORKDIR /app-prod
+
+# Don't run production as root
+RUN addgroup --system --gid 1001 nodejs
+RUN adduser --system --uid 1001 nextjs
+USER nextjs
+
+COPY --from=installer /app/apps/web/next.config.mjs .
+COPY --from=installer /app/apps/web/package.json .
+COPY --from=installer /app/node_modules ./node_modules
+COPY --from=installer /app/apps/web/package.json ./apps/web/package.json
+
+# Automatically leverage output traces to reduce image size
+# https://nextjs.org/docs/advanced-features/output-file-tracing
+COPY --from=prod-builder --chown=nextjs:nodejs /app/apps/web/.next/standalone ./apps
+COPY --from=prod-builder --chown=nextjs:nodejs /app/apps/web/.next/static ./apps/web/.next/static
+COPY --from=prod-builder --chown=nextjs:nodejs /app/apps/web/public ./apps/web/public
+
+ARG PORT=3000
+ENV PORT $PORT
+ENV NODE_ENV=production
 EXPOSE 3000
 
-CMD pnpm start
+CMD HOSTNAME="0.0.0.0" node apps/web/server.js -p $PORT
--- a/apps/web/docker/run-production.sh
+++ b/apps/web/docker/run-production.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+export DATABASE_URL="postgres://latitude:secret@db:5432/latitude_development"
+
+# Redis
+export REDIS_PORT=6379
+export REDIS_HOST=redis
+export REDIS_PASSWORD=secret
+
+# Elastic Search
+export ELASTIC_URL=elastic
+export ELASTIC_USERNAME=latitude
+export ELASTIC_PASSWORD=secret
+
+docker compose run -it \
+  -p 3006:3006 \
+  -e DATABASE_URL="$DATABASE_URL" \
+  -e REDIS_HOST="$REDIS_HOST" \
+  -e REDIS_PORT="$REDIS_PORT" \
+  -e ELASTIC_URL="$ELASTIC_URL" \
+  -e ELASTIC_USERNAME="$ELASTIC_USERNAME" \
+  -e ELASTIC_PASSWORD="$ELASTIC_PASSWORD" \
+  -e PORT="3006" \
+  web apps/web/server.js -p 3006
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -1,4 +1,55 @@
-/** @type {import('next').NextConfig} */
-const nextConfig = {};
+import path from 'path'
+import { fileURLToPath } from 'url'
 
-export default nextConfig;
+const currentFileURL = import.meta.url
+const __filename = fileURLToPath(currentFileURL)
+const __dirname = path.dirname(__filename)
+
+const PACKAGES = ['env', 'core', 'jobs']
+const PACKAGE_NAMES = PACKAGES.map((pkg) => `@latitude-data/${pkg}`)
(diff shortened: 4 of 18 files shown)


apps/web/next.config.mjs
```diff
@@ -1,4 +1,42 @@
...
+
+const PACKAGES = ['env', 'core', 'jobs']
+const PACKAGE_NAMES = PACKAGES.map((pkg) => `@latitude-data/${pkg}`)
+
+function buildAlias({ localPackages }) {
+  return localPackages.reduce((aliases, pkg) => {
+    aliases[`$${pkg}`] = path.resolve(__dirname, `../../packages/${pkg}/src`)
+    return aliases
+  }, {})
+}
```
andresgutgon, line 16: If we go for the Rollup build by package in production we check NODE_ENV and avoid setting these alias. That's all it needs I think

apps/web/next.config.mjs
```diff
@@ -1,4 +1,42 @@
...
+
 /** @type {import('next').NextConfig} */
-const nextConfig = {};
+const nextConfig = {
+  output: 'standalone',
+  transpilePackages: Object.values(PACKAGE_NAMES),
```
andresgutgon, line 21: Also, these would be only for development. We're not ready for real production so I'm happy with this for now.

apps/web/package.json
```diff
@@ -10,10 +10,6 @@
     "tc": "tsc --noEmit",
     "prettier": "prettier --write \"**/*.{ts,tsx,md}\""
   },
-  "packageManager": "pnpm@8.6.3",
-  "engines": {
-    "node": ">=20.0.0"
-  },
```
andresgutgon, line 16: This lives in the root package.json

apps/web/src/jobs/index.ts
```diff
@@ -1,11 +1,24 @@
 import { setupJobs } from '@latitude-data/jobs'
+import env from '$/env'
 
-import env from '../env'
+/**
+ * In theory Nextjs should not pre-render pages that export
+ *   export const dynamic = 'force-dynamic'
```
andresgutgon, line 6: This is kind of a lie. I'm avoiding calling Redis while building. But not sure this is working tbh. Again only an issue for production. Let's keep this way for now

docker/Dockerfile.base
```diff
@@ -7,7 +7,7 @@ RUN apk add --no-cache libc6-compat
 # Install pnpm
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
-RUN corepack enable && corepack prepare pnpm@8.6.3 --activate
+RUN corepack enable && corepack prepare pnpm@8.15.6 --activate
```
andresgutgon, line 10: And improvement for the future is pass this as build argument. To have more centralized

packages/jobs/src/workers/worker-definitions/exampleWorker.ts
```diff
@@ -4,6 +4,7 @@ import { Job, Processor } from 'bullmq'
 
 type ExampleResult = { job: Job; patata: string }
 const processor: Processor<ExampleJobData, ExampleResult> = async (job) => {
+  console.log('JOB received', job.id, job.data)
```
andresgutgon, line 7: Leave for now. The first person implementing a real job will review this example worker

.npmrc
```diff
@@ -0,0 +1,3 @@
+auto-install-peers=false
+strict-peer-dependencies=false
+dedupe-peer-dependents=false
```
geclos, line 3: why do we need all this

apps/web/docker/Dockerfile
```diff
@@ -1,25 +1,63 @@
-FROM latitudedata/base AS installer
-
-ARG NODE_ENV=development
+FROM latitudedata/base AS base
 
 RUN turbo prune @latitude-data/web --docker
 
-FROM installer AS runner
+FROM base AS prunner
```
geclos, line 5: empty build stage

apps/web/docker/Dockerfile
```diff
@@ -1,25 +1,63 @@
...
 
-COPY .gitignore .gitignore
-COPY --from=installer /app/out/full .
+# COMMON stage (installer)
+# ------------------------------------------------------
+FROM base AS installer
+COPY --from=prunner /app/out/json/ .
```
geclos, line 10: why do you copy the json folder?

apps/web/docker/Dockerfile
```diff
@@ -1,25 +1,63 @@
...
 
-RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
-    if [ "$NODE_ENV" = "production" ]; then \
-        pnpm install --prod --frozen-lockfile; \
-    else \
-        pnpm install; \
-    fi
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install \
+  --frozen-lockfile \
+  --filter "@latitude-data/web..."
```
geclos, line 17: do you need this filter? prune is already setting up the dependencies so that npm install only installs the deps for web afaik

apps/web/docker/Dockerfile
geclos, line 10: 👾 pleas

apps/web/docker/Dockerfile
```diff
@@ -1,25 +1,63 @@
...
 
+# DEVELOPMENT stage
+FROM installer AS dev-runner
 WORKDIR /app/apps/web
```
geclos, line 21: 👾 after FROM declaration please

apps/web/docker/Dockerfile
```diff
@@ -1,25 +1,63 @@
...
+ENV NODE_ENV=development
+CMD pnpm dev
+
+# PRODUCTION stager
+# ------------------------------------------------------
+FROM installer AS prod-deps
+WORKDIR /app
```
geclos, line 28: ditot 👾

apps/web/docker/Dockerfile
```diff
@@ -1,25 +1,63 @@
...
+COPY --from=prunner /app/out/full/ .
 
-RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
-    if [ "$NODE_ENV" = "production" ]; then \
-        pnpm install --prod --frozen-lockfile; \
-    else \
-        pnpm install; \
-    fi
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install \
+  --frozen-lockfile \
```
geclos, line 16: missing --prod flag for install?

apps/web/docker/Dockerfile
```diff
@@ -1,25 +1,63 @@
...
+ENV NODE_ENV=development
+CMD pnpm dev
+
+# PRODUCTION stager
+# ------------------------------------------------------
+FROM installer AS prod-deps
+WORKDIR /app
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install \
+  --prod \
+  --frozen-lockfile
```
geclos, line 31: another install?

apps/web/docker/Dockerfile
```diff
@@ -1,25 +1,63 @@
...
+# Automatically leverage output traces to reduce image size
+# https://nextjs.org/docs/advanced-features/output-file-tracing
+COPY --from=prod-builder --chown=nextjs:nodejs /app/apps/web/.next/standalone ./apps
+COPY --from=prod-builder --chown=nextjs:nodejs /app/apps/web/.next/static ./apps/web/.next/static
+COPY --from=prod-builder --chown=nextjs:nodejs /app/apps/web/public ./apps/web/public
 
+ARG PORT=3000
+ENV PORT $PORT
+ENV NODE_ENV=production
 EXPOSE 3000
```
geclos, line 61: you should expose $PORT instead of 3000 otherwise if someone sets up port to something else other than 3000 next won't be accesible from host

apps/web/docker/run-production.sh
```diff
@@ -0,0 +1,23 @@
...
+docker compose run -it \
+  -p 3006:3006 \
+  -e DATABASE_URL="$DATABASE_URL" \
+  -e REDIS_HOST="$REDIS_HOST" \
+  -e REDIS_PORT="$REDIS_PORT" \
+  -e ELASTIC_URL="$ELASTIC_URL" \
+  -e ELASTIC_USERNAME="$ELASTIC_USERNAME" \
+  -e ELASTIC_PASSWORD="$ELASTIC_PASSWORD" \
+  -e PORT="3006" \
+  web apps/web/server.js -p 3006
```
geclos, line 23: this script is no longer needed no?

apps/web/next.config.mjs
geclos: yes please 🙏🏼

apps/web/src/jobs/index.ts
```diff
@@ -1,11 +1,24 @@
...
+ *
+ * But in practice it does, so we need to check if we are building
+ * the image and avoid connecting to Redis
+ */
+function setup() {
+  if (process.env.BUILDING_CONTAINER) return null
```
geclos, line 12: why do we need this

apps/web/docker/Dockerfile
andresgutgon: What about `api` app? How pnpm knows it only have to install `web` and its dependencies?

apps/web/docker/Dockerfile
andresgutgon: They're different stages. But let's talk offline. Not sure if it's the best approach.

apps/web/docker/run-production.sh
andresgutgon: Uhmm. So production app would work inside docker compose just pointing to apps/web/server.js. I mean while debuging the production build

apps/web/src/jobs/index.ts
andresgutgon: The app tries to connect to Redis while building the bundle

apps/web/docker/Dockerfile
geclos: because you run the prune for @latitude-data/api previously afaik

andresgutgon: Ok, I'll remove and try

.npmrc
andresgutgon: To do what it says. Is clear no?

I'm joking, I have no idea. I was trying

apps/web/docker/run-production.sh
geclos: talked offline he wants to keep it for debugging production builds if we ever need to

apps/web/docker/Dockerfile
andresgutgon: It does try to build api without this file
image

andresgutgon: Without this install fails. And even if it didn't failed it does not hurt. Does it?

apps/web/docker/Dockerfile
```diff
@@ -1,25 +1,61 @@
...
 
-CMD pnpm start
+CMD HOSTNAME="0.0.0.0" node apps/web/server.js -p $PORT
```
geclos, line 61: beautiful

PR #30, opened 2024-07-09
Setup web UI fundamentals

andresgutgon:
# What?
Setup ui package `@latitude-data/web-ui`. This package holds the Design System and Sections which are structures more complex than simple components. 

This package would also hold shared hooks.

### DIFF
--- a/packages/web-ui/package.json
+++ b/packages/web-ui/package.json
@@ -1,0 +1,39 @@
+{
+  "name": "@latitude-data/web-ui",
+  "version": "0.0.1",
+  "description": "Latitude web ui elements",
+  "private": true,
+  "scripts": {
+    "lint": "eslint ./src",
+    "tc": "tsc --noEmit",
+    "prettier": "prettier --write src/**/*.ts"
+  },
+  "main": "src/index.ts",
+  "exports": {
+    ".": "./src/index.ts",
+    "./styles.css": "./styles.css",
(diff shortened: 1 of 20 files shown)


geclos (dismissed):

packages/web-ui/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+/** @type {import("eslint").Linter.Config} */
+module.exports = {
+  extends: ['./node_modules/@latitude-data/eslint-config/library.js'],
+  env: {
+    node: true,
+    browser: true,
+  },
+}
```
geclos, line 8: This won't work on eslint 9. Steps to fix:
- rename it to .eslintrc.json (fix the config so that it's json also ofc)
- run npx @eslint/migrate-config .eslintrc.json
- run  globals @eslint/js @eslint/eslintrc to install the missing dependencies for eslint 9

packages/web-ui/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@latitude-data/web-ui",
+  "version": "0.0.1",
+  "description": "Latitude web ui elements",
+  "private": true,
+  "scripts": {
+    "lint": "eslint . --max-warnings 0 --ext .ts src",
```
geclos, line 7: just leave it at `eslint ./src`

PR #31, opened 2024-07-09
Feature/rollup build core 2

geclos:


packages/core/src/schema/index.ts
```diff
@@ -1,18 +1,13 @@
...
+export { commits } from './models/commits'
 export { convos, convoRelations, type Convo } from './models/convos'
 export { users, type User } from './models/users'
 export { accounts, type Account } from './models/accounts'
 export { sessions, type Session } from './models/sessions'
 export { workspaces, type Workspace } from './models/workspaces'
 export { apiKeys, type ApiKey } from './models/apiKeys'
+
+export * from './models/types'
+export * from './models/relations'
```
geclos, line 13: there were several circular dependencies in our models. This is why this exists.

PR #33, opened 2024-07-10
Make setup and login screens

csansoon:
# What?
Prepare web app with setup and login screens

## TODO
- [x] Setup screen with input validation
- [x] When setup is ok set session
- [x] Check if a user already exists in the system when setup is done
- [x] Store current workspace when login is successful
- [x] Redirect to `/setup` if the workspace is not found.
- [x] Redirect to `/login` when access to `/` if no valid session is present.
- [x] Login screen
- [x] Implement Logout 
- [x] QA production build works

## Next PR
- [ ] When setup is done create API tokens in DB


apps/web/src/app/setup/page.tsx
```diff
@@ -0,0 +1,65 @@
+'use client'
```
andresgutgon, line 1: Is better to move this to an inner component and let the routes always be server side rendered

apps/web/src/lib/auth/getCurrentUser.ts
```diff
@@ -0,0 +1,21 @@
+import { cache } from 'react'
+
+import db from '$/db/database'
+import { getCurrentUserFromDB } from '$core/data-access'
+import { Result } from '$core/lib'
+import { getSession } from './getSession'
+
+export const getCurrentUser = cache(async () => {
```
andresgutgon, line 8: React [cache](https://react.dev/reference/react/cache) is like in Rails something that lives through the whole request. So if this method is called 8 times is only executed once

package.json
```diff
@@ -25,5 +26,9 @@
   "packageManager": "pnpm@8.15.6",
   "engines": {
     "node": ">=20"
+  },
+  "dependencies": {
+    "drizzle-orm": "0.31.4",
+    "pg": "8.12.0"
```
andresgutgon, line 32: Dependencies that are shared by `packages/core` and `apps/web` but start different instances gives these error
```
Argument of type 'NodePgDatabase<typeof import("/Users/andres/code/latitude/llm/packages/core/src/schema/index")>' is not assignable to parameter of type 'PgDatabase<any, any, any>'.
  The types of '_.session' are incompatible between these types.
    Type 'PgSession<NodePgQueryResultHKT, typeof import("/Users/andres/code/latitude/llm/packages/core/src/schema/index"), ExtractTablesWithRelations<typeof import("/Users/andres/code/latitude/llm/packages/core/src/schema/index")>>' is not assignable to type 'PgSession<any, any, any>'.
      Property 'dialect' is protected but type 'PgSession<TQueryResult, TFullSchema, TSchema>' is not a class derived from 'PgSession<TQueryResult, TFullSchema, TSchema>'.

```

image

Having it in the root `package.json` remove these errors

packages/core/drizzle/0000_true_falcon.sql
```diff
@@ -1,22 +1,5 @@
...
-	"access_token" text,
-	"expires_at" integer,
-	"token_type" text,
-	"scope" text,
-	"id_token" text,
-	"session_state" text,
-	"created_at" timestamp DEFAULT now() NOT NULL,
-	"updated_at" timestamp DEFAULT now() NOT NULL,
-	CONSTRAINT "accounts_provider_provider_account_id_pk" PRIMARY KEY("provider","provider_account_id")
-);
```
andresgutgon, line 18: This was a `next-auth` table. Not needed with [Lucia](https://lucia-auth.com/)

packages/core/src/data-access/users.ts
```diff
@@ -0,0 +1,98 @@
...
+export type SessionData = {
+  user: SafeUser
+  workspace: { id: number; name: string }
+}
+
+export async function getUserFromCredentials(
+  credentials: { email: unknown; password: unknown },
+  { db }: AppContext,
+): PromisedResult<SessionData> {
+  const result = loginSchema.safeParse(credentials)
```
andresgutgon, line 29: No need for this schema. Is comes already validated

apps/web/src/app/(private)/page.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const dynamic = 'force-dynamic'
+
+export default async function Home() {
+  return (
+    <FocusLayout
+      header={
+        <FocusHeader
+          title='Inside the APP 💪'
+          description="Your're in let's kick those random AI's butts!"
+        />
```
csansoon, line 13: 💪

apps/web/src/actions/user/setupAction.ts
```diff
@@ -0,0 +1,39 @@
...
+import { createServerAction } from 'zsa'
+
+export const setupAction = createServerAction()
+  .input(
+    z.object({
+      name: z.string().min(1, { message: 'Name is a required field' }),
+      email: z.string().email(),
+      password: z.string().refine((pwd) => strongPassword(pwd), {
+        message:
+          'Password must be at least 8 characters, contain at least 1 uppercase letter, 1 lowercase letter, 1 number, and 1 special character (@$!%*?&).',
```
geclos, line 19: can... we relax all this requirements? It's just bad XP imo, consult it with product 🙏🏼

apps/web/src/lib/auth/getSession.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  try {
+    if (result.session && result.session.fresh) {
+      const sessionCookie = lucia.createSessionCookie(result.session.id)
+      cookies().set(
+        sessionCookie.name,
+        sessionCookie.value,
+        sessionCookie.attributes,
+      )
+    }
```
geclos, line 29: Explain to me why do we need to set the session cookie again if it has already been validated in line 19?

apps/web/src/lib/auth/getSession.ts
```diff
@@ -0,0 +1,46 @@
...
+
+      cookies().set(
+        sessionCookie.name,
+        sessionCookie.value,
+        sessionCookie.attributes,
+      )
+    }
+  } catch {
+    // ignore
+    // next.js throws when you attempt to set cookie when rendering page
```
geclos, line 42: so this is the recommended way of doing sessions in nextjs with lucia? looks hacky to me

apps/web/src/lib/auth/index.ts
```diff
@@ -0,0 +1,36 @@
+import { sessions, users } from '@latitude-data/core'
+import { DrizzlePostgreSQLAdapter } from '@lucia-auth/adapter-drizzle'
+import db from '$/db/database'
```
geclos, line 3: hmmm I've removed this code in my PR, since we shouldn't export db to consumers imo

packages/core/src/data-access/users.ts
```diff
@@ -0,0 +1,98 @@
...
+  password: z.string(),
+})
+
+export type SessionData = {
+  user: SafeUser
+  workspace: { id: number; name: string }
+}
+
+export async function getUserFromCredentials(
+  credentials: { email: unknown; password: unknown },
```
csansoon, line 26: How are these unknown? Aren't they strings?

csansoon (commented):
👍

geclos (dismissed):
looks good 👌🏼 !

apps/web/src/lib/auth/getSession.ts
andresgutgon: I followed [lucia docs](https://lucia-auth.com/tutorials/username-and-password/nextjs-app)
image

packages/core/src/data-access/users.ts
andresgutgon: Not anymore

apps/web/src/lib/auth/getSession.ts
andresgutgon: Yep https://github.com/latitude-dev/latitude-llm/pull/33#discussion_r1677487956

PR #34, opened 2024-07-10
feature: sidebar

geclos:
This commits implements a wakky first version of the dashboard Sidebar
and all the plumbing required:

- client-side stores
- db operations
- server/client component splits where needed
- server REST API
- Closure table for hierarchy

Missing:
- tenancy
- authentication in api
- proper error handling in api
- tests

PR #36, opened 2024-07-11
Document editor section

csansoon:
## Section for document editor.

There is currently **no login** in the component other than parsing errors and identifying parameters. The Playground UI is being revised, and its logic will depend on the LLM generation, which is yet to be implemented.

The `DocumentEditor` component currently receives a `document` as the plain document string, and displays it. There is no document retrieval logic yet either, as it depends on the current Commit.

https://github.com/user-attachments/assets/1d6e78a2-fefa-4697-856f-275673c474a8


packages/env/src/env/development.env
```diff
@@ -1,11 +1,11 @@
...
+DATABASE_URL=postgres://latitude:secret@localhost:5432/latitude_development
 
 ELASTIC_URL=https://elastic:9200
 ELASTIC_USERNAME=latitude
 ELASTIC_PASSWORD=secret
 
 REDIS_PORT=6379
-REDIS_HOST=redis
+REDIS_HOST=localhost
 REDIS_PASSWORD=secret
```
csansoon, line 11: Running `web` locally because otherwise my mac takes ~20 seconds to hot-refresh any changes 😭

packages/web-ui/styles.css
csansoon, line 1: I need to specify the theme values in runtime as a hex string on Monaco. I can just use the computed styles, but for that I need to have them defined as hex strings too.

packages/env/src/env/development.env
andresgutgon: We need to think how to enable host run nextjs for Carlos in a way that can be commited. Maybe a `$USE_WEB_IN_HOST` env variable that’s  sets this to `localhost` or maybe better load the `development.localhost.env` file. Do a template file

packages/web-ui/package.json
```diff
@@ -29,17 +32,18 @@
...
+    "tailwind-merge": "^2.4.0",
+    "tailwindcss-animate": "^1.0.7"
   },
   "devDependencies": {
     "@latitude-data/eslint-config": "workspace:*",
     "@latitude-data/typescript-config": "workspace:*",
     "@types/react": "18.3.0",
     "@types/react-dom": "18.3.0",
     "autoprefixer": "^10.4.19",
+    "monaco-editor": "^0.50.0",
```
andresgutgon, line 46: Shouldn't be this a dependency instead of devDependency?

packages/web-ui/src/ds/atoms/DocumentTextEditor/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  import('./editor').then(
+    (module) =>
+      ({ default: module.DocumentTextEditor }) as {
+        default: React.ComponentType<DocumentTextEditorProps>
+      },
+  ),
+)
+
+function EditorWrapper(props: DocumentTextEditorProps) {
+  if (typeof window === 'undefined') return null
```
andresgutgon, line 23: Leave a comment explaining why this is needed.

packages/web-ui/src/sections/DocumentEditor/index.tsx
```diff
@@ -0,0 +1,34 @@
+'use client'
+
+import { useEffect, useState } from 'react'
+
+import { ConversationMetadata, readMetadata } from '@latitude-data/compiler'
+import { DocumentTextEditor, Text } from '$ui/ds/atoms'
```
andresgutgon, line 6: Talked offline. Move out of `ds` this is not a design system element.

packages/web-ui/src/sections/DocumentEditor/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+          <Text.Mono color='foreground'>Prompt editor</Text.Mono>
+        </div>
+        <DocumentTextEditor
+          value={value}
+          metadata={metadata}
+          onChange={setValue}
+        />
+      </div>
+      <div className='flex flex-col flex-1'>
+        <div className='flex flex-row space-between'>
```
andresgutgon, line 28: Is amazing how having abstracted the ui elements all it remains for tailwind is to do do the flex thing

packages/web-ui/tsconfig.json
```diff
@@ -2,8 +2,8 @@
   "extends": "@latitude-data/typescript-config/base.json",
   "compilerOptions": {
     "jsx": "react-jsx",
-    "moduleResolution": "NodeNext",
-    "module": "NodeNext",
+    "moduleResolution": "Bundler",
+    "module": "ESNext",
```
andresgutgon, line 6: Let's review this decision carefully before merging. I know you did to make it work but let's thing deeper why this is necessary

packages/env/src/index.ts
```diff
@@ -1,20 +1,21 @@
...
+  dotenv.populate(process.env as DotenvPopulateInput, {
+    NODE_ENV: env,
+    DATABASE_URL: `postgres://latitude:secret@${dbHost}:5432/latitude_development`,
+    ELASTIC_URL: 'https://elastic:9200',
+    ELASTIC_USERNAME: 'latitude',
+    ELASTIC_PASSWORD: 'secret',
+    REDIS_PORT: '6379',
+    REDIS_HOST: redisHost,
+    REDIS_PASSWORD: 'secret',
+  })
```
csansoon, line 20: Did a bit of DRY and added a way to configure the correct host depending on whether the package is being ran locally or on a docker image

packages/web-ui/src/ds/tokens/font.ts
```diff
@@ -6,7 +6,7 @@ export const font = {
     h5: 'text-sm leading-5', // 14px/20px
     h4: 'text-normal leading-6', // 16px/24px
     h3: 'text-xl leading-8', // 20px/32px
-    h2: 'text-h2 leading-10', // 26px/40px
+    h2: 'text-[26px] leading-10', // 26px/40px
```
andresgutgon, line 9: Put in tailwind

andresgutgon (dismissed):

apps/web/src/app/layout.tsx
```diff
@@ -31,7 +37,7 @@ export default function RootLayout({
         <meta name='viewport' content='width=device-width, initial-scale=1' />
         <link rel='icon' href='/favicon.svg' />
       </head>
-      <body className={`${fontSans.variable} ${fontMono.variable}`}>
+      <body className={`${fontSans.variable} ${fontMono.variable} font-sans`}>
```
andresgutgon, line 40: Why `font-sans` here?

packages/compiler/package.json
```diff
@@ -14,8 +21,6 @@
     "lint": "eslint src",
     "tc": "tsc --noEmit"
   },
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
```
andresgutgon, line 18: It's really necessary to change this?

apps/web/src/app/layout.tsx
csansoon: Otherwise any text that does not use this class would not be assigned any font, and would just use the browser's fallback.

An example where this happens is the label for form's inputs.

packages/compiler/package.json
csansoon: Now that there’s no ⁠dist, everything comes directly from the original ⁠src file. Otherwise, TypeScript tries to load the package from that file, but the linter workflow hasn’t built it yet and it fails.

apps/web/src/app/layout.tsx
andresgutgon: > An example where this happens is the label for form's inputs.

This is because label don't use `<Text />` component.

apps/web/tsconfig.json
```diff
@@ -10,7 +10,8 @@
       "$/*": ["./src/*"],
       "$core/*": ["../../packages/core/src/*"],
       "$jobs/*": ["../../packages/jobs/src/*"],
-      "$ui/*": ["../../packages/web-ui/src/*"]
+      "$ui/*": ["../../packages/web-ui/src/*"],
+      "$compiler/*": ["../../packages/compiler/src/*"]
```
andresgutgon, line 14: Is compiler used by web? Or is used by core? If it's used by core put this in core tsconfig.json

csansoon: It's used by web-ui, and will be used by core too in the future.

andresgutgon: But does it web need to know about it here?

csansoon: Yes, in the same way that `web` needs to install all `web-ui` dependencies too. Since it is importing the original `src/index.ts` instead of a dist build, it's the parent package who needs to resolve all those imports; in this case, the `web` package.

andresgutgon: Can you try this in web-ui/tsconfig.json for compiler?
https://github.com/latitude-dev/latitude-llm/pull/34/files#r1677555026

Maybe that way this relation is not necesary

`web -> web-ui -> compiler`. Web and compiler shouldn't know about each other

csansoon: Does not work. Web still needs to resolve `$compiler`

andresgutgon (approved):

PR #38, opened 2024-07-15
Sidebar fix server actions

andresgutgon:


apps/web/src/actions/documents/create.ts
```diff
@@ -19,6 +25,13 @@ export const createDocumentVersionAction = authProcedure
...
   )
-  .handler(async ({ input }) => createDocumentVersion(input))
+  .handler(async ({ input }) => {
+    await createDocumentVersion(input)
+    await materializeDocumentsAtCommit({
+      commitUuid: input.commitUuid,
+      staged: true,
+    })
+
+    revalidatePath(ROUTES.root)
```
geclos, line 36: can we keep it a regular pure function that executes a write and returns the result? In general actions that perform side effects are more complex to maintain.

apps/web/src/components/Sidebar/DocumentTree/index.tsx
```diff
@@ -20,13 +22,14 @@ export function CreateNode({ parentId }: { parentId?: number }) {
...
   return (
     <button
-      onClick={() =>
-        create({
+      onClick={async () =>
+        await execute({
+          commitUuid: 'HEAD',
           parentId,
-          documentType: 'folder' as DocumentType.Folder,
+          documentType: DocumentType.Folder,
```
geclos, line 32: this fails because DocumentType is declared in a file that has other non-client deps like drizzle. You'll have to move the declaration of documenttype if u wanna make it work

apps/web/src/actions/documents/create.ts
andresgutgon: Yes agree. I was just trying someting

apps/web/src/actions/documents/create.ts
```diff
@@ -9,16 +9,14 @@ export const createDocumentVersionAction = authProcedure
...
+      commitUuid: z.string(),
       parentId: z.number().optional(),
+      documentType: z.nativeEnum(DocumentType).optional(),
     }),
     { type: 'json' },
   )
-  .handler(async ({ input }) => createDocumentVersion(input))
+  .handler(async ({ input }) => {
+    const result = await createDocumentVersion(input)
+    return result.unwrap()
```
andresgutgon, line 21: This was one of the issues. If the server actions return data to the frontend it has to be serializable
```
Error: Only plain objects, and a few built-ins, can be passed to Server Actions. Classes or null prototypes are not supported. 
```

apps/web/src/stores/documentVersions.ts
```diff
@@ -1,8 +1,11 @@
 'use client'
 
-import { DocumentVersion, HEAD_COMMIT } from '@latitude-data/core'
+/* import { HEAD_COMMIT, type DocumentType } from '@latitude-data/core' */
```
andresgutgon, line 3: I think this was other problem. Core package contains server code and when we do import things from it in client it fails. 

This is very surprising to be honest

packages/core/src/data-access/commits.ts
```diff
@@ -1,4 +1,6 @@
-import { commits, database, HEAD_COMMIT } from '@latitude-data/core'
+import { database } from '$core/client'
+import { HEAD_COMMIT } from '$core/constants'
+import { commits } from '$core/schema'
```
andresgutgon, line 3: ☝️ Reference core in core does not makes sense

packages/core/src/services/documentVersions/create.ts
```diff
@@ -42,7 +42,9 @@ export async function createDocumentVersion({
     }
 
     const res = await createCommit(tx)
-    if (!res.ok) return res
+    if (res.error) return res
+
+    const commitId = res.value?.id!
```
andresgutgon, line 47: No change but I think is more clear

apps/web/src/actions/documents/create.ts
```diff
@@ -9,16 +9,14 @@ export const createDocumentVersionAction = authProcedure
...
-      documentType: z
-        .enum([
-          'folder' as DocumentType.Folder,
-          'document' as DocumentType.Document,
-        ])
-        .optional(),
       name: z.string(),
+      commitUuid: z.string(),
       parentId: z.number().optional(),
+      documentType: z.nativeEnum(DocumentType).optional(),
```
andresgutgon, line 15: You can pass typescript enums to zod this way

apps/web/src/stores/documentVersions.ts
```diff
@@ -1,8 +1,9 @@
 'use client'
 
-import { DocumentVersion, HEAD_COMMIT } from '@latitude-data/core'
+import { HEAD_COMMIT, type DocumentType } from '@latitude-data/core/browser'
```
andresgutgon, line 3: I split core in server `@latitude-data/core` and browser `@latitude-data/core/browser`

packages/core/src/browser.ts
csansoon, line 1: Why is this `browser` required in `core`? Is core not agnostic to its environment?

andresgutgon: No, server vs backend split is there and we need to embrace it. There are reasons to want frontend consume core entities like constants or types

apps/web/src/stores/documentVersions.ts
```diff
@@ -24,17 +25,19 @@ export default function useDocumentVersions(
...
   const create = async (payload: {
     commitUuid?: string
     name: string
-    documentType?: DocumentVersion['documentType']
+    documentType?: DocumentType
     parentId?: number
   }) => {
     try {
-      const doc = await createDocumentVersionAction({
-        ...payload,
```
andresgutgon, line 35: I try to avoid destructuring. Is hard to see all that's passed

packages/core/src/browser.ts
geclos: it cannot be that you can import the same thing from core and browser though.

andresgutgon: What do you propose?

geclos: You can either split core into browser and server and not sharee exports between them or you can keep core and browser and don't export browser from core. In any case the bottom line is that you shouldn't be able to export some browser-safe code from supposedly server-specific stuff.

packages/core/src/schema/index.ts
```diff
@@ -1,12 +1,18 @@
 export * from './db-schema'
+
+// Closure tree hierarchy
+// This relies on a SQL TRIGGER. Check the code in ../../drizzle folder
```
geclos, line 4: already added docs, no need for these

andresgutgon: At least link it here no?

apps/web/src/stores/documentVersions.ts
```diff
@@ -1,9 +1,14 @@
...
-import { DocumentVersion, HEAD_COMMIT } from '@latitude-data/core'
+import { useCallback } from 'react'
+
+import { DocumentVersion } from '@latitude-data/core'
+import { HEAD_COMMIT, type DocumentType } from '@latitude-data/core/browser'
 import { createDocumentVersionAction } from '$/actions/documents/create'
 import useSWR, { SWRConfiguration } from 'swr'
+import { useServerAction } from 'zsa-react'
 
+const FIXME_HARDCODED_PROJECT_ID = 1
```
csansoon, line 11: ☝️

apps/web/tsconfig.json
```diff
@@ -12,6 +12,9 @@
     "baseUrl": ".",
     "paths": {
       "$/*": ["./src/*"],
+      "@latitude-data/core": ["../../packages/core/src/*"],
+      "@latitude-data/jobs": ["../../packages/jobs/src/*"],
+      "@latitude-data/web-ui": ["../../packages/web-ui/src/*"],
```
csansoon, line 17: wat why

apps/web/src/app/api/commits/[commitUuid]/documents/route.ts
```diff
@@ -7,9 +7,9 @@ export async function GET(
 ) {
   try {
     const staged = Boolean(req.nextUrl.searchParams.get('staged') || false)
-    const nodes = await materializeDocumentsAtCommit({ commitUuid, staged })
+    const documents = await materializeDocumentsAtCommit({ commitUuid, staged })
 
-    return NextResponse.json(nodes)
+    return NextResponse.json(documents)
```
Comment (andresgutgon) [inline apps/web/src/app/api/commits/[commitUuid]/documents/route.ts:12]: Proposal. Instead of returning the documents here I would return the tree and call this API something like 
`/api/sidebar/commitUuid` 

Let's data over the wire

docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
     depends_on:
       - base
       - db
-    command: pnpm run dev
+    command: pnpm run dev:debug
```
csansoon, line 29: Should this be kept on?

apps/web/src/stores/documentVersions.ts
```diff
@@ -19,30 +24,37 @@ export default function useDocumentVersions(
...
-      mutate([...data, doc])
+      const prev = documents ?? []
 
-      return doc
-    } catch (err) {
-      console.error(err)
-    }
-  }
+      if (document) {
+        mutate([...prev, document])
```
andresgutgon, line 51: Instead of dealing with documents and then the UI having to do the `toTree` this end point could return the tree directly

packages/core/src/data-access/projects.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+export async function findProject({
+  projectId,
+  workspaceId,
+}: {
+  projectId: number
+  workspaceId: number
+}) {
```
csansoon, line 18: Do we need a workspaceId to find the project?

apps/web/tsconfig.json
andresgutgon, line 17: I hope this improve the auto-import in our LSPs

packages/core/src/schema/models/commits.ts
```diff
@@ -44,10 +38,10 @@ export const commits = latitudeSchema.table(
 
 export const commitRelations = relations(commits, ({ one, many }) => ({
   snapshots: many(documentSnapshots, { relationName: 'snapshots' }),
-  workspace: one(workspaces),
+  workspace: one(projects, {
+    fields: [commits.projectId],
+    references: [projects.id],
+  }),
```
csansoon, line 44: project*

packages/core/drizzle/0002_windy_wind_dancer.sql
```diff
@@ -1,20 +1,15 @@
-CREATE SCHEMA "latitude";
---> statement-breakpoint
 DO $$ BEGIN
- CREATE TYPE "public"."document_type" AS ENUM('document', 'folder');
+ CREATE TYPE "latitude"."document_type" AS ENUM('document', 'folder');
```
andresgutgon, line 2: Wrong schema

packages/core/src/services/commits/create.ts
```diff
@@ -1,16 +1,23 @@
...
+}: {
+  projectId: number
+  db?: Database
+}) {
+  return Transaction.call<Commit>(async (tx) => {
+    const result = await tx.insert(commits).values({ projectId }).returning()
+    const commit = result[0]
+
+    return Result.ok(commit!)
+  }, db)
```
andresgutgon, line 22: Use `Transactions` and DRY the creation of document. Now is way clear

csansoon (approved):

apps/web/src/stores/documentVersions.ts
csansoon: Nah just don't forget to fix this later

packages/core/drizzle/0002_windy_wind_dancer.sql
csansoon: Isn't this auto generated?

packages/core/src/data-access/projects.ts
andresgutgon: Yes. We only want to find the right combination of project Y that belongs to workspace X

packages/core/src/schema/models/commits.ts
andresgutgon: 💯 Good catch

packages/core/src/services/documentVersions/create.ts
```diff
@@ -1,63 +1,67 @@
...
 }: {
+  projectId: number
   name: string
   commitUuid: string
   documentType?: DocumentType
   parentId?: number
 }) {
-  const data = { name, parentId, documentType }
+  let commit = await findCommit({ uuid: commitUuid })
+  return Transaction.call<DocumentVersion>(async (tx) => {
```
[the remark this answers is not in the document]
geclos: thanks

PR #39, opened 2024-07-15
Changed web font

csansoon:
Changed web font to IMB Plex Mono (both for the monospace and the regular sans version)


apps/web/src/app/layout.tsx
```diff
@@ -1,24 +1,141 @@
...
+  ],
   variable: '--font-sans',
 })
-const fontMono = IBM_Plex_Mono({
-  subsets: ['latin'],
-  weight: ['400', '500', '600', '700'],
-  style: ['normal'],
-  display: 'swap',
+
+const fontMono = localFont({
```
andresgutgon, line 76: Extract somewhere else please.

andresgutgon (approved):

apps/web/src/app/layout.tsx
csansoon: I tried! But I got an error where `Font loader values must be explicitly written literals.`

geclos (approved):

PR #40, opened 2024-07-16
Current commit management

csansoon:
Now the user will be able to select which commit to load into the app. This will let them not only watch older versions of the project, but also create new draft commits and work on them, while other users can simultaneously work on their own drafts.

This requires some changes on the code.
- created `commitContext` to store in the app's localStorage the current selected commit. Only draft commits are editable, so another `isEditable` property is also exposed.

## TODO
The currently selected commit should be scoped by project, not the whole client.


apps/web/src/actions/commits/index.ts
```diff
@@ -0,0 +1,19 @@
...
+  .input(
+    z.object({
+      commitUuid: z.string(),
+    }),
+    { type: 'json' },
+  )
+  .handler(async ({ input }) => {
+    const res = await isCommitDraft({ uuid: input.commitUuid })
+    return res.unwrap()
+  })
```
geclos, line 19: does this need to be a server action? you are gonna need the list of commits in the frontend regardless so you can check there is a commit is draft or not

packages/core/src/data-access/commits.ts
```diff
@@ -27,6 +34,25 @@ export async function listCommits() {
...
+
+  if (commit.nextCommitId) return Result.ok(false)
+
+  const commitsReferencingCommit = await tx
+    .select({ id: commits.id })
+    .from(commits)
+    .where(eq(commits.nextCommitId, commit.id))
+
+  return Result.ok(commitsReferencingCommit.length === 0)
 }
```
geclos, line 58: data access should not answer booleans or perform business logic they just an interface to the database. Think of them as repositories.

packages/core/src/schema/models/commits.ts
```diff
@@ -44,7 +44,10 @@ export const commits = latitudeSchema.table(
 
 export const commitRelations = relations(commits, ({ one, many }) => ({
   snapshots: many(documentSnapshots, { relationName: 'snapshots' }),
-  workspace: one(workspaces),
+  workspace: one(workspaces, {
+    fields: [commits.workspaceId],
+    references: [workspaces.id],
+  }),
```
geclos, line 50: this is gonna change so i'd probably wouldn't touch it (andres is moving tenancy of commits to projects)

apps/web/src/stores/commitContext.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const updateCurrentCommit = ({
+    commitUuid,
+    isEditable,
+  }: {
+    commitUuid?: string
+    isEditable: boolean
+  }) => {
+    setCurrentCommit(commitUuid)
+    setIsCommitEditable(isEditable)
```
geclos, line 55: this is derived state so it should be a memoized computation, don't duplicate the state.

packages/web-ui/src/providers/commitProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+}: {
+  children: ReactNode
+} & CommitContextType) => {
+  return (
+    <CommitContext.Provider value={context}>{children}</CommitContext.Provider>
+  )
+}
+
+const useCurrentCommit = () => {
+  return useContext(CommitContext)
```
andresgutgon, line 31: In useContext is good practice put this check
```
  if (!context) {
    throw new Error(
      'useCurrentWorkspace must be used within a WorkspaceProvider',
    )
  }

```

andresgutgon:
> store in the app's localStorage the current selected commit

Thinking about this we could store it in a cookie so the selection of the commit happens on the server

packages/core/vitest.config.mjs
andresgutgon, line 1: Do a PR only with the vitest setup. Also setup the CI to run tests. In that PR do a dummy test `expect(true).toBe(true)` to make sure everything is working

packages/web-ui/src/providers/commitProvider.tsx
```diff
@@ -0,0 +1,38 @@
+'use client'
+
+import { createContext, ReactNode, useContext } from 'react'
+
+import { HEAD_COMMIT } from '@latitude-data/core/browser'
+
+// const HEAD_COMMIT = 'HEAD'
+
+interface CommitContextType {
+  commitUuid: string | typeof HEAD_COMMIT
```
andresgutgon, line 10: Move this type to `core/src/constants` It will be handy for more cases. Also call the type somethiing like `HeadCommit`

packages/web-ui/src/providers/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './commitProvider'
+export * from './projectProvider'
```
andresgutgon, line 2: I'm doing the project provider. We'll have a little collision but it is fine.

packages/web-ui/src/providers/projectProvider.tsx
andresgutgon, line 1: Providers a Uppper case by convention. `ProjectProvider.ts`

packages/core/src/data-access/commits.ts
```diff
@@ -1,26 +1,78 @@
...
+): Promise<TypedResult<Date | null, LatitudeError>> {
+  if (commitUuid === HEAD_COMMIT) {
+    const result = await tx
+      .select({ mergedAt: commits.mergedAt })
+      .from(commits)
+      .where(and(eq(commits.projectId, projectId), isNotNull(commits.mergedAt)))
+      .orderBy(desc(commits.mergedAt))
+      .limit(1)
+
+    if (!result.length)
```
andresgutgon, line 62: Use `{}` for the ifs when are more than one line please

packages/core/src/data-access/commits.ts
```diff
@@ -1,26 +1,78 @@
...
+      return Result.error(new NotFoundError('No head commit found'))
+    const headCommit = result[0]!
+    return Result.ok(headCommit.mergedAt!)
+  }
+
+  const result = await tx
+    .select({ mergedAt: commits.mergedAt })
+    .from(commits)
+    .where(eq(commits.uuid, commitUuid))
+    .limit(1)
```
andresgutgon, line 72: For the findOne, it is better to use `findFirst`. You don't need to do the `limit(1)[0]` thing.

I believe something like this works:
```javascript
tx.query.commits.findFirst({
  with: { mergedAt: true },
  where: eq(commits.uuid, commitUuid)
})
```

packages/core/src/schema/models/commits.ts
```diff
@@ -20,19 +20,19 @@ export const commits = latitudeSchema.table(
...
+      .references(() => projects.id, { onDelete: 'cascade' }),
+    mergedAt: timestamp('merged_at'),
     ...timestamps(),
   },
   (table) => ({
-    nextCommitIdx: index('commit_next_commit_idx').on(table.nextCommitId),
+    projectCommitOrderIdx: index('project_commit_order_idx').on(
+      table.mergedAt,
+      table.projectId,
+    ),
```
andresgutgon, line 35: Does this change the schema?

packages/core/src/tests/factories/projects.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      return Result.ok(newProject)
+    }, db)
+  }
+
+  const result = await createProjectFn({
+    name: name ?? randomName,
+    workspaceId: workspace.id,
+  })
+  const project = result.unwrap()
```
andresgutgon, line 54: Don't do unwraps in services. Used in other services it will 💥 . Is better always return the Result

packages/core/src/tests/factories/projects.ts
```diff
@@ -0,0 +1,57 @@
...
+    const newWorkspace = await createWorkspace(workspaceData)
+    workspace = newWorkspace.workspace
+    user = newWorkspace.userData
+  }
+
+  const randomName = faker.commerce.department()
+  const { name } = projectData
+
+  // TODO: Replace with actual service
+  const createProjectFn = ({
```
andresgutgon, line 31: At least move out of this function

packages/web-ui/src/providers/commitProvider.tsx
```diff
@@ -0,0 +1,38 @@
+'use client'
+
+import { createContext, ReactNode, useContext } from 'react'
+
+import { HEAD_COMMIT } from '@latitude-data/core/browser'
+
+// const HEAD_COMMIT = 'HEAD'
```
andresgutgon, line 7: 🔥

packages/core/src/data-access/commits.ts
csansoon: They were originally one-liners but prettier apparently didn't like it 😭

packages/core/src/tests/factories/projects.ts
csansoon: This is just a test factory

csansoon:
☝️ Redistributed as multiple different PRs

PR #41, opened 2024-07-16
Fix production build broken by missing env DATABASE_URL at build time 

andresgutgon:
# What?
We moved DB initialization to core: https://github.com/latitude-dev/latitude-llm/pull/34 This breaks production build because at build time DATABASE_URL is not and should never be defined

## TODO
- [x] QA the build works

### DIFF
--- a/apps/web/docker/Dockerfile
+++ b/apps/web/docker/Dockerfile
@@ -5,9 +5,11 @@ COPY . .
 
 RUN turbo prune @latitude-data/web --docker
 
-# COMMON stage (installer)
+# BUILDER stage
 # ------------------------------------------------------
-FROM base AS installer
+FROM base AS builder
+WORKDIR /app
+
 COPY --from=base /app/out/json/ .
 COPY --from=base /app/out/pnpm-lock.yaml ./pnpm-lock.yaml
 COPY --from=base /app/out/pnpm-workspace.yaml ./pnpm-workspace.yaml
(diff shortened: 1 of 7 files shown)


apps/web/docker/Dockerfile
```diff
@@ -17,38 +18,37 @@ RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install \
...
+
+# Unpriviledged user
 RUN addgroup --system --gid 1001 nodejs
 RUN adduser --system --uid 1001 nextjs
 USER nextjs
 
 COPY --from=installer /app/apps/web/next.config.mjs .
 COPY --from=installer /app/apps/web/package.json .
-COPY --from=installer /app/node_modules ./node_modules
-COPY --from=installer /app/apps/web/package.json ./apps/web/package.json
```
andresgutgon, line 40: Double check the 2 lines are not necessary. Not sure the way pnpm works we at least need the root `app/node_modules`

docker-compose.yml
```diff
@@ -21,7 +23,26 @@ services:
...
       - '6379:6379'
+
+  web:
+    build:
+      context: .
+      target: production
+      dockerfile: apps/web/docker/Dockerfile
+    # This enables to have this service for testing production build
+    # It will not run when doing docker compose up
+    profiles: [building]
```
andresgutgon, line 39: Docker compose allow to have profiles. So no profile runs when `docker compose up` which is perfect for this case. I only want for the combinience of developing Dockerfile production and run `docker compose build web`

packages/core/package.json
```diff
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "@latitude-data/env": "workspace:^",
+    "@t3-oss/env-core": "^0.10.1",
```
andresgutgon, line 27: I have the perfect excuse for this sorry.

packages/core/src/env.ts
```diff
@@ -1,16 +1,15 @@
 import z from 'zod'
+import { createEnv } from "@t3-oss/env-core"
 
 import '@latitude-data/env'
 
-const envvars = z.object({
```
andresgutgon, line 5: Doing this breaks production build because it tries to find `DATABASE_URL` at build time. With t3-env this is easy fixable by using `skipValidation` with `BUILDING_CONTAINER` flag as we did in web.

packages/core/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@latitude-data/typescript-config/base.json",
   "compilerOptions": {
+    "moduleResolution": "nodenext",
```
andresgutgon, line 4: I needed this to make it work t3-env, default in base is `node`. 😂 Let's hope is fine.

apps/api/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@latitude-data/typescript-config/base.json",
   "compilerOptions": {
+    "moduleResolution": "Bundler",
```
andresgutgon, line 4: > 'bundler' for use with bundlers. Like node16 and nodenext, this mode supports package.json "imports" and "exports", but unlike the Node.js resolution modes, bundler never requires file extensions on relative paths in imports.

[Reference](https://www.typescriptlang.org/tsconfig/#moduleResolution)

I think `Bundler` is ok

packages/core/src/env.ts
geclos: why? I don't get it why is this code evaluated at build time?

packages/core/package.json
geclos: plz... `if (process.env.BUILDING_CONTAINER == 'true') return`

packages/core/src/env.ts
andresgutgon: Because it's imported by nextjs bundler. And because is imported zod `parse` function is run

packages/core/package.json
andresgutgon: image

apps/api/tsconfig.json
geclos: why do u needed this change

packages/core/package.json
andresgutgon: I don't understand why you don't want a tool [built for this propose](https://env.t3.gg/docs/introduction#support-for-multiple-environments)

apps/api/tsconfig.json
andresgutgon: For `t3-env/core`. It use a module system that requires the more modern version. But the thing is we're already using `Bundler` in NextJS app.

andresgutgon: An I think `bundler` makes a lot of sense for us because we're using one only final bundler in nextjs or api and not having by package bundlers

packages/core/package.json
geclos: I actually like it for NextJS for the client-side safety measures that you mention ☝🏼 . I hate that we use it in api because it brings 0 value there and adding a dependency is just overhead.

packages/core/src/env.ts
geclos: next bundler runs the code of its dependencies at compile time? :drunk:

packages/core/package.json
andresgutgon: Also this package is type save. When you do `env.SOMETHING` you're sure that's a number or string. Doing the old way you still use untyped process.env

PR #42, opened 2024-07-17
feature: api fetch prompts by commit and path

geclos:
This commit tweaks queries and implements an example on how to find a
prompt by path and commit uuid. It also removes unnecessary table
documentSnapshots.


apps/api/src/routes/api/v1/commits/routes.ts
```diff
@@ -0,0 +1,39 @@
+import { materializeDocumentsAtCommit, Node, toTree } from '@latitude-data/core'
+import HttpStatusCodes from '@src/common/HttpStatusCodes'
```
andresgutgon, line 2: Use `$` for paths we're using it in all projects in this monorepo

apps/web/src/components/Sidebar/DocumentTree/index.tsx
```diff
@@ -54,31 +83,56 @@ function TreeNode({ node, level = 0 }: { node: Node; level?: number }) {
...
+      <div className='flex flex-col gap-4'>
+        <div className='flex flex-row align-items justify-between'>
+          <h2 className='font-bold'>Create document</h2>
+          <CreateNode commit={commit} />
+        </div>
+        <TreeNode commit={commit} node={rootNode} />
+      </div>
+    </div>
+  )
 }
```
geclos, line 138: most of this ui will probably get removed so don't pay too much attention to it.

apps/web/src/components/Sidebar/index.tsx
```diff
@@ -1,22 +1,17 @@
...
-      <div className='flex flex-row align-items justify-between'>
-        <h2>Prompts</h2>
-        <div className='flex flex-row gap-2 align-items'>
-          <CreateNode />
-        </div>
-      </div>
-      <DocumentTree documents={documents} />
-    </div>
-  )
+  const commits = await listCommits()
```
geclos, line 14: it's puzzling that read queries are not wrapped in result but writes are, we have to wrap all in result

packages/core/src/services/documentVersions/create.ts
```diff
@@ -1,25 +1,28 @@
...
+    commitId,
+    parentId,
+    documentType,
+  }: {
+    name: string
+    commitId: number
+    parentId?: number
+    documentType?: DocumentType
+  },
+  db = database,
```
geclos, line 24: as a convention all db accessors accept a db as an optional last parameter in order to pass transaction contexts around

packages/core/src/services/documentVersions/materializeAtCommit.ts
```diff
@@ -1,28 +1,30 @@
...
-    if (!staged) return snapshots
+    .groupBy(documentVersions.documentUuid, documentVersions.id, commits.id)
+    .orderBy(desc(commits.mergedAt))
 
-    const versions = await listStagedDocuments()
-    return uniqBy([...versions, ...snapshots], (doc) => doc.documentUuid)
-  } else {
-    return await getDocumentsAtCommit(commitUuid)
-  }
+  return Result.ok(docsInCommits.map((doc) => doc.document_versions))
```
geclos, line 29: this query gets you all documents in a main branch commit, no staged documents. Useful for production use cases where you wanna target only main branch commits.

apps/web/src/components/Sidebar/index.tsx
andresgutgon: I hard agree on this one

packages/core/src/services/documentVersions/create.ts
andresgutgon: We had that. Look for `AppContext`

apps/web/src/components/Sidebar/DocumentTree/index.tsx
andresgutgon: My next task is to do the sidebar styles.

apps/web/src/stores/commits.ts
```diff
@@ -0,0 +1,39 @@
...
+  const key = '/api/commits'
+  const { execute } = useServerAction(createCommitAction)
+  const {
+    data = [],
+    mutate,
+    ...rest
+  } = useSWR<Commit[]>(
+    key,
+    async () => {
+      const [commits] = await getCommitsAction({ projectId })
```
andresgutgon, line 22: We should implement a convention on how to deal with errors on actions in the frontend. We have errors in the returned value here on outside. Let's pick one and stick to it. Specially for actions that are "GET" like this

apps/web/src/stores/commits.ts
```diff
@@ -0,0 +1,39 @@
...
+      return commits!
+    },
+    opts,
+  )
+
+  const create = useCallback(async () => {
+    const [commit] = await execute({
```
andresgutgon, line 29: I guess same apply to writes

apps/web/src/stores/documentVersions.ts
```diff
@@ -20,35 +20,55 @@ export default function useDocumentVersions(
...
+      const [document, err] = await createDocumentVersionAction({
         ...payload,
         projectId,
         name: payload.name!,
         commitUuid: commitUuid || HEAD_COMMIT,
       })
-      const prev = documents ?? []
+
+      if (err) {
+        console.error(err)
```
andresgutgon, line 59: You have `toast` which is nice. Talk with product to see how we display errors

packages/core/drizzle/0000_nice_daimon_hellstrom.sql
```diff
@@ -1,10 +0,0 @@
-CREATE SCHEMA "latitude";
---> statement-breakpoint
-CREATE TABLE IF NOT EXISTS "latitude"."document_hierarchies" (
-	"id" bigserial PRIMARY KEY NOT NULL,
-	"parent_id" bigint,
-	"depth" integer NOT NULL,
-	"child_id" bigint NOT NULL,
-	"created_at" timestamp DEFAULT now() NOT NULL,
-	"updated_at" timestamp DEFAULT now() NOT NULL
-);
```
andresgutgon, line 10: What happened?

andresgutgon: Why you removed an existing migration?

packages/core/drizzle/0003_curious_anthem.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE "latitude"."commits" ADD COLUMN "merged_at" timestamp;--> statement-breakpoint
+ALTER TABLE "latitude"."commits" DROP COLUMN IF EXISTS "title";--> statement-breakpoint
```
andresgutgon, line 2: Why? We're already using this column

packages/core/src/data-access/commits.ts
```diff
@@ -26,13 +26,14 @@ export async function findHeadCommit(
...
-  projectId: number
+  projectId?: number
 }
 export async function findCommit(
   { projectId, uuid }: FindCommitProps,
   tx = database,
 ): Promise<TypedResult<Commit, LatitudeError>> {
-  if (uuid === HEAD_COMMIT) return findHeadCommit({ projectId }, tx)
+  if (uuid === HEAD_COMMIT && projectId)
+    return findHeadCommit({ projectId }, tx)
```
andresgutgon, line 36: Brackets

packages/core/drizzle/0000_nice_daimon_hellstrom.sql
geclos: struggles with conflicting migrations 🤷🏼 I don't know how to solve them without a drop and generate but that changes all migrations

packages/core/src/data-access/commits.ts
geclos: that's prettier better not to touch it

andresgutgon: Prettier doesn't complain if you put the brackets

geclos: Sorry what I mean is that we cannot rely on andres linter to catch all prettier formattings on PRs that we don't like we should change the prettier config if we don't like it.

PR #44, opened 2024-07-18
Merged commits management

csansoon:
- Added `mergedAt` property to commits
- Created `ProjectProvider` and `CommitProvider` components
- Modified `getDocumentsAtCommit` and other methods to new schema.

Note: Many functions require both `commitUuid` and `projectId`, since the `commitUuid` can be 'HEAD', and we will require the project to know which commit does that represent.


apps/web/src/app/(private)/layout.tsx
```diff
@@ -14,12 +13,5 @@ export default async function PrivateLayout({
...
-  return (
-    <main className='flex flex-row w-full'>
-      <div className='w-[280px]'>
-        <Sidebar />
-      </div>
-      <div className='flex-1'>{children}</div>
-    </main>
-  )
+  return children
 }
```
csansoon, line 17: Sidebar has been moved to `/projects/[projectId]`

packages/web-ui/src/providers/CommitProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface CommitContextType {
+  commitUuid: string | typeof HEAD_COMMIT
+  isDraft: boolean
+}
+
+const CommitContext = createContext<CommitContextType>({
+  commitUuid: HEAD_COMMIT,
+  isDraft: false,
+})
```
csansoon, line 15: `useCurrentCommit` will return both the `commitUuid` and an `isDraft` property, which can be used to allow or forbid modifications depending on the commit.

packages/core/src/services/documentVersions/create.ts
```diff
@@ -36,32 +41,37 @@ function createDocument({
...
+  const commitResult = await findCommit({ commitUuid, projectId })
+  const commit = commitResult.unwrap()
 
-      commit = resultCommit.value
-    }
+  if (commit.mergedAt !== null) {
+    return Result.error(
+      new ForbiddenError('Cannot create a document version in a merged commit'),
+    )
+  }
```
csansoon, line 67: Added this restriction on `createDocumentVersion`.

andresgutgon (approved):

packages/core/src/services/documentVersions/create.ts
geclos: hmm would be cool to have model level validations like activerecord, @andresgutgon do u know if there's something for this in drizzle land?

andresgutgon: I saw this kind of repository pattern example:
https://gist.github.com/cayter/49d5c256a885d90c399ca6c1eca19f51

We could implement it and hook into `beforeCreate` to validate the row
https://gist.github.com/cayter/49d5c256a885d90c399ca6c1eca19f51#file-repository-ts-L357

But I didn't see anything official. 


I'm a fan of ActiveRecord but Drizzle is super functional. I don't think it is wrong to have a service to create documents with the validation inside. As long as we use it as unique way for creating documents

PR #45, opened 2024-07-19
Document path

csansoon:


packages/core/src/schema/models/documentVersions.ts
```diff
@@ -1,44 +1,48 @@
...
+      .references(() => commits.id),
+    deletedAt: timestamp('deleted_at'),
+    ...timestamps(),
+  },
+  (table) => ({
+    uniqueDocumentUuidCommitId: unique('unique_document_uuid_commit_id').on(
+      table.documentUuid,
+      table.commitId,
+    ),
+    uniquePathCommitId: unique('unique_path_commit_id').on(
```
andresgutgon, line 36: Call ending in Idx

andresgutgon (dismissed):
We're going in the right path here

PR #47, opened 2024-07-19
Change sidebar tree creation to use path instead of parentId. Also move sidebar to web-ui

andresgutgon:
# What?
We decided to start using [document.path](https://github.com/latitude-dev/latitude-llm/pull/45) as source of truth. We need to change sidebar tree creation

### DIFF
--- a/packages/web-ui/src/sections/DocumentsSidebar/useTree/index.ts
+++ b/packages/web-ui/src/sections/DocumentsSidebar/useTree/index.ts
@@ -1,0 +1,117 @@
+import { useMemo } from 'react'
+
+export type SidebarDocument = {
+  path: string
+  doumentUuid: string
+}
+
+export class Node {
+  public id: string
+  public name: string
+  public isRoot: boolean = false
+  public children: Node[] = []
+  public doc?: SidebarDocument
+
+  constructor({
+    id,
+    doc,
+    children = [],
+    isRoot = false,
+    name = '',
+  }: {
+    id: string
+    doc?: SidebarDocument
+    children?: Node[]
+    isRoot?: boolean
+    name?: string
+  }) {
+    this.id = id
+    this.name = isRoot ? 'root' : name
+    this.isRoot = isRoot
+    this.children = children
+    this.doc = doc
+  }
+
+  toJSON(): object {
+    return {
+      id: this.id,
+      name: this.name,
+      isRoot: this.isRoot,
+      children: this.children.map((child) => child.toJSON()),
+      doc: this.doc,
+    }
+  }
+}
+
+function sortByPathDepth(a: SidebarDocument, b: SidebarDocument) {
+  const depth1 = (a.path.match(/\//g) || []).length
+  const depth2 = (b.path.match(/\//g) || []).length
+  return depth1 - depth2
+}
+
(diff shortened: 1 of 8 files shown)


packages/web-ui/src/sections/DocumentsSidebar/useTree/index.ts
```diff
@@ -0,0 +1,118 @@
...
+    name?: string
+  }) {
+    this.id = id
+    this.name = isRoot ? 'root' : name
+    this.isRoot = isRoot
+    this.children = children
+    this.doc = doc
+  }
+
+  // Useful for testing
```
geclos, line 35: no need for the comment

packages/web-ui/src/sections/DocumentsSidebar/useTree/index.ts
```diff
@@ -0,0 +1,118 @@
...
+  public doc?: SidebarDocument
+
+  constructor({
+    id,
+    doc,
+    children = [],
+    isRoot = false,
+    name = '',
+  }: {
+    id: string
```
geclos, line 22: what is this id? do we need it?

packages/web-ui/src/sections/DocumentsSidebar/useTree/index.ts
```diff
@@ -0,0 +1,118 @@
...
+  generateNodeId = defaultGenerateNodeUuid,
+}: {
+  documents: SidebarDocument[]
+  generateNodeId?: typeof defaultGenerateNodeUuid
+}) {
+  return useMemo(() => {
+    const root = new Node({ id: generateNodeId(), children: [], isRoot: true })
+    const nodeMap = new Map<string, Node>()
+    nodeMap.set('', root)
+    const sorted = documents.slice().sort(sortByPathDepth)
```
geclos, line 113: minor gripe but I'm sure you can do this without this pre sorting step

packages/web-ui/src/sections/DocumentsSidebar/useTree/index.ts
andresgutgon: for `key` react elements. Is nice to do `key={node.id}`

packages/web-ui/src/sections/DocumentsSidebar/useTree/index.ts
```diff
@@ -0,0 +1,117 @@
...
+        })
+        nodeMap.set(path, node)
+
+        const parentPath = path.split('/').slice(0, -1).join('/')
+
+        // We force TypeScript to check that the parentPath is not empty
+        // We pre-sorted documents by path depth, so we know
+        // that the parent node exists
+        const parent = nodeMap.get(parentPath)!
+        parent.children.push(node)
```
andresgutgon, line 93: Before pushing the node find their place alphabetically. if it's a folder goes first.

csansoon (approved):
👍

packages/web-ui/src/sections/DocumentsSidebar/useTree/index.ts
andresgutgon: I'll do this in another PR Carlos wants this code

PR #48, opened 2024-07-19
Create and modify documents

csansoon:
Created new `createNewDocument` and `modifyExistingDocument` methods.

Modifying a document now recomputes all documents in the commit to find changes in hash, as they might be referenced by others.


packages/core/src/data-access/documentVersions.ts
```diff
@@ -3,21 +3,39 @@ import {
...
+  if (commitResult.error) return commitResult
+
+  return Result.ok(commitResult.value.mergedAt)
+}
+
 export async function getDocumentsAtCommit(
-  { commitUuid, projectId }: { commitUuid: string; projectId: number },
+  {
+    commitId,
+    commitMergedAt,
```
geclos, line 27: this api is unintuitive as it accepts both a commitId for a commit and another merged_at property which also belongs to commit. 

You could split it into getDocumentsAtDate and getDocumentsAtCommit methods, for example.

packages/core/src/data-access/documentVersions.ts
```diff
@@ -64,44 +82,39 @@ export async function getDocumentsAtCommit(
...
-      documentsAtDraft.find((d2) => d2.documentUuid !== d.documentUuid),
+    .filter(
+      (d) =>
+        documentsAtDraft.find((d2) => d2.documentUuid === d.documentUuid) ===
+        undefined,
     )
     .concat(documentsAtDraft)
 
   return Result.ok(totalDocuments)
 }
```
geclos, line 108: hmm hadn't seen this method before. It's a clever query! You should be able to improve code by splitting in smaller chunks. You really only need three operations for getting documents at a particular commit afaict:

- fetch main branch documents
- fetch commit documents
- merge main branch and draft documents

Can we split this method in these 3 concepts?

packages/core/src/services/documents/create.ts
```diff
@@ -0,0 +1,50 @@
...
+    commitId,
+    commitMergedAt: commitResult.value.mergedAt,
+  })
+  if (currentDocuments.error) return currentDocuments
+
+  if (
+    existsAnotherDocumentWithSamePath({
+      documents: currentDocuments.value,
+      path,
+    })
```
geclos, line 32: hmm we are still doing validations in services.. drizzle is really lacking compared to active record

packages/core/src/services/documents/create.ts
```diff
@@ -0,0 +1,50 @@
...
+  ) {
+    return Result.error(
+      new LatitudeError('A document with the same path already exists'),
```
geclos, line 35: use a badrequest error instead

packages/core/src/services/documents/modify.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    }).then((r) => r.unwrap())
+
+    expect(changedDocuments.length).toBe(2)
+    expect(
+      changedDocuments.find((d) => d.path === 'referenced/doc'),
+    ).toBeDefined()
+    expect(changedDocuments.find((d) => d.path === 'unmodified')).toBeDefined()
+  })
+
+  it('renaming a document creates a change to all other documents that reference it', async (ctx) => {
```
geclos, line 97: also to all its children given you have to change the paths

packages/core/src/services/documents/modify.ts
```diff
@@ -0,0 +1,136 @@
...
+  path?: string
+  content?: string | null
+  deletedAt?: Date | null
+}) {
+  const commitResult = await getEditableCommit(commitId)
+  if (commitResult.error) return commitResult
+
+  const currentDocuments = await getDocumentsAtCommit({
+    commitId,
+    commitMergedAt: commitResult.value.mergedAt,
```
geclos, line 60: hmm yeah it's not clear what this will do

packages/core/src/services/documents/modify.ts
```diff
@@ -0,0 +1,136 @@
...
+  })
+  if (currentDocuments.error) return currentDocuments
```
geclos, line 62: hmmm we this result error check in a million places already I wonder if we can improve. How about a try catch for this whole method and you just wrap whatever error you get in a Result.error?

packages/core/src/services/documents/modify.ts
```diff
@@ -0,0 +1,136 @@
...
+            set: newDocumentVersion,
+          })
+          .returning()
+          .then((r) => r[0]!)
+      }),
+    )
+
+    return Result.ok(results.find((r) => r.documentUuid === documentUuid)!)
+  })
+}
```
geclos, line 136: hard to follow what is happening in this method can we split it in smaller chunks? Can we do it with less filters, destructings and finds and less if else statements?

packages/core/src/services/documents/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+  findCommitById,
+  Result,
+  TypedResult,
+} from '@latitude-data/core'
+import { ForbiddenError, LatitudeError } from '$core/lib/errors'
+
+export async function getEditableCommit(
+  commitId: number,
+): Promise<TypedResult<Commit, LatitudeError>> {
+  const commit = await findCommitById({ id: commitId })
```
geclos, line 13: all other finds return a result this one should too

packages/core/src/data-access/documentVersions.ts
csansoon: I only added `mergedAt` for performance, as it is needed for retrieving the documents, and several parts of the code that use this function already have this information so I didn't want to fetch this if it's not necessary. But I agree that it's unintuitive.

packages/core/src/services/documents/modify.test.ts
csansoon: This is just testing a change of the path of a single document. This does not affect any other document.

The action of renaming a folder will be done as a different service, where it will rename all paths with the same folder.

packages/core/src/services/documents/create.ts
andresgutgon: I always thought model validation level was not the right approach. In the end you usually have one or two services that change the entity. I think is better to have in those services the validation.

apps/web/src/actions/documents/create.ts
```diff
@@ -15,6 +15,13 @@ export const createDocumentVersionAction = withProject
...
   .handler(async ({ input }) => {
-    const result = await createDocumentVersion(input)
+    const commit = await findCommitByUuid({
+      projectId: input.projectId,
+      uuid: input.commitUuid,
+    }).then((r) => r.unwrap())
+    const result = await createNewDocument({
+      commitId: commit.id,
+      path: input.path,
+    })
```
andresgutgon, line 25: The finding and saving should be all inside `createNewDocument`. The less code we put inside actions the better

apps/web/src/components/Sidebar/index.tsx
```diff
@@ -9,9 +9,9 @@ export default async function Sidebar({
   commitUuid: string
   projectId: number
 }) {
+  const commit = await findCommitByUuid({ projectId, uuid: commitUuid })
```
andresgutgon, line 12: I feel this `findCommit` it's going to be used on several components on the server at the same time. https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/app/(private)/_data-access/index.ts#L11C32-L11C37

packages/core/src/data-access/documentVersions/getDocumentByPath.ts
```diff
@@ -16,7 +17,9 @@ export async function getDocumentByPath(
   db = database,
 ) {
   try {
-    const result = await getDocumentsAtCommit({ projectId, commitUuid }, db)
+    const commit = await findCommitByUuid({ projectId, uuid: commitUuid }, db)
+    if (commit.error) return commit
+    const result = await getDocumentsAtCommit({ commitId: commit.value.id }, db)
```
andresgutgon, line 22: 👾

packages/core/src/data-access/documentVersions/getDocumentsAtCommit.test.ts
```diff
@@ -1,8 +1,10 @@
 import { HEAD_COMMIT } from '$core/constants'
+import { modifyExistingDocument } from '$core/services'
```
andresgutgon, line 2: Why not a more standard name. `updateDocument`

packages/core/src/services/documents/modify.test.ts
```diff
@@ -0,0 +1,131 @@
+import { listCommitChanges } from '$core/data-access'
+import useTestDatabase from '$core/tests/useTestDatabase'
+import { describe, expect, it } from 'vitest'
+
+import { mergeCommit } from '../commits/merge'
+import { modifyExistingDocument } from './modify'
+
+useTestDatabase()
```
andresgutgon, line 8: Why this?

packages/core/src/services/documents/modify.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+describe('modifyExistingDocument', () => {
```
andresgutgon, line 10: DItto use more standard `update`

packages/core/src/data-access/documentVersions/getDocumentByPath.ts
csansoon: ?

andresgutgon (approved):

packages/core/src/services/documents/create.ts
geclos: Validations should be closest to the source of truth as possible so to make it difficult to bypass them, this means validations in db when possible, validations closest to db when db validations aren't possible. Services are two steps removed from the db, model just one. Compose your model with validations and you'll be safer.

packages/core/src/data-access/documentVersions/getDocumentsAtCommit.ts
```diff
@@ -61,33 +60,60 @@ export async function getDocumentsAtCommit(
       ),
     )
 
-  const documentsAtPreviousMergedCommits =
-    documentsAtPreviousMergedCommitsResult.map((d) => d.document_versions)
+  return documentsFromMergedCommits
+}
```
geclos, line 64: fantastic improvement 👌🏼 !

PR #51, opened 2024-07-22
feature: tenancy

geclos:
Refactor data accessors to repositories which enforce tenancy. There's two guiding principles:
- All data access should happen via repositories so to ensure tenancy. In the cases that tenancy must not be enforced data accessors have been prepended with `unsafely` to make things clear
- All data writes – mainly everything under `services` folder – should receive **object instances** instead of object ids. This is to ensure consumers have to enforce tenancy before writing to database.


packages/core/src/repositories/projectsRepository.ts
```diff
@@ -0,0 +1,39 @@
...
+      .select()
+      .from(projects)
+      .where(eq(projects.workspaceId, this.workspaceId))
+      .as('projectsScope')
+  }
+
+  async getProjectById(id: number) {
+    const result = await this.db
+      .select()
+      .from(this.scope)
```
andresgutgon, line 21: QUE NICE

packages/core/src/repositories/documentVersionsRepository.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  async getDocumentById(documentId: number) {
+    const res = await this.db
+      .select()
+      .from(this.scope)
+      .where(eq(documentVersions.id, documentId))
+
+    // NOTE: I hate this
+    const document = res[0]?.document_versions
```
andresgutgon, line 26: This can be solved. Drizzle has a `getColumns` method

``` javascript
const columns = getColumns(db.documentVersions)
const res = await this.db.select(columns)...
```
And now you have a nice array of document versions

```

File: apps/web/src/actions/documents/getContentByPath.ts [hunk 3]
```diff
@@ -14,9 +15,13 @@ export const getDocumentContentByPathAction = withProject
...
   )
-  .handler(async ({ input }) => {
+  .handler(async ({ input, ctx }) => {
+    const commitsScope = new CommitsRepository(ctx.project.workspaceId)
+    const commit = await commitsScope
+      .getCommitById(input.commitId)
+      .then((r) => r.unwrap())
     const document = await getDocumentByPath({
-      commitId: input.commitId,
+      commit,
```
geclos, line 24: as a general rule make all db readers/writers receive objects instead of identifiers. This will force the consumer to ensure tenancy and will reduce the amount of redundants roundtrips to the database. It results in fatter controllers but that is ok, tenancy is ensured at the entrypoints of our backend.

apps/web/src/actions/procedures/index.ts
```diff
@@ -16,11 +16,11 @@ export const authProcedure = createServerActionProcedure().handler(async () => {
...
   .input(z.object({ projectId: z.number() }))
   .handler(async ({ input, ctx }) => {
+    const { workspace } = ctx
+    const projectScope = new ProjectsRepository(workspace.id)
     const project = (
-      await findProject({
-        projectId: input.projectId,
-        workspaceId: ctx.workspace.id,
-      })
+      await projectScope.getProjectById(input.projectId)
```
andresgutgon, line 22: Would be nice that this is a class call it `find` when is by ID

apps/web/src/actions/documents/getContentByPath.ts
```diff
@@ -14,9 +15,13 @@ export const getDocumentContentByPathAction = withProject
     }),
     { type: 'json' },
   )
-  .handler(async ({ input }) => {
+  .handler(async ({ input, ctx }) => {
+    const commitsScope = new CommitsRepository(ctx.project.workspaceId)
+    const commit = await commitsScope
+      .getCommitById(input.commitId)
+      .then((r) => r.unwrap())
```
csansoon, line 22: Why do we need to use the `workspaceId` to get the `commitScope`? Shouldn't we use the `commitId` instead?

apps/web/src/data-access/users.ts
```diff
@@ -60,7 +57,7 @@ export async function getCurrentUserFromDB({
 }: {
   userId: string | undefined
 }): PromisedResult<SessionData, NotFoundError> {
-  const user = await getUser(userId)
+  const user = await unsafelyGetUser(userId)
```
andresgutgon, line 60: Why this?

package.json
```diff
@@ -34,12 +34,9 @@
...
       "allowedVersions": {
-        "react": "18.x",
-        "react-dom": "18.x",
-        "react": "19.x",
-        "react-dom": "19.x",
-        "eslint": "8.x",
-        "eslint": "9.x"
+        "react": ">=18.x",
+        "react-dom": ">=18.x",
+        "eslint": ">=8.x"
```
andresgutgon, line 39: Thanks

packages/core/src/repositories/documentVersionsRepository.ts
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  async getDocumentById(documentId: number) {
+    const res = await this.db
+      .select()
+      .from(this.scope)
+      .where(eq(this.scope.id, documentId))
+
+    // NOTE: I hate this
+    const document = res[0]
```
andresgutgon, line 36:
```javascript
this.db.documents.findFirst({
  where: eq(this.scope.id, documentId)
})
```

apps/web/src/actions/procedures/index.ts
geclos: 🤷🏼 can change if we prefer it

andresgutgon (approved):

apps/web/src/actions/documents/getContentByPath.ts
geclos: nein, commitId alone is not scoped by tenant, you have to first scope by tenant (the workspace) before you can do any  search of commits

apps/web/src/data-access/users.ts
geclos: From the repo description:

> All data access should happen via repositories so to ensure tenancy. In the cases that tenancy must not be enforced data accessors have been prepended with unsafely to make things clear

packages/core/src/repositories/documentVersionsRepository.ts
geclos: nein, that's one of the negative effects of using subqueries you cannot use drizzle query dsl

packages/core/src/repositories/documentVersionsRepository.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    const lastVersionOfEachDocument = this.db
+      .$with('lastVersionOfDocuments')
+      .as(
+        this.db
+          .select({
+            documentUuid: documentVersions.documentUuid,
+            mergedAt: max(commits.mergedAt).as('maxMergedAt'),
+          })
+          .from(documentVersions)
```
andresgutgon, line 169: This is not using the scope no?

packages/core/src/repositories/documentVersionsRepository.ts
```diff
@@ -0,0 +1,211 @@
...
+          .innerJoin(commits, eq(commits.id, documentVersions.commitId))
+          .where(and(filterByMaxMergedAt(), eq(commits.projectId, projectId)))
+          .groupBy(documentVersions.documentUuid),
+      )
+
+    const documentsFromMergedCommits = await this.db
+      .with(lastVersionOfEachDocument)
+      .select(getTableColumns(documentVersions))
+      .from(documentVersions)
```
andresgutgon, line 178: This is also not using the scope I think

andresgutgon: Plese don't fix before I merge my PR. I'm sick of resolving conflicts 🙏

PR #53, opened 2024-07-22
Recompute changes on commit to find all indirect changes and syntax errors

csansoon:
Modifying, or even adding a new document, can trigger a change in other documents for the same commit. To know the actual new state of the commit, we need to recompute the metadata of all of them. Since this is an expensive action, it can now be done on demand with the `recomputeChanges` method.

This method is ran automatically in the `mergeCommit` method, for two reasons:
1. Merging the commit must contain all changes, as we never should add changes to an already merged commit.
2. Merging should fail when any of the documents contains syntax errors.


packages/core/src/services/documents/create.test.ts
```diff
@@ -58,4 +58,26 @@ describe('createNewDocument', () => {
       'Cannot create a document version in a merged commit',
     )
   })
+
+  it('modifies other documents if it is referenced by another document', async (ctx) => {
+    const { project } = await ctx.factories.createProject({
+      documents: {
+        main: '<ref prompt="referenced/doc" />',
+      },
+    })
```
csansoon, line 67: Added a new attribute to the project factory to start with an initial document structure :)

packages/core/src/services/documents/create.test.ts
```diff
@@ -58,4 +58,26 @@ describe('createNewDocument', () => {
...
+
+    const { commit } = await ctx.factories.createDraft({ project })
+    await createNewDocument({
+      commitId: commit.id,
+      path: 'referenced/doc',
+    })
+
+    const changes = await listCommitChanges({ commitId: commit.id }).then((r) =>
+      r.unwrap(),
+    )
```
geclos, line 77: result pattern on async await has trash ergonomics tbh

geclos (commented):
👌🏼

geclos:
> Recalculate hashes for all documents in new state
Compare with HEAD to obtain only the new changes
Remove all document versions from draft

This is a very expensive op to do on each write specially considering we are single threaded. Consider doing the operation asynchronously (update the doc atomically and compute any other changes async), using node's worker threads or changing the UX (maybe compute changes to related docs prior to commit merge and inform the user "these other docs will also be changed"?)

packages/core/src/services/documents/create.ts
```diff
@@ -1,16 +1,12 @@
...
-  existsAnotherDocumentWithSamePath,
-} from './utils'
+  getDraft,
+  getMergedAndDraftDocuments,
+  replaceCommitChanges,
+  resolveDocumentChanges,
+} from './shared'
 
 export async function createNewDocument({
   commitId,
```
geclos, line 12: in order to ensure tenancy pass a commit instance instead of instance id, this will force the consumer to ensure tenancy before calling any writer (and facilitate the refactor [here](https://github.com/latitude-dev/latitude-llm/pull/51))

packages/core/src/services/documents/update.ts
```diff
@@ -1,78 +1,13 @@
...
-  return Result.ok(documentsWithUpdatedHash)
-}
+import {
+  getDraft,
+  getMergedAndDraftDocuments,
+  replaceCommitChanges,
+  resolveDocumentChanges,
+} from './shared'
 
 export async function updateDocument({
```
geclos, line 12: same here instead of documentuuid pass the document instance

packages/core/src/services/commits/merge.ts
```diff
@@ -35,6 +36,16 @@ export async function mergeCommit(
...
 
+    const recomputedResults = await recomputeChanges({ commitId }, tx)
+    if (recomputedResults.error) return recomputedResults
+    if (Object.keys(recomputedResults.value.errors).length > 0) {
+      return Result.error(
+        new LatitudeError(
+          'There are errors in the updated documents in this commit',
+        ),
+      )
+    }
```
csansoon, line 47: Mergeing will always recompute changes for two reasons:
- To always have ALL changes.
- Validate that none of the changes contain errors.

packages/core/src/services/commits/merge.ts
```diff
@@ -35,6 +36,16 @@ export async function mergeCommit(
       )
     }
 
+    const recomputedResults = await recomputeChanges({ commitId }, tx)
+    if (recomputedResults.error) return recomputedResults
+    if (Object.keys(recomputedResults.value.errors).length > 0) {
+      return Result.error(
+        new LatitudeError(
+          'There are errors in the updated documents in this commit',
```
andresgutgon, line 44: What about returning the errors?

packages/core/src/services/documents/utils.ts
```diff
@@ -29,3 +71,84 @@ export function existsAnotherDocumentWithSamePath({
...
+  }
+
+  const newDocumentsWithUpdatedHash = await Promise.all(
+    newDocuments.map(async (d) => {
+      const metadata = await readMetadata({
+        prompt: d.content ?? '',
+        referenceFn: getDocumentContent,
+      })
+      if (metadata.errors.length > 0) {
+        errors[d.documentUuid] = metadata.errors
```
andresgutgon, line 102: What shape have these errors?

packages/core/src/data-access/documentVersions/getDocumentsAtCommit.ts
```diff
@@ -107,6 +110,36 @@ export async function getDocumentsAtCommit(
...
+  if (documentInCommit !== undefined) return Result.ok(documentInCommit)
+
+  const documentsAtCommit = await getDocumentsAtCommit({ commitId }, tx)
+  if (documentsAtCommit.error) return Result.error(documentsAtCommit.error)
+
+  const document = documentsAtCommit.value.find(
+    (d) => d.documentUuid === documentUuid,
+  )
+
+  if (!document) return Result.error(new LatitudeError('Document not found'))
```
andresgutgon, line 138: This should be a `NotFoundError` no?

packages/core/src/services/documents/utils.ts
csansoon: They're instances of `CompileError`, which is an `Error` object with some extra stuff.

packages/core/src/services/commits/merge.ts
csansoon: Why though? It's not being used for anything

andresgutgon: What about explaining the errors to the user in the UI?

csansoon: That should be shown in the UI **before** trying to actually perform the merge, shouldn't it? 
This should just be the last validation before doing the actual merge action.

andresgutgon (approved):

PR #55, opened 2024-07-23
Fix open folders in sidebar

andresgutgon:
# What?
The old approach did not work because navigating to a new document would re-render the tree. We now try to hold that state in Zustand so it can persist across navigations

### DIFF
--- a/packages/web-ui/src/sections/Document/Sidebar/Files/useOpenPaths/index.test.ts
+++ b/packages/web-ui/src/sections/Document/Sidebar/Files/useOpenPaths/index.test.ts
@@ -1,0 +1,32 @@
+import { act, renderHook } from '@testing-library/react'
+import { describe, expect, it } from 'vitest'
+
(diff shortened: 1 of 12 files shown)


packages/web-ui/src/sections/Document/Sidebar/Files/useOpenPaths/index.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    const { result } = renderHook(() => useOpenPaths((state) => state))
+    act(() => {
+      result.current.togglePath('some-folder/nested-folder/doc1')
+    })
+
+    act(() => {
+      result.current.togglePath('some-folder/nested-folder')
+    })
+    expect(result.current.openPaths).toEqual(['', 'some-folder'])
+  })
```
csansoon, line 31: 👌

csansoon (approved):

PR #58, opened 2024-07-24
Resolve prompt references

csansoon:
Replaced the `hash` property in compiler's metadata with a `resolvedPrompt`, which represents the complete prompt, with all of the references replaced with the actual content.

This leads to a better optimization of prompt compiling, since there is no need to obtain referenced prompts when compiling the parent.

### DIFF
--- a/packages/compiler/src/compiler/readMetadata.ts
+++ b/packages/compiler/src/compiler/readMetadata.ts
@@ -1,9 +1,8 @@
-import { createHash } from 'crypto'
-
 import {
   CUSTOM_MESSAGE_ROLE_ATTR,
   CUSTOM_MESSAGE_TAG,
   REFERENCE_PROMPT_ATTR,
+  REFERENCE_PROMPT_TAG,
 } from '$compiler/constants'
 import CompileError, { error } from '$compiler/error/error'
 import errors from '$compiler/error/errors'
@@ -18,6 +17,7 @@ import type {
 } from '$compiler/parser/interfaces'
 import { Config, ConversationMetadata, MessageRole } from '$compiler/types'
 import { Node as LogicalExpression } from 'estree'
+import yaml from 'yaml'
 
 import { ReferencePromptFn } from './compile'
 import { readConfig } from './config'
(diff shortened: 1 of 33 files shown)


packages/core/src/schema/models/documentVersions.ts
```diff
@@ -21,7 +21,7 @@ export const documentVersions = latitudeSchema.table(
     documentUuid: uuid('document_uuid').notNull().defaultRandom(),
     path: varchar('path').notNull(),
     content: text('content').notNull().default(''),
-    hash: varchar('hash'),
+    resolvedContent: text('resolved_content'),
```
csansoon, line 24: `resolvedContent` is nullable. If it does have content, it is safe to use. Otherwise it's required to compute it.

In order to guarantee that this attribute is safe, it is converted to NULL each time a document from the same commit is updated / created.

packages/core/src/services/documents/update.ts
```diff
@@ -69,6 +70,12 @@ export async function updateDocument({
       return Result.error(new NotFoundError('Document does not exist'))
     }
 
+    // Invalidate all resolvedContent for this commit
+    await tx
+      .update(documentVersions)
+      .set({ resolvedContent: null })
+      .where(eq(documentVersions.commitId, commitId))
+
```
csansoon, line 78: As mentioned before, all `resolvedContent` values from the same commit are invalidated when updating a document.

packages/compiler/src/compiler/readMetadata.ts
```diff
@@ -419,13 +425,20 @@ export class ReadMetadata {
...
+          resolvedRefPrompt = `/* <${REFERENCE_PROMPT_TAG} ${REFERENCE_PROMPT_ATTR}="${refPromptPath}" /> */`
         }
+
+        /* Replace the reference tag with the actual referenced prompt */
+        const start = node.start! + this.resolvedPromptOffset
+        const end = node.end! + this.resolvedPromptOffset
+        const pretext = this.resolvedPrompt.slice(0, start)
+        const posttext = this.resolvedPrompt.slice(end)
+        this.resolvedPrompt = pretext + resolvedRefPrompt + posttext
+        this.resolvedPromptOffset += resolvedRefPrompt.length - (end - start)
```
andresgutgon, line 440: This could be a nice method with a nice name

csansoon: agree. Will add it to a next pr

andresgutgon (approved):

PR #59, opened 2024-07-25
Compile reorganization

csansoon:
General reorganization of the Compile logic.

Before this change, the main `compile` function was defined by a large if-else statement, where the logic for each node type (If, Each, Comment, Text, ElementTag, MustacheTag...) was crammed inside. 

Now, each node type has its logic in its own file, making organization much better.

Some of these functions need to modify the global state of the compiler. Previously, they did this by accessing properties via the ⁠this keyword. Now, since they are not defined within the Compile class, they no longer have access to it. To fix this, we wrap every required method into a “context” that is passed to these functions, allowing them to use the context as needed.

### DIFF
--- a/packages/compiler/src/compiler/base/nodes/config.ts
+++ b/packages/compiler/src/compiler/base/nodes/config.ts
@@ -1,0 +1,7 @@
+import { Config } from '$compiler/parser/interfaces'
+
+import { CompileNodeContext } from '../types'
+
+export async function compile(_: CompileNodeContext<Config>) {
+  /* do nothing */
+}
--- a/packages/compiler/src/compiler/base/nodes/tag.ts
+++ b/packages/compiler/src/compiler/base/nodes/tag.ts
@@ -1,0 +1,104 @@
+import {
+  isContentTag,
+  isMessageTag,
+  isRefTag,
+  isToolCallTag,
+} from '$compiler/compiler/utils'
+import errors from '$compiler/error/errors'
+import {
+  ContentTag,
+  ElementTag,
+  MessageTag,
+  ReferenceTag,
+  ToolCallTag,
+} from '$compiler/parser/interfaces'
+
+import { CompileNodeContext } from '../types'
+import { compile as resolveContent } from './tags/content'
+import { compile as resolveMessage } from './tags/message'
+import { compile as resolveRef } from './tags/ref'
+import { compile as resolveToolCall } from './tags/toolCall'
+
+async function resolveTagAttributes({
+  node: tagNode,
+  scope,
+  resolveExpression,
+}: CompileNodeContext<ElementTag>): Promise<Record<string, unknown>> {
+  const attributeNodes = tagNode.attributes
+  if (attributeNodes.length === 0) return {}
+
+  const attributes: Record<string, unknown> = {}
+  for (const attributeNode of attributeNodes) {
+    const { name, value } = attributeNode
+    if (value === true) {
+      attributes[name] = true
+      continue
+    }
+
+    let totalValue: string = ''
+    for await (const node of value) {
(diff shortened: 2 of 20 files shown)


packages/compiler/src/compiler/compile.ts
```diff
@@ -170,383 +96,98 @@ export class Compile {
...
+      addMessage: this.addMessage.bind(this),
+      addStrayText: this.addStrayText.bind(this),
+      popStrayText: this.popStrayText.bind(this),
+      groupStrayText: this.groupStrayText.bind(this),
+      addContent: this.addContent.bind(this),
+      popContent: this.popContent.bind(this),
+      groupContent: this.groupContent.bind(this),
+      addToolCall: this.addToolCall.bind(this),
+      popToolCalls: this.popToolCalls.bind(this),
+    }
```
csansoon, line 170: The context with all required functions to read/modify the global state from the new external methods, as explained in the PR description.

packages/compiler/src/compiler/compile.test.ts
```diff
@@ -304,7 +304,7 @@ describe('message contents', async () => {
 })
 
 describe('reference tags', async () => {
-  it('can reference a prompt', async () => {
+  it('always fails', async () => {
```
csansoon, line 307: Compiler now cannot resolve references, it will always fail with a custom error message. Instead, now it is required to resolve all references before compiling.

All tests about references in compiling have been removed.

packages/compiler/src/compiler/base/nodes/config.ts
```diff
@@ -0,0 +1,7 @@
+import { Config } from '$compiler/parser/interfaces'
+
+import { CompileNodeContext } from '../types'
+
+export async function compile(_: CompileNodeContext<Config>) {
+  /* do nothing */
+}
```
andresgutgon, line 7: Are these necessary? You could implement a generic `noopFn` function

packages/compiler/src/compiler/base/nodes/tag.ts
```diff
@@ -0,0 +1,104 @@
...
+  MessageTag,
+  ReferenceTag,
+  ToolCallTag,
+} from '$compiler/parser/interfaces'
+
+import { CompileNodeContext } from '../types'
+import { compile as resolveContent } from './tags/content'
+import { compile as resolveMessage } from './tags/message'
+import { compile as resolveRef } from './tags/ref'
+import { compile as resolveToolCall } from './tags/toolCall'
```
andresgutgon, line 20: Not to change but it would be better if from the file you export these names. No need for doing this and also when you are in the file you understand what type of compile is

packages/compiler/src/compiler/base/nodes/tag.ts
```diff
@@ -0,0 +1,104 @@
...
+  const {
+    node,
+    scope,
+    isInsideContentTag,
+    isInsideMessageTag,
+    resolveBaseNode,
+    baseNodeError,
+    groupStrayText,
+  } = props
+  groupStrayText()
```
andresgutgon, line 70: Where is defined this?

csansoon: Everything from the context is defined in the `Compile` class

packages/compiler/src/compiler/base/nodes/tag.ts
csansoon: Hmm I like that all nodes just need to define a generic `compile` method, same name and arguments.

andresgutgon: Ok if you like it no more discussion xD

andresgutgon (approved):

PR #60, opened 2024-07-25
Create and Delete documents (not folders yet) in the sidebar

andresgutgon:
# What
☝️ That

### TODO
- [x] Add file IMPLEMENT server action
- [x] When creating a file reseting tmpFolders zustand store
- [x] Implement `destroyOrSoftDeleteDocuments` (used by both destroy document or destroy folder)
- [x] Destroy document server action
- [x] DRY File node display in File tree

## Next PR
- [ ] Destroy folder server action
- [ ] Resizable sidebar 
- [ ] Make the sidebar files tree vertically scrollable. Custom scrollbar styles
- [ ] Sidebar header with File + | Folder +
- [ ] Review constraint in document version by commit and path. Now that documents can be soft deleted it will be possible to have same path in 2 different docs


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/ClientFilesTree/index.tsx
```diff
@@ -32,12 +33,17 @@ export default function ClientFilesTree({
...
+    fallbackData: serverDocuments,
+  })
 
   return (
     <FilesTree
       documents={documents}
       currentPath={documentPath}
       navigateToDocument={navigateToDocument}
+      createFile={createFile}
+      destroyFile={destroyFile}
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/ClientFilesTree/index.tsx:46]: Add `destroyFolder` here

packages/core/src/services/documents/create.test.ts
```diff
@@ -3,6 +3,9 @@ import { describe, expect, it } from 'vitest'
 
 import { mergeCommit } from '../commits/merge'
 import { createNewDocument } from './create'
+import useTestDatabase from '$core/tests/useTestDatabase'
+
+useTestDatabase()
```
andresgutgon, line 8: I'm not happy with this. I'm not sure the rollback is working on each test. We need to investigate

packages/core/src/tests/factories/projects.ts
```diff
@@ -80,8 +82,10 @@ export async function createProject(projectData: Partial<ICreateProject> = {}) {
       })
       documents.push(updatedDoc.unwrap())
     }
+    commit = draft
+
     await mergeCommit({ commitId: draft.id }).then((r) => r.unwrap())
   }
 
-  return { project, user, workspace, documents }
+  return { project, user, workspace, documents, commit: commit! }
```
andresgutgon, line 90: dry

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
```diff
@@ -384,7 +420,11 @@ export function FilesTree({
   currentPath,
   documents,
   navigateToDocument,
+  createFile,
+  destroyFile,
 }: {
+  createFile: (args: { path: string }) => Promise<void>
+  destroyFile: (documentUuid: string) => Promise<void>
```
andresgutgon, line 427: Add `destroyFolder` here

packages/web-ui/src/sections/Document/Editor/index.tsx
```diff
@@ -81,8 +81,11 @@ export default function DocumentEditor({
           <div className='flex flex-col gap-3'>
             <Text.H6M>Inputs</Text.H6M>
             {inputs.length > 0 ? (
-              inputs.map((param) => (
-                <div className='flex flex-row gap-4 w-full items-center'>
+              inputs.map((param, idx) => (
+                <div
+                  key={idx}
```
andresgutgon, line 86: Always put your keys in react loops kids

apps/web/src/app/(private)/_data-access/index.ts
```diff
@@ -54,20 +54,28 @@ export const getDocumentByUuid = cache(
     documentUuid: string
     commit: Commit
   }) => {
-    const workspace = await findWorkspaceFromCommit(commit)
```
andresgutgon, line 57: This pattern for finding the workspace on commit can be solved in a more generic way when returning from a repository `.find` method. Instead of returning just the model we return the model with their relations. At least some of then. 

What we're doing now is doing 1 more SQL query to get again the Workspace when we already got it to build the scope for that repository. I think this is doable. I might try in another PR.

apps/web/src/app/(private)/_data-access/index.ts
```diff
@@ -54,20 +55,28 @@ export const getDocumentByUuid = cache(
...
-    const scope = new DocumentVersionsRepository(workspace!.id)
+    const { workspace } = await getCurrentUser()
+    const scope = new DocumentVersionsRepository(workspace.id)
     const result = await scope.getDocumentAtCommit({ documentUuid, commit })
+    if (result.error) {
+      const error = result.error
+      if (error instanceof NotFoundError) {
+        return notFound()
```
andresgutgon, line 64: When is a not found we don't raise the error. We use NextJS `notFound()` method that display the `not-found.tsx` page

apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/documents/route.ts
```diff
@@ -23,7 +23,7 @@ export async function GET(
     const commit = await commitsScope
       .getCommitByUuid({ uuid: commitUuid, project })
       .then((r) => r.unwrap())
-    const documents = await scope.getDocumentsAtCommit(commit)
+    const documents = await scope.getDocumentsAtCommit({ commit })
```
Comment (andresgutgon) [inline apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/documents/route.ts:26]: We have discussion offline about using or not positional arguments. The conclusion we arrive is that we hate it. No, I'm joking (kind of). 

The conclusion was that we're going to have methods with positional arguments when we're sure this method will never have more arguments. This is an exception because I did this refactor before arriving to that conclusion and I'm sick of having this PR open. Feel free to change this one if you want.

apps/web/src/stores/documentVersions.ts
```diff
@@ -0,0 +1,118 @@
...
+export default function useDocumentVersions(
+  { currentDocument }: { currentDocument: SidebarDocument | undefined },
+  opts?: SWRConfiguration,
+) {
+  const { toast } = useToast()
+  const router = useRouter()
+  const { project } = useCurrentProject()
+  const { commit } = useCurrentCommit()
+  const { mutate, data, ...rest } = useSWR<DocumentVersion[]>(
+    ['documentVersions', project.id, commit.id],
```
andresgutgon, line 29: Let's agree on using Arrays for keys in the stores. URL like strings are harder to manipulate

packages/core/src/repositories/documentVersionsRepository/getDocumentsAtCommit.test.ts
andresgutgon, line 1: Huge refactoring of these tests. Please don't touch this file before I merge this PR 🙏 

Resolving conflicts was taught

packages/core/src/repositories/documentVersionsRepository/index.ts
```diff
@@ -51,8 +63,8 @@ export class DocumentVersionsRepository extends Repository {
       .from(this.scope)
       .where(
         and(
-          eq(documentVersions.commitId, commit.id),
-          eq(documentVersions.documentUuid, documentUuid),
+          eq(this.scope.commitId, commit.id),
+          eq(this.scope.documentUuid, documentUuid),
```
andresgutgon, line 67: Now that we use scopes we always have to use the scope for picking the columns of the table. Otherwise it doesn't work

image

packages/core/src/repositories/documentVersionsRepository/index.ts
```diff
@@ -145,6 +144,32 @@ export class DocumentVersionsRepository extends Repository {
     return Result.ok(changedDocuments)
   }
 
+  private async getAllDocumentsAtCommit({ commit }: { commit: Commit }) {
```
andresgutgon, line 147: Returning all documents with deleted documents included is private. The moment we want to show then to for example "Recover from trash" we can implement `getAllDeletedDocumentsAtCommit`

packages/core/src/repositories/documentVersionsRepository/index.ts
```diff
@@ -145,6 +144,32 @@ export class DocumentVersionsRepository extends Repository {
...
+      })
+
+    if (commit.mergedAt !== null) {
+      // Referenced commit is merged. No additional documents to return.
+      return Result.ok(documentsFromMergedCommits)
+    }
+
+    const documentsFromDraft = await this.db
+      .select(getTableColumns(documentVersions))
+      .from(documentVersions)
```
andresgutgon, line 161: This is wrong. We should use `this.scope`. We'll fix in another PR

packages/core/src/services/commits/create.ts
```diff
@@ -8,19 +8,23 @@ import {
...
 export async function createCommit({
-  commit,
+  commit: { projectId, title, mergedAt },
   db = database,
 }: {
-  commit: Omit<Partial<Commit>, 'id'>
+  commit: {
+    projectId: number
+    title?: string
+    mergedAt?: Date
```
andresgutgon, line 17: I prefer to be explicit about what means creating an object. This way people using don't include not permitted columns.

packages/core/src/services/documents/destroyDocument.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export async function destroyDocument({
+  document,
+  commit,
+  workspaceId,
+  db = database,
+}: {
+  document: DocumentVersion
+  commit: Commit
+  workspaceId: number
```
andresgutgon, line 16: This is not ideal as I [commented here](https://github.com/latitude-dev/latitude-llm/pull/60/files#r1693971751) a commit passed here was found with a repository so the workspace was already there. 

There has to be a nice way of having `commit.workspace.id` so object includes relations already resolved

packages/core/src/services/documents/destroyOrSoftDeleteDocuments.ts
```diff
@@ -0,0 +1,156 @@
...
+  commitId: number,
+  tx = database,
+) {
+  return tx
+    .update(documentVersions)
+    .set({ resolvedContent: null })
+    .where(eq(documentVersions.commitId, commitId))
+}
+
+/**
```
andresgutgon, line 121: Thanks to Carlos for helping me with this. A lot is going on here. I tried my best to make as understandable as possible and added tests

packages/core/src/services/documents/update.ts
```diff
@@ -1,51 +1,56 @@
...
-  document,
-  path,
-  content,
-  deletedAt,
-}: {
-  commit: Commit
-  document: DocumentVersion
-  path?: string
-  content?: string | null
-  deletedAt?: Date | null
```
andresgutgon, line 22: Removed `deletedAt` from here now that we specific services for deleting documents.

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
```diff
@@ -1,98 +1,21 @@
...
-      )}
-    >
-      <IndentationBar
-        indentation={indentation}
-        hasChildren={open && node.children.length > 0}
-      />
-      {children}
-    </div>
-  )
-})
```
andresgutgon, line 94: Now ☝️ a document and a folder share all the presentation layers and they only pass the necessary props. This way design is unified. This was a leftover refactor I had pending from initial file tree implementation

packages/core/src/services/commits/create.ts
csansoon: 👌

packages/core/src/repositories/documentVersionsRepository/index.ts
geclos: 💯 good catch

packages/core/src/services/documents/destroyOrSoftDeleteDocuments.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      commit: draft,
+      document,
+      content: 'Doc 1 (version 2)',
+    }).then((r) => r.unwrap())
+
+    // Fake cached content exists to prove the method invalidate cache
+    await database
+      .update(documentVersions)
+      .set({
+        resolvedContent: '[CHACHED] Doc 1 (version 1)',
```
csansoon, line 69: chached

apps/web/src/stores/documentVersions.ts
```diff
@@ -0,0 +1,118 @@
...
+              .detail({ id: project.id })
+              .commits.detail({ uuid: commit.uuid }).documents.root,
+          )
+        }
+      }
+    },
+    [executeDestroyDocument, mutate, data],
+  )
+
+  return { ...rest, documents: data ?? [], createFile, destroyFile, mutate }
```
geclos, line 117: minor nitpicks: I'd prefer to keep it consistent between stores and always return data instead of a different-named prop that contains the data for each store. Also no need for the suffix "file" in actions. create and destroy is enough and beautifully concise.

apps/web/src/stores/documentVersions.ts
```diff
@@ -0,0 +1,118 @@
...
+}
```
geclos, line 118: fantastic

packages/core/src/repositories/documentVersionsRepository/getDocumentAtCommit.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      documentUuid,
+    })
+    const document = result.unwrap()
+
+    expect(omit(document, 'id', 'updatedAt')).toEqual({
+      ...omit(doc, 'id', 'updatedAt'),
+      resolvedContent: 'VERSION_1',
+    })
+  })
+})
```
geclos, line 32: nice

packages/core/src/services/documents/destroyOrSoftDeleteDocuments.ts
```diff
@@ -0,0 +1,156 @@
...
+      commitId,
+    })
+    const toBeSoftDeleted = getToBeSoftDeleted({ documents, existingUuids })
+
+    await Promise.all([
+      hardDestroyDocuments({ documents, existingUuids, tx }),
+      createDocumentsAsSoftDeleted({ toBeSoftDeleted, commitId, tx }),
+      updateDocumetsAsSoftDeleted({ toBeSoftDeleted, commitId, tx }),
+      invalidateDocumentsCacheInCommit(commitId, tx),
+    ])
```
csansoon, line 152: How is it you're calling both `createDocumentsAsSoftDeleted` and `updateDocumentsAsSoftDeleted` with the same array?

I see that you're filtering them inside each method, but I think it should be more clear to filter it here instead. The `create…` and `update…` services just receive instructions, and if they fail that's responsibility of the caller.

packages/core/src/services/commits/create.ts
```diff
@@ -8,19 +8,23 @@ import {
 } from '@latitude-data/core'
 
 export async function createCommit({
-  commit,
+  commit: { projectId, title, mergedAt },
   db = database,
 }: {
-  commit: Omit<Partial<Commit>, 'id'>
+  commit: {
+    projectId: number
```
geclos, line 15: leave a todo this should be a project instead of projectid

packages/core/src/tests/factories/documents.ts
```diff
@@ -16,6 +18,13 @@ function makeRandomDocumentVersionData() {
   }
 }
 
+export async function markAsSoftDelete(documentUuid: string, tx = database) {
+  return tx
+    .update(documentVersions)
+    .set({ deletedAt: new Date() })
+    .where(eq(documentVersions.documentUuid, documentUuid))
+}
+
```
csansoon, line 27: Add a commitId, otherwise you're marking it as deleted in the whole file history. A document should not be created being deleted (?)

packages/core/src/services/documents/destroyDocument.ts
geclos: just pass a workspace to this method instead of workspaceId

packages/web-ui/src/sections/Document/Sidebar/Files/NodeHeaderWrapper/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  // by the click event in the menu item that created this node.
+  useEffect(() => {
+    const timeout = setTimeout(() => {
+      inputRef.current?.focus(), 100
+    })
+
+    return () => {
+      clearTimeout(timeout)
+    }
+  }, [inputRef])
```
csansoon, line 95: You could probably use `1` or even `0` ms and it should still work, right?

packages/core/src/services/documents/destroyOrSoftDeleteDocuments.test.ts
```diff
@@ -0,0 +1,87 @@
...
+import { documentVersions } from '$core/schema'
+import { updateDocument } from '$core/services/documents/update'
+import * as factories from '$core/tests/factories'
+import { and, eq } from 'drizzle-orm'
+import { describe, expect, it } from 'vitest'
+
+import { destroyOrSoftDeleteDocuments } from './destroyOrSoftDeleteDocuments'
+
+describe('destroyOrSoftDeleteDocuments', () => {
+  it('hardDestroyDocuments', async () => {
```
geclos, line 11: can we make the titles self explanatory? why does it hard destroy in this test case but it creates in the next test case? It would be awesome if i could know the rule by just reading this line

packages/core/src/services/documents/destroyOrSoftDeleteDocuments.ts
geclos: I agree. Also the array can be conditionally computed since some of the methods do not require it.

geclos (approved):
awesome

apps/web/src/stores/documentVersions.ts
andresgutgon: We'll have `createFolder` in this store

PR #62, opened 2024-07-25
Compile iterator

csansoon:
## Overview
A new tag, provisionally named `step`, has been introduced. This tag pauses the compilation process until a new message is generated with the current messages. Once this new message is generated, it is added to the conversation, allowing the compilation to proceed.

### Compile Method:

- The compile method is now an `Iterator` instead of a simple function, which has a single `next` method.
- Each call to `next` returns an object `{ completed, conversation }`, informing wether it has finished or not, and the current compiled conversation up to this point.
- If `completed` is false, the next call to next expects an `AssistantMessage` as an argument, which is added to the conversation replacing the `step` tag, to continue the compilation from where it paused.

### Step-by-Step Execution:

- The compilation process is repeatedly called for each step.
- To avoid recompiling every node at each step, the status of nodes is saved in the AST, which is updated in the `Iterator`
- A node marks its `status.completed` property as `true` once its execution is complete.
- In subsequent steps, completed nodes are skipped to optimize performance.

### Handling Interrupted Nodes

- If a node's execution is interrupted, it sets `status.completed` to `false` and saves the current scope values in `status.scopePointers`.
- This approach ensures that variable changes from skipped nodes do not cause incorrect scopes in subsequent nodes.

### DIFF
--- a/packages/compiler/src/compiler/base/nodes/config.ts
+++ b/packages/compiler/src/compiler/base/nodes/config.ts
@@ -1,7 +1,10 @@
-import { Config } from '$compiler/parser/interfaces'
+import { Config as ConfigNode } from '$compiler/parser/interfaces'
 
 import { CompileNodeContext } from '../types'
 
-export async function compile(_: CompileNodeContext<Config>) {
-  /* do nothing */
+export async function compile({
+  node,
+  setConfig,
+}: CompileNodeContext<ConfigNode>): Promise<void> {
+  setConfig(node.value)
 }
--- a/packages/compiler/src/compiler/base/nodes/each.ts
+++ b/packages/compiler/src/compiler/base/nodes/each.ts
@@ -2,7 +2,13 @@ import { hasContent, isIterable } from '$compiler/compiler/utils'
 import errors from '$compiler/error/errors'
 import { EachBlock } from '$compiler/parser/interfaces'
 
-import { CompileNodeContext } from '../types'
+import { CompileNodeContext, TemplateNodeWithStatus } from '../types'
+
+type EachNodeWithStatus = TemplateNodeWithStatus & {
+  status: TemplateNodeWithStatus['status'] & {
+    loopIterationIndex: number
+  }
+}
 
 export async function compile({
   node,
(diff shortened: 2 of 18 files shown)


packages/compiler/src/compiler/chain.ts
```diff
@@ -0,0 +1,81 @@
...
+  async step(response?: AssistantMessage): Promise<ChainStep> {
+    if (this.completed) {
+      throw new Error('The chain has already completed')
+    }
+    if (!this.didStart && response !== undefined) {
+      throw new Error('A response is not allowed before the chain has started')
+    }
+    if (this.didStart && response === undefined) {
+      throw new Error('A response is required to continue a chain')
+    }
```
csansoon, line 49: `step` method requirements.

packages/compiler/src/compiler/base/nodes/each.ts
```diff
@@ -54,14 +73,22 @@ export async function compile({
...
       })
     }
+
     i++
   }
+
+  nodeWithStatus.status = {
+    ...nodeWithStatus.status,
+    loopIterationIndex: 0,
+  }
```
csansoon, line 93: The `{{#each …}}` tag stores additional information in its status to know which iterations to skip when performing a new step.

packages/compiler/src/compiler/base/nodes/config.ts
```diff
@@ -1,7 +1,10 @@
-import { Config } from '$compiler/parser/interfaces'
+import { Config as ConfigNode } from '$compiler/parser/interfaces'
 
 import { CompileNodeContext } from '../types'
 
-export async function compile(_: CompileNodeContext<Config>) {
-  /* do nothing */
+export async function compile({
```
geclos, line 5: render instead of compile 🙏🏼

geclos, line 5: render instead of compile plz

csansoon: Done

geclos (approved):

PR #63, opened 2024-07-26
feature: update workspace name

geclos:
Implements first version of settings page with workspace name update


geclos:
🐮 👦🏼

packages/web-ui/src/sections/Settings/WorkspaceName/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const name = ev.target.value
+    if (!name) return
+
+    return update({ name: ev.target.value })
+  }
+  const debouncedChange = useDebouncedCallback(onChange, 500, {
+    trailing: true,
+  })
+
+  // TODO: i18n
```
andresgutgon, line 25: Really?

packages/web-ui/src/sections/Settings/WorkspaceName/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <div className='flex flex-col gap-4 max-w-[50%]'>
```
andresgutgon, line 27: A simple form like this is easier with normal forms + revalidatePath. You could remove a lot of the code done here

andresgutgon: No necessary to debounce because you don't manage input value. You let uncontrolled and on form submit action goes to the server like a plain old rails app

packages/web-ui/src/sections/Settings/WorkspaceName/index.tsx
geclos: it adds little friction and adding i18n later is a massive pita

packages/web-ui/src/sections/Settings/WorkspaceName/index.tsx
geclos: let's talk offline

packages/web-ui/src/sections/Settings/WorkspaceName/index.tsx
andresgutgon: But this is an app for people who are prompting in English. Why would it need to be translated into other languages? 

Adding i18n is always a pita. Less now than later but still maintaining translations is a burden that we might avoid in this app. My 2 cents

PR #64, opened 2024-07-26
feature: crud provider apikeys

geclos:
Implements basic crud for apikeys

image


apps/web/src/actions/providerApiKeuys/create.ts
andresgutgon, line 1: typo `s/Keuys/Keys/g`

apps/web/src/actions/providerApiKeuys/create.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Providers } from '@latitude-data/core/browser'
+import { z } from 'zod'
+
+import { authProcedure } from '../procedures'
+
+export const createProviderApiKeyAction = authProcedure
+  .createServerAction()
+  .input(
+    z.object({
+      provider: z.string(),
```
andresgutgon, line 13: This smells like a [native enum](https://zod.dev/?id=native-enums)

apps/web/src/actions/workspaces/update.ts
```diff
@@ -17,7 +17,7 @@ export const updateWorkspaceAction = authProcedure
     const userId = ctx.session.userId
     const workspacesScope = new WorkspacesRepository(userId)
     const workspace = await workspacesScope
-      .getWorkspaceById(input.workspaceId)
+      .find(input.workspaceId)
```
andresgutgon, line 20: noice

apps/web/src/app/(private)/settings/page.tsx
```diff
@@ -2,9 +2,22 @@
...
 export default function SettingsPage() {
   const { data, update } = useCurrentWorkspace()
+  const { data: apiKeys, create, destroy } = useProviderApiKeys()
+  const { data: users } = useUsers()
 
-  return <Settings workspace={data!} updateWorkspace={update} />
+  return (
+    <Settings
+      users={users}
+      workspace={data}
```
andresgutgon, line 16: Workspace is in the context

apps/web/src/stores/users.ts
```diff
@@ -0,0 +1,27 @@
+import { User } from '@latitude-data/core'
+import { useToast } from '@latitude-data/web-ui'
+import { getUsersActions } from '$/actions/users/fetch'
+import useSWR, { SWRConfiguration } from 'swr'
+
+export default function useUsers(opts?: SWRConfiguration) {
+  const { toast } = useToast()
+
+  const fetcher = async () => {
+    const [data, error] = await getUsersActions()
```
andresgutgon, line 10: I'll open a light debate here. For simple crud screens do we need to store data in the client? Couldn't we just `revaltidatePath` in the action?

packages/core/src/repositories/providerApiKeysRepository.ts
```diff
@@ -0,0 +1,32 @@
...
+      .as('providerApiKeysScope')
+  }
+
+  async find(id: number) {
+    const result = await this.db
+      .select()
+      .from(this.scope)
+      .where(eq(this.scope.id, id))
+    if (!result.length) {
+      return Result.error(new NotFoundError('ProviderApiKey not found'))
```
andresgutgon, line 22: Be aware that in the server component you have to handle this with [notFound()](https://nextjs.org/docs/pages/api-reference/functions/get-server-side-props#notfound) to show users the not found error boundary. Otherwise this is treated like an exception

packages/core/src/repositories/usersRepository.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import Repository from './repository'
+
+export class UsersRepository extends Repository {
+  get scope() {
+    return this.db
+      .select(getTableColumns(users))
+      .from(users)
+      .innerJoin(
+        memberships,
```
andresgutgon, line 13: Should we introduce `role` at some point? Otherwise anyone can do anything

packages/core/src/schema/models/providerApiKeys.ts
```diff
@@ -0,0 +1,47 @@
...
+  'provider_api_keys',
+  {
+    id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
+    name: varchar('name').notNull(),
+    token: varchar('token').notNull(),
+    provider: providersEnum('provider').notNull(),
+    authorId: text('author_id')
+      .notNull()
+      .references(() => users.id),
+    workspaceId: bigint('workspace_id', { mode: 'number' })
```
andresgutgon, line 33: Add the relations for author and workspace

packages/core/src/services/providerApiKeys/create.ts
```diff
@@ -0,0 +1,36 @@
...
+    const result = await tx
+      .insert(providerApiKeys)
+      .values({
+        workspaceId: workspace.id!,
+        provider,
+        token,
+        name,
+        authorId,
+      })
+      .returning()
```
andresgutgon, line 32: I know we want to return anything here. However, we must be mindful because this can leak to the frontend columns, which might not be necessary or desirable for the client. We solved this before with presenters or POJOs. Not sure if it's worth it just commenting

packages/web-ui/src/sections/Settings/ProviderApiKeys/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+              token: defaultGenerateNodeUuid(),
+              name: defaultGenerateNodeUuid(),
+            })
+          }
+        >
+          Add new API Key
+        </Button>
+      </div>
+      <div className='flex flex-col gap-2'>
+        <table>
```
andresgutgon, line 40: Smells like a UI DS component maybe

geclos: will happen in another pr

packages/core/src/services/providerApiKeys/create.ts
geclos: Yup,

any serialization concerns should happen at the boundaries (actions/controllers)

apps/web/src/stores/users.ts
geclos: I don't think it's a good idea to introduce more than one way data flows between frontend and backend, let's keep it simple, data flows always via actions.

apps/web/src/app/(private)/settings/page.tsx
geclos: Long story short I can't make it work with sessions without some serious refactor I think we have to rethink the whole web-ui/web divide. Not being able to use stores in web-ui components is a massive pita.

PR #66, opened 2024-07-28
Implement folder deletion in sidebar

andresgutgon:
# What?
☝️ That

## Next PR
- [x] Destroy folder server action
- [x] ~~Review constraint in document version by commit and path. Now that documents can be soft deleted it will be possible to have same path in 2 different docs. This only happens when a doc existed in a merge commit and you remove it in the draft commit. Then try to create it again. Instead of removing the foreing key constraint. Change delete method to see if there is already a soft deleted document with that path.~~ **ADDED DELETED_AT To the constraint**. This way is not possible to have 2 different docs with same COMMIT + same PATH + DELETED_AT=null
- [x] ~~When deleting a folder remove also tmp folders inside it. Refresh zustand store~~ We can live without this for now. Not doing it
- [x] Implement a confirmation folder for deleting documents or folders

image

## Next PR
- [x] Resizable sidebar 
- [ ] Make the sidebar files tree vertically scrollable. Custom scrollbar styles
- [ ] Sidebar header with File + | Folder +


apps/web/src/stores/documentVersions.ts
```diff
@@ -111,8 +119,46 @@ export default function useDocumentVersions(
...
+      })
+
+      if (error) {
+        toast({
+          title: 'Error deleting folder',
+          description: error.formErrors?.[0] || error.message,
+          variant: 'destructive',
+        })
+      } else {
+        mutate()
```
geclos, line 140: you are not waiting for this mutation before redirecting could lead to jumpy ui

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
```diff
@@ -276,41 +277,100 @@ function FileNode({
...
+              : `You're deleting ${deletableNode?.name} prompt`
+          }
+          confirm={{
+            isConfirming: isDestroying,
+            label: deletingFolder ? 'Delete folder' : 'Delete prompt',
+            description: `Deleting this ${deletingFolder ? 'folder' : 'prompt'} will also delete all the prompts it contains. This action cannot be undone.`,
+          }}
+        />
+      ) : null}
+    </>
```
geclos, line 374: huge component consider splitting un smaller chunks

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
```diff
@@ -276,41 +277,100 @@ function FileNode({
...
+    useState<DeletableElement<DeletableType> | null>(null)
 
   useEffect(() => {
     if (currentPath) {
       togglePath(currentPath)
     }
   }, [currentPath, togglePath])
+  const onConfirmDelete = useCallback(
+    async <T extends DeletableType>(deletable: DeletableElement<T>) => {
+      if (deletable.type === 'file') {
```
geclos, line 322: magic string

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
```diff
@@ -276,41 +277,100 @@ function FileNode({
...
+        await destroyFile(deletable.documentUuid)
+      } else if (deletable.type === 'folder') {
```
geclos, line 324: magic string

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
```diff
@@ -276,41 +277,100 @@ function FileNode({
...
+        await destroyFolder(deletable.path)
+      }
 
+      setDeletable(null)
+    },
+    [destroyFile, destroyFolder, deletableNode, setDeletable],
+  )
+
+  const deletingFolder = deletableNode?.type === 'folder'
```
geclos, line 333: ditto

apps/web/src/stores/documentVersions.ts
```diff
@@ -111,8 +119,46 @@ export default function useDocumentVersions(
...
+
+        if (currentDocument?.path?.startsWith?.(path)) {
+          router.push(
+            ROUTES.projects
+              .detail({ id: project.id })
+              .commits.detail({ uuid: commit.uuid }).documents.root,
+          )
+        }
```
csansoon, line 148: Edge case: Add a final `/` to `path` if it doesn't have it already; otherwise you'll be redirecting the user when removing a folder with the same name and path as another file.

```
a/
  b    <-- user is here
  b/   <-- user removes this folder
```

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
```diff
@@ -276,41 +277,100 @@ function FileNode({
...
-    </FileTreeProvider>
+    <>
+      <FileTreeProvider
+        currentPath={currentPath}
+        onNavigateToDocument={navigateToDocument}
+        onCreateFile={async (path) => {
+          createFile({ path })
+        }}
+        onDeleteFile={({ node, documentUuid }) => {
+          setDeletable({ type: 'file', documentUuid, name: node.name })
```
geclos, line 343: ditto, don't let ferran fool you, enums are good for u

packages/web-ui/src/ds/atoms/Modal/index.tsx
```diff
@@ -37,10 +48,65 @@ export function Modal({
...
+            isLoading={confirm.isConfirming}
+            variant={type}
+            onClick={onConfirm}
+          >
+            {confirm.label}
+          </Button>
+        </div>
+      }
+    >
+      {confirm.description || confirm.title ? (
```
geclos, line 101: add parenthesis to make clear the order of evaluation, not sure if prettier is gonna remove them though

csansoon (approved):
Minor comment with an edge case. Solve it before merging

geclos (approved):
looks good 👌🏼

apps/web/src/stores/documentVersions.ts
andresgutgon: Good catch

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
andresgutgon: Is ultra componentized already with hooks and wrappers

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
andresgutgon: Is a type if you put other thing Typescript fails

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
andresgutgon: ditto

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
andresgutgon: ditto

apps/web/src/stores/documentVersions.ts
andresgutgon: This is `/documents/` path is not a folder path

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
andresgutgon: lol

packages/web-ui/src/ds/atoms/Modal/index.tsx
andresgutgon: Prettier removed the parenthesis

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
geclos: ferran is wrong enums are easier to refactor and in practice you'll never hit any of the issues he mentions (you haven't had a problem with enums in 10 years of using them)

andresgutgon:
Fix creating

```
a/
  b/
    file
```

If Now I create 
```
  a/
   b (this is a file)
```

existing `b` folder disappears from sidebar. It's in the DB so it's a frontend bug

PR #68, opened 2024-07-29
minor: move settings to web

geclos:
This commit moves settings section to web. In the future, I will migrate
documents section to web.


apps/web/src/app/(private)/settings/ProviderApiKeys/index.tsx
```diff
@@ -1,23 +1,13 @@
...
-}: {
-  apiKeys: ProviderApiKey[]
-  users: User[]
-  createApiKey: (payload: {
-    provider: Providers
-    token: string
-    name: string
-  }) => Promise<ProviderApiKey | undefined>
-  destroyApiKey: (id: number) => Promise<ProviderApiKey | undefined>
-}) {
```
geclos, line 20: now we can compose with stores so apis become much simpler

apps/web/src/app/(private)/settings/ProviderApiKeys/index.tsx
```diff
@@ -65,7 +55,7 @@ export default function ProviderApiKeys({
                   <Button
                     size='small'
                     variant='destructive'
-                    onClick={() => destroyApiKey(apiKey.id)}
+                    onClick={() => destroy(apiKey.id)}
```
andresgutgon, line 58: I will give a try to simple action in forms for these simple UIs. You avoid having to use local stores

PR #69, opened 2024-07-29
feat: stop it

geclos:
let's stop fighting over this 🙏🏼 

We are a small team such that we don't need to establish hard coding rules (with few exceptions). We should trust our common sense to maintain good code quality. Feel free to use positional or named arguments when it makes things better, we can discuss the particulars of each case in PRs.


csansoon (approved):
☮️

andresgutgon (approved):
☮️

andresgutgon:
I like the conclusion we arrived the other day. 

1. Positional arguments when is only and will always be only one argument
2. Otherwise use object arguments. 

Future prediction is very hard so I will go almost always with object arguments

PR #72, opened 2024-07-29
feature: presenters

geclos:
Implements the first presenter


packages/core/src/services/users/createUser.ts
```diff
@@ -28,11 +28,7 @@ export async function createUser(
         name,
         encryptedPassword,
       })
-      .returning({
-        id: users.id,
-        email: users.email,
-        name: users.name,
-      })
+      .returning()
```
geclos, line 31: moved the presentation logic to a presenter

csansoon (approved):

PR #74, opened 2024-07-29
Fix bug creating a file with the same name of an existing folder. When

andresgutgon:
# What?
Fix the bug creating a file with the same name as an existing folder. When is created the folder with that name disappears from the UI. Looks like a client bug

## Also this visual bug
Fix this when the sidebar is expanded
image

### DIFF
--- a/packages/web-ui/src/sections/Document/Sidebar/Files/FolderHeader/index.tsx
+++ b/packages/web-ui/src/sections/Document/Sidebar/Files/FolderHeader/index.tsx
@@ -1,0 +1,110 @@
+import { useCallback, useMemo } from 'react'
+
+import { Icons } from '$ui/ds/atoms'
+import { MenuOption } from '$ui/ds/atoms/DropdownMenu'
+import { cn } from '$ui/lib/utils'
+import { useFileTreeContext } from '$ui/sections/Document/Sidebar/Files/FilesProvider'
+import NodeHeaderWrapper, {
+  ICON_CLASS,
+  IndentType,
+} from '$ui/sections/Document/Sidebar/Files/NodeHeaderWrapper'
+import { useOpenPaths } from '$ui/sections/Document/Sidebar/Files/useOpenPaths'
+import { useTempNodes } from '$ui/sections/Document/Sidebar/Files/useTempNodes'
+
+import { Node } from '../useTree'
+
+export default function FolderHeader({
+  node,
+  open,
+  indentation,
+  onToggleOpen,
+}: {
+  node: Node
+  open: boolean
+  indentation: IndentType[]
+  onToggleOpen: () => void
+}) {
+  const { onDeleteFolder } = useFileTreeContext()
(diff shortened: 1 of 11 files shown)


packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
andresgutgon, line 1: From 383 LOC -> 220 LOC

How big to be considered big?

packages/web-ui/src/sections/Document/Sidebar/Files/useTree/index.test.ts
```diff
@@ -47,6 +47,49 @@ describe('useTree', () => {
     expect(root.children).toHaveLength(5)
   })
 
+  it('renders folder with same name as file in the same level', async () => {
+    const { result } = renderHook(() =>
+      useTree({
+        documents: [
+          { path: 'a', documentUuid: '1' },
+          { path: 'a/b', documentUuid: '2' },
```
andresgutgon, line 55: This was the bug. When a folder and file was named the same the tree was not rendering the folder `a` and the file `b`

packages/web-ui/src/sections/Document/Sidebar/Files/FolderHeader/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const actions = useMemo<MenuOption[]>(
+    () => [
+      { label: 'New folder', onClick: onAddNode({ isFile: false }) },
+      { label: 'New Prompt', onClick: onAddNode({ isFile: true }) },
+      {
+        label: 'Delete folder',
+        type: 'destructive',
+        onClick: () => {
+          if (node.isPersisted) {
+            onDeleteFolder({ node, path: node.path })
```
geclos, line 73: leaving just as a comment but I feel there's a lot of needless indirection with the write methods. This onDeleteFolder comes from a context hook which gets the ondeletefolder method as a component prop in the parent component, which depends on a deletefile method that is also received as a prop in the parent of the parent, which eventually resolves to the deletefolder method of the documentversions store.  

If we move this to web we should be able to compose the store wherever needed instead of drilling down these methods in several layers of components.

geclos (approved):

packages/web-ui/src/sections/Document/Sidebar/Files/useTree/index.test.ts
geclos: nice

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
geclos: Whatever size you feel makes it easy to read

andresgutgon: It depends on what's on the file I guess. In this case it was one entry component and then a header component that split in 2 (these 2 were the ones I extracted). For me in this case is equally easy to read

PR #78, opened 2024-07-30
Implement commits selector

andresgutgon:
# What?
Implement commits selector in the sidebar 

## TODO
- [x] Test paginated queries
- [x] ~Change users action fetch to be members~ We want to relate commits with users not with memberships
- [x] Add `userId` relationship to commit
- [x] Use fetch by project commits action
- [x] Test action fetch by project
- [x] Expose all SCHEMA types to nextjs client in `core/browser`

image

image

## Next PR
- [ ] UI for selector
- [ ] DISABLE In sidebar actions when a commit is `live`
- [ ] Fix sidebar resizable. It does not save the width and when browser resize it resize the sidebar without user intervention


packages/core/src/repositories/repository.ts
```diff
@@ -10,5 +12,19 @@ export default abstract class Repository {
...
+   */
+  static async paginateQuery<T extends PgSelect>({
+    query,
+    page = 1,
+    pageSize = 20,
+  }: {
+    query: T
+  } & PaginationArgs) {
+    return query.limit(pageSize).offset((page - 1) * pageSize)
+  }
```
andresgutgon, line 27: Super nice and simple generic pagination with drizzle

apps/web/src/actions/users/fetch.test.ts
```diff
@@ -22,9 +26,12 @@ describe('getUsersAction', () => {
 
   describe('authorized', () => {
     beforeEach(async () => {
-      const session = await createWorkspace()
-      ;(getSession as Mock).mockReturnValue({
-        user: session.userData,
+      const { workspace, userData } = await factories.createWorkspace({
+        name: 'test',
+      })
+      mocks.getSession.mockReturnValue({
```
andresgutgon, line 32: This is a nicer way of mocking

apps/web/src/actions/users/fetch.test.ts
```diff
@@ -1,12 +1,16 @@
-import { getSession } from '$/services/auth/getSession'
-import { createWorkspace } from '$core/tests/factories/workspaces'
```
andresgutgon, line 2: Please review your imports. This was not using the factory but the real service. Always import from `@latitude-data/core`

apps/web/package.json
```diff
@@ -36,8 +36,8 @@
     "swr": "^2.2.5",
     "use-debounce": "^10.0.1",
     "zod": "^3.23.8",
-    "zsa": "^0.5.1",
-    "zsa-react": "^0.2.2"
+    "zsa": "file:../../../../opensource/zsa/packages/zsa",
+    "zsa-react": "file:../../../../opensource/zsa/packages/zsa-react"
```
andresgutgon, line 40: ⚠️ change this before merge

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+  open,
+  setOpen,
+}: {
+  open: boolean
+  setOpen: (open: boolean) => void
+}) {
+  const { data, mutate } = useCommits()
+  const { project } = useCurrentProject()
+  const router = useRouter()
+  const { error, executeFormAction, input } = useServerAction(
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx:26]: @geclos This is the new version of `zsa-react` Look how nicely it works with the form. The new `input` field gives me the values present in the form. I'm not convinced we should always use our hook instead of this one. For simple forms like this works super nice

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/page.tsx
```diff
@@ -30,7 +30,7 @@ export default async function CommitRoot({
     <DocumentDetailWrapper
       resizableId={resizableId}
       resizableSizes={layoutData}
-      sidebar={<Sidebar commit={commit} />}
+      sidebar={<Sidebar project={project} commit={commit} />}
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/page.tsx:33]: Check `/versions/:uuid/` page to see if sidebar breaks.

apps/gateway/src/index.ts
```diff
@@ -24,8 +24,16 @@ app.route(jetPaths(ROUTES).Api.V1.Documents.Base, documentsRouter)
...
 app.use(errorHandlerMiddleware())
 
-serve(app, (info) => {
-  console.log(`Listening on http://localhost:${info.port}`)
-})
+serve(
+  {
+    fetch: app.fetch,
+    overrideGlobalObjects: undefined,
+    port: parseInt(process.env.GATEWAY_PORT || '4000', 10),
```
andresgutgon, line 31: I had nextjs app running on 3000 and running test of gateway conflicted

geclos: can't be default gateway port is 8787

apps/web/src/actions/commits/fetchCommitsByProjectAction.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    getSession: vi.fn(),
+  }
+})
+vi.mock('$/services/auth/getSession', () => ({
+  getSession: mocks.getSession,
+}))
+
+let workspace: Workspace
+describe('getUsersAction', () => {
+  useTestDatabase()
```
geclos, line 19: you don't need it

apps/web/src/actions/users/fetch.test.ts
geclos: it was using the factory no?

apps/web/src/actions/users/fetch.ts
```diff
@@ -10,7 +10,7 @@ export const getUsersActions = authProcedure
   .handler(async ({ ctx }) => {
     const usersScope = new UsersRepository(ctx.workspace.id)
 
-    return await usersScope
```
geclos, line 13: remove the async declaration then also

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+        mutate([...data, draft])
+        router.push(
+          ROUTES.projects
+            .detail({ id: project.id })
+            .commits.detail({ uuid: draft.uuid }).root,
+        )
+        setOpen(false)
+      },
+    },
+  )
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx:46]: put this write method inside commits store and use useFormAction instead

```
const { createDraft } = useCommits
const { action, error, isPending } = useFormAction(createDraft) 
```

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+            name='title'
+            errors={formattedErrors?.title}
+            defaultValue={input?.title ?? undefined}
+            placeholder='A title for the version'
+          />
+          <TextArea
+            label='Description'
+            name='description'
+            errors={formattedErrors?.description}
+            defaultValue={input?.description ?? undefined}
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx:82]: with useFormAction you don't need input just use data

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+        Math.max(triggerWidth + TRIGGER_X_PADDING_PX, MIN_WIDTH_SELECTOR_PX),
+      )
+    })
+    if (ref.current) {
+      resizeObserver.observe(ref.current)
+    }
+    return () => {
+      resizeObserver.disconnect()
+    }
+  }, [ref])
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/index.tsx:98]: put the whole width-related shenanigans on its own hook please

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/index.tsx
```diff
@@ -1,29 +1,56 @@
...
+      header={
+        <CommitSelector
+          headCommitId={headCommit.id}
+          currentCommit={commit}
+          draftCommits={draftCommits}
+        />
+      }
+      tree={
+        <ClientFilesTree
+          currentDocument={currentDocument}
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/index.tsx:50]: hmm minor nitpick you usually want this to be an id and fetch the current document from the documents array within the ClientFilesTree component so to avoid duplicated data sync issues (stuff like currentDocument not being present in documents array, or currentDocument having different data than document).

apps/web/src/presenters/commitPresenter.ts
```diff
@@ -0,0 +1,5 @@
+import type { Commit } from '@latitude-data/core/browser'
+
+export default function commitPresenter(commit: Commit) {
+  return commit
+}
```
geclos, line 5: useless presenter are you planning on using it for something in a future PR? otherwise remove please

packages/core/src/schema/models/commits.ts
```diff
@@ -25,6 +25,9 @@ export const commits = latitudeSchema.table(
     projectId: bigint('project_id', { mode: 'number' })
       .notNull()
       .references(() => projects.id, { onDelete: 'cascade' }),
+    userId: text('user_id')
```
geclos, line 28: I think we are calling it author_id/authorId in other places maybe we wanna maintain consistency

packages/core/src/schema/types.ts
```diff
@@ -0,0 +1,25 @@
...
+export type SafeUser = Pick<User, 'id' | 'name' | 'email'>
+export type Session = InferSelectModel<typeof sessions> & {
+  user: User
+}
+export type Membership = InferSelectModel<typeof memberships>
+export type ProviderApiKey = InferSelectModel<typeof providerApiKeys>
+export type ApiKey = InferSelectModel<typeof apiKeys>
+export type Commit = InferSelectModel<typeof commits>
+export type DocumentVersion = InferSelectModel<typeof documentVersions>
+export type Project = InferSelectModel<typeof projects>
```
geclos, line 25: hmm you shouldn't need this if you only import types from a schema file it does not get bundled client-side and everything works fine, problem starts when you import anything other than types from a schema file

packages/core/src/services/commits/create.ts
```diff
@@ -1,21 +1,23 @@
...
 
 export async function createCommit({
   project,
-  data: { title, mergedAt },
+  user,
+  data: { title, description, mergedAt },
   db = database,
 }: {
   project: Project
+  user: SafeUser
```
geclos, line 17: what's the difference between user and safeuser?

packages/web-ui/src/ds/atoms/TextArea/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    size: {
+      normal: INPUT_VARIANT_SIZE.normal,
+      small: INPUT_VARIANT_SIZE.small,
+    },
+  },
+  defaultVariants: {
+    size: 'normal',
+  },
+})
+export type TextAreaProps = TextareaAutosizeProps &
```
geclos, line 22: carlos adds blanklines like a machine gun and you do the exact opposite 😂

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
```diff
@@ -211,7 +211,9 @@ export function FilesTree({
           confirm={{
             isConfirming: isDestroying,
             label: deletingFolder ? 'Delete folder' : 'Delete prompt',
-            description: `Deleting this ${deletingFolder ? 'folder' : 'prompt'} will also delete all the prompts it contains. This action cannot be undone.`,
+            description: deletingFolder
+              ? 'Deleting this folder will also delete all the prompts it contains. This action cannot be undone.'
+              : 'Deleting this prompt might affect your production app and other prompts referencing it. This action cannot be undone.',
```
geclos, line 216: Hmm I don't get the second warning. Deleting anything does not affect production untill you merge the commit no? Also it could be undone no? Documents are soft deleted

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx:82]: This makes uncontrolled input. There is no need for debounce. Also feels better this way

packages/core/src/schema/types.ts
andresgutgon: I sick of dealing with this problem. I think is clear if we put all schema types in one file that we now we can call from anywhere

packages/core/src/services/commits/create.ts
andresgutgon: Safe user only exposes email,name,id.

packages/web-ui/src/sections/Document/Sidebar/Files/index.tsx
andresgutgon: If you delete some document and you merge it without checking reference will break production

packages/core/src/schema/models/commits.ts
andresgutgon: Is it super important? I would rather not make this change.

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/index.tsx:50]: Indeed this object is only pass down to do Id comparison

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/CommitSelector/NewDraftCommitModal/index.tsx:46]: I find this way more concise. Let's talk offline I'm not convince `useFormAction` is better than this

apps/web/src/actions/users/fetch.ts
andresgutgon: I'll do but does not matter no? This method keeps being async

apps/web/src/actions/commits/fetchCommitsByProjectAction.test.ts
andresgutgon: Ok but I keep watching data in test db after test run. We have something wrong. Doesn't matter if I put this here or not

PR #79, opened 2024-07-30
Put new API key in a separate modal in their own route path

andresgutgon:
# What?
The idea is that the settings page is the layout and nested routes are children of the settings dashboard. The only thing a bit weird is that the actual page for `/settings` is empty because all its content is in the layout. But I think it makes sense

## TODO
- [x] It works 🎉 

image


apps/web/src/app/(private)/settings/page.tsx
```diff
@@ -1,13 +1,4 @@
...
-  return (
-    <div className='w-full py-6 flex flex-col items-center'>
-      <div className='flex flex-col w-[1024px] gap-[40px]'>
-        <WorkspaceName />
-        <ProviderApiKeys />
-      </div>
-    </div>
-  )
+  // All the content is in ./layout.tsx
+  return null
```
andresgutgon, line 3: This weird but you think about it is exactly how works React Router I think

apps/web/package.json
```diff
@@ -31,6 +31,7 @@
     "lucia": "^3.2.0",
     "monaco-editor": "^0.50.0",
     "next": "^14.3.0-canary.87",
+    "nextjs-toploader": "^1.6.12",
```
andresgutgon, line 34: This is just one file package. We could do our selves but is works out of the box

andresgutgon:
Not the right idea

PR #80, opened 2024-07-31
feature: gateway

geclos:
Implements Latitude API Gateway using Hono with the nodejs adapter


apps/gateway/test/setup.ts
```diff
@@ -0,0 +1,5 @@
+// vitest-env.d.ts
+
+import useTestDatabase from '$core/tests/useTestDatabase'
+
+useTestDatabase()
```
andresgutgon, line 5: Makes sense to put this here if we have to put inside describes anyway?

apps/gateway/test/vitest-env.d.ts
```diff
@@ -0,0 +1,9 @@
+import 'vitest'
+
+import * as factories from './factories'
```
andresgutgon, line 3: I guess copy & pasta

apps/gateway/vitest.config.ts
```diff
@@ -0,0 +1,9 @@
+import tsconfigPaths from 'vite-tsconfig-paths'
```
andresgutgon, line 1: No need, remember vitest already handle this nicely

apps/gateway/test/setup.ts
geclos: you don't need to put it in describe :)

packages/core/src/schema/models/apiKeys.ts
```diff
@@ -19,7 +19,7 @@ export const apiKeys = latitudeSchema.table(
     token: uuid('token')
       .notNull()
       .unique()
-      .$defaultFn(() => crypto.randomUUID()),
+      .default(sql`gen_random_uuid()`),
```
csansoon, line 22: Why this change?

apps/gateway/test/setup.ts
andresgutgon: Ah no. Ok nice

packages/core/src/schema/models/apiKeys.ts
andresgutgon: This is generated natively by postgreSQL

geclos: better to let the db handle defaults when possible instead of our code.

apps/gateway/test/setup.ts
geclos: fixed web

andresgutgon (approved):

PR #81, opened 2024-07-31
feature: api run prompt endpoint

geclos:


packages/core/package.json
```diff
@@ -23,9 +23,11 @@
     "prettier": "prettier --write src/**/*.ts"
   },
   "dependencies": {
+    "@ai-sdk/openai": "^0.0.40",
     "@latitude-data/compiler": "workspace:^",
     "@latitude-data/env": "workspace:^",
     "@t3-oss/env-core": "^0.10.1",
+    "ai": "^3.2.42",
```
andresgutgon, line 30: Does it need to go in core?

geclos: yeah it will be used in wen and api

apps/gateway/src/common/env.ts
```diff
@@ -0,0 +1,26 @@
+import '@latitude-data/env'
+
+import { createEnv } from '@t3-oss/env-nextjs'
```
andresgutgon, line 3: Nextjs no, I believe they have an agnostic package

apps/gateway/src/server.ts
```diff
@@ -0,0 +1,16 @@
+import { serve } from '@hono/node-server'
+
+import app from '.'
+import env from './common/env'
```
andresgutgon, line 4: Do we have alias in the tsconfig of this package?

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      await mergeCommit(commit).then((r) => r.unwrap())
+
+      // TODO: We refetch the document because merging a commit actually replaces the
+      // draft document with a new one. Review this behavior.
+      const docsScope = new DocumentVersionsRepository(workspace.id)
+      const documentVersion = await docsScope
+        .getDocumentByPath({ commit, path })
+        .then((r) => r.unwrap())
+
+      const route = `/api/v1/projects/${project!.id}/commits/${commit!.uuid}/documents/run`
```
andresgutgon, line 63: Use a routes.ts object like in web to avoid repeating all the paths

packages/core/package.json
csansoon: If we use the sdk to generate stuff in web, then we only need `ai` in api.

geclos: I actually don't think it's a good idea to use the sdk from web, since the sdk points to the production gateway and you don't want to have users do tests in production systems.

andresgutgon: But users in our web are also in production no?

apps/web/src/actions/documents/updateContent.ts
```diff
@@ -26,7 +26,7 @@ export const updateDocumentContentAction = withProject
       .then((r) => r.unwrap())
     const docsScope = new DocumentVersionsRepository(ctx.project.workspaceId)
     const document = await docsScope
-      .getDocumentByUuid({
+      .getDocumentAtCommit({
```
geclos, line 29: this was failing for me without this change

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/_shared.ts
```diff
@@ -0,0 +1,46 @@
...
+  workspace: Workspace
+  projectId: number
+  commitUuid: string
+  documentPath: string
+}) => {
+  const projectsScope = new ProjectsRepository(workspace.id)
+  const commitsScope = new CommitsRepository(workspace.id)
+  const docsScope = new DocumentVersionsRepository(workspace.id)
+
+  // get project, commit, and document
```
csansoon, line 31: 🔥

packages/compiler/src/compiler/base/nodes/tags/message.ts
```diff
@@ -95,7 +95,7 @@ function buildMessage(
   if (role === MessageRole.system) {
     return {
       role,
-      content,
+      content: content[0]?.value,
```
geclos, line 98: little change now system message returns string instead of messagecontent

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.test.ts
```diff
@@ -0,0 +1,141 @@
...
+
+          controller.close()
+        },
+      })
+
+      const response = new Promise((resolve) => {
+        resolve({ text: 'Hello', usage: {} })
+      })
+
+      ;(streamText as Mock).mockReturnValue(
```
andresgutgon, line 64: Do this way please. Is less dirty
https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/actions/commits/fetchCommitsByProjectAction.test.ts#L8-L15

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.test.ts
```diff
@@ -0,0 +1,141 @@
...
+              response,
+            }),
+          )
+        }),
+      )
+
+      const { workspace, user, project } = await factories.createProject()
+      const apikey = await database.query.apiKeys.findFirst({
+        where: eq(apiKeys.workspaceId, workspace.id),
+      })
```
andresgutgon, line 78: Pick apiKey from `createProject` factory

packages/core/package.json
geclos: but you are doing tests with new prompts that might cause shenanigans, i prefer to sandbox local tests in the dashbaord rather than using the production gateway that affects all users

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.test.ts
geclos: ? createProject factory creates an apikey?

apps/web/src/actions/documents/updateContent.ts
andresgutgon: @csansoon already fixed this

packages/core/src/services/documents/streamText.ts
```diff
@@ -0,0 +1,172 @@
...
+  name,
+}: {
+  scope: ProviderApiKeysRepository
+  name: string
+}) {
+  const apiKeyResult = await scope.findByName(name)
+  if (apiKeyResult.error) throw apiKeyResult.error
+
+  return apiKeyResult.value!
+}
```
andresgutgon, line 172: This file could have some tests on its own

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.test.ts
geclos: doesn't export it

andresgutgon: But it could

tools/typescript/base.json
```diff
@@ -2,24 +2,24 @@
...
-    "noUncheckedIndexedAccess": true,
+    "noEmit": true,
     "noFallthroughCasesInSwitch": true,
+    "noUncheckedIndexedAccess": true,
+    "noUnusedLocals": true,
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "noUnusedLocals": true,
     "target": "ES2022"
```
andresgutgon, line 23: What changed here?

geclos: nah it's just reordering

andresgutgon (approved):

PR #86, opened 2024-08-01
Chain: do not add response as message when storing it as a variable

csansoon:
## Without `as` attribute:
(See the `Assistant` message between `Foo` and `Bar`)
image

## With `as` attribute:
(See the lack of `Assistant` message between `Foo` and `Bar`)
image

### DIFF
--- a/packages/compiler/src/compiler/readMetadata.ts
+++ b/packages/compiler/src/compiler/readMetadata.ts
@@ -514,11 +514,11 @@ export class ReadMetadata {
 
         if (attributes.has('as')) {
           const asAttribute = node.attributes.find((a) => a.name === 'as')!
-          const asValue = (asAttribute.value as TemplateNode[])
-            .map((n) => n.value)
-            .join('')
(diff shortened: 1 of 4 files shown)


packages/compiler/src/compiler/readMetadata.ts
```diff
@@ -514,11 +514,12 @@ export class ReadMetadata {
...
         if (attributes.has('as')) {
           const asAttribute = node.attributes.find((a) => a.name === 'as')!
-          const asValue = (asAttribute.value as TemplateNode[])
-            .map((n) => n.value)
-            .join('')
+          if (asAttribute.value !== true) {
+            const asValue = asAttribute.value.map((n) => n.data).join('')
 
-          scopeContext.definedVariables.add(asValue)
+            console.log('adding', asValue, 'to scope')
```
andresgutgon, line 520: 🔥

andresgutgon (approved):

geclos (approved):

PR #93, opened 2024-08-05
Fix: edit document

csansoon:
Editing documents from a new draft was always failing because it tried to perform the update on the HEAD commit instead.


apps/web/src/stores/documentVersions.ts
```diff
@@ -150,11 +159,33 @@ export default function useDocumentVersions(
         }
       }
     },
-    [executeDestroyFolder, mutate, currentDocument?.path],
+    [executeDestroyFolder, mutate, currentDocument?.path, commit.id],
   )
 
-  const { execute: updateContent } = useLatitudeAction(
```
andresgutgon, line 156: Why this? We should use `useLatitudeAction` to have a unique way of doing actions

csansoon: Every other action in here uses the regular `useServerAction`

PR #94, opened 2024-08-05
Sidebar resizable and scrollable

andresgutgon:
# What?
Fix resizable sidebar and make it vertically scrollable


packages/web-ui/src/ds/atoms/ResizablePane/index.tsx
```diff
@@ -1,58 +0,0 @@
-'use client'
-
-import { ComponentProps } from 'react'
-
-import { cn } from '$ui/lib/utils'
-import * as ResizablePrimitive from 'react-resizable-panels'
```
andresgutgon, line 6: @geclos I tried something new and it was not better than what we used before 😂 . I'll stick with what I know from now on

geclos (approved):

PR #97, opened 2024-08-05
Fix: AI generation in Playground

csansoon:
Fixes on AI generation in playground:

- References were not being resolved before generating response.
- AI was being called 3x the times it required, and adding several responses to the same conversation thread.
- Renamed `apiKey` config option to `provider`.


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx
```diff
@@ -232,7 +224,7 @@ export default function Chat({
         />
         <ChatTextArea
           placeholder='Enter followup message...'
-          disabled={responseStream !== undefined}
+          disabled={responseStream !== undefined || !!error}
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx:227]: I explicitly left it enabled when erroring. We want users to retry if things fail.

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx
```diff
@@ -138,10 +136,9 @@ export default function Chat({
       })
 
       return response
-    } catch (_) {
-      // do nothing
-    } finally {
+    } catch (error) {
       setResponseStream(undefined)
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx:140]: You are not setting the stream to undefined if things go well now. Which was the previous behavior.

packages/core/src/services/documents/runDocumentVersion.ts
```diff
@@ -1,32 +1,80 @@
...
+    const metadata = await readMetadata({
+      prompt: document.content,
+      fullPath: document.path,
+      referenceFn,
+    })
+
+    resolvedContent = metadata.resolvedPrompt
+  }
+
+  const chain = createChain({ prompt: resolvedContent, parameters })
```
geclos, line 77: So you have to pass resolved prompt to the chain? Seems unexpected. I'd expect the chain to receive the reference function and the prompt and the parameters and do all the shenanigans internally. 

(Totally forgot about references lol)

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/index.tsx
```diff
@@ -24,14 +24,19 @@ import {
...
 import useDocumentVersions from '$/stores/documentVersions'
 import { useDebouncedCallback } from 'use-debounce'
 
 import { Header } from './Header'
 import Playground from './Playground'
 
 export const DocumentEditorContext = createContext<
-  { runAction: Function; streamTextAction: Function } | undefined
+  | {
+      runAction: Function
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/index.tsx:36]: We don't need this action anymore no?

packages/core/src/services/documents/runDocumentVersion.ts
```diff
@@ -1,32 +1,80 @@
...
 
 export async function runDocumentVersion({
-  document,
+  documentUuid,
+  commit,
   parameters,
 }: {
-  document: DocumentVersion
+  documentUuid: string
+  commit: Commit
```
geclos, line 34: Nein keep the api as it was. The document version service receives the document version, and you can resolve all the rest from there. You don't need scopes it's not the service responsibility to ensure tenancy.

apps/web/src/actions/documents/streamTextAction.ts
```diff
@@ -1,37 +1,47 @@
...
+
+type StreamTextActionResponse = Promise<{ output: StreamableValue }>
+export type StreamTextOutputAction = (
+  _: StreamTextActionProps,
+) => StreamTextActionResponse
 
 export async function streamTextAction({
-  id,
+  config,
   messages,
```
geclos, line 26: Much better!

andresgutgon: Gerard why we're not using `zsa` here?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx:140]: Yes, it is inside the `try`, above the last `addMessage`. It was being called twice before.

packages/core/src/services/documents/runDocumentVersion.ts
csansoon: Yeah, I get this is not the expected usage. But chains require having references resolved, and for that it needs a `referenceFn` to know how to resolve them. We could add it to the `createChain` function, if that makes it easier to understand.

packages/core/src/services/documents/runDocumentVersion.ts
csansoon: You can't resolve the `commit`. I need to know which `commit` you're running the prompt on, because the `ref` tags have to resolve to documents in the same commit.

The `document` object can contain an older `commitId`, if that's when it was last modified, but if a referenced document has been edited after that commit, those changes would not appear if I just resolve the references using the documents from the `getDocumentsAtCommit(document.commitId)`

apps/web/src/actions/documents/streamTextAction.ts
geclos: it didn't support streaming afaik

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx
Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx:140]: ah lol put it in finally that is the point

packages/core/drizzle/meta/0008_snapshot.json
```diff
@@ -660,7 +660,7 @@
         "provider": {
           "name": "provider",
           "type": "provider",
-          "typeSchema": "public",
+          "typeSchema": "latitude",
```
geclos, line 663: :\ I don't get why this happened

andresgutgon: The first time was added as public. https://github.com/latitude-dev/latitude-llm/commit/8602afddbebafb18559a38595d62f96c05018389#diff-218de992584a1201775b354fc2048c6fa4555dcab5196b1745d3f24c1e85375cR18

andresgutgon: And we all had from that point forward the same enum in latitude and public schemas

packages/core/drizzle/0008_mean_wallop.sql
```diff
@@ -1,5 +1,5 @@
 DO $$ BEGIN
- CREATE TYPE "public"."provider" AS ENUM('openai', 'anthropic');
```
geclos, line 2: hmm this shouldn't be the way to do it you are changing the past here

packages/core/src/services/documents/runDocumentVersion.ts
geclos: buf ya

packages/core/drizzle/meta/0008_snapshot.json
geclos: sure but don't change the past, do a migration to change the enum schema

geclos:
@csansoon closing this

PR #101, opened 2024-08-06
Document logs

csansoon:


packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -97,12 +99,32 @@ async function iterate({
...
+    const result = await ai(
+      {
+        messages: conversation.messages,
+        apiKey: apiKey.token,
+        provider: apiKey.provider,
+        model: config.model,
+      },
+      {
+        onFinish: (event: FinishCallbackEvent) => {
+          logProviderGeneration({
```
andresgutgon, line 111: Use verbs for services. `generateProviderLog`

packages/core/drizzle/0026_latitude_rocks.sql
```diff
@@ -0,0 +1,51 @@
+CREATE TABLE IF NOT EXISTS "latitude"."document_logs" (
+	"id" bigserial PRIMARY KEY NOT NULL,
+	"document_uuid" uuid NOT NULL,
+	"commit_id" bigint NOT NULL,
+	"resolved_content" text NOT NULL,
+	"parameters" json NOT NULL,
+	"custom_identifier" text,
```
andresgutgon, line 7: Just identifier if it's a plain text anyway

packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -136,13 +143,22 @@ async function iterate({
 
       return response
     } else {
+      controller.enqueue({
+        event: LATITUDE_EVENT,
+        data: {
+          type: ChainEventTypes.StepComplete,
+          response,
+        },
+      })
```
andresgutgon, line 152: Why this?

packages/core/src/services/providerLogs/index.ts
```diff
@@ -0,0 +1,58 @@
...
+  model: string
+  config: Record<string, unknown>
+  messages: Message[]
+  responseText: string
+  toolCalls?: ToolCall[]
+  tokens: number
+  duration: number
+}
+
+export async function logProviderGeneration(
```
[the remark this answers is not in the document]
csansoon: I'm using 'log' as a verb here. The provider performs a generation, and we want to log it. We "Log the Provider's Generation" (`logProviderGeneration`)

packages/core/drizzle/0026_latitude_rocks.sql
csansoon: I don't want to have two columns with the same meaning ("id" and "identifier"). This is a _custom_ field for the user to use any value as their own _identifier_.

Another field name for this could be `tag` or `custom_tag` or something like that. What do you think?

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.ts
```diff
@@ -15,28 +21,47 @@ const runSchema = z.object({
...
-
-      const result = await runDocumentAtCommit({
-        documentUuid: document.documentUuid,
-        commit,
-        parameters,
-      }).then((r) => r.unwrap())
-
-      await pipeToStream(stream, result.stream)
-    })
+  (c) => {
```
andresgutgon, line 24: I did this change in the other PR

andresgutgon (approved):

PR #102, opened 2024-08-06
feature: update last used of provider api key

geclos:
Makes usage of our jobs system to update the last time a provider api key was used.

### DIFF
--- a/apps/gateway/src/middlewares/errorHandler.ts
+++ b/apps/gateway/src/middlewares/errorHandler.ts
@@ -9,6 +9,7 @@ const errorHandlerMiddleware = () =>
     if (!err) return next()
 
     if (process.env.NODE_ENV !== 'test') {
+      console.error('=== ERROR ===')
       console.error(err.message)
     }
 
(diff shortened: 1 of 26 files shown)


packages/core/src/services/documents/create.ts
```diff
@@ -22,6 +28,7 @@ export async function createNewDocument({
 
     const workspace = await findWorkspaceFromCommit(commit, tx)
     const docsScope = new DocumentVersionsRepository(workspace!.id, tx)
+    const path = sanitizeDocumentPath(inputPath)!
```
geclos, line 31: example of why validations at service level is a bad idea. Besides services having multiple responsibilities, I can't confidently refactor and add a validation to the model on every write. Is updating the create/update services enough? are the models written somewhere else?

apps/gateway/src/middlewares/errorHandler.ts
```diff
@@ -9,6 +9,7 @@ const errorHandlerMiddleware = () =>
     if (!err) return next()
 
     if (process.env.NODE_ENV !== 'test') {
+      console.error('=== ERROR ===')
```
csansoon, line 12: 🔥

csansoon:
Implemented these changes in #101

PR #103, opened 2024-08-06
Refactor run document at commit types to make easy to use in SDK

andresgutgon:
# What?
We want to use the existing gateway API in our JS (TS) SDK. We also want to dog-food this same SDK to do the initial run prompt on the app playground. 

### TODO
- [x] Get the latitude API Key token and set it in project provider
- [x] Refactor Chain types to make more precise
- [x] Remove in role assistant the possibility of `content` attribute be an object. Now it can only be a string

### Next PR
- [ ] Replace the existing `runChain` in `Chat.tsx` by `sdk.runDocument` method.

This PR should ship the `run` method responsible for running a latitude document on a nodejs/browser environment.


packages/sdks/typescript/package.json
```diff
@@ -1,24 +1,36 @@
...
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    }
+  },
+  "dependencies": {
+    "@t3-oss/env-core": "^0.10.1",
+    "eventsource": "^2.0.2",
```
andresgutgon, line 25: EventSource is a browser thing. NodeJS does not have this concept.

packages/web-ui/src/providers/ProjectProvider.tsx
```diff
@@ -13,11 +15,14 @@ const ProjectContext = createContext<IProjectContextType>(
 const ProjectProvider = ({
   children,
   project,
+  latitudeApiKey,
 }: {
   children: ReactNode
+  latitudeApiKey: string
 } & IProjectContextType) => {
+  const sdkRef = useRef(new LatitudeSdk({ latitudeApiKey }))
```
andresgutgon, line 23: ⚠️ This is not the right approach. The SDK has to be used on the server. So I'll move it to a server action to be used in the playground. 

Is not good idea to use in client because we expose the api key. Inside playground is less important but still a security concern.

packages/sdks/typescript/src/utils/types.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+export type GetDocumentParams = {
+  projectId: string
+  commitUuid: string
+  documentPath: string
+}
+
+export enum HandlerType {
+  RunDocument = 'run-document',
+  GetDocument = 'get-document',
```
andresgutgon, line 13: What's the use case for get document?

apps/gateway/package.json
```diff
@@ -4,6 +4,7 @@
   "scripts": {
     "build": "tsc --build tsconfig.prod.json --verbose",
     "dev": "tsx watch src/server",
+    "dev:debug": "tsx watch --inspect-brk src/server",
```
andresgutgon, line 7: Not very helpful for now. 🔥

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/_shared.ts
```diff
@@ -37,7 +36,7 @@ export const getData = async ({
   const document = await docsScope
     .getDocumentByPath({
       commit,
-      path: toDocumentPath(documentPath),
+      path: documentPath,
```
andresgutgon, line 39: This is wrong no? The API should always be sent the full document path so there is no reason to prefix with `/`

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx
```diff
@@ -87,13 +90,13 @@ export default function Chat({
           const { event, data } = serverEvent as ChainEvent
 
           if (
-            event === PROVIDER_EVENT &&
-            data.type === ChainEventTypes.TextDelta
+            event === StreamEventTypes.Provider &&
+            data.type === 'text-delta'
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx:94]: No magic word. This is a type from `ai` package

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/_shared.ts
geclos: well this is shared with get endpoint which can't include the leading / because document path being a url parameter.

packages/sdks/typescript/src/utils/types.ts
geclos: If we open source the compiler this is useful so users can get the document, compile on their end and do whatever they want with it. Otherwise it's pretty useless.

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/_shared.ts
andresgutgon: So it would have to be unshared 😂

packages/compiler/src/types/message.ts
```diff
@@ -47,9 +47,10 @@ export type UserMessage = IMessage & {
   name?: string
 }
 
-export type AssistantMessage = IMessage & {
+export type AssistantMessage = {
   role: MessageRole.assistant
   toolCalls: ToolCall[]
+  content: string
 }
```
csansoon, line 54: 👍

packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -105,26 +116,35 @@ async function iterate({
...
-        event: PROVIDER_EVENT,
+      enqueueEvent(controller, {
+        event: StreamEventTypes.Provider,
         data: value,
       })
     }
 
+    // TODO: The type on `ai` package return a different definition for `toolCalls`
+    // than `@latitude-data/compiler` package. We have to unify the naming.
+    // For now no toolCalls
```
csansoon, line 127: The naming in `ai` is not only ugly, but does not follow the openai schema.

andresgutgon: I don't understand this comment

PR #104, opened 2024-08-07
feature: support structured outputs from AI and used them in evaluations

csansoon:
Also implemented optionally transactional provider log generation upon
calling ai service


packages/core/package.json
```diff
@@ -88,12 +90,13 @@
     "@latitude-data/mailers": "workspace:^",
     "@sindresorhus/slugify": "^2.2.1",
     "@t3-oss/env-core": "^0.11.1",
-    "ai": "^3.2.42",
+    "ai": "^3.3.3",
```
andresgutgon, line 93: El buen version bump

packages/core/src/constants.ts
```diff
@@ -81,12 +98,13 @@ type LatitudeEventData =
...
       type: ChainEventTypes.StepComplete
-      response: ChainStepCallResponse
+      response: ChainCallResponse
     }
   | {
       type: ChainEventTypes.Complete
       config: Config
-      messages: Message[]
+      messages?: Message[]
+      object?: any
```
andresgutgon, line 107: We should have a `AnyObject` and avoid `any` always

andresgutgon (approved):

PR #105, opened 2024-08-07
feature: prepare web for deployment

geclos:
Prepares web to be deployed to fly.io

### DIFF
--- a/apps/web/docker/Dockerfile
+++ b/apps/web/docker/Dockerfile
@@ -1,4 +1,15 @@
-FROM latitudedata/base AS base
+FROM node:20-alpine AS base
+
+# Check https://github.com/nodejs/docker-node/tree/b4117f9333da4138b03a546ec926ef50a31506c3#nodealpine to understand why libc6-compat might be needed.
+RUN apk update
+RUN apk add --no-cache libc6-compat
+
+# Install pnpm
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable && corepack prepare pnpm@8.15.6 --activate
+
+RUN pnpm i -g turbo
 
 WORKDIR /app
 COPY . .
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -34,15 +34,13 @@ services:
...
       dockerfile: apps/web/docker/Dockerfile
-    # This enables to have this service for testing production build
-    # It will not run when doing docker compose up
-    profiles: [building]
-    volumes:
-      - ./packages/env:/app/packages/env
-      - ./packages/core:/app/packages/core
-      - ./packages/jobs:/app/packages/jobs
-      - ./packages/web-ui:/app/packages/web-ui
-      - ./apps/web:/app/apps/web
```
andresgutgon, line 45: I find this helpful

docker-compose.yml
```diff
@@ -34,15 +34,13 @@ services:
...
+    ports:
+      - 3000:3000
+    environment:
+      - NODE_ENV=production
+      - DATABASE_URL=postgres://latitude:secret@db:5432/latitude_production
+      - REDIS_PORT=6379
+      - REDIS_HOST=redis
+      - REDIS_PASSWORD=secret
```
andresgutgon, line 44: Why is this necessary?

apps/web/docker/Dockerfile
```diff
@@ -1,4 +1,15 @@
-FROM latitudedata/base AS base
+FROM node:20-alpine AS base
```
andresgutgon, line 1: With only these changes the image builds. I find it better to have an install of dependencies for build and then another stage for `--prod` dependencies. Take into account if you change this

docker-compose.yml
geclos: why? are u using the dockerfile for development?

docker-compose.yml
geclos: not needed just using it to test the production image locally

apps/web/src/data-access/users.ts
```diff
@@ -34,7 +34,7 @@ export async function getUserFromCredentials({
   if (!user) return notFound()
 
   const validPassword = await verifyPassword(password, user.encryptedPassword)
-  if (!validPassword) notFound()
+  if (!validPassword) return notFound()
```
csansoon, line 37: nice catch

csansoon (approved):

PR #106, opened 2024-08-08
Document logs

csansoon:
Create `documentLogs` when running a chain through `runDocumentAtCommit`, and associate all the generated `providerLogs` from the chain.


packages/core/src/services/documentLogs/create.ts
```diff
@@ -0,0 +1,58 @@
...
+  Transaction,
+} from '@latitude-data/core'
+import { DocumentLog } from '$core/browser'
+
+import { assignDocumentLogToProviderLog } from '../providerLogs'
+
+export type CreateDocumentLogProps = {
+  logUuid: string
+  documentUuid: string
+  commitId: number
```
andresgutgon, line 14: Pass the `commit` not the id. This service should be used after commit is found

packages/core/src/services/documents/getResolvedContent.ts
```diff
@@ -0,0 +1,53 @@
...
+import { LatitudeError, NotFoundError, Result, TypedResult } from '$core/lib'
+import { DocumentVersionsRepository } from '$core/repositories'
+
+export async function getResolvedContent({
+  workspace,
+  documentUuid,
+  commit,
+}: {
+  // TODO: review, we shouldn't need to pass workspace around so much
+  workspace: Workspace
```
andresgutgon, line 14: You don't. Why you don't pass the document from a scope

packages/core/src/services/providerLogs/assignDocumentLog.ts
```diff
@@ -0,0 +1,46 @@
...
+    providerLogUuids,
+  }: {
+    documentLogUuid: string
+    providerLogUuids: string[]
+  },
+  db = database,
+) {
+  return Transaction.call(async (trx) => {
+    const documentLog = await trx.query.documentLogs.findFirst({
+      where: eq(documentLogs.uuid, documentLogUuid),
```
andresgutgon, line 23: This does not have scope

PR #107, opened 2024-08-08
Use SDK response in playground

andresgutgon:
# What?
Use SDK to fetch data from gateway and use it in Playground

## TODO
- [x] SDK accumulates conversation messages and return at the end
- [x] Add tests for `runDocument`
- [x] Handle errors for example wrong model name

### DIFF
--- a/apps/web/src/actions/sdk/runDocumentAction.ts
+++ b/apps/web/src/actions/sdk/runDocumentAction.ts
@@ -1,6 +1,6 @@
 'use server'
 
-import { LatitudeSdk } from '@latitude-data/sdk-js'
+import { LatitudeSdk, type ChainEvent } from '@latitude-data/sdk-js'
 import { getLatitudeApiKey } from '$/app/(private)/_data-access/latitudeApiKey'
 import { createStreamableValue, StreamableValue } from 'ai/rsc'
 
@@ -8,8 +8,9 @@ type RunDocumentActionProps = {
   documentPath: string
   projectId: number
   commitUuid: string
+  parameters: Record<string, unknown>
 }
-type RunDocumentResponse = Promise<{ output: StreamableValue }>
+type RunDocumentResponse = Promise<{ output: StreamableValue<ChainEvent> }>
 export type RunDocumentActionFn = (
   _: RunDocumentActionProps,
 ) => RunDocumentResponse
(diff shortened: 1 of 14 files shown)


packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -179,10 +180,29 @@ async function iterate({
...
+    controller.close()
+    return {
+      text: error.message,
+      usage: {
+        promptTokens: 0,
+        completionTokens: 0,
+        totalTokens: 0,
+      },
+      toolCalls: [],
+    }
```
andresgutgon, line 205: This way we catch the error from AI apis providers
image

packages/sdks/typescript/src/test/chunks-example.ts
```diff
@@ -0,0 +1,174 @@
...
+          promptTokens: 114,
+          completionTokens: 352,
+          totalTokens: 466,
+        },
+      },
+    },
+  },
+]
+
+export const FINAL_RESPONSE = {
```
andresgutgon, line 153: @csansoon this is the output of `runDocument`

apps/web/src/actions/sdk/runDocumentAction.ts
```diff
@@ -22,26 +22,30 @@ export async function runDocumentAction({
...
+          name: error.name,
+          message: error.message,
+          stack: error.stack,
+        })
+      },
+      onFinished: () => stream.done(),
+    })
+  }
+
+  runSdk()
```
csansoon, line 48: Why is this defined as a function? Can't you just do it directly?

csansoon (dismissed):

apps/web/src/actions/sdk/runDocumentAction.ts
andresgutgon: Removed

packages/sdks/typescript/src/index.ts
```diff
@@ -1,12 +1,24 @@
...
 import { RouteResolver } from '$sdk/utils'
-import {
-  ChainEvent,
-  ChainEventTypes,
-  HandlerType,
-  StreamEventTypes,
-} from '$sdk/utils/types'
+import { ChainEvent, HandlerType } from '$sdk/utils/types'
 
+type Message = {
```
andresgutgon, line 5: My OCD is going to explode. We need to DRY all these types

PR #113, opened 2024-08-13
Link document logs and provider logs by UUID

andresgutgon:
# What?
A document run produces 1 Document log and N Provider logs. Until now, these logs were linked by a `documentLogId` foreign key in provider logs. The problem is that in some situations like when a document runs the logs of the providers are generated before document log. To overcome this asyncrony problem we decided to link all these logs by `documentLogUUID` string that's generated before start and is passed down so users can attach new provider logs to an existing document log.

## TODO
- [x] `runDocumentAtCommit` return `documentLogUuid`. 
- [x] document log use returned `documentLogUuid`
- [x] Fix provider logs respository to query by `documentLogUuid`
- [x] `providerLogHandler` is called with `documentLogUuid`
- [x] QA playground
- [x] Test `runDocumentAtCommit` adds `docummentLogUuid` when chain is completed.


packages/core/src/services/documents/getResolvedContent.ts
```diff
@@ -1,26 +1,41 @@
...
+  const documentScope = new DocumentVersionsRepository(workspaceId)
+  const docs = await documentScope
+    .getDocumentsAtCommit(commit)
+    .then((r) => r.unwrap())
 
-  const document = docs.value.find((d) => d.documentUuid === documentUuid)
-  if (!document) return Result.error(new NotFoundError('Document not found'))
+  const docInCommit = docs.find((d) => d.documentUuid === document.documentUuid)
+
+  if (!docInCommit) {
```
andresgutgon, line 34: Now if workspaceId and document don't match it throws an error

packages/core/src/repositories/providerApiKeysRepository.ts
```diff
@@ -9,7 +9,7 @@ export class ProviderApiKeysRepository extends Repository {
     return this.db
       .select(getTableColumns(providerApiKeys))
       .from(providerApiKeys)
-      .innerJoin(workspaces, eq(workspaces.id, providerApiKeys.workspaceId))
+      .where(eq(providerApiKeys.workspaceId, this.workspaceId))
```
andresgutgon, line 12: I believe this was wrong. `workspaceId` is a column in provider apis table

packages/core/src/repositories/providerLogsRepository.ts
```diff
@@ -17,19 +17,11 @@ export class ProviderLogsRepository extends Repository {
       .as('providerLogsScope')
   }
 
-  async findByUuids(uuids: string[]) {
```
andresgutgon, line 20: 🔥 Was not used

packages/core/src/schema/models/providerLogs.ts
```diff
@@ -24,6 +23,7 @@ export const logSourcesEnum = latitudeSchema.enum('log_source', [
 export const providerLogs = latitudeSchema.table('provider_logs', {
   id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
   uuid: uuid('uuid').notNull().unique(),
+  documentLogUuid: uuid('document_log_uuid'),
```
andresgutgon, line 26: In provider_logs table document UUID is nullable. We might want to have provider logs without a linked document log. 

Please confirm this assumption is true

packages/core/src/services/ai/index.ts
andresgutgon, line 1: This service is core and is fully untested. I spent quite some time testing other services and I am tired for now. Please let's add some basic tests to core services when we introduce it 🙏

packages/core/src/services/commits/runDocumentAtCommit.test.ts
andresgutgon, line 1: This was fully untested and still is untested all the streaming flow but the parts that I changed in this PR are covered

packages/core/src/services/commits/runDocumentAtCommit.test.ts
```diff
@@ -0,0 +1,232 @@
...
+        commit,
+        parameters: {},
+        // @ts-ignore
+        providerLogHandler: mocks.providerLogHandler,
+        // @ts-ignore
+        generateUUID: mocks.uuid,
+      }).then((r) => r.unwrap())
+
+      await consumeStream(stream)
+      expect(mocks.runAi).toHaveBeenCalledWith(
```
andresgutgon, line 210: Mocks are hard but on top of that I was not waiting for the stream to be consumed so the test was not passing 🤯

packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -5,57 +5,92 @@ import {
...
-  documentUuid: string
+  workspaceId: Workspace['id']
+  document: DocumentVersion
   commit: Commit
   parameters: Record<string, unknown>
-  logHandler: (log: AILog) => void
+  providerLogHandler: (log: AILog) => void
+  generateUUID?: typeof uuid
 }) {
-  const workspace = await findWorkspaceFromCommit(commit)
```
andresgutgon, line 28: Now we don't pass workspace because is passed in the API gateway. I know we should pass object to service and the external layer validates ownership (which is already doing). But this service is already doing their own check inside `getResolvedContent` so this was a redundant Db call

packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -5,57 +5,92 @@ import {
...
+      })
+        .then(responseResolve)
+        .catch(responseReject)
+    },
   })
 
+  // Dummy handling of the response
+  // This is helpful for not throwing the error
+  // when no one is listening to the promise
+  response.then(() => {}).catch(() => {})
```
andresgutgon, line 87: Response is ready when stream finish. We do this here because if an error is thrown and the consumer is not using the response like we are doing in API gateway the error is raised and appears on API logs. This way it does not appear and also consumer can `catch` the error if using the `response`.

packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -5,57 +5,92 @@ import {
...
+  let responseReject: (reason?: any) => void
 
-  await new Promise<void>((resolve) => {
-    stream = new ReadableStream<ChainEvent>({
-      start(controller) {
-        response = iterate({ chain, scope, controller, logHandler })
+  const response = new Promise<ChainCallResponse>((resolve, reject) => {
+    responseResolve = resolve
+    responseReject = reject
+  })
```
andresgutgon, line 67: Much better this way I think. Thanks Carlos for the help

packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -69,22 +104,24 @@ function enqueueEvent(
 async function iterate({
   chain,
   previousApiKey,
-  scope,
+  allApiKeys,
```
andresgutgon, line 107: Now instead of doing N SQL queries we do on query for all api keys in workspace which usually will be a small collection. From here we just check in a `Map` if the key is found by `name`

packages/core/src/services/documentLogs/create.ts
```diff
@@ -4,33 +4,31 @@ import {
...
-    uuid,
-    documentUuid,
+    data: {
+      uuid,
+      documentUuid,
+      resolvedContent,
+      parameters,
+      customIdentifier,
+      duration,
+    },
```
andresgutgon, line 30: Is nicer in services to have `data` and other parameters that are models like `commit`. This way is clear what's data for the new entity and what are other entities.

packages/core/drizzle/0028_complex_leopardon.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE "latitude"."provider_logs" DROP CONSTRAINT "provider_logs_document_log_id_document_logs_id_fk";
+--> statement-breakpoint
+ALTER TABLE "latitude"."provider_logs" ADD COLUMN "document_log_uuid" uuid;--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "document_log_uuid_idx" ON "latitude"."document_logs" USING btree ("document_uuid");--> statement-breakpoint
```
andresgutgon, line 4: I added an index to provider_logs for `document_log_uuid`. We'll be doing a lot of queries by this column

packages/sdks/typescript/src/index.ts
```diff
@@ -7,6 +7,7 @@ type Message = {
   content: string
 }
 type ChainCallResponse = {
```
andresgutgon, line 9: These types are repeated. We need to do a common shared package to share types and constants common to 

1. Web
2. Gateway
3. SDK
4. Compiler

All these packages should use this common package that would have only the constants and the base types.

packages/web-ui/src/ds/molecules/Chat/MessageList/index.tsx
```diff
@@ -16,21 +17,20 @@ export function MessageList({
   separator?: boolean
 }) {
   return messages.map((message, index) => (
-    <>
```
andresgutgon, line 19: Open the browser in a bigger screen and check Dev tools 🙏 
Otherwise all the warnings like missing key in a loop pass unnoticed

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx
```diff
@@ -148,13 +153,17 @@ export default function Chat({
...
-    const { output } = await runDocumentAction({
+    const { response: actionResponse, output } = await runDocumentAction({
       projectId: project.id,
       documentPath: document.path,
       commitUuid: commit.uuid,
       parameters,
     })
 
+    actionResponse.then((res) => {
+      setDocumentLogUuid(res.response.documentLogUuid)
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Chat.tsx:164]: When SDK (used in our server action) finishes it returns `documentLogUuid` which we set in the state and use when the user keeps talking in the chat.

turbo.json
```diff
@@ -20,7 +20,7 @@
     "build": {
       "dependsOn": ["^build"],
       "inputs": ["$TURBO_DEFAULT$", ".env*"],
-      "outputs": [".next/**", "!.next/cache/**"]
+      "outputs": [".next/**", "!.next/cache/**", "dist/**"]
```
andresgutgon, line 23: I think this fix random CI fails when `@latitude-data/sdk-js` appears not to be build cc c @csansoon

PR #115, opened 2024-08-14
LLM Providers cost estimate

csansoon:
Listed pricing per 1M tokens for each one of our supported Providers.

These list do not contain all models or all model versions so, in addition to the listed prices, I added a function to fallback to a "similar" model when providing one that's not present. As pointed, this is just an estimate.

Note: Azure pricing are just the same as OpenAI's.

### DIFF
--- a/apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.ts
+++ b/apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.ts
@@ -51,6 +51,8 @@ export const runHandler = factory.createHandlers(
         },
       }).then((r) => r.unwrap())
 
+      await pipeToStream(stream, result.stream)
+
       queues.defaultQueue.jobs.enqueueCreateDocumentLogJob({
         commit,
         data: {
@@ -61,8 +63,6 @@ export const runHandler = factory.createHandlers(
           duration: Date.now() - startTime,
         },
       })
-
-      await pipeToStream(stream, result.stream)
     })
   },
 )
--- a/apps/web/src/actions/sdk/runDocumentAction.ts
+++ b/apps/web/src/actions/sdk/runDocumentAction.ts
@@ -28,13 +28,12 @@ export async function runDocumentAction({
   commitUuid,
   parameters,
 }: RunDocumentActionProps) {
-  const result = await getLatitudeApiKey()
-  if (result.error) return result
+  const latitudeApiKey = await getLatitudeApiKey().then((r) => r.unwrap())
 
   const stream = createStreamableValue<ChainEvent, Error>()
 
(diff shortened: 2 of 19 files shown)


packages/core/src/services/ai/index.ts
```diff
@@ -161,3 +161,5 @@ export function validateConfig(config: Record<string, unknown>): Config {
 
   return configSchema.parse(config)
 }
+
+export { estimateCost } from './costEstimate'
```
andresgutgon, line 165: estimateCost -> costEstimate. Why not same func and filename? : )

csansoon: you got a point there sir

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.ts
```diff
@@ -54,6 +54,8 @@ export const runHandler = factory.createHandlers(
         },
       }).then((r) => r.unwrap())
 
+      await pipeToStream(stream, result.stream)
```
andresgutgon, line 57: Why this?

csansoon: Because the `createDocumentLog` job was being enqueued before the stream was even started. This caused it to not assign their `providerLogs` correctly, and to have wrong information such as duration.
Now we await the `pipeToStream` to finish, and then create the document log afterwards.

andresgutgon: Add a fix test

apps/web/src/actions/sdk/runDocumentAction.ts
```diff
@@ -28,13 +28,12 @@ export async function runDocumentAction({
   commitUuid,
   parameters,
 }: RunDocumentActionProps) {
-  const result = await getLatitudeApiKey()
-  if (result.error) return result
```
andresgutgon, line 32: Is better to return `Result`s from services. Then in the API endpoints or actions we call `unwrap`. But if for some reason we want to use this service inside another service we would had to deal with `try/catch` which is discusting

csansoon: But this is a server action, not a service

andresgutgon: Then me callo 😶

apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/run.ts
andresgutgon: No test 😢

PR #116, opened 2024-08-16
/api/chat/add-message API endpoint + sdk method

andresgutgon:
# What?
We want to allow users to keep talking with a document. To link document logs with future messages in a chat-like interface they have to pass document log uuid so we can reference the right document log and from there extract the right version and document config to feed the underlying LLM and add a new provider log to the queue

## TODO
- [x] Implement `api/logs/:uuid/add-message` gateway endpoint
- [x] Implement sdk `addMessage` method (Users will be able to add more than one message)
- [x] Use `sdk.addMessage` in playground (temporarily to test e2e the flow)
- [x] Use sdk in an action
- [x] QA it works
- [x] QA it tag logs as `playground`
- [x] QA it put together document logs + message (provider) logs
- [x] Add tests to `addMessages` provider function
- [x] Add tests to `addMessages` gateway API handler 
- [x] Add tests to `addMessages` SDK function


apps/gateway/src/routes/api/v1/projects/:projectId/commits/:commitUuid/documents/handlers/addMessage.ts
```diff
@@ -0,0 +1,93 @@
...
+    queues.defaultQueue.jobs.enqueueCreateDocumentLogJob({
+      commit,
+      data: {
+        uuid: result.documentLogUuid,
+        documentUuid: document.documentUuid,
+        resolvedContent: result.resolvedContent,
+        parameters,
+        duration: Date.now() - startTime,
+      },
+    })
```
csansoon, line 74: No need for a new `documentLog`

andresgutgon: Wip copy & paste I'll 🔥

apps/web/src/actions/documents/streamTextAction.ts
```diff
@@ -1,79 +0,0 @@
...
-          provider: apiKey,
-          messages,
-          config: rest,
-        },
-        {
-          providerLogHandler: (log) => {
-            queues.defaultQueue.jobs.enqueueCreateProviderLogJob({
-              ...log,
-              documentLogUuid,
-              source: LogSources.Playground,
```
andresgutgon, line 52: TODO: tag `sdk.addMessages` as `LogSources.Playground`

apps/gateway/src/common/routes.ts
```diff
@@ -4,6 +4,10 @@ const ROUTES = {
     Base: '/api',
     V1: {
       Base: '/v1',
+      DocumentLogs: {
+        Base: '/document-logs',
+        AddMessage: '/:documentLogUuid/add-message',
+      },
```
csansoon, line 10: We could expose this just as `/chat`, user does not need to know this is a "document log".

apps/gateway/src/routes/api/v1/document-logs/:documentLogUuid/handlers/addMessage.ts
```diff
@@ -0,0 +1,42 @@
...
+            source,
+            apiKeyId: apiKey.id,
+          })
+        },
+      }).then((r) => r.unwrap())
+
+      await pipeToStream(stream, result.stream)
+    })
+  },
+)
```
csansoon, line 42: If we can add new messages to a chat, I think it may be interesting to mark which of the `providerLogs` is the last one of the chan, and which are from an afterwards added message.

An example of this is "time until response", which is the time spent until getting the stream of the latest step in a chain. To estimate this, I am currently doing `documentLogDuration` - `lastProviderLogDuration`, but this would no longer work when new messages are added afterwards.

packages/core/src/services/documentLogs/addMessages/index.ts
```diff
@@ -0,0 +1,151 @@
...
+      })
+        .then(responseResolve)
+        .catch(responseReject)
+    },
+  })
+
+  // Dummy handling of the response
+  // This is helpful for not throwing the error
+  // when no one is listening to the promise
+  response.then(() => {}).catch(() => {})
```
csansoon, line 77: 👌

apps/gateway/src/common/routes.ts
andresgutgon: `/chat/:documentLogUuid/add-message`? or `chat/add-message` and `documentLogUuid` as a body param?

packages/core/src/schema/models/providerLogs.ts
```diff
@@ -31,10 +33,10 @@ export const providerLogs = latitudeSchema.table('provider_logs', {
...
     }),
   model: varchar('model'),
-  config: json('config').notNull(),
-  messages: json('messages').notNull(),
-  responseText: text('response_text'),
-  toolCalls: json('tool_calls'),
+  config: json('config').$type<PartialConfig>().notNull(),
+  messages: json('messages').$type<Message[]>().notNull(),
+  responseText: text('response_text').$type<string>().default(''),
+  toolCalls: json('tool_calls').$type<ToolCall[]>().default([]),
```
andresgutgon, line 39: This is awesome. Drizzle can type `json` columns

apps/gateway/src/routes/api/v1/document-logs/:documentLogUuid/handlers/addMessage.ts
andresgutgon: Captured here https://github.com/latitude-dev/latitude-llm/issues/120

apps/gateway/src/common/routes.ts
csansoon: > `/chat/:documentLogUuid/add-message`?

yeah, just a change in the name so that it is more straightforward to the user. I think 'chat' is easier to understand than 'document-logs' to the user, since this is the only use they'll give it through the api.

packages/core/src/schema/models/providerLogs.ts
csansoon: oh wow

packages/core/src/services/documentLogs/addMessages/index.test.ts
```diff
@@ -170,7 +170,8 @@ describe('addMessages', () => {
         messages: [
           ...providerLog.messages,
           {
-            role: MessageRole.assistant,
+            /* role: MessageRole.assistant, */
+            role: 'caracoles',
```
csansoon: repámpanos

andresgutgon: Me has pillado, a ver si pillo yo el faking bug

packages/core/src/tests/factories/documentLogs.ts
```diff
@@ -61,8 +63,11 @@ export async function createDocumentLog({
       return acc + content.length
     }, 0)
     const completionTokens = mockedResponse.length
+    iteration++
+    generatedAt = new Date(generatedAt.getTime() + MS_IN_ONE_DAY * iteration)
```
andresgutgon, line 67: One test where 2 provider logs where created where created with the same timestamp. We add here an artificial day span between logs

packages/core/src/services/ai/index.ts
```diff
@@ -126,6 +128,7 @@ export async function ai(
     onFinish: (event) => {
       providerLogHandler({
         uuid: uuidv4(),
+        generatedAt: logGeneratedAt,
```
andresgutgon, line 131: Introduced `generatedAt` to avoid depending on `createdAt`. I already had issues on CI where `createdAt` was identical and test was failing.

packages/core/src/schema/models/providerLogs.ts
andresgutgon: I added defaults in `responseText` and `toolCalls` I think it doesn't hurt

packages/core/src/schema/models/providerLogs.ts
```diff
@@ -47,6 +50,7 @@ export const providerLogs = latitudeSchema.table('provider_logs', {
       onUpdate: 'cascade',
     },
   ),
+  generatedAt: timestamp('generated_at', { mode: 'date' }).notNull(),
```
andresgutgon, line 53: Now we order provider logs by `generatedAt` instead of by `createdAt`

PR #117, opened 2024-08-19
Evaluations schema

csansoon:
Initial schema for evaluations.

### DIFF
--- a/packages/core/drizzle/0030_evaluations_schema.sql
+++ b/packages/core/drizzle/0030_evaluations_schema.sql
@@ -1,0 +1,81 @@
+CREATE TABLE IF NOT EXISTS "latitude"."evaluations" (
+	"id" bigserial PRIMARY KEY NOT NULL,
+	"name" varchar(256) NOT NULL,
+	"description" text NOT NULL,
+	"prompt" text NOT NULL,
+	"workspace_id" bigint NOT NULL,
+	"template_id" bigint,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"updated_at" timestamp DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."connected_evaluations" (
+	"id" bigserial PRIMARY KEY NOT NULL,
+	"document_uuid" uuid NOT NULL,
+	"evaluation_id" bigint NOT NULL,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"updated_at" timestamp DEFAULT now() NOT NULL,
+	CONSTRAINT "connected_evaluations_unique" UNIQUE("document_uuid","evaluation_id")
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."evaluation_results" (
+	"id" bigserial PRIMARY KEY NOT NULL,
+	"evaluation_id" bigint NOT NULL,
+	"document_log_id" bigint NOT NULL,
+	"provider_log_id" bigint NOT NULL,
+	"result" text NOT NULL,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"updated_at" timestamp DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+CREATE TABLE IF NOT EXISTS "latitude"."evaluations_templates" (
+	"id" bigserial PRIMARY KEY NOT NULL,
+	"name" varchar(256) NOT NULL,
+	"description" text NOT NULL,
+	"prompt" text NOT NULL,
+	"created_at" timestamp DEFAULT now() NOT NULL,
+	"updated_at" timestamp DEFAULT now() NOT NULL
+);
+--> statement-breakpoint
+DO $$ BEGIN
+ ALTER TABLE "latitude"."evaluations" ADD CONSTRAINT "evaluations_workspace_id_workspaces_id_fk" FOREIGN KEY ("workspace_id") REFERENCES "latitude"."workspaces"("id") ON DELETE cascade ON UPDATE no action;
+EXCEPTION
+ WHEN duplicate_object THEN null;
+END $$;
+--> statement-breakpoint
+DO $$ BEGIN
+ ALTER TABLE "latitude"."evaluations" ADD CONSTRAINT "evaluations_template_id_evaluations_templates_id_fk" FOREIGN KEY ("template_id") REFERENCES "latitude"."evaluations_templates"("id") ON DELETE no action ON UPDATE no action;
+EXCEPTION
+ WHEN duplicate_object THEN null;
+END $$;
+--> statement-breakpoint
+DO $$ BEGIN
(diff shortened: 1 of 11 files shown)


packages/core/src/schema/models/evaluations.ts
```diff
@@ -0,0 +1,29 @@
...
+    prompt: text('prompt').notNull(),
+    workspaceId: bigint('workspace_id', { mode: 'number' })
+      .notNull()
+      .references(() => workspaces.id, { onDelete: 'cascade' }),
+    ...timestamps(),
+  },
+  (table) => ({
+    evaluationWorkspaceIdx: index('workspace_idx').on(table.workspaceId),
+  }),
+)
```
csansoon, line 22: **Evaluations** — The evaluations defined in the workspace.

`prompt` – Here's where the evaluation document is stored. Not sure if we should let the user define this directly or help them in some way. Also, should this be editable afterwards?

packages/core/src/schema/models/connectedEvaluations.ts
```diff
@@ -0,0 +1,38 @@
...
+      .notNull()
+      .references(() => evaluations.id),
+    ...timestamps(),
+  },
+  (table) => ({
+    connectedEvaluationsPk: primaryKey({
+      columns: [table.evaluationId, table.documentUuid],
+    }),
+  }),
+)
```
csansoon, line 24: **Connected evaluations** — Relation between a document and an evaluation. Defines which evaluations are connected to which documents, independently of document's version.

Not much to say here. The `documentUuid` foreign key may not be possible since it is not a primary key from the `documentVersion` table

packages/core/src/schema/models/evaluations.ts
andresgutgon: Questions for #product

packages/core/src/schema/models/evaluationResults.ts
```diff
@@ -0,0 +1,51 @@
...
+  (table) => ({
+    evaluationIdx: index('evaluation_idx').on(table.evaluationId),
+    evaluationResultDocumentLogIdx: index('document_log_idx').on(
+      table.documentLogId,
+    ),
+    evaluationResultProviderLogIdx: index('provider_log_idx').on(
+      table.providerLogId,
+    ),
+  }),
+)
```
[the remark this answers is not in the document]
andresgutgon: Yes, I am not sure if adding different types of result fields (boolean, numeric, text,...) or even different results table or maybe not maybe a unique text field as you have here is enough 

🧠 Think what we miss if we have every kind of possible result store as text

packages/core/src/schema/models/connectedEvaluations.ts
geclos: Add id primary key to the table. I know you have the compound pk but having a dedicated autoincremental column facilitates working with the table directly when running data migrations, debugging prod db issues, etc..

geclos: Add also indexes for document uuid and relation is columns

packages/core/src/schema/models/evaluationResults.ts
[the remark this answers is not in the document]
csansoon: Talked with product. We're going to store only text for now. We're not going to perform aggregations other than group counts, so that's fine. Will have to revisit this in the future though.

packages/core/src/schema/models/connectedEvaluations.ts
csansoon: hmm, I'll add it, but I don't actually see the need for it. What advantages does it have over the composed pk?

packages/core/src/schema/models/evaluations.ts
csansoon: Talked with product. We may want to add version control in the future, but not for now. This will just be a `prompt` field, and can be modified after its creation.

packages/core/src/schema/models/evaluationResults.ts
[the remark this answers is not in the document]
geclos: Hmm I think it's fine for now? My only concern is other types of evaluations that do not involve prompting an llm for evaluation results (like user submitted evaluations) which suggest a polymorphic approach instead. Bring it up with product to see if this is something we wanna support in the short to mid term and if so then I would consider using polymorphism from the get go.

packages/core/src/schema/models/connectedEvaluations.ts
geclos: When working with tables directly, it is much easier to work with a single colum id number than having to work with two columns ids to find stuff. I know it seems a small thing but it becomes a pita when you are debugging production issues for a user for a whole day.

packages/core/src/schema/index.ts
```diff
@@ -17,3 +17,8 @@ export * from './models/providerApiKeys'
 // Log tables
 export * from './models/documentLogs'
 export * from './models/providerLogs'
+
+// Evaluations tables
+export * from './models/evaluations'
+export * from './models/connectedEvaluations'
+export * from './models/evaluationResults'
```
andresgutgon, line 24: Remember export types

packages/core/src/schema/models/evaluationResults.ts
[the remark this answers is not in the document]
csansoon: Talked with product. Not something we're going to be adding in short/mid-term. All evaluations will be performed with LLMs for now, even dumber ones.

packages/core/src/schema/index.ts
csansoon: True!

packages/core/drizzle/0030_evaluations_schema.sql
csansoon, line 1:
```
pnpm db:generate --name "custom_name"
```
May help organizing :)

packages/core/src/schema/models/connectedEvaluations.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { documentVersions, latitudeSchema } from '..'
+import { timestamps } from '../schemaHelpers'
+import { evaluations } from './evaluations'
+
+export const connectedEvaluations = latitudeSchema.table(
+  'connected_evaluations',
+  {
+    id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
+    documentUuid: uuid('document_uuid').notNull(),
```
csansoon, line 12: Had to remove the foreign key from `documentUuid`, as it is not a unique/pk field in the referenced table.

packages/core/src/schema/models/evaluationTemplates.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const evaluationTemplates = latitudeSchema.table(
+  'evaluations_templates',
+  {
+    id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
+    name: varchar('name', { length: 256 }).notNull(),
+    description: text('description').notNull(),
+    prompt: text('prompt').notNull(),
+    ...timestamps(),
+  },
```
csansoon, line 14: **Evaluation Templates** — Library of our predefined evaluation templates. Users can either create a new evaluation from scratch, or just select one from our library.

andresgutgon: How do we track template changes over time? Should we have template versions? 

(I'm joking just wanted to annoy you)

andresgutgon (approved):
I think is fine.

PR #119, opened 2024-08-19
Create new evaluations - core

csansoon:
Added basic services, factories and tests for the creation of Evaluations and Evaluation Results.


packages/core/src/services/evaluations/importTemplate.ts
```diff
@@ -0,0 +1,33 @@
...
+type Props = {
+  workspace: Partial<Workspace>
+  templateId: number
+}
+export async function importEvaluationTemplate(
+  { workspace, templateId }: Props,
+  db = database,
+) {
+  const templateResult = await findEvaluationTemplateById(templateId)
+  if (!templateResult.ok) return templateResult
```
andresgutgon, line 16: We usually use `result.error` more clear without negation

PR #121, opened 2024-08-20
Create evaluations UI

csansoon:
UI for creating new evaluations. Managing, modifying and testing evaluations will be a different PR.

https://github.com/user-attachments/assets/0cb4a460-16fb-4442-8320-0b8e7b37d513


packages/core/src/schema/models/evaluationTemplates.ts
```diff
@@ -9,6 +9,7 @@ export const evaluationTemplates = latitudeSchema.table(
     id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
     name: varchar('name', { length: 256 }).notNull(),
     description: text('description').notNull(),
+    category: varchar('category', { length: 256 }).notNull(),
```
andresgutgon, line 12: Are we sure we don't want to have some categories table to make filters easier?

csansoon: Could do, but that would add `JOIN` statements to the selects. I do like it more as foreign keys, but I'm not sure if we really do care about it, given that it affects (little) to the query performance.

andresgutgon: I mean unstructured category not sure if it's that helpful. Also how is category loaded? Maybe is not necessary on evaluation detail or can be loaded lazy

And lastly if in the future we want to to have multiple categories it would be easier if this is already a table

csansoon: 👌

PR #122, opened 2024-08-21
Implement deploy of gateway API and Redis Workers apps 

andresgutgon:
# What?
In this PR we setup the Dockerfiles for building the nodejs images of these 2 apps that were missing in fly for testing e2e all the flow of creating prompts and running the AI

### TODO
- [x] Pass gateway host and port configuration to `apps/web` to use SDK
- [x] Fix `@latitude-data/compiler` build.
- [x] Implement Dockerfile for `apps/gateway`
- [x] Implement `Dockerfile` for `apps/workers` (The jobs server)
- [x] QA `apps/gateway` with their `docker/run-production.sh`
- [x] QA `apps/workers` with their `docker/run-production.sh`
~~- [ ] Implement fly deploy for `apps/gateway`~~
~~- [ ] Implement fly deploy for `packages/workers`~~

## About monorepo dependencies
Why the `*` in the package's version in all the internal packages [this one](https://www.reddit.com/r/webdev/comments/146rzbh/monorepo_dependency_management_pnpm_turborepo/) and [this one](https://dev.to/adamgoth/understanding-package-dependencies-within-a-pnpm-monorepo-19ge)


apps/gateway/src/common/env.ts
```diff
@@ -1,5 +1,3 @@
-import '@latitude-data/env'
```
andresgutgon, line 1: We don't need this package for now here

packages/core/src/services/documents/recomputeChanges/getHeadDocumentsAndDraftDocuments.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+async function getDocumentsAtCommit(
+  { commit, workspaceId }: { commit: Commit; workspaceId: number },
+  tx = database,
+) {
+  const projectResult = await getProjectFromCommit({ commit, workspaceId }, tx)
+  if (projectResult.error) return projectResult
+
+  const commitsScope = buildCommitsScope(workspaceId, tx)
```
andresgutgon, line 37: Decouple commits scope from CommitsRepository. 
This file is used inside `recomputeChanges` that is used in `CommitsRepository` resulting in Circular dependency warnings in Rollup ♻️

apps/gateway/docker/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+  --frozen-lockfile \
+  --filter "@latitude-data/gateway..."
+
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
+  BUILDING_CONTAINER=true \
+  pnpm turbo build --filter='@latitude-data/gateway...'
+
+FROM builder AS prod-deps
+WORKDIR /app
+# pnpm deploy: https://pnpm.io/cli/deploy
```
andresgutgon, line 42: This is awesome

apps/gateway/package.json
```diff
@@ -1,33 +1,38 @@
 {
   "name": "@latitude-data/gateway",
-  "type": "module",
+  "version": "1.0.0",
+  "description": "Gateway for Latitude Data",
+  "private": true,
   "scripts": {
-    "build": "tsc --build tsconfig.prod.json --verbose",
+    "build": "tsup --config tsup.config.ts",
```
andresgutgon, line 7: `tsup` is a real bundler that allow us to transpile internal packages and mark the rest as externals

package.json
```diff
@@ -29,7 +29,30 @@
...
+    "zod": "^3.23.8",
+    "acorn": "^8.9.0",
+    "code-red": "^1.0.3",
+    "locate-character": "^3.0.0",
+    "yaml": "^2.4.5",
+    "dotenv": "^16.4.5",
+    "bullmq": "^5.8.3",
+    "ioredis": "^5.4.1",
+    "zod": "^3.23.8",
+    "@faker-js/faker": "^8.4.1"
```
andresgutgon, line 55: We treat root package.json like a monolitic Rails app `Gemfile`. I think this can work. Explained in the PR description

packages/core/package.json
```diff
@@ -22,32 +22,36 @@
...
+    "lodash-es": "*",
+    "uuid": "*",
+    "zod": "*"
   },
   "devDependencies": {
     "@faker-js/faker": "^8.4.1",
     "@latitude-data/eslint-config": "workspace:*",
     "@latitude-data/typescript-config": "workspace:*",
+    "@rollup/plugin-alias": "^5.1.0",
+    "@rollup/plugin-typescript": "^11.1.6",
```
andresgutgon, line 46: Undo this. I think is not needed

packages/core/rollup.config.mjs
andresgutgon, line 1: Undo this I think is not needed because core is an internal package that's build by the `apps/*`

packages/core/src/repositories/commitsRepository/index.ts
```diff
@@ -1,22 +1,19 @@
...
   ModifiedDocumentType,
 } from '$core/constants'
-import { NotFoundError, Result } from '$core/lib'
-import { commits, projects } from '$core/schema'
-import { recomputeChanges, RecomputedChanges } from '$core/services'
-import { assertCommitIsDraft } from '$core/services/documents/utils'
+import { InferedReturnType, NotFoundError, Result } from '$core/lib'
+import { assertCommitIsDraft } from '$core/lib/assertCommitIsDraft'
+import { buildCommitsScope } from '$core/repositories/commitsRepository/utils/buildCommitsScope'
+import { getHeadCommitForProject } from '$core/repositories/commitsRepository/utils/getHeadCommit'
```
andresgutgon, line 11: These changes were done because when doing the rollup it complained about circular dependencies. Although I don't need to do the build of core I would keep these changes because at the end I managed to generate a clean rollup build without Circular dependencies warnings.

packages/core/src/repositories/commitsRepository/index.ts
```diff
@@ -38,7 +35,7 @@ function filterByStatusQuery({
   status,
 }: {
   status: CommitStatus
-  scope: typeof CommitsRepository.prototype.scope
+  scope: InferedReturnType<typeof buildCommitsScope>
```
andresgutgon, line 38: Infer return of a function. Very handy in this case so we can type scopes

packages/core/src/schema/models/commits.ts
```diff
@@ -44,14 +46,3 @@ export const commits = latitudeSchema.table(
...
-export const commitRelations = relations(commits, ({ one }) => ({
-  project: one(projects, {
-    fields: [commits.projectId],
-    references: [projects.id],
-  }),
-  user: one(users, {
-    fields: [commits.userId],
-    references: [users.id],
-  }),
-}))
```
andresgutgon, line 57: Circular dependencies solved by moving all relations to one file

packages/core/src/services/ai/estimateCost/index.ts
```diff
@@ -1,4 +1,4 @@
-import { Providers } from '@latitude-data/core'
```
andresgutgon, line 1: 🙏 Let's try not to use package name inside the same package

packages/core/src/services/documentLogs/create.ts
```diff
@@ -1,10 +1,7 @@
-import {
-  database,
-  documentLogs,
-  Result,
-  Transaction,
-} from '@latitude-data/core'
```
andresgutgon, line 6: Same core inside core ⛔

packages/jobs/src/queues/index.ts
```diff
@@ -34,7 +34,11 @@ function setupQueue({
         params: JobDefinition[typeof name][typeof jobName]['data'],
         options?: JobsOptions,
       ) => {
-        return queue.add(jobName, params, options)
+        try {
+          queue.add(jobName, params, options)
+        } catch (e) {
+          console.error('ERROR', e)
+        }
```
andresgutgon, line 41: ⚠️ Undo. Is to check something

packages/jobs/src/job-definitions/providerLogs/createJob.ts
```diff
@@ -6,5 +6,6 @@ export type CreateProviderLogJobData = CreateProviderLogProps
 export const createProviderLogJob = async (
   job: Job<CreateProviderLogJobData>,
 ) => {
+  console.log('createProviderLogJob', job.data)
```
andresgutgon, line 9: 🔥

packages/jobs/src/job-definitions/documentLogs/createJob.ts
```diff
@@ -6,5 +6,6 @@ export type CreateDocumentLogJobData = CreateDocumentLogProps
 export const createDocumentLogJob = async (
   job: Job<CreateDocumentLogJobData>,
 ) => {
+  console.log('createDocumentLogJob', job.data)
```
andresgutgon, line 9: 🔥

apps/gateway/docker/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+ARG HOSTNAME="0.0.0.0"
+ENV HOSTNAME $HOSTNAME
+
+ENV NODE_ENV=production
+
+EXPOSE 8787
+
+# Bcrypt is not compiled for some reason
+# https://github.com/kelektiv/node.bcrypt.js/issues/800
+RUN  npm rebuild bcrypt
```
andresgutgon, line 74: @geclos if you want feel free to look into this

packages/core/src/repositories/commitsRepository/index.ts
geclos: hmm I usually use ReturnType, not sure what's the diff

packages/core/src/services/ai/estimateCost/index.ts
geclos: it all depends on the order of path alias declarations in the tsconfig file. If "@latitude-data/core" path alias is ever declared before "core/*", then the lsp will suggest the first match it finds which is in "@latitude-data/core", changing the path alias declaration order fixes the issue but it's very brittle.

apps/gateway/docker/Dockerfile
geclos: replaced bcrypt with argon2 which is more modern and supports typescript types out of the box

PR #123, opened 2024-08-22
Evaluations editor

csansoon:
<details>
  <summary>Look at this really cool dashboard (click here)</summary>
  image
</details>


packages/compiler/src/compiler/readMetadata.ts
```diff
@@ -64,20 +65,25 @@ export class ReadMetadata {
   constructor({
     document,
     referenceFn,
+    withParameters,
   }: {
     document: Document
     referenceFn?: ReferencePromptFn
-    configSchema?: object
+    withParameters?: string[]
   }) {
```
csansoon, line 73: `ReadMetadata` now accepts a `withParameters` optional prop.
When it is not defined, all variables that are being used in the code but have not been defined will be included in the returned metadata as `parameters`.
When defined, only variables included in this array can be used in the code. If a variable is used but was not previously defined and is not included in the array, an error is added to the returned metadata.

packages/core/src/services/evaluations/compiler/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export async function readMetadataFromEvaluation(
+  evaluation: Evaluation,
+): Promise<TypedResult<ConversationMetadata, LatitudeError>> {
+  const metadata = await readMetadata({
+    prompt: evaluation.prompt,
+    withParameters: Object.keys(PARAMETERS_FROM_LOG),
+  })
+  return Result.ok(metadata)
+}
```
csansoon, line 41: New `readMetadataFromEvaluation` and `buildEvaluationChain` methods have been created to compile using only the log's data as parameters. Which data from the log is used and how to extract it from it is defined in `PARAMETERS_FROM_LOG`

packages/core/src/services/evaluations/compiler/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+    const assistantMessage = {
+      role: MessageRole.assistant,
+      content: providerLog.responseText,
+      toolCalls: providerLog.toolCalls,
+    } as AssistantMessage
+    return [...providerLog.messages, assistantMessage]
+  },
+  response: ({ providerLog }: { providerLog: ProviderLog }) =>
+    providerLog.responseText,
+}
```
csansoon, line 20: `PARAMETERS_FROM_LOGS` currently contains only two parameters: `messages` and `response`.

This means that using any other variable name as if it was a parameter will result in an error.
(Of course, the user can still define their own variables within the code and use it later)

apps/web/src/app/(private)/evaluations/[evaluationUuid]/editor/_components/EvaluationEditor/Editor/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+    setValue(value)
+    debouncedSave(value)
+  }, [])
+
+  useEffect(() => {
+    readMetadata({
+      prompt: value,
+      withParameters: Object.keys(PARAMETERS_FROM_LOG),
+    }).then(setMetadata)
+  }, [value])
```
Comment (csansoon) [inline apps/web/src/app/(private)/evaluations/[evaluationUuid]/editor/_components/EvaluationEditor/Editor/index.tsx:66]: Currently, the Evaluation Editor is using `PARAMETERS_FROM_LOG` directly instead of the premade `readMetadataFromEvaluation` function. This is because this must be run directly from the browser, instead of being a service.

The compiling should be done in the server, though, as it will require running the AI.

packages/core/src/schema/models/evaluations.ts
```diff
@@ -20,25 +24,24 @@ export const evaluations = latitudeSchema.table(
...
     description: text('description').notNull(),
-    prompt: text('prompt').notNull(),
     workspaceId: bigint('workspace_id', { mode: 'number' })
       .notNull()
       .references(() => workspaces.id, { onDelete: 'cascade' }),
-    templateId: bigint('template_id', { mode: 'number' }).references(
-      () => evaluationTemplates.id,
-    ),
+    metadataId: bigint('metadata_id', { mode: 'number' }).notNull(),
+    metadataType: metadataTypesEnum('metadata_type').notNull(),
```
geclos, line 31: I've implemented the polymorphic pointer approach before things get too crazy

packages/core/src/repositories/evaluationsRepository.ts
```diff
@@ -1,25 +1,48 @@
...
+      'metadata_metadata_id',
+    ),
+    ...omit(getTableColumns(llmAsJudgeEvaluationMetadatas), [
+      'id',
+      'metadataType',
+      'createdAt',
+      'updatedAt',
+    ]),
+  },
+}
```
geclos, line 23: since we will always want the metadata with the evaluation i've opted for including it in the scope return value so you can always do `evaluation.metadata` bonus point:  it will force u to use the repository which will enforce tenancy as a nice side effect

packages/core/src/services/evaluations/update.ts
```diff
@@ -0,0 +1,41 @@
...
+    description,
+    metadata = {},
+  }: {
+    evaluation: EvaluationDto
+    name?: string
+    description?: string
+    metadata: Record<string, unknown>
+  },
+  trx = database,
+) {
```
geclos, line 22: @csansoon never pass ids to services that write to db, always full models. The reason is to force the consumer to use repositories and thus enforce tenancy.

apps/web/src/actions/evaluations/create.ts
```diff
@@ -11,7 +12,15 @@ export const createEvaluationAction = authProcedure
...
-      prompt: z.string().optional(),
+      type: z
+        .nativeEnum(EvaluationMetadataType)
+        .optional()
+        .default(EvaluationMetadataType.LlmAsJudge),
+      metadata: z
+        .object({
+          prompt: z.string(),
+        })
+        .optional(),
```
geclos, line 23: from the FE pov evaluations always come with their metadata included, the FE should never have to deal with metadata tables

PR #133, opened 2024-09-04
feat: google AI provider

geclos:
Adds support for Gemini and related models in Latitude

### DIFF
--- a/packages/core/drizzle/meta/0042_snapshot.json
+++ b/packages/core/drizzle/meta/0042_snapshot.json
@@ -1,0 +1,1841 @@
+{
+  "id": "51c93352-5bd5-4667-8a9f-0a97df72c155",
+  "prevId": "63b62478-8b10-4a13-a030-e3344e1db99d",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "latitude.users": {
+      "name": "users",
+      "schema": "latitude",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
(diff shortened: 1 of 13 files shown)


packages/core/src/services/providerLogs/create.ts
```diff
@@ -51,14 +51,7 @@ export async function createProviderLog(
...
       .values({
-        // TODO: Review if wrapping with a `new Date` is necessary.
-        // `generatedAt` is already a `Date` object, so this should work but it doesn't.
-        // I saw this workouround here:
-        // https://github.com/drizzle-team/drizzle-orm/issues/1113#issuecomment-2220076371
-        //
-        // Docs for timestamp with `{ mode: 'date' }`
-        // https://orm.drizzle.team/docs/column-types/pg#timestamp
-        generatedAt: new Date(generatedAt),
+        generatedAt: generatedAt,
```
andresgutgon, line 54: Not necessary?

packages/core/drizzle/meta/0042_snapshot.json
andresgutgon, line 1: This metadata json in drizzle are terrible. I wish there were another way.

packages/core/src/services/providerLogs/create.ts
geclos: no, my suspicion is that since this happened in a job before, whoever implemented this forgot that properties get serialized into serializable values for the job system to work (so dates get transformed into strings). Since this happens in sync now, no need for this.

andresgutgon: That was me. Makes sense

PR #134, opened 2024-09-04
Unify blank slates, change header styles export env package as ESM

andresgutgon:
# What?
☝️ That

I saw these new header style in Figma
image

## Unifiy blank slates
image


geclos (dismissed):
awesome!

PR #137, opened 2024-09-05
Implement datasets data model

andresgutgon:
# What?
☝️ Table for storing data related with uploaded datasets

## What?
- [x] Store file metadata
- [x] Store headers
- [x] Store number of rows
- [x] List of datasets 
- [x] Destroy dataset modal
- [x] Add uploaded by column
- [x] Add Created at column
- [x] Drag & Drop do not detect the file uploaded
- [x] Remove file when dataset is destroyed
- [x] Add index to `workspace_id` and `author_id`. 
- [x] Make unique `datasets.name` in combination with `workspace_id`
- [x] Add CSV delimiter `[',', '\t', ' ', ';']`. We need to know this in order to read the CSV. Nothing is easy 😂 

## Next TODO
- [x] Preview dataset content modal (500 first rows)
- [x] Table loading skeleton


packages/core/src/services/datasets/destroy.ts
```diff
@@ -0,0 +1,31 @@
...
+  {
+    dataset,
+    disk,
+  }: {
+    dataset: Dataset
+    disk: DiskWrapper
+  },
+  db = database,
+) {
+  const deleteResult = await disk.delete(dataset.fileKey)
```
andresgutgon, line 19: This deletes the file in disk (Filesystem or S3 or other clouds). Do you think we should do in another way, or is blocking the request fine?

geclos: fine

andresgutgon: Let me try in production with S3 enabled and we can move it to a job

apps/web/src/actions/datasets/create.ts
```diff
@@ -6,12 +6,6 @@ import { z } from 'zod'
 
 import { authProcedure } from '../procedures'
 
-const ACCEPTED_FILE_TYPES = [
-  'text/csv',
-  'application/vnd.ms-excel',
-  'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
-  'application/vnd.oasis.opendocument.spreadsheet',
-]
 const MAX_SIZE = 3
```
geclos, line 9: 3mb only? that's very small

apps/web/src/stores/datasets.ts
```diff
@@ -0,0 +1,72 @@
...
+      })
+
+      mutate(data.filter((ds) => ds.id === dataset.id))
+    },
+  })
+
+  return {
+    data,
+    mutate,
+    createFormAction,
```
geclos, line 66: you shouldn't use this as it won't be compatible with the json-based server action

packages/core/src/lib/readCsv.ts
```diff
@@ -0,0 +1,34 @@
...
+      delimiter,
+      relax_column_count: true,
+      trim: true,
+      skip_empty_lines: true,
+      columns: true,
+      info: true,
+    }) as {
+      record: Record<string, string>
+      info: { columns: { name: string }[] }
+    }[] // not typed
```
geclos, line 22: is this comment a TODO?

packages/core/src/repositories/datasetsRepository.ts
```diff
@@ -0,0 +1,34 @@
...
+  Dataset
+> {
+  get scope() {
+    const foo = this.db
+      .select(datasetColumns)
+      .from(datasets)
+      .leftJoin(users, eq(users.id, datasets.authorId))
+      .where(eq(datasets.workspaceId, this.workspaceId))
+      .as('datasetsScope')
+    return foo
```
geclos, line 32: hey

packages/core/src/schema/relations.ts
```diff
@@ -29,7 +30,11 @@ export const userRelations = relations(users, ({ many }) => ({
   magicLinkTokens: many(magicLinkTokens),
 }))
 
-export const workspaceRelations = relations(workspaces, ({ many }) => ({
+export const workspaceRelations = relations(workspaces, ({ one, many }) => ({
+  creator: one(users, {
```
geclos, line 34: call it author no? to keep it consistent

packages/core/src/services/datasets/create.ts
```diff
@@ -22,15 +24,47 @@ export const createDataset = async (
     }
     disk: DiskWrapper
```
geclos, line 25: this is called disk but it can be other file storage no? like s3 or others

apps/web/src/actions/datasets/create.ts
andresgutgon: 25mb of text?

geclos (commented):
minor comments 👌🏼

packages/core/src/lib/readCsv.ts
andresgutgon: Removing in other PR I just meant that this package doesn't type return

packages/core/src/schema/models/datasets.ts
```diff
@@ -0,0 +1,23 @@
...
+  workspaceId: bigint('workspace_id', { mode: 'number' })
+    .notNull()
+    .references(() => workspaces.id, { onDelete: 'cascade' }),
+  authorId: text('author_id').references(() => users.id, {
+    onDelete: 'set null',
+  }),
+  fileKey: varchar('file_key', { length: 256 }).notNull(),
+  fileMetadata: jsonb('file_metadata').$type<FileMetadata>().notNull(),
+  ...timestamps(),
+})
```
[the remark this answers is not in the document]
andresgutgon: Good catch 💯

packages/core/src/services/datasets/create.ts
andresgutgon: This is how they called in flydrive 
image

apps/web/src/actions/datasets/create.ts
geclos: well the theory is that they want to evaluate on a large body of data no?, 3mb doesn't push our infra hard at all and it can be limiting for no reason

andresgutgon: 25 then. But node server will handle it? 

We'll have to do this https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#bodysizelimit

Although the best approach would be to direct upload to S3 and then process the file when S3 responds. But we need to change how upload is done

apps/web/src/stores/datasets.ts
andresgutgon: I need this to submit a multi-part form with attachment.

apps/web/src/actions/datasets/create.ts
andresgutgon: I'll start with 15MB and see if someone complains

apps/web/src/stores/datasets.ts
geclos: True. Edge case.

packages/core/drizzle/0044_needy_the_hood.sql
```diff
@@ -0,0 +1,27 @@
...
+--> statement-breakpoint
+DO $$ BEGIN
+ ALTER TABLE "latitude"."datasets" ADD CONSTRAINT "datasets_author_id_users_id_fk" FOREIGN KEY ("author_id") REFERENCES "latitude"."users"("id") ON DELETE set null ON UPDATE no action;
+EXCEPTION
+ WHEN duplicate_object THEN null;
+END $$;
+--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "datasets_workspace_idx" ON "latitude"."datasets" USING btree ("workspace_id");--> statement-breakpoint
+CREATE INDEX IF NOT EXISTS "datasets_author_idx" ON "latitude"."datasets" USING btree ("author_id");--> statement-breakpoint
+CREATE UNIQUE INDEX IF NOT EXISTS "datasets_workspace_id_name_index" ON "latitude"."datasets" USING btree ("workspace_id","name");
```
andresgutgon, line 27: ☝️ Added 3 indexes

1. Workspace
2. Author 
3. Workspace + Name 

I'll add UI validation for workspace + name uniqueness. This was a recommendation from @csansoon . He said it's nice to don't repeat names in datasets and I agree.

PR #143, opened 2024-09-06
Dataset preview data

andresgutgon:
# What?
Preview dataset CSV data 

## TODO
- [ ] Load data from CSV (100 first rows)
- [ ] Table skeleton


packages/core/src/services/datasets/preview.ts
```diff
@@ -0,0 +1,21 @@
+import { Dataset } from '../../browser'
+import { DiskWrapper } from '../../lib/disk'
+import { syncReadCsv } from '../../lib/readCsv'
+
+/**
+ * This service pick the first N rows of a CSV file
```
andresgutgon, line 6: csv lib we use also has `from` so we can implement some kind of pagination in the future

apps/web/src/actions/datasets/create.ts
```diff
@@ -1,41 +1,88 @@
...
+              },
+              {
+                message:
+                  'This name was already used, plese use something different',
+              },
+            ),
+          csvDelimiter: z.enum(DELIMITERS_KEYS, {
+            message: 'Choose a valid delimiter option',
+          }),
+          csvCustomDelimiter: z.string(),
```
csansoon, line 49: Shouldn't `csvCustomDelimiter` be `optional`?

andresgutgon: It's ok, we always send an empty string because the field is always sent. Is in the validation that we check user picked custom and this field has content

csansoon (approved):
minor comments

apps/web/src/actions/datasets/create.ts
```diff
@@ -1,41 +1,88 @@
...
+] as const
+const DELIMITER_VALUES = {
+  comma: ',',
+  semicolon: ';',
+  tab: '\t',
+  space: ' ',
+}
+
 const MAX_SIZE = 3
 const MAX_UPLOAD_SIZE_IN_MB = 3 * 1024 * 1024
```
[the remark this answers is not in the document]
andresgutgon: No good catch 💯 . This needs to be changed to 15MB now. I forgot thanks

PR #147, opened 2024-09-09
Infra for setting S3 bucket for storing datasets

andresgutgon:
# What?
Set S3 bucket for storing datasets in latitude workspaces

### TODO
- [x] Setup Pulumi config and AWS secrets for AWS access and secret 
- [x] ~~Setup S3 Policy in bucket~~ Not necessary. User from `AWS_ACCESS_KEY` already has these permissions.
- [x] Setup bucket in the `eu-central-1` region
- [x] Run `pulumi up` to create the bucket
- [x] Ask Gerard how we can avoid removing the bucket even if we run pulumi destroy

## Next PR
- [ ] Test the files are uploaded to S3 in development by setting the Driver as S3. Not working doing the fix in another PR to merge the Pulumi state


packages/core/src/lib/disk.ts
```diff
@@ -1,24 +1,42 @@
+import { debuglog } from 'node:util'
```
andresgutgon, line 1: This is super nice. It prints debug traces scoped by a string. I saw Flydrive used it

packages/core/src/lib/disk.ts
```diff
@@ -1,24 +1,42 @@
...
 
+import { HeadBucketCommand, S3Client } from '@aws-sdk/client-s3'
 import { Result } from '@latitude-data/core/lib/Result'
 import { env } from '@latitude-data/env'
 import { Disk, errors } from 'flydrive'
 import { FSDriver } from 'flydrive/drivers/fs'
 import { S3Driver } from 'flydrive/drivers/s3'
 import { WriteOptions } from 'flydrive/types'
 
+var debug_default = debuglog('flydrive:s3')
```
andresgutgon, line 12: In this case I tag these logs (temp hack to debug) as FlyDrive S3

geclos: var? are we back to 2006 javascript? :trollface:

packages/core/src/lib/disk.ts
```diff
@@ -39,6 +57,27 @@ export class DiskWrapper {
...
+    const input = {
+      Bucket: env.S3_BUCKET,
+      ExpectedBucketOwner: 'TODO_FIND_ID',
+    }
+    const command = new HeadBucketCommand(input)
+    try {
+      await client.send(command)
+    } catch (error) {
+      debug_default('error pinging bucket %s', bucket)
+      debug_default('error pinging bucket %s', error)
```
geclos, line 77: shouldn't we throw or do something with the exception here besides printing?

geclos (dismissed):

packages/core/src/lib/disk.ts
geclos: i think carlos hasn't seen a var in real life, too young

apps/web/src/services/auth/index.ts
```diff
@@ -3,7 +3,12 @@ import { sessions, users } from '@latitude-data/core/schema'
 import { DrizzlePostgreSQLAdapter } from '@lucia-auth/adapter-drizzle'
 import { Lucia } from 'lucia'
 
-const adapter = new DrizzlePostgreSQLAdapter(database, sessions, users)
+const adapter = new DrizzlePostgreSQLAdapter(
+  // @ts-expect-error - No idea why this is happening
+  database,
```
andresgutgon, line 8: No idea why started failing. Also in `main`. I'll try after merge and see if I can remove it

PR #154, opened 2024-09-09
Feature: Run evaluations in batch

geclos:
This commit implements evaluation runs in batch.


apps/web/src/actions/datasets/preview.ts
```diff
@@ -2,10 +2,10 @@
 
 import { DatasetsRepository } from '@latitude-data/core/repositories'
 import { previewDataset } from '@latitude-data/core/services/datasets/preview'
-import disk from '$/lib/disk'
 import { z } from 'zod'
 
 import { authProcedure } from '../procedures'
+import { disk } from '@latitude-data/core/lib/disk'
```
andresgutgon, line 8: Joder : /

packages/core/src/lib/disk.ts
```diff
@@ -147,3 +151,15 @@ export class DiskWrapper {
...
 }
+
+const PUBLIC_PATH = 'uploads'
+
+export const disk = new DiskWrapper({
+  local: {
+    publicPath: PUBLIC_PATH,
+    location: path.join(
+      path.dirname(fileURLToPath(import.meta.url)),
+      `../../public/${PUBLIC_PATH}`,
```
andresgutgon, line 162: This path is ok now? Before was `apps/web/public/uploads`

packages/jobs/src/job-definitions/batchEvaluations/runBatchEvaluationJob.ts
```diff
@@ -0,0 +1,88 @@
...
+  } = job.data
+  const workspace = await findWorkspaceFromDocument(document)
+  if (!workspace) throw new NotFoundError('Workspace not found')
+
+  const commit = await new CommitsRepository(workspace.id)
+    .find(document.commitId)
+    .then((r) => r.unwrap())
+  const fileMetadata = dataset.fileMetadata
+  const runCount = jobRunCount || fileMetadata.rowCount
+  const result = await previewDataset({
```
andresgutgon, line 49: CSV parser has a stream version. I think we should use that version here to avoid loading all the file in memory 
https://csv.js.org/parse/api/stream/

packages/jobs/src/job-definitions/batchEvaluations/runDocumentJob.ts
```diff
@@ -0,0 +1,54 @@
...
+  batchId: string
+}
+
+export const runDocumentJob = async (job: Job<RunDocumentJobData>) => {
+  const { workspaceId, document, commit, parameters, evaluationId, batchId } =
+    job.data
+
+  const progressTracker = new ProgressTracker(connection, batchId)
+
+  try {
```
andresgutgon, line 29: Why try/catch if the service is using a result?

geclos: there's 3 awaits that could throw in here, i wanna make sure we handle those

packages/core/src/lib/disk.ts
geclos: hmm i'm not sure this is a valid approach anyways since this file gets bundled in the production bundle in a completely different location so...

andresgutgon: Production doesn't matter if we us S3. But in develop the app using the disk has to have access to the uploads directory. Give a try, just asking maybe this way works

apps/web/src/actions/datasets/preview.ts
```diff
@@ -1,8 +1,8 @@
 'use server'
 
+import { disk } from '@latitude-data/core/lib/disk'
 import { DatasetsRepository } from '@latitude-data/core/repositories'
 import { previewDataset } from '@latitude-data/core/services/datasets/preview'
-import disk from '$/lib/disk'
```
andresgutgon, line 5: What's the reason to do this?. For using in jobs? I think if you manage to make it work is fine. But also fine that each app init their own disk singleton

geclos: we need it in jobs yes

packages/core/src/lib/disk.ts
geclos: it has to work in production OSS product anyone could be using local disk driver in production

apps/web/src/actions/evaluations/runBatch.ts
```diff
@@ -0,0 +1,58 @@
...
+  .input(
+    z.object({
+      datasetId: z.number(),
+      projectId: z.number(),
+      documentUuid: z.string(),
+      commitUuid: z.string(),
+      runCount: z.number(),
+      offset: z.number().optional().default(0),
+      parameters: z.record(z.number()).optional(),
+      evaluationId: z.number(),
```
andresgutgon, line 23: Shouldn't be this an array?

apps/web/src/actions/evaluations/runBatch.ts
```diff
@@ -0,0 +1,58 @@
...
+export const runBatchAction = authProcedure
+  .createServerAction()
+  .input(
+    z.object({
+      datasetId: z.number(),
+      projectId: z.number(),
+      documentUuid: z.string(),
+      commitUuid: z.string(),
+      runCount: z.number(),
+      offset: z.number().optional().default(0),
```
andresgutgon, line 21: Also limit no?

andresgutgon: And also a boolean to pick all the rows
image

apps/web/src/actions/evaluations/runBatch.ts
geclos: yes

apps/web/src/actions/evaluations/runBatch.ts
geclos: yes

geclos: it's gonna be fromLine and toLine btw

geclos: > And also a boolean to pick all the rows

better to handle this in frontend, don't send the limits to the backend if the users wants everything

apps/web/src/actions/evaluations/runBatch.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const runBatchAction = authProcedure
+  .createServerAction()
+  .input(
+    z.object({
+      datasetId: z.number(),
+      projectId: z.number(),
+      documentUuid: z.string(),
+      commitUuid: z.string(),
+      runCount: z.number(),
```
andresgutgon, line 20: What is this?

apps/web/src/actions/evaluations/runBatch.ts
```diff
@@ -0,0 +1,60 @@
...
+      offset: z.number().optional().default(0),
+      parameters: z.record(z.number()).optional(),
```
andresgutgon, line 22: Why is this a record of numbers?

geclos: it maps parameters to the index column of rows in a csv, are you doing it differently in frontend?

andresgutgon: But you also need the name of the parameters in the documents no? It's a map between document params and csv headers or I'm missunderstanding?
image

geclos: yes the key is the parameter name and the value is the column index

geclos: so in this case let's assume the csv has last_name and name in columns 3 and 4, the parameters map would be

{ last_name: 3, name: 4} the keys being the document parameters and the values being the csv collumn indeces

andresgutgon: It's true, headers is an array 👍

PR #167, opened 2024-09-12
feature: evaluation result types and configuration

geclos:
Evaluations/templates now have a configuration to configure the type of result we are expecting, stored in a jsonb column. EvaluationResults now have a polymorphic table to store the result depending on its type – This is so that we can aggregate results in SQL later on.

### DIFF
--- a/apps/web/src/actions/evaluations/create.ts
+++ b/apps/web/src/actions/evaluations/create.ts
@@ -1,6 +1,9 @@
 'use server'
 
-import { EvaluationMetadataType } from '@latitude-data/core/browser'
+import {
+  EvaluationMetadataType,
+  EvaluationResultableType,
+} from '@latitude-data/core/browser'
 import { createEvaluation } from '@latitude-data/core/services/evaluations/create'
 import { z } from 'zod'
 
@@ -16,6 +19,12 @@ export const createEvaluationAction = authProcedure
         .nativeEnum(EvaluationMetadataType)
         .optional()
         .default(EvaluationMetadataType.LlmAsJudge),
+      configuration: z.object({
+        type: z.nativeEnum(EvaluationResultableType),
+        detail: z
+          .object({ range: z.object({ from: z.number(), to: z.number() }) })
+          .optional(),
+      }),
       metadata: z
         .object({
           prompt: z.string(),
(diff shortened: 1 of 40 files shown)


apps/web/src/actions/evaluations/create.ts
```diff
@@ -25,11 +34,14 @@ export const createEvaluationAction = authProcedure
     { type: 'json' },
   )
   .handler(async ({ input, ctx }) => {
+    console.log(input)
+
```
csansoon, line 38: 🔥

packages/core/src/repositories/connectedEvaluationsRepository/getConnectedDocumentsWithMetadata.test.ts
```diff
@@ -185,7 +185,7 @@ describe('getConnectedDocumentsWithMetadata', () => {
     expect(result2.length).toEqual(1)
   })
 
-  it('does not return removed documents', async () => {
+  it.only('does not return removed documents', async () => {
```
csansoon, line 188: 🔥

apps/web/src/app/(private)/evaluations/_components/CreateEvaluationModal/index.tsx
```diff
@@ -102,7 +110,103 @@ export default function CreateEvaluationModal({
...
+                        },
+                      }
+                    }
+
+                    return next
+                  })
+                }
+              />
+            </div>
+          </FormField>
```
andresgutgon, line 208: HUGE onChange. Maybe this could be a hook
```javascript
const { onMinInputChange, onMaxInputChange } = useInputRangeValidation()
```

packages/jobs/vitest.config.mjs
```diff
@@ -0,0 +1,12 @@
...
+
+export default defineConfig({
+  test: {
+    globals: true,
+    testTimeout: 5000,
+    environment: 'node',
+    setupFiles: ['./tests/setup.ts'],
+    include: ['./src/**/*.test.ts'],
+  },
+})
```
andresgutgon, line 12: No tests in jobs no?

geclos: yes I added some a few days ago

andresgutgon (dismissed):

PR #171, opened 2024-09-13
Setup websockets communication between web and workers

andresgutgon:
# What?
We want to keep the client informed when evaluations are running. How many logs are created, and how many logs fail
This PR set up a new Express server that set up a Socket.IO websocket server with 2 endpoints /websocket and worker-websocket. The one for the web has secure cookies token auth and the one for workers share a secret token between Sockets server and workers.

I think this setup makes sense

https://github.com/user-attachments/assets/c38bd897-7871-42ee-ac0c-1945c2562a10

## TODO
- [x] Setup Sockets server
- [x] Setup setting token and refresh token cookies in `web` when user login with magic link.
- [x] Setup Sockets client in worker so it can emit events to the websockets server
- [x] Setup DOMAIN_NAME env variable.
- [x] Setup `<WebsocketsProvider />` in React client so it can start listening to events
- [x] Make Websocket connection work `WEB`
- [x] Make Websocket connection work `WORKERS`
- [x] Run evaluation and see if it triggers a websocket to the frontend
- [x] Build UI to show evaluation progress.
- [x] WHY first time a evaluation is run `total` is 0? REVIEW this
- [x] Setup refresh websocket token server action
- [x] Review ENV variables to avoid breaking deploy with this PR
- [x] Implement `logout` lol we didn't have it
- [x] Delete Websocket cookies when `logout`

## Next PR
- [x] Review all necessary env variables
- [ ] Think about `ws` or `wss` (Secure). Maybe is not necessary because is behind a proxy cc @geclos 
- [ ] Talk about private/public keys by workspace cc @geclos 
- [ ] @socket.io/redis-adapter for pub/sub with Redis. Maybe? I don't think we need now if it works without it cc @geclos


packages/jobs/src/job-definitions/batchEvaluations/runEvaluationJob.ts
```diff
@@ -45,6 +45,18 @@ export const runEvaluationJob = async (job: Job<RunEvaluationJobData>) => {
     }
 
     await progressTracker.incrementErrors()
+  } finally {
     await progressTracker.decrementTotal()
```
andresgutgon, line 49: decrementTotal was only the catch but I think is something we want to do always no?

apps/web/package.json
```diff
@@ -27,10 +27,12 @@
     "@sentry/nextjs": "^8",
     "@t3-oss/env-nextjs": "^0.10.1",
     "ai": "^3.2.42",
+    "socket.io-react-hook": "^2.4.5",
```
andresgutgon, line 30: Not very complicated to integrate but this package is small and nice

apps/web/package.json
```diff
@@ -42,6 +43,8 @@
     "nprogress": "^0.2.0",
     "react": "19.0.0-rc-f994737d14-20240522",
     "react-dom": "19.0.0-rc-f994737d14-20240522",
+    "socket.io-client": "^4.7.5",
```
andresgutgon, line 46: 🔥 Included in the hook

apps/websockets/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "cookie-parser": "^1.4.6",
+    "express": "^4.21.0",
+    "socket.io": "^4.7.5"
+  },
+  "devDependencies": {
+    "@latitude-data/eslint-config": "workspace:^",
+    "@latitude-data/typescript-config": "workspace:^",
+    "@types/cookie-parser": "^1.4.7",
+    "@types/express": "^4.17.21",
+    "@types/jsonwebtoken": "^9.0.6",
```
andresgutgon, line 28: 🔥

apps/workers/src/server.ts
```diff
@@ -6,7 +6,7 @@ const workers = setupWorkers()
 
 console.log('Workers started')
 
-const port = 3000
+const port = 3001
```
andresgutgon, line 9: This was conflicting with apps/web @geclos

packages/env/src/index.ts
```diff
@@ -66,9 +71,18 @@ export const env = createEnv({
...
+    APP_DOMAIN: z.string(),
+    APP_SUBDOMAIN: z.string().optional(),
   },
   runtimeEnv: {
     ...process.env,
+    WEBSOCKETS_SERVER:
+      process.env.WEBSOCKETS_SERVER ?? 'https://ws.latitude.so',
+    APP_DOMAIN: process.env.APP_DOMAIN ?? 'latitude.so',
+    APP_SUBDOMAIN: process.env.APP_SUBDOMAIN ?? 'app',
     FILE_PUBLIC_PATH: process.env.FILE_PUBLIC_PATH ?? FILE_PUBLIC_PATH,
```
andresgutgon, line 86: FAKE SECRETS in this PR. Not necessary until we deploy it

packages/core/src/websockets/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+  const secret = new TextEncoder().encode(env.WEBSOCKET_SECRET_TOKEN_KEY)
+  const token = await new SignJWT(payload)
+    .setProtectedHeader({ alg: 'HS256' })
+    .setExpirationTime(config.maxAge.stringValue)
+    .sign(secret)
+  return {
+    token,
+    cookiesOptions: {
+      secure: isProd,
+      domain: isProd ? `.${env.APP_DOMAIN}` : 'localhost',
```
andresgutgon, line 25: CORS for all the subdomains of the app domain. This way we can share cookies between web client and websockets server as auth method

apps/web/src/env.ts
```diff
@@ -23,5 +24,6 @@ export default createEnv({
     GATEWAY_HOSTNAME: process.env.GATEWAY_HOSTNAME,
     GATEWAY_PORT: process.env.GATEWAY_PORT,
     GATEWAY_SSL: process.env.GATEWAY_SSL,
+    WEBSOCKETS_SERVER: process.env.WEBSOCKETS_SERVER,
```
geclos, line 27: no hay default aqui

apps/websockets/src/server.ts
```diff
@@ -0,0 +1,158 @@
...
+
+const isProd = process.env.NODE_ENV === 'production'
+const server = http.createServer(app)
+const subdomain = env.APP_SUBDOMAIN ? `${env.APP_SUBDOMAIN}.` : ''
+const io = new Server(server, {
+  path: '/websocket',
+  cors: {
+    origin: isProd
+      ? `https://${subdomain}${env.APP_DOMAIN}`
+      : 'http://localhost:3000',
```
geclos, line 47: env var instead of isprod check would be cooler

apps/websockets/src/server.ts
```diff
@@ -0,0 +1,158 @@
...
+
+const app = express()
+app.use(cookieParser())
+app.get('/health', (_, res) => {
+  res.json({ status: 'Websockets server running' })
+})
+
+const isProd = process.env.NODE_ENV === 'production'
+const server = http.createServer(app)
+const subdomain = env.APP_SUBDOMAIN ? `${env.APP_SUBDOMAIN}.` : ''
```
geclos, line 41: we have LATITUDE_URL and LATITUDE_DOMAIN env vars can we work with those?

packages/jobs/src/job-definitions/batchEvaluations/runEvaluationJob.ts
geclos: hmm no we only wanna do it when there's an error, if we succeed we don't wanna decrement the total

apps/web/src/env.ts
andresgutgon: It's on `packages/env/src/index.ts`

geclos (approved):

PR #197, opened 2024-09-17
feature: api key settings section

geclos:


apps/web/src/presenters/apiKeyPresenter.ts
andresgutgon, line 1: Empty file

packages/core/src/services/apiKeys/destroy.ts
```diff
@@ -0,0 +1,14 @@
+import { eq } from 'drizzle-orm'
+
+import { ApiKey } from '../../browser'
+import { database } from '../../client'
+import { Result, Transaction } from '../../lib'
+import { apiKeys } from '../../schema'
+
+export async function destroyApiKey(apiKey: ApiKey, db = database) {
+  return Transaction.call(async (trx) => {
+    await trx.delete(apiKeys).where(eq(apiKeys.id, apiKey.id))
```
andresgutgon, line 10: You check in frontend. But in backend do we check if it's last API key?

geclos: will add it a future pr

Issue #180, opened 2024-09-16
The range selector when creating a batch evaluation doesn't seem to have an effect


cesr:
I tried to use only 10 rows but it used all of the rows in the dataset.

PR #202, opened 2024-09-17
Fix batch range line selector and also allow to create logs on draft commits

andresgutgon:
# What?
Line picker selector was not working properly
Issue: https://github.com/latitude-dev/latitude-llm/issues/180

### 🐛 Evaluation logs not created on DRAFT commits
While fixing the batch line range selector we saw that evaluation logs where not created when in a draft commit. The reason is that we were enqueueing jobs with `document.commitId` instead of draft commit id. Now we pass `commitUuid` that is in the URL of the draft commit.

## What?
- [x] Fix draft commits not creating evaluation logs
- [x] Fix batch range line selector

### DIFF
--- a/apps/web/src/components/EditorHeader/index.tsx
+++ b/apps/web/src/components/EditorHeader/index.tsx
@@ -53,12 +53,14 @@ export default function EditorHeader({
   metadata,
   onChangePrompt,
   rightActions,
+  disabledMetadataSelectors = false,
 }: {
   title: string
   metadata: ConversationMetadata | undefined
   prompt: string
   onChangePrompt: (prompt: string) => void
   rightActions?: ReactNode
+  disabledMetadataSelectors?: boolean
 }) {
   const promptMetadata = useMemo<PromptMeta>(() => {
     const config = metadata?.config
@@ -204,13 +206,16 @@ export default function EditorHeader({
         <Select
           name='provider'
           label='Provider'
-          placeholder='Select a provider'
+          placeholder='Using default provider'
           options={providerOptions}
           value={selectedProvider}
+          disabled={
+            disabledMetadataSelectors || isLoading || !providerOptions.length
+          }
           onChange={onSelectProvider}
         />
         <Select
-          disabled={!selectedProvider}
+          disabled={disabledMetadataSelectors || !selectedProvider}
           name='model'
           label='Model'
           placeholder='Select a model'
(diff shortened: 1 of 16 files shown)


packages/env/src/index.ts
```diff
@@ -37,7 +37,7 @@ if (environment !== 'production') {
       WEBSOCKET_REFRESH_SECRET_TOKEN_KEY: 'refresh-refresh-token-key',
       WORKERS_WEBSOCKET_SECRET_TOKEN: 'workers-secret-token',
       DRIVE_DISK: 'local',
-      DEFAULT_PROVIDER_ID: '1',
+      DEFAULT_PROVIDER_ID: '10000000000000000000', // Crazy number to avoid conflicts
```
csansoon, line 40: What about -1?

andresgutgon: Not sure if it's a good idea to pass to postgresql a -1

csansoon (dismissed):

apps/web/src/components/EditorHeader/index.tsx
```diff
@@ -204,13 +206,16 @@ export default function EditorHeader({
         <Select
           name='provider'
           label='Provider'
-          placeholder='Select a provider'
+          placeholder='Using default provider'
           options={providerOptions}
           value={selectedProvider}
+          disabled={
+            disabledMetadataSelectors || isLoading || !providerOptions.length
+          }
```
geclos, line 214: does the placeholder show when disabled? we should tell the user they are using the default provider if none selected

packages/jobs/src/job-definitions/batchEvaluations/runBatchEvaluationJob.ts
```diff
@@ -43,7 +47,7 @@ export const runBatchEvaluationJob = async (
   if (!workspace) throw new NotFoundError('Workspace not found')
 
   const commit = await new CommitsRepository(workspace.id)
-    .find(document.commitId)
```
geclos, line 46: I don't get this, how can document.commitid be different than commitUuid? the documentversion always poins to its commit no?

geclos (approved):

apps/web/src/components/EditorHeader/index.tsx
andresgutgon: If non selected it says "default provider" yes

packages/jobs/src/job-definitions/batchEvaluations/runBatchEvaluationJob.ts
andresgutgon: Before it was the commitId the document was created (live). Now is the commit that the user is seeing. live or a draft commit. So, the logs are attached to that commit. Which is what we want

PR #204, opened 2024-09-17
Import default project

csansoon:
Import a project by default when creating a Workspace


packages/core/src/services/projects/import.ts
```diff
@@ -0,0 +1,76 @@
...
+    user,
+  }: {
+    workspace: Workspace
+    user: User
+  },
+  db = database,
+) {
+  if (!('DEFAULT_PROJECT_ID' in env) || env.DEFAULT_PROJECT_ID === undefined) {
+    return Result.ok(null)
+  }
```
csansoon, line 26: Just do nothing if there's no DEFAULT_PROJECT_ID env variable.

packages/core/src/services/projects/import.ts
```diff
@@ -0,0 +1,76 @@
...
+  {
+    workspace,
+    user,
+  }: {
+    workspace: Workspace
+    user: User
+  },
+  db = database,
+) {
+  if (!('DEFAULT_PROJECT_ID' in env) || env.DEFAULT_PROJECT_ID === undefined) {
```
geclos, line 24: I mean just make the env var mandatory please 😂 all this null checking is unnecessary

packages/core/src/services/projects/import.ts
```diff
@@ -0,0 +1,76 @@
...
+    return Result.ok(null)
+  }
+
+  const defaultProjectId = env.DEFAULT_PROJECT_ID
```
geclos, line 28: needless assignemnt

packages/core/src/services/projects/import.ts
```diff
@@ -0,0 +1,76 @@
...
+    })
+  if (defaultDocuments.error) return defaultDocuments
+
+  return Transaction.call<Project>(async (tx) => {
+    const project = await createProject(
+      {
+        workspace,
+        user,
+        name: defaultProject!.name,
+      },
```
geclos, line 49: if this project needs to be created with a merged commit you need to pass a mergedAt option here

packages/core/src/services/projects/import.ts
geclos: wait no actually it's fine that is optional since the OSS can't depend on our db state

packages/env/src/index.ts
```diff
@@ -41,6 +41,7 @@ if (environment !== 'production') {
       DEFAULT_PROVIDER_MODEL: 'gpt-4o',
       FILE_PUBLIC_PATH,
       FILES_STORAGE_PATH,
+      DEFAULT_PROJECT_ID: '1',
```
andresgutgon, line 44: We should do a seed data. Otherwise what happen when you reset dev database?

packages/core/src/services/projects/import.ts
csansoon: I'm using the `createNewDocument` service, which requires an unmerged commit. I'm merging it afterwards

packages/core/src/services/projects/import.ts
```diff
@@ -0,0 +1,76 @@
...
+      tx,
+    ).then((r) => r.unwrap())
+
+    const commit = await createCommit({
+      project,
+      user,
+      data: { title: 'Initial version' },
+      db: tx,
+    }).then((r) => r.unwrap())
```
geclos, line 58: pa que if the project already gets created with a commit

PR #209, opened 2024-09-18
Fix YAML compiler

csansoon:
The config section in documents is written using YAML. Currently, its parsing has some quirks:
- It must be at the top of the document, not even linebreaks or comments are allowed
- If any part of the config is not parseable (this is really common mid-editing it), the whole section appears as an error
- It does not show whether the config values are valid or not.

This came with some bugs too:
 - Adding a comment above the config section seemed to work, but modifying it would crash the editor
 - Referencing another document with a config section that had comments on top would import the config as if it was plain text
 - some more

This PR fixes all of that :)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/index.tsx
```diff
@@ -119,12 +122,30 @@ export default function DocumentEditor({
...
+    const providerSchema = z
+      .string({
+        required_error: `You must select a provider.\nFor example: 'provider: ${availableProviders[0] ?? '<your-provider-name>'}'`,
+      })
+      .refine((p) => availableProviders.includes(p), {
+        message:
+          availableProviders.length === 0
+            ? `You don't have any providers configured.`
+            : `Provider not available. You must use one of the following:\n${availableProviders.map((p) => `'${p}'`).join(', ')}`,
+      })
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/index.tsx:136]: This would be also needed in evaluations editor. So do a hook to share it

packages/compiler/src/compiler/readMetadata.test.ts
```diff
@@ -253,22 +254,94 @@ describe('config', async () => {
     })
   })
 
-  it('does not compile the prompt as YAML when it is not the first element in the prompt', async () => {
-    const prompt = `
+  it('fails when the config is not at the top of the prompt', async () => {
```
andresgutgon, line 257: I thought we wanted to allow comments for the sample project

geclos: product won't like it but i actually agree that frontmatter should be in... the front

csansoon: Oh nono, we allow comments. We just don't allow actual content like messages or text.

packages/compiler/src/parser/index.ts
```diff
@@ -57,7 +57,7 @@ export class Parser {
           code: `unclosed-block`,
           message: `Block was left open`,
         },
-        current.start!,
+        current.start! + 1,
```
andresgutgon, line 60: Magic number. What means this 1 in this context?

packages/core/src/helpers.ts
```diff
@@ -1,7 +1,38 @@
...
+    })
+    .refine((p) => providers.find((provider) => provider.name === p), {
+      message:
+        providerNames.length === 0
+          ? `You don't have any providers configured. Create a provider in your workspace settings.`
+          : `Provider not available. You must use one of the following:\n${providerNames.map((p) => `'${p}'`).join(', ')}`,
+    })
+
+  return z.object({
+    provider: providers.length > 0 ? providerSchema : providerSchema.optional(),
```
andresgutgon, line 32: Now is always required no?

packages/core/src/services/documents/recomputeChanges/index.ts
```diff
@@ -137,10 +148,15 @@ export async function recomputeChanges(
       documentsInDrafCommit,
     })
 
+    const providersScope = new ProviderApiKeysRepository(workspaceId)
+    const providers = await providersScope.findAll()
+    if (!providers) return providers
```
andresgutgon, line 153: What's doing this if? Maybe check `!providers.length`

packages/compiler/src/parser/index.ts
csansoon: It's just that the `current` index represents the start of the node, not the position of the first character (which is always 1 character after it). So... not really a magic number ig? I don't think it's super relevant in this case. Thoughts?

packages/core/src/helpers.ts
csansoon: True!

andresgutgon (dismissed):

csansoon:
Fixed all broken tests from this PR. Other failing tests come from previous broken commits >:(

PR #215, opened 2024-09-18
fix: Uncaught Error: Minified React error

andresgutgon:
# What?
Moving fetching of documents to server in `/dashboard` project listing to avoid hidratation issues with project `createdAt` missmatch. More about it here: https://github.com/vercel/next.js/discussions/43921 which points to this one https://github.com/vercel/next.js/issues/37489

image

Doing all SSR will fix the issue I think

### DIFF
--- a/packages/core/src/repositories/projectsRepository.ts
+++ b/packages/core/src/repositories/projectsRepository.ts
@@ -1,8 +1,16 @@
-import { eq, getTableColumns, isNull } from 'drizzle-orm'
+import {
+  and,
+  count,
+  eq,
+  getTableColumns,
+  isNotNull,
+  isNull,
+  max,
+} from 'drizzle-orm'
 
 import { Project } from '../browser'
 import { NotFoundError, Result } from '../lib'
-import { projects } from '../schema'
+import { commits, documentVersions, projects } from '../schema'
 import Repository from './repository'
 
 const NOT_FOUND_MSG = 'Project not found'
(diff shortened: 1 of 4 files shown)


packages/core/src/repositories/projectsRepository.ts
```diff
@@ -48,4 +56,57 @@ export class ProjectsRepository extends Repository<typeof tt, Project> {
...
+      .with(aggredatedData)
+      .select({
+        id: this.scope.id,
+        name: this.scope.name,
+        createdAt: this.scope.createdAt,
+        updatedAt: this.scope.updatedAt,
+        workspaceId: this.scope.workspaceId,
+        deletedAt: this.scope.deletedAt,
+        documentCount: aggredatedData.documentCount,
+        lastCreatedAtDocument: aggredatedData.lastCreatedAtDocument,
```
andresgutgon, line 105: Here is a dilema. Before it was 2 calls. 1 in server for projects and othe in frontend for documents and do the match. All is fine with that approach except the hidratation issue with the dates.

But what do you think about doing the thing ssr?

geclos: repository shouldn't be conditioned by frontend serialization concerns tbh, that's the function of presenters.

andresgutgon: I'll do an spread here. In the end we were already sending all from projects

geclos (dismissed):

csansoon (approved):

PR #217, opened 2024-09-18
Fix: Run batch modal

csansoon:
Run batch modal was calculating metadata from the document in server-side, then sending it back to the client component. Now that `ConversationMetadata` contains a function, it is no longer serializable


andresgutgon (approved):
Nice

PR #227, opened 2024-09-19
Fix useMetadata leak

csansoon:
Fixed memory leak from `useMetadata` hook.

Also improved performance by calculating metadata only once while the document loads, instead of 50+ times :)

### DIFF
--- a/apps/web/src/hooks/useMetadata.ts
+++ b/apps/web/src/hooks/useMetadata.ts
@@ -1,0 +1,20 @@
+'use client'
+
+import { DependencyList, useEffect, useState } from 'react'
+
+import { ConversationMetadata, readMetadata } from '@latitude-data/compiler'
+
+type Props = Parameters<typeof readMetadata>[0]
+export function useMetadata(props: Props, deps: DependencyList) {
+  const [isLoading, setIsLoading] = useState(true)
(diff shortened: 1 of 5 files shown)


apps/web/src/stores/providerApiKeys.ts
```diff
@@ -24,7 +26,7 @@ export default function useProviderApiKeys(opts?: SWRConfiguration) {
     return data || []
   }
   const {
-    data = [],
+    data = EMPTY_ARRAY,
```
csansoon, line 29: This guy was being was being set as a different "empty array" instance on every single render, and thus triggering anything that subscribed to it as a dependency. Thanks to this, opening a new document would compute the document's metadata 50+ times while this store was still loading its data.

apps/web/src/hooks/useMetadata.ts
```diff
@@ -0,0 +1,20 @@
+'use client'
+
+import { DependencyList, useEffect, useState } from 'react'
+
+import { ConversationMetadata, readMetadata } from '@latitude-data/compiler'
+
+type Props = Parameters<typeof readMetadata>[0]
+export function useMetadata(props: Props, deps: DependencyList) {
```
andresgutgon, line 8: Talked offline. I think it would be better to define all the props individually

PR #228, opened 2024-09-19
Fix: Config was not actually being sent to AI

csansoon:
Configuration options like 'temperature' were not being actually sent to the AI providers.

### DIFF
--- a/packages/core/src/services/ai/index.ts
+++ b/packages/core/src/services/ai/index.ts
@@ -1,3 +1,5 @@
+import { omit } from 'lodash-es'
+
 import { Message } from '@latitude-data/compiler'
 import { env } from '@latitude-data/env'
 import { setupJobs } from '@latitude-data/jobs'
@@ -76,6 +78,7 @@ export async function ai({
   const model = config.model
   const m = createProvider({ provider, apiKey, config })(model)
   const commonOptions = {
+    ...omit(config, ['schema']),
     model: m,
     prompt,
     messages: messages as CoreMessage[],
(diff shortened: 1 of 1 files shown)


packages/core/src/services/ai/index.ts
```diff
@@ -79,6 +79,7 @@ export async function ai({
   const m = createProvider({ provider, apiKey, config })(model)
 
   const commonOptions = {
+    ...config,
```
geclos, line 82: careful this will include schema and output to all requests to ai

csansoon: output?

csansoon: If `schema` exists in the config object, we're replacing it afterwards anyways, so no problem

geclos: Even though this is correct I'm worried this is a gun waiting to be fired on our foot, could you exclude schema and output props from this config before doing the spread? just to be on the safe side

geclos (dismissed):
minor comment and 🚀

packages/core/src/services/ai/index.ts
```diff
@@ -95,10 +98,10 @@ export async function ai({
...
       ...commonOptions,
       schema: jsonSchema(schema),
-      // @ts-expect-error - output is valid but depending on the type of schema
+      // output is valid but depending on the type of schema
       // there might be a mismatch (e.g you pass an object schema but the
       // output is "array"). Not really an issue we need to defend atm.
-      output,
+      output: output as any,
       onFinish,
     })
```
csansoon, line 106: For some reason building this now fails with the following error message:
`../../core/src/services/ai/index.ts(101,7): error TS2578: Unused '@ts-expect-error' directive.`
But removing it introduces a type error 😭

geclos: just add the ts-expect-error at the bottom of the comment instead of the top (right before the output argument)

PR #231, opened 2024-09-19
Rename files

csansoon:
Rename files and folders


packages/core/src/services/documents/renameDocumentPaths.ts
```diff
@@ -0,0 +1,51 @@
...
+    newPath: string
+  },
+  db = database,
+): Promise<TypedResult<DocumentVersion[], Error>> {
+  return await Transaction.call(async (tx) => {
+    if (commit.mergedAt !== null) {
+      return Result.error(new BadRequestError('Cannot modify a merged commit'))
+    }
+
+    const workspace = await findWorkspaceFromCommit(commit, tx)
```
andresgutgon, line 26: Pass workspace, don't use this

geclos: this is valid actually, no need to pass workspace 🤷🏼 the equivalent of doing `commit.workspace` in rails

andresgutgon: uff ya. ok

csansoon: more performance to just pass it, anyways. right?

geclos: yup but it results in an API that has invalid cases (I could pass one workspace and a commit from another workspace) and is non-obvious (if I'm already passing a commit from a workspace, why do i also have to pass the workspace?). The better solution is to fetch the tenant information together with the commit in the commits repository, so that getting the tenant from a commit does not result in an extra roundtrip to the db. later on. 

In any case we are splitting hairs here.

package.json
```diff
@@ -8,7 +8,8 @@
     "tc": "turbo tc",
     "prettier": "prettier --write \"**/*.{ts,tsx,md}\"",
     "prettier:check": "prettier --check \"**/*.{ts,tsx,md}\"",
-    "test": "turbo test"
+    "test": "turbo test",
+    "catchup": "pnpm i && pnpm build --filter=\"./packages/*\" && pnpm --filter \"@latitude-data/core\" db:migrate"
```
csansoon, line 12: A useful script to install and setup everything up to date in the branch!

packages/web-ui/src/sections/Document/Sidebar/Files/TreeToolbar/index.tsx
```diff
@@ -54,7 +54,9 @@ export function TreeToolbar() {
           open={false}
           hasChildren={false}
           isFile={isFile}
-          name=' '
+          name=''
+          isEditing={true}
+          setIsEditing={() => {}}
```
andresgutgon, line 59: Why this?

andresgutgon (approved):

PR #233, opened 2024-09-19
feature: improvements to evaluations results table

geclos:
Fixes queue ordering as evaluations were being run after all document runts had completed which is not the best experience.

Also added interval fetching of the table data to simulate real time experience.

Finally, added collapsable convos in the log messages section in order to make that section more readable.

To further improve this we need:
1. Decouple expensive queries from this view so that we can query table logs more freely
2. Add an independent queue dedicated to running evaluations in order to increase parallelism between document and evaluation runs

### DIFF
--- a/apps/web/src/actions/procedures/index.ts
+++ b/apps/web/src/actions/procedures/index.ts
@@ -54,7 +54,7 @@ export const withProject = createServerActionProcedure(authProcedure)
     return { ...ctx, project }
   })
 
-export const widthDocument = createServerActionProcedure(withProject)
+export const withDocument = createServerActionProcedure(withProject)
   .input(z.object({ commitUuid: z.string(), documentUuid: z.string() }))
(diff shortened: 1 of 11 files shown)


apps/web/src/actions/procedures/index.ts
```diff
@@ -54,7 +54,7 @@ export const withProject = createServerActionProcedure(authProcedure)
     return { ...ctx, project }
   })
 
-export const widthDocument = createServerActionProcedure(withProject)
+export const withDocument = createServerActionProcedure(withProject)
```
geclos, line 57: typo @andresgutgon 😂

geclos:
🐮 👦🏼

apps/web/src/actions/procedures/index.ts
andresgutgon: Damm

PR #242, opened 2024-09-20
chore: a bunch of small improvements

geclos:


packages/jobs/src/job-definitions/batchEvaluations/runEvaluationJob.ts
```diff
@@ -64,7 +64,6 @@ export const runEvaluationJob = async (job: Job<RunEvaluationJobData>) => {
           documentUuid,
           status: finished ? 'finished' : 'running',
           ...progress,
-          completed: progress.completed < 1 ? 1 : progress.completed,
```
andresgutgon, line 67: Why this?

geclos: it's a frontend concern

andresgutgon: And did you handle in frontend?

geclos: i don't think it's particularly important to show 1/10 vs 0/10

andresgutgon: Ok then

PR #246, opened 2024-09-20
chore: report default project not getting created in case of error

geclos:
It also improves setup service and cleans workspace creation service which had too many responsibilities

### DIFF
--- a/packages/core/src/repositories/projectsRepository.ts
+++ b/packages/core/src/repositories/projectsRepository.ts
@@ -6,6 +6,7 @@ import {
   isNotNull,
   isNull,
   max,
+  sql,
 } from 'drizzle-orm'
 
 import { Project } from '../browser'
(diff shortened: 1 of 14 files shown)


andresgutgon (dismissed):

packages/core/src/repositories/projectsRepository.ts
```diff
@@ -96,7 +97,10 @@ export class ProjectsRepository extends Repository<typeof tt, Project> {
       .with(aggredatedData)
       .select({
         ...this.scope._.selectedFields,
-        documentCount: aggredatedData.documentCount,
+        documentCount:
+          sql<number>`CAST(CASE WHEN ${aggredatedData.documentCount} IS NULL THEN 0 ELSE ${aggredatedData.documentCount} END AS INTEGER)`.as(
+            'documentCount',
+          ),
```
geclos, line 103: tests detected this bug

packages/core/src/services/projects/import.ts
```diff
@@ -28,12 +28,14 @@ export async function importDefaultProject(
 
   const defaultProjectDocumentsScope = new DocumentVersionsRepository(
     defaultProject!.workspaceId,
+    db,
```
geclos, line 31: tests detected this bug

PR #249, opened 2024-09-22
chore: add max free runs per day

geclos:


packages/core/src/services/ai/index.ts
```diff
@@ -131,15 +131,17 @@ const checkDefaultProviderUsage = async ({
...
-      `workspace:${workspace.id}:defaultProviderRunCount`,
-    )
+  if (provider.token !== env.DEFAULT_PROVIDER_API_KEY) return
 
-    if (value > DEFAULT_PROVIDER_MAX_RUNS) {
-      throw new Error('You have exceeded your maximum number of free runs')
-    }
+  const c = await cache()
+  const date = new Date()
+  const key = `workspace:${workspace.id}:${date.toISOString().slice(0, 10)}:defaultProviderRunCount`
```
csansoon, line 138: Ohh, so these are **daily** limits? I thought they were overall.

csansoon: We could limit by cost. This way users have more runs if they use cheaper models, or by reducing their context.

Having limited runs without limited models may encourage to use the top and most expensive models to better squeeze their usage.

geclos: yup those are all good ideas

PR #250, opened 2024-09-22
Feature/publish typescript sdk

geclos:


apps/gateway/src/routes/api/v1/chats/handlers/addMessage.ts
```diff
@@ -1,19 +1,20 @@
...
 import { z } from 'zod'
 
+import { chainEventPresenter } from '../../projects/:projectId/commits/:commitUuid/documents/handlers/_shared'
+
 const factory = new Factory()
 
 const schema = z.object({
   messages: z.array(messageSchema),
-  documentLogUuid: z.string(),
-  source: z.nativeEnum(LogSources).optional().default(LogSources.API),
```
geclos, line 16: public API has to be the simplest possible

apps/gateway/src/routes/api/v1/chats/handlers/addMessage.ts
```diff
@@ -22,17 +23,26 @@ export const addMessageHandler = factory.createHandlers(
...
+        const { uuid, messages } = c.req.valid('json')
         const workspace = c.get('workspace')
 
         const result = await addMessages({
           workspace,
-          documentLogUuid,
+          documentLogUuid: uuid,
           messages,
-          source,
+          source: LogSources.API,
```
geclos, line 33: I know this is not technically correct because we use ts sdk in playground but honestly i don't care

apps/gateway/src/routes/api/v1/chats/handlers/addMessage.ts
```diff
@@ -22,17 +23,26 @@ export const addMessageHandler = factory.createHandlers(
...
         }).then((r) => r.unwrap())
 
-        await pipeToStream(stream, result.stream)
+        let id = 0
+        for await (const event of streamToGenerator(result.stream)) {
+          const data = chainEventPresenter(event)
```
geclos, line 38: internally nothing has really changed but from the public API perspective I don't want to expose users to concepts such as providerlogs or document logs, and thus I'm introducing event presenters to filter this information out

apps/gateway/src/server.ts
```diff
@@ -21,3 +23,11 @@ function gracefulShutdown() {
...
 process.on('SIGTERM', gracefulShutdown)
 process.on('SIGINT', gracefulShutdown)
+
+process.on('uncaughtException', function (err) {
+  captureException(err)
+})
+
+process.on('unhandledRejection', (reason: string) => {
+  captureMessage(reason)
+})
```
geclos, line 33: important otherwise gateway doesn't recover from any unexpected error (hono is a bit crap honestly)

apps/web/src/actions/sdk/addMessagesAction.ts
```diff
@@ -26,12 +26,14 @@ export async function addMessagesAction({
...
+  const stream = createStreamableValue<
+    { event: StreamEventTypes; data: ChainEventDto },
+    Error
+  >()
 
-  const response = sdk.addMessages({
-    params: { documentLogUuid, messages, source: LogSources.Playground },
-    onMessage: (chainEvent) => {
-      stream.update(chainEvent)
+  const response = sdk.chat(documentLogUuid, messages, {
```
geclos, line 34: I'm changing the public API to follow the pattern of positional argument for mandatory arguments and a last hash object for any optional argument. I think it's a better dev experience.

geclos: I'm also changing the two main methods from runDocument and addMessages to `run` and `chat`. It's easier and faster to write at the cost of sliiiightly less expresiveness.

apps/web/src/actions/sdk/runDocumentAction.ts
```diff
@@ -28,18 +25,17 @@ export async function runDocumentAction({
...
-    params: {
-      projectId,
-      commitUuid,
-      documentPath,
-      parameters,
-      source: LogSources.Playground,
-    },
-    onMessage: (chainEvent) => {
-      stream.update(chainEvent)
+  const sdk = await createSdk(projectId).then((r) => r.unwrap())
```
geclos, line 28: projectid can now be a constructor argument so you don't have to repeat it everywhere

apps/web/src/app/(private)/_lib/createSdk.ts
```diff
@@ -15,15 +15,13 @@ async function getLatitudeApiKey() {
...
+export async function createSdk(projectId?: number) {
   const result = await getLatitudeApiKey()
   if (result.error) return result
 
   const latitudeApiKey = result.value.token
-  const gateway = {
-    host: env.GATEWAY_HOSTNAME,
-    port: env.GATEWAY_PORT,
-    ssl: env.GATEWAY_SSL,
-  }
```
geclos, line 27: shouldn't be needed

packages/core/src/constants.ts
```diff
@@ -44,32 +44,27 @@ export type ChainStepTextResponse = {
   text: string
   usage: CompletionTokenUsage
   toolCalls: ToolCall[]
-  documentLogUuid: string
-  providerLog: undefined
+  documentLogUuid?: string
```
geclos, line 47: these can always be optional since not all provider calls are made in the context of a documentloguuid

packages/sdks/typescript/src/index.ts
```diff
@@ -1,163 +1,212 @@
...
-      documentLogUuid: string
-      messages: Message[]
-      source?: LogSources
+      return this.handleStream({
+        stream: response.body!,
+        onEvent,
+        onFinished,
+        onError,
+      })
+    } catch (err) {
```
geclos, line 95: run and chat methods should not return a promise that can fail as otherwise it makes the dev experience kind of shit. In theory we should tell devs the way to handle any error from the sdk is through the onError callback.

Later on, we will introduce a no stream option that will return a promise and those methods can actually return _failable_ promises

packages/sdks/typescript/src/index.ts
```diff
@@ -1,163 +1,212 @@
...
+        if (json.type === 'chain-complete') {
+          uuid = json.uuid!
+          chainResponse = json.response!
+        }
+      }
 
-    const conversation: Message[] = []
-    let chainResponse: ChainCallResponse
+      onEvent?.({ event: event.event as StreamEventTypes, data: json })
+    }
```
geclos, line 172: bunch of improvements mostly related to error handling

packages/sdks/typescript/src/index.ts
```diff
@@ -1,163 +1,212 @@
...
+          return
         }
 
-        onMessage?.(chunk)
-      })
+        consumeEvent({ event, data })
+      }
     }
 
+    if (!uuid || !chainResponse) return
```
geclos, line 196: it means the streaming failed at some point. Should never be true but better safe than sorry.

PR #254, opened 2024-09-23
feature: improve performance when navigating to a project

geclos:
- Moved document versions fetching to a proper GET endpoint so that we can leverage browser caching
- Reduced the amount of redirects when entering a project
- Added the documentUuid to the latest seen cookie state so that we can go there directly upon reentry
- Added a blank slate in the documents view when no document is selected instead of doing a random redirect
- Added missing db indices and simplified the getDocumentsAtCommit query a bit

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/page.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/page.tsx
@@ -10,7 +10,7 @@ import { ROUTES } from '$/services/routes'
 import { cookies } from 'next/headers'
 import { notFound, redirect } from 'next/navigation'
 
-import { getCommitUrl } from './utils'
+import { getRedirectUrl } from './utils'
 
 const PROJECT_ROUTE = ROUTES.projects.detail
 
@@ -19,10 +19,8 @@ export type ProjectPageParams = {
 }
 
 export default async function ProjectPage({ params }: ProjectPageParams) {
-  const cookieStore = cookies()
-  const lastSeenCommitUuid = cookieStore.get(
-    lastSeenCommitCookieName(Number(params.projectId)),
-  )
+  const { commitUuid: lastSeenCommitUuid, documentUuid: lastSeenDocumentUuid } =
+    getLastSeenDataFromCookie(params.projectId)
 
   let session: SessionData
   let project: Project
@@ -34,15 +32,15 @@ export default async function ProjectPage({ params }: ProjectPageParams) {
       projectId: Number(params.projectId),
       workspaceId: session.workspace.id,
     })
-
     const commits = await findCommitsByProjectCached({
       projectId: project.id,
     })
 
-    url = getCommitUrl({
+    url = getRedirectUrl({
       commits,
       projectId: project.id,
-      lastSeenCommitUuid: lastSeenCommitUuid?.value,
+      lastSeenCommitUuid,
+      lastSeenDocumentUuid,
       PROJECT_ROUTE,
     })
   } catch (error) {
@@ -54,4 +52,17 @@ export default async function ProjectPage({ params }: ProjectPageParams) {
   }
 
   return redirect(url)
+}
+
+function getLastSeenDataFromCookie(projectId: string) {
+  const cookieStore = cookies()
+  const data = cookieStore.get(lastSeenCommitCookieName(Number(projectId)))
+  if (!data?.value) return {}
+
+  try {
+    const { commitUuid, documentUuid } = JSON.parse(data.value)
+    return { commitUuid, documentUuid }
+  } catch (_) {
+    return { commitUuid: data.value as string }
+  }
 }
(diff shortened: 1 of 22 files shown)


apps/web/src/app/(private)/projects/[projectId]/page.tsx
```diff
@@ -55,3 +59,16 @@ export default async function ProjectPage({ params }: ProjectPageParams) {
...
+function getLastSeenDataFromCookie(projectId: string) {
+  const cookieStore = cookies()
+  const data = cookieStore.get(lastSeenCommitCookieName(Number(projectId)))
+  if (!data?.value) return {}
+
+  try {
+    const { commitUuid, documentUuid } = JSON.parse(data.value)
+    return { commitUuid, documentUuid }
+  } catch (e) {
+    return data.value as string
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/page.tsx:72]: so it's backwards compatible with the old cookies

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/layout.tsx
```diff
@@ -38,13 +37,6 @@ export default async function DocumentPage({
       </DocumentVersionProvider>
     )
   } catch (error) {
-    // TODO: Show a 404 page within the documents layout, while still showing
-    // the sidebar and stuff For now, we just redirect to documents root if
-    // document is not found instead for a cleaner UX
-    if (env.NODE_ENV === 'development') {
-      console.error(error)
-    }
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/layout.tsx:46]: actually redirecting to the root of documents is the right behaviour here

packages/core/src/schema/models/documentVersions.ts
```diff
@@ -27,14 +28,14 @@ export const documentVersions = latitudeSchema.table(
     ...timestamps(),
   },
   (table) => ({
-    uniqueDocumentUuidCommitId: unique('unique_document_uuid_commit_id').on(
```
geclos, line 30: unique is not an index always use uniqueIndex

packages/core/src/schema/models/documentVersions.ts
```diff
@@ -27,14 +28,14 @@ export const documentVersions = latitudeSchema.table(
...
-    ),
+    uniqueDocumentUuidCommitId: uniqueIndex(
+      'document_versions_unique_document_uuid_commit_id',
+    ).on(table.documentUuid, table.commitId),
+    uniquePathCommitIdDeletedAt: uniqueIndex(
+      'document_versions_unique_path_commit_id_deleted_at',
+    ).on(table.path, table.commitId, table.deletedAt),
+    commitIdIdx: index('document_versions_commit_id_idx').on(table.commitId),
+    deletedAtIdx: index('document_versions_deleted_at_idx').on(table.deletedAt),
+    pathIdx: index('document_versions_path_idx').on(table.path),
```
geclos, line 39: compound indices are ok but having dedicated indices per-column is actually better for performance

packages/core/src/repositories/documentVersionsRepository/index.ts
```diff
@@ -233,16 +233,14 @@ export class DocumentVersionsRepository extends Repository<
...
+          .where(and(filterByMaxMergedAt(), filterByProject()))
           .groupBy(this.scope.documentUuid),
       )
 
     const documentsFromMergedCommits = await this.db
       .with(lastVersionOfEachDocument)
       .select(this.scope._.selectedFields)
       .from(this.scope)
-      .innerJoin(documentVersions, eq(this.scope.id, documentVersions.id))
-      .innerJoin(projects, eq(this.scope.projectId, projects.id))
```
geclos, line 245: i think these two aren't needed, tests pass correctly and cannot see anything wrong when testing locally, let me know if I'm missing something @andresgutgon @csansoon

packages/core/src/schema/models/commits.ts
```diff
@@ -40,9 +40,12 @@ export const commits = latitudeSchema.table(
       table.mergedAt,
       table.projectId,
     ),
-    uniqueCommitVersion: unique('unique_commit_version').on(
```
geclos, line 43: uniqueIndex please

apps/web/src/app/(private)/projects/[projectId]/page.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/page.tsx:72]: Return `{ lastSeenCommitUuid:  data.value as string, lastSeenDocumentUuid: undefined }` this way you can destructure safely later

packages/core/src/schema/models/commits.ts
```diff
@@ -40,9 +40,12 @@ export const commits = latitudeSchema.table(
...
+    uniqueCommitVersion: uniqueIndex('unique_commit_version').on(
       table.version,
       table.projectId,
     ),
+    userIdx: index('user_idx').on(table.userId),
+    mergedAtIdx: index('merged_at_idx').on(table.mergedAt),
```
geclos, line 48: super important we are querying with this column constantly

andresgutgon (dismissed):

PR #274, opened 2024-09-25
feature: dynamic docs

geclos:
Adds documentation in the prompt section so that users can easily integrate their prompt in their environments via sdk.


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentationModal/_components/APIUsage.tsx
```diff
@@ -0,0 +1,48 @@
...
+        .map((key) => `      "${key}": ""`)
+        .join(',\n')
+
+      bodyContent.push(`    "parameters": {
+${parameterEntries}
+    }`)
+    }
+
+    return bodyContent.join(',\n')
+  }
```
Comment (csansoon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentationModal/_components/APIUsage.tsx:30]: Weird hardcodded spacing in the strings

Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentationModal/_components/APIUsage.tsx:30]: I get that's the indentation for the body, but could you just add the indentation afterwards in a function so that it's simpler? Or just add it to an object and stringify later?

packages/web-ui/src/ds/atoms/CodeBlock/CopyButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+      })
+    })
+  }
+
+  return (
+    <Button onClick={handleCopy} variant='ghost' size='small'>
+      <Icon name='clipboard' />
+    </Button>
+  )
+}
```
andresgutgon, line 30: Move out of CodeBlock. looks something we use in other places already no?

csansoon (approved):
minor comments 👍

packages/web-ui/package.json
```diff
@@ -21,7 +21,8 @@
   },
   "dependencies": {
     "@latitude-data/compiler": "workspace:^",
-    "@latitude-data/core": "workspace:^"
+    "@latitude-data/core": "workspace:^",
+    "react-syntax-highlighter": "^15.5.0"
```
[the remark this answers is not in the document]
geclos: hmm very hard to do since there's lots of local state that can't be done in server (usestate y tal)

packages/web-ui/src/ds/atoms/CodeBlock/CopyButton.tsx
geclos: hmm will do in another pr

Issue #271, opened 2024-09-25
Missing link to original provider log in the evaluation log


geclos:
<img width="1101" alt="image" src="https://github.com/user-attachments/assets/40687bb3-4947-46ad-b2d5-84e99fb9d9a7">

PR #277, opened 2024-09-25
Show original document log from a evaluation log

andresgutgon:
# What? 
Solves this issue https://github.com/latitude-dev/latitude-llm/issues/271

## Todo
- [x] Open modal with document log info
- [x] Defend from not found document log

### Off-topic: Refactor `useRefetchStats`
The problem we had is that the catch all `useRefetchStats` hook we had was triggering for example numeric mean in non numeric evaluations. The reason is the way SWR stores work by default the store fetch data when it's mounted. This could be solved by disabling fetching
on mount but I think is more correct to put each metric a listener to the `evaluationStatus` websocket event


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx
```diff
@@ -1,14 +1,85 @@
...
+      toast({
+        title: 'Failed to fetch document log',
+        description: error.message,
+        variant: 'destructive',
+      })
+    }
+  }, [documentLogId])
+
+  return { fetching, fetchDocumentLog }
+}
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx:81]: why aren't u using regular stores?

geclos (approved):
👍🏼

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx:81]: Changed

PR #278, opened 2024-09-25
Usage Indicator

csansoon:
A usage indicator showing the number of runs that have been used since the last renewal.

Usage is calculated by adding the number of `DocumentLog` and `EvaluationResult` created since the last renewal date.

Currently, all workspaces have a maximum of 2k runs, which is renewed in a monthly basis since the workspace was created.


apps/web/src/components/layouts/AppLayout/Header/UsageIndicator/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  return children
+}
+
+function descriptionText({ usage, max }: WorkspaceUsage) {
+  const ratio = (max - usage) / max
+  if (ratio <= 0)
+    return "You've reached the maximum number of runs. Your app will continue working, the Latitude limit reached means you'll no longer be able to run tests, view new logs or evaluation results."
+  if (ratio < 0.25)
```
andresgutgon, line 79: The brackeeets

packages/core/src/constants.ts
```diff
@@ -19,6 +19,8 @@ export const LATITUDE_SLACK_URL =
   'https://trylatitude.slack.com/join/shared_invite/zt-17dyj4elt-rwM~h2OorAA3NtgmibhnLA#/shared-invite/email'
 export const LATITUDE_HELP_URL = LATITUDE_SLACK_URL
 export const HEAD_COMMIT = 'live'
+export const MAX_FREE_RUNS = 2000
+
```
csansoon, line 23: Since there is no concept of plan yet, I'm just hardcoding this value. Database schema is the exact same for now.

packages/core/src/services/workspaces/usage.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const createdAtDate = workspace.createdAt
+  const currentDate = new Date(Date.now())
+
+  const day = createdAtDate.getDate()
+  const month =
+    (currentDate.getMonth() - (currentDate.getDate() < day ? 1 : 0)) % 12
+  const year = currentDate.getFullYear()
+
+  const latestRenewalDate = new Date(year, month, day)
```
csansoon, line 30: Free plan is renewed in a monthly basis on the same day of creation each month (not every 30 days). This is how I calculate which day was it last renewed based on the creation date.

packages/core/src/events/handlers/documentLogs/createJob.ts
```diff
@@ -31,4 +34,14 @@ export const createDocumentLogJob = async ({
...
+
+  websockets.emit('documentLogCreated', {
+    workspaceId,
+    data: {
+      workspaceId,
+      documentUuid: event.data.documentUuid,
+      commitUuid: event.data.commitUuid,
+      documentLogId: documentLog.id,
+    },
+  })
```
geclos, line 46: side effect should be in an event handler, we already have the documentLogCreated event in fact. Leave a TODO if you don't wanna do it now.

packages/core/src/services/documentLogs/create.ts
```diff
@@ -12,6 +12,7 @@ export type CreateDocumentLogProps = {
     resolvedContent: string
     customIdentifier?: string
     duration: number
+    createdAt?: Date
```
geclos, line 15: why? should be automatically populated otherwise we have a problem

csansoon: This is only used for testing. I need to create old logs in a test environment.

packages/core/src/services/workspaces/usage.ts
```diff
@@ -0,0 +1,61 @@
...
+import { PromisedResult, Result } from '../../lib'
+import { EvaluationResultsRepository } from '../../repositories'
+import { DocumentLogsRepository } from '../../repositories/documentLogsRepository'
+
+export function getLatestRenewalDate(firstRenewalDate: Date, targetDate: Date) {
+  if (targetDate.getTime() < firstRenewalDate.getTime()) {
+    return firstRenewalDate
+  }
+
+  const day = firstRenewalDate.getDate() // The day of the month is always mantained.
```
geclos, line 14: careful with utc shenanigans if any of your dates is not in utc you could have unexpected variations

geclos (approved):
looks good 👌🏼

packages/core/src/events/handlers/documentLogs/createJob.ts
csansoon: But this is the `DocumentRunEvent`'s event handler.

geclos: sure but this is a side effect of document log created event, instead of having a single job that handles all side effects it's more maintainable and it scales better to have an event handler for each side effect

PR #279, opened 2024-09-25
feature: live evaluations

geclos:


apps/web/src/actions/connectedEvaluations/update.ts
```diff
@@ -0,0 +1,33 @@
+'use server'
+
+import { ConnectedEvaluationsRepository } from '@latitude-data/core/repositories'
+import { updateConnectedEvaluation } from '@latitude-data/core/services/connectedEvaluations/update'
+import { z } from 'zod'
+
+import { authProcedure } from '../procedures'
+
+export const updateConnectedEvaluationAction = authProcedure
```
andresgutgon, line 9: This is a more `toggleLiveAction` so we can track individual events when it happens

apps/web/src/stores/connectedEvaluations.ts
```diff
@@ -1,51 +1,81 @@
...
 
+  const { execute: update, isPending: isUpdating } = useLatitudeAction(
+    updateConnectedEvaluationAction,
+    {
+      onSuccess: ({
+        data: updatedEvaluation,
+      }: {
+        data: ConnectedEvaluation
+      }) => {
+        toast({
```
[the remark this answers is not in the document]
geclos: yes but i wanted a more specific toast for this use case

apps/web/src/actions/connectedEvaluations/update.ts
geclos: hmm it's a good point, will do in another pr also

PR #281, opened 2024-09-25
Upgrade Next.js & enable Turbopack

timneutkens:
- Upgraded to the latest canary version
- Enabled Turbopack for development

I manually checked a bunch of routes but obviously you will know the application better so please have a go at double checking that everything still works correctly 👍


apps/web/package.json
```diff
@@ -38,13 +38,14 @@
     "lucia": "^3.2.0",
     "monaco-editor": "^0.50.0",
     "nanoid": "^5.0.7",
-    "next": "^14.3.0-canary.87",
+    "next": "15.0.0-canary.168",
     "next-themes": "^0.3.0",
     "nextjs-toploader": "^1.6.12",
     "nprogress": "^0.2.0",
+    "oslo": "1.2.0",
```
andresgutgon, line 45: was this a missing dependency?

timneutkens: Yeah Turbopack actively shows that the dependency is missing when it's external, it would fail at runtime currently without this being installed directly.

timneutkens:

geclos:
Thanks!

PR #282, opened 2024-09-25
chore: add origin to evaluation results table

geclos:


packages/web-ui/src/ds/atoms/Tooltip/index.tsx
```diff
@@ -85,9 +86,7 @@ function Tooltip({
       delayDuration={delayDuration}
       disableHoverableContent={disableHoverableContent}
     >
-      <TooltipTrigger asChild onClick={(e) => e.preventDefault()}>
```
geclos, line 88: this prevent default is messing with links, not sure why it's there and there's no comment explaining why so i'm just removing it

packages/web-ui/src/ds/atoms/SplitPane/index.tsx
```diff
@@ -114,7 +114,7 @@ function HorizontalSplit({
   const [paneWidth, setPaneWidth] = useState<number>(initialWidth)
   const width = typeof window !== 'undefined' ? paneWidth : initialWidth
   return (
-    <div className='h-full w-full grid grid-cols-[auto,1fr]'>
+    <div className='min-h-full w-full grid grid-cols-[auto,1fr]'>
```
geclos, line 117: mega bug, you couldn't scroll in any project section

PR #286, opened 2024-09-26
quickfix: allow to import logs from any document in the project

geclos:
Also fixes the issue where importing logs sometimes did not work

### DIFF
--- a/apps/web/src/app/api/documents/[projectId]/for-import/route.test.ts
+++ b/apps/web/src/app/api/documents/[projectId]/for-import/route.test.ts
@@ -1,0 +1,94 @@
+import {
+  DocumentVersion,
+  Providers,
+  User,
+  Workspace,
+} from '@latitude-data/core/browser'
+import { createProject, helpers } from '@latitude-data/core/factories'
+import { NextRequest } from 'next/server'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { GET } from './route'
+
+const mocks = vi.hoisted(() => {
+  return {
+    getSession: vi.fn(),
+  }
+})
+vi.mock('$/services/auth/getSession', () => ({
+  getSession: mocks.getSession,
+}))
+
+describe('GET handler for documents/[projectId]/for-import', () => {
+  let mockRequest: NextRequest
+  let mockParams: { projectId: string }
+  let mockWorkspace: Workspace
+  let mockDocuments: DocumentVersion[]
+  let mockUser: User
+
+  beforeEach(async () => {
+    mockRequest = new NextRequest('http://localhost:3000')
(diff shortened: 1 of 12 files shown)


apps/web/src/app/api/documents/[projectId]/for-import/route.ts
```diff
@@ -0,0 +1,34 @@
...
+      }: {
+        params: { projectId: string }
+        workspace: Workspace
+      },
+    ) => {
+      const { projectId } = params
+
+      if (!projectId) {
+        throw new BadRequestError('Project ID is required')
+      }
```
Comment (csansoon) [inline apps/web/src/app/api/documents/[projectId]/for-import/route.ts:24]: Can we use zod for input validation in API routes like we do with server actions?

packages/core/src/repositories/documentVersionsRepository/index.ts
```diff
@@ -252,4 +252,16 @@ export class DocumentVersionsRepository extends Repository<
...
+    const documents = await this.db
+      .selectDistinct({
+        documentUuid: this.scope.documentUuid,
+        path: this.scope.path,
+      })
+      .from(this.scope)
+      .where(eq(this.scope.projectId, projectId))
+
+    return Result.ok(documents)
+  }
```
csansoon, line 266: What if the document has been deleted? This list could get annoyingly long on long-lived projects

csansoon: A nice problem to have, though

csansoon (dismissed):

apps/web/src/app/api/documents/[projectId]/for-import/route.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        },
+      },
+    })
+    mockParams = { projectId: String(project.id) }
+
+    mockWorkspace = workspace
+    mockDocuments = documents
+  })
+
+  it('should return 400 if projectId is missing', async () => {
```
Comment (andresgutgon) [inline apps/web/src/app/api/documents/[projectId]/for-import/route.test.ts:38]: What about checking user logged in?

andresgutgon (dismissed):

packages/core/src/repositories/documentVersionsRepository/index.ts
geclos: 💯 good point!

apps/web/src/app/api/documents/[projectId]/for-import/route.ts
Comment (geclos) [reply apps/web/src/app/api/documents/[projectId]/for-import/route.ts:24]: we would need to create a middleware for that, totally doable

apps/web/src/app/api/documents/[projectId]/for-import/route.test.ts
Comment (geclos) [reply apps/web/src/app/api/documents/[projectId]/for-import/route.test.ts]: 💯

PR #297, opened 2024-09-27
Video in Evaluation blankslate

csansoon:
image


geclos (dismissed):

packages/web-ui/src/ds/atoms/Badge/index.tsx
```diff
@@ -28,7 +28,7 @@ const badgeVariants = cva(
       },
       shape: {
         default: 'px-2 py-0.5',
-        square: 'p-2',
+        square: 'max-h-5 py-2 px-[6px]',
```
andresgutgon, line 31: This is `p-something` don't use px if possible

PR #302, opened 2024-09-27
quickfix: evaluarion results repository scope missed a filter

geclos:
Most of the times it does not matter because we filter in almost all methods by other means. But in the credits calculations it does matter.

### DIFF
--- a/packages/core/src/repositories/evaluationResultsRepository/index.ts
+++ b/packages/core/src/repositories/evaluationResultsRepository/index.ts
@@ -1,5 +1,4 @@
 import { eq, getTableColumns, sql } from 'drizzle-orm'
-import { NodePgDatabase } from 'drizzle-orm/node-postgres'
 
 import {
   Commit,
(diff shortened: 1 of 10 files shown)


packages/core/src/repositories/evaluationResultsRepository/index.ts
```diff
@@ -42,15 +41,9 @@ export class EvaluationResultsRepository extends Repository<
   typeof evaluationResultDto,
   EvaluationResultDto
 > {
-  static baseQuery(
```
geclos, line 45: basequery is just a bad idea

packages/core/src/repositories/evaluationResultsRepository/index.ts
```diff
@@ -68,12 +61,8 @@ export class EvaluationResultsRepository extends Repository<
         evaluationResultableTexts,
         sql`${evaluationResults.resultableType} = ${EvaluationResultableType.Text} AND ${evaluationResults.resultableId} = ${evaluationResultableTexts.id}`,
       )
-  }
-
-  get scope() {
-    return EvaluationResultsRepository.baseQuery(this.db).as(
-      'evaluationResultsScope',
-    )
+      .where(eq(evaluations.workspaceId, this.workspaceId))
```
geclos, line 64: if i see a scope without a where clause i will yell at u

andresgutgon (dismissed):

geclos:
Merging even with tests failing in CI. They are failing because redis service container [is down](https://hub.docker.com/_/redis) at the moment and our tests CI relies on this service. I've run all the tests locally and I'm confident they all pass.

PR #307, opened 2024-09-29
feature: providerlog context is no longer a json

geclos:
We think it's better if context is just a string with the full conversation excluding the response.

### DIFF
--- a/packages/core/src/repositories/documentVersionsRepository/index.test.ts
+++ b/packages/core/src/repositories/documentVersionsRepository/index.test.ts
@@ -47,8 +47,9 @@ describe('DocumentVersionsRepository', () => {
 
       const documents = result.unwrap()
       expect(documents).toHaveLength(2)
-      const paths = documents.map((d) => d.path).sort()
-      expect(paths).toEqual(['bar', 'foo'])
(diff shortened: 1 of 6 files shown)


packages/core/src/repositories/documentVersionsRepository/index.test.ts
```diff
@@ -47,7 +47,10 @@ describe('DocumentVersionsRepository', () => {
 
       const documents = result.unwrap()
       expect(documents).toHaveLength(2)
-      expect(documents.map((d) => d.path)).toEqual(['foo', 'bar'])
+      expect(documents.map((d) => d.path).sort()).toEqual(['bar', 'foo'])
+      expect(new Set(documents.map((d) => d.path))).toEqual(
+        new Set(['foo', 'bar']),
+      )
```
andresgutgon, line 53: git conflict! I fixed it already

geclos: let's see who merges faster :trollface:

andresgutgon: I said first that I fixed it if that counts ;) I feel I always lost this game for some reason

csansoon (approved):

PR #311, opened 2024-09-30
Fix evaluation blank slate not being scrollable

andresgutgon:
# What?
☝️ That


packages/web-ui/src/ds/atoms/Table/index.tsx
```diff
@@ -35,7 +35,7 @@ const Table = forwardRef<HTMLTableElement, TableProps>(
       className='flex flex-col relative w-full rounded-lg border'
     >
       <div
-        className={cn('relative w-full flex-grow', overflow, {
+        className={cn('relative w-full flex-grow', 'overflow-hidden', {
```
geclos, line 38: overflow hidden on tables now? dangerous are we sure about this?

geclos (approved):

PR #312, opened 2024-09-30
Compiler fixes

csansoon:
Some minor (but important) fixes for the compiler:

- Now compiler ONLY detects 3 consecutive dashes as a config section. Using less or more will not be treated as such
```
---
is_valid: true
---

-------------------------
☝️ this is just text

----
☝️ still text

--
☝️ text

---
☝️ whoops, you're starting a config

\---
☝️ nah, this is just text
```

- Adding content inside a `<response>` tag will now result in an error, which will explain that the user needs to put the content BEFORE the tag in order to include context to the AI response.
- Other minor bugfixes.


geclos (approved):
🚀

andresgutgon (approved):

Issue #182, opened 2024-09-16
Add user events


cesr:
### Implementation

I'd love to use Posthog for this as it gives us a lot of precooked metrics

### Events

- [x] Sign up
- [x] Log in
- [x] Create project
- [x] Create evaluation
- [x] Run evaluation prompt
- [x] Create dataset
- [x] Add provider api key
- [x] Invite user
- [x] Create new project version
- [x] Publish new project version
- [ ] Create folder
- [x] Create file
- [x] Run document prompt
- [x] Connect evaluation to document
- [x] Run batch evaluation
- [ ] Visit page
  - [ ] Path in metadata

geclos:
#226

cesr:
Reopening to move these events to the backend

PR #313, opened 2024-09-30
feature: more user events

geclos:
#182 

In a later PR I will address the problem of side effects in transactions not respecting transaction commits.


packages/core/src/events/handlers/index.ts
```diff
@@ -20,7 +24,7 @@ import { sendMagicLinkJob } from './sendMagicLinkHandler'
 
 type LatitudeEventGeneric<
   U extends keyof typeof EventHandlers,
-  T extends Record<string, unknown>,
+  T extends { userEmail?: string; workspaceId?: number; [key: string]: any },
 > = {
   type: U
   data: T
```
csansoon, line 30: Why adding `userEmail` to the generic type?
Is this just to humanize the events info?

csansoon (dismissed):

packages/core/src/events/handlers/index.ts
geclos: right at the beginning those weren't optional so that's why but you are totally right 👌🏼

PR #318, opened 2024-10-01
fix: heights

geclos:
We had some shenanigans with heights in documents with lots of inputs. This fixes it.


andresgutgon (dismissed):

apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/EvaluationEditor/Editor/index.tsx
```diff
@@ -92,7 +92,7 @@ export default function EvaluationEditor({
           />
         </Suspense>
       </div>
-      <div className='flex flex-col flex-1 gap-2 min-w-0 max-w-1/2'>
+      <div className='flex flex-col flex-1 gap-2 min-w-0 max-w-1/2 overflow-y-auto max-h-[calc(100vh-150px)]'>
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/EvaluationEditor/Editor/index.tsx:95]: This is super weak. Will break with any design change. I would with JS for this. But let's see

PR #323, opened 2024-10-02
chore: tweak user events

geclos:
Only send events with email to analytics. Added a couple of sdk events to track document runs.

### DIFF
--- a/packages/jobs/src/job-definitions/events/publishToAnalyticsJob.ts
+++ b/packages/jobs/src/job-definitions/events/publishToAnalyticsJob.ts
@@ -1,5 +1,6 @@
 import { LatitudeEvent } from '@latitude-data/core/events/handlers/index'
 import { PostHogClient } from '@latitude-data/core/services/posthog'
+import { env } from '@latitude-data/env'
 import { Job } from 'bullmq'
 
 export const publishToAnalyticsJob = async (job: Job<LatitudeEvent>) => {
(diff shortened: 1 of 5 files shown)


packages/jobs/src/job-definitions/events/publishToAnalyticsJob.ts
```diff
@@ -12,28 +13,24 @@ export const publishToAnalyticsJob = async (job: Job<LatitudeEvent>) => {
     workspaceId = event.data.workspaceId
   }
 
-  const client = PostHogClient()
+  if (!userEmail) return
+
+  if (env.NODE_ENV !== 'production') {
+    console.log('Analytics event captured:', event.type)
```
andresgutgon, line 19: Do we need this console.log

andresgutgon (approved):

packages/jobs/src/job-definitions/events/publishToAnalyticsJob.ts
geclos: It's useful for local development tbh

andresgutgon: But not when not developing events. I would add when working on it. The rest of the time is just noise

geclos: ok will remove then

andresgutgon: Whe have this 
https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/lib/debug.ts

If you use `debug` this console logs will only appear when you run 

```json
NODE_DEBUG=latitude:debug next dev --turbo
```

Which now is always but we can move this to `pnpm dev:debug`

All done easy just use `debug` instead of console.log

andresgutgon: This is the standard node way of doing console.logs

PR #324, opened 2024-10-02
feature: default provider banner and some other impros

geclos:
- Redirect to the evaluation editor from evaluation page
- Preload providers in server for better editor experience
- Add default provider banner in evaluation page


geclos:
🐮 👦🏼 simple

PR #326, opened 2024-10-02
feature: add free run counter in UI

geclos:
image

### DIFF
--- a/apps/web/src/components/EditorHeader/index.tsx
+++ b/apps/web/src/components/EditorHeader/index.tsx
@@ -11,11 +11,16 @@ import {
   AppLocalStorage,
   DropdownMenu,
   FormFieldGroup,
+  Icon,
   Select,
   Text,
+  Tooltip,
   useLocalStorage,
 } from '@latitude-data/web-ui'
+import { envClient } from '$/envClient'
+import { ROUTES } from '$/services/routes'
 import useProviderApiKeys from '$/stores/providerApiKeys'
+import Link from 'next/link'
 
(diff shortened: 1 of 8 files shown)


apps/web/src/components/EditorHeader/index.tsx
```diff
@@ -234,6 +241,55 @@ export default function EditorHeader({
...
       </FormFieldGroup>
+      {selectedProvider === envClient.NEXT_PUBLIC_DEFAULT_PROJECT_ID && (
+        <div>
+          {freeRunsCount !== undefined ? (
+            <Text.H6 color='foregroundMuted'>
+              You have consumed{' '}
+              <Tooltip
+                trigger={
+                  <strong className='text-primary'>
+                    {freeRunsCount} of 100 daily free runs.
```
csansoon, line 252: I think there's a global constant for the limit? Use it here too.

andresgutgon (dismissed):

apps/web/src/components/EditorHeader/index.tsx
geclos: the constant and the number are different (constant is 1k, but we wanna say 100 here)

geclos: that constant and this number are different

csansoon: What? Why? Isn't it the daily limit for free runs?

geclos: talked offline

PR #327, opened 2024-10-02
Data model for document run errors

andresgutgon:
# What?
We want to capture errors that happen before AI provider runs and also errors that the AI reports


packages/core/src/constants.ts
```diff
@@ -73,6 +73,16 @@ export enum LogSources {
...
   Evaluation = 'evaluation',
 }
+
+export enum DocumentRunErrorCodes {
+  Unknown = 'unknown_error',
+  DefaultProviderExceededQuota = 'default_provider_exceeded_quota_error',
+  DocumentConfigError = 'document_config_error',
+  MissingProvider = 'missing_provider_error',
+  ChainCompileError = 'chain_compile_error',
+  AIRunError = 'ai_run_error',
```
andresgutgon, line 83: Can you think of any other kind of errors in this catalog?

packages/core/src/schema/models/documentRunErrors.ts
```diff
@@ -0,0 +1,46 @@
...
+    details: jsonb('details').$type<Record<string, unknown>>(),
+    ...timestamps(),
+  },
+  (table) => ({
+    documentLogUuidIdx: index('document_run_errors_uuid_idx').on(
+      table.documentUuid,
+    ),
+    commitIdIdx: index('document_run_errors_commit_id_idx').on(table.commitId),
+  }),
+)
```
csansoon, line 46: What if the error comes from an evaluation? That's not a document run but still has the same behaviour.
I think this should have a polymorphic relation with either `documentLog` or `evaluationResult`, since they're the two sources of provider logs and AI execution.

csansoon: This is similar as how `ProviderLogs` should have a polymorphic relationship with either `DocumentLog` or `EvaluationResult` too, instead of our current approach.

andresgutgon: I think that makes a lot of sense. Changing

packages/core/src/schema/models/evaluationResults.ts
```diff
@@ -26,8 +26,8 @@ export const evaluationResults = latitudeSchema.table(
     providerLogId: bigint('provider_log_id', { mode: 'number' })
       .notNull()
       .references(() => providerLogs.id),
-    resultableType: evaluationResultTypes('resultable_type').notNull(),
-    resultableId: bigint('resultable_id', { mode: 'number' }).notNull(),
```
andresgutgon, line 30: Talked with @csansoon we want to capture errors and also persist the document logs or evaluation results no matter what. This way we document that something happened also we document the duration, what document and with what parameters that error occurred.

andresgutgon: For that reason we need to make from now on results nullable

packages/core/src/schema/models/runErrors.ts
```diff
@@ -0,0 +1,49 @@
...
+      {
+        onDelete: 'set null',
+        onUpdate: 'cascade',
+      },
+    ),
+    message: text('message').notNull(),
+    details: jsonb('details').$type<Record<string, unknown>>(),
+    ...timestamps(),
+  },
+  (table) => ({
```
andresgutgon, line 40: 👇 Review index please

PR #329, opened 2024-10-02
Ensure evaluation results and document logs only gives results without errors.

andresgutgon:
# What?
Also splitting run errors repository in two. The use case is to get evaluation result errors in some UI and in a different UI get document log errors.

## TODO
- [x] Test document log scope repository
- [x] Test document log error scope repository
- [x] Test evaluation result scope repository
- [x] Test evaluation result error scope repository


packages/core/src/schema/models/runErrors.ts
```diff
@@ -19,26 +18,23 @@ export const runErrorEntities = latitudeSchema.enum('run_error_entity_enum', [
...
+    : never
+
 export const runErrors = latitudeSchema.table(
   'run_errors',
   {
     id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
     code: errorCodeEnum('code').notNull(),
-    errorableType: runErrorEntities('errorable_type'),
-    errorableId: bigint('errorable_id', { mode: 'number' }),
-    providerLogId: bigint('provider_log_id', { mode: 'number' }).references(
```
andresgutgon, line 29: I decided that I'd add this pointer in the other direction. A provider log would have a nullable error

packages/core/src/schema/models/runErrors.ts
```diff
@@ -19,26 +18,23 @@ export const runErrorEntities = latitudeSchema.enum('run_error_entity_enum', [
...
+    ? { compileCode: string; message: string }
+    : never
+
 export const runErrors = latitudeSchema.table(
   'run_errors',
   {
     id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
     code: errorCodeEnum('code').notNull(),
-    errorableType: runErrorEntities('errorable_type'),
-    errorableId: bigint('errorable_id', { mode: 'number' }),
```
andresgutgon, line 28: This was wrong. It has to be `.notNull`

csansoon (dismissed):

geclos (approved):
fantastic!

PR #335, opened 2024-10-03
Evaluation results by document content + diff editor apparently

csansoon:
> From lil train 🚂 #320

Fetch evaluation results that match the document content, instead of the commit.
This allows to display results more accurately thanks to:
 - Include all results from **any** commit since the document was last modified, which allows to include more organic results.
 - Only include results since the last time the document was modified within the same draft commit, which removes results that are no longer applied.


apps/web/src/stores/evaluationResultsByDocumentContent.ts
```diff
@@ -0,0 +1,75 @@
...
+    commitUuid: string
+    projectId: number
+  },
+  { fallbackData }: SWRConfiguration = {},
+) {
+  const { toast } = useToast()
+
+  const fetcher = useCallback(async () => {
+    const response = await fetch(
+      `/api/documents/${projectId}/${commitUuid}/${documentUuid}/evaluationResultsByDocumentContent/${evaluationId}`,
```
andresgutgon, line 34: Let's start today doing `API_ROUTES`. There are 3 cases. Please start this good practice 🙏

packages/core/src/schema/models/documentLogs.ts
```diff
@@ -34,5 +34,8 @@ export const documentLogs = latitudeSchema.table(
   (table) => ({
     documentLogUuidIdx: index('document_log_uuid_idx').on(table.documentUuid),
     commitIdIdx: index('document_logs_commit_id_idx').on(table.commitId),
+    resolvedContentIdx: index('document_logs_resolved_content_idx').on(
```
andresgutgon, line 37: Wouldn't be too much tralla for an index?

apps/web/src/services/routes.ts
```diff
@@ -27,6 +27,38 @@ export const ROUTES = {
       root: '/backoffice/rewards',
     },
   },
+  api: {
+    documentLogs: {
+      detail: ({ id }: { id: number }) => ({
+        root: `/api/documentLogs/${id}`,
```
andresgutgon, line 33: Do dry of `/api` prefix

apps/web/src/stores/documentLogWithMetadata.ts
```diff
@@ -11,7 +12,7 @@ export default function useDocumentLogWithMetadata(
       if (!documentLogUuid) return []
 
       const response = await fetch(
-        `/api/documentLogs/uuids/${documentLogUuid}`,
+        ROUTES.api.documentLogs.uuids.detail({ uuid: documentLogUuid }).root,
```
andresgutgon, line 15: Remove `.uuids.` doesn't add value and is confusing

packages/core/src/schema/models/documentLogs.ts
csansoon: yes... I added a new column instead called `contentHash`, which serves the same purpose but with a shorter content.

apps/web/src/stores/documentLogWithMetadata.ts
geclos: yeah i would actually wouldn't have used detail  which is the really confusing part

packages/core/src/lib/hashContent.ts
```diff
@@ -0,0 +1,5 @@
+import { hash } from 'crypto'
+
+export function hashContent(content: string) {
+  return hash('md5', content)
```
geclos, line 4: any reason for using md5?

packages/core/src/repositories/documentVersionsRepository/getDocumentAtCommit.test.ts
```diff
@@ -32,8 +32,10 @@ describe('getDocumentAtCommit', () => {
     })
     const document = result.unwrap()
 
-    expect(omit(document, 'id', 'updatedAt', 'resolvedContent')).toEqual({
-      ...omit(doc, 'id', 'updatedAt', 'resolvedContent'),
+    expect(
+      omit(document, 'id', 'updatedAt', 'resolvedContent', 'contentHash'),
+    ).toEqual({
+      ...omit(doc, 'id', 'updatedAt', 'resolvedContent', 'contentHash'),
```
geclos, line 38: menuda expectation just use expect.objectContaining and remove all this mambo

apps/web/src/services/routes/api.ts
```diff
@@ -0,0 +1,41 @@
...
+            },
+          }
+        },
+        forImport: {
+          root: `${documentsAtProject}/for-import`,
+        },
+      }
+    },
+  },
+}
```
geclos, line 41: usually not the best idea to do these kind of tangential refactors when you are in a mega ultra massive PR already, a tip for next times :)

apps/web/src/stores/documentLogWithMetadata.ts
andresgutgon: detail is what we use in the other routes

geclos: yeah but this is rest paths in which a detail noun makes no really sense, but anyhow we are really just discussing minor stuff here

packages/core/src/lib/hashContent.ts
csansoon: same as postgres

apps/web/src/services/routes/api.ts
csansoon: This was asked for in another smaller PR. This PR is just the whole trenecito ready to merge. It was split into several smaller ones for easier review.

PR #340, opened 2024-10-03
Move validation of chain step to StepValidator

andresgutgon:
# What? 
Extract step validation of a chain step and return a result. The idea is to categorize better the kind of errors that can happen while running a prompt.

## TODO
- [ ] Fix tests
- [ ] QA run prompt
- [ ] QA run evaluation
- [ ] QA chat API


apps/web/src/actions/datasets/generateDataset.ts
```diff
@@ -1,7 +1,7 @@
 'use server'
 
 import {
-  ChainObjectResponse,
+  ChainStepResponse,
```
andresgutgon, line 4: Make this generic to simplify types

packages/core/src/services/chains/run.ts
```diff
@@ -96,106 +95,120 @@ async function iterate({
...
-    const step = await computeStepData({
-      chain,
-      previousResponse,
-      apikeys,
-      apiKey: previousApiKey,
-      sentCount: previousCount,
-    })
+  const prevText = previousResponse?.text
+  const stepValidator = new ChainValidator({ prevText, chain, providersMap })
+  const step = await stepValidator.call().then((r) => r.unwrap())
```
andresgutgon, line 107: For now we keep failing as before. Only difference is that all the errors that come from validator are already categorized and ready to be saved in `run_errors` table. 

I'm thinking on handling the saving on the `.catch` in the top promise

csansoon (approved):

packages/core/src/constants.ts
```diff
@@ -46,27 +46,30 @@ export const HELP_CENTER = {
...
-}
-export type ChainObjectResponse = ChainStepObjectResponse & {
-  providerLog: ProviderLog
-}
-export type ChainCallResponse = ChainTextResponse | ChainObjectResponse
+export type ChainStepResponse<T extends StreamType> = T extends 'text'
+  ? ChainStepTextResponse
+  : T extends 'object'
+    ? ChainStepObjectResponse
+    : never
```
geclos, line 72: lovely

packages/core/src/services/chains/ProviderProcessor/saveOrPublishProviderLogs.ts
```diff
@@ -13,7 +11,7 @@ export type LogData<T extends StreamType> = T extends 'text'
   ? Awaited<TextProviderLogsData>
   : T extends 'object'
     ? Awaited<ObjectProviderLogsData>
-    : never
+    : unknown
```
geclos, line 14: hmm why? should never happen no?

packages/core/src/services/chains/run.ts
```diff
@@ -42,18 +40,20 @@ export async function runChain({
...
   generateUUID?: () => string
   source: LogSources
-  apikeys: CachedApiKeys
+  providersMap: CachedApiKeys
   configOverrides?: ConfigOverrides
 }) {
   const documentLogUuid = generateUUID()
 
-  let responseResolve: (value: ChainCallResponse) => void
-  let responseReject: (reason?: any) => void
```
geclos, line 51: nice

packages/core/src/services/chains/run.ts
```diff
@@ -96,106 +95,120 @@ async function iterate({
...
-        previousCount: step.sentCount,
-        previousResponse: response as ChainTextResponse,
+        previousCount: sentCount,
+        previousResponse: response,
         configOverrides,
       })
     }
-  } catch (error) {
-    handleIterationError(controller, error)
+  } catch (e: unknown) {
```
geclos, line 155: can we keep the code in the catch in its own method? this method is hard enough to read already

packages/core/src/services/evaluations/run.ts
```diff
@@ -101,52 +101,50 @@ export const runEvaluation = async (
...
-  // Call the new method to handle the promise
-  chainResult.value.response.then((response) =>
-    handleEvaluationResponse(response, documentUuid, evaluation, documentLog),
+  const response = run.response as Promise<ChainStepResponse<'object'>>
+  response.then((res) =>
+    handleEvaluationResponse(res, documentUuid, evaluation, documentLog),
   )
 
-  return chainResult
+  return Result.ok(run)
```
geclos, line 123: why

packages/core/src/services/evaluations/run.ts
```diff
@@ -101,52 +101,50 @@ export const runEvaluation = async (
...
     },
   })
 
-  if (chainResult.error) return chainResult
-
-  // Call the new method to handle the promise
-  chainResult.value.response.then((response) =>
-    handleEvaluationResponse(response, documentUuid, evaluation, documentLog),
+  const response = run.response as Promise<ChainStepResponse<'object'>>
+  response.then((res) =>
```
geclos, line 119: hmm there's where I display my promises lack of knowledge but shouldn't you need to return res here so that if anyone is listening to this promise they keep getting the response?

geclos (approved):

packages/core/src/services/chains/ProviderProcessor/saveOrPublishProviderLogs.ts
andresgutgon: `tsc` failed with `never` https://chatgpt.com/share/e/66ffb37b-d09c-800a-8d8c-45618c0b8ddd

packages/core/src/services/chains/run.ts
andresgutgon: I'm working on that on the next PR

packages/core/src/services/evaluations/run.ts
andresgutgon: I removed `Result` from `runChain` but this service was returning Result and don't want to introduce more entropy

packages/core/src/services/evaluations/run.ts
andresgutgon: `response` is included in `run` so anyone consuming it can attach their `.then`. You did this xD

PR #342, opened 2024-10-03
Fix dataset generation not working in development

andresgutgon:
# What?
GENERATE_DATASET_WORKSPACE_API_KEY was hardcoded to a specific key in one dev's machine. We need to set this from env variables so anyone can configure it

### DIFF
--- a/apps/web/src/env.ts
+++ b/apps/web/src/env.ts
@@ -10,10 +10,12 @@ export default createEnv({
     NODE_ENV: z.string(),
     DATABASE_URL: z.string(),
     WEBSOCKETS_SERVER: z.string(),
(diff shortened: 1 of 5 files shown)


apps/web/src/env.ts
```diff
@@ -10,10 +10,12 @@ export default createEnv({
     NODE_ENV: z.string(),
     DATABASE_URL: z.string(),
     WEBSOCKETS_SERVER: z.string(),
+    DATASET_GENERATOR_WORKSPACE_APIKEY: z.string(),
```
geclos, line 13: nein, declare it in packages/env please 🙏🏼 , I'm in the process of removing this env file

PR #347, opened 2024-10-04
Move validation of chain step to StepValidator (SECOND TRY)

andresgutgon:
# What?
Second attend for this [PR](https://github.com/latitude-dev/latitude-llm/pull/340)

## TODO
- [x] Fix JSON generation

### DIFF
--- a/packages/core/src/services/chains/run.ts
+++ b/packages/core/src/services/chains/run.ts
@@ -1,25 +1,23 @@
-import { Chain, MessageRole } from '@latitude-data/compiler'
+import { Chain, Message, MessageRole } from '@latitude-data/compiler'
 import { CoreTool, ObjectStreamPart, TextStreamPart } from 'ai'
 import { JSONSchema7 } from 'json-schema'
 import { v4 } from 'uuid'
-import { ZodError } from 'zod'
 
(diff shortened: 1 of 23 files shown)


packages/core/src/services/chains/run.ts
```diff
@@ -218,44 +231,33 @@ async function streamAIResult(
...
+  response: ChainStepResponse<StreamType>,
 ) {
-  const eventData = {
-    type: ChainEventTypes.Complete,
-    config: stepResult.conversation.config as Config,
-    documentLogUuid: response.documentLogUuid,
-    response,
-  } as const
-
-  if ('text' in response) {
```
andresgutgon, line 231: This was the bug. In any case an AI response sent the same. It was always entering this `if` because an object response also has `.text` attribute. So the else condition was never used. 

What I did in my first attempt was to use my fancy `streamType` thinking this code worked as expeted
```javascript
if (response.streamType === 'text') {
} else if (response.streamType === 'object') {
  // Sending this instead of `messages` was the 🐛 
  Object.assign(eventData, {
    object: (response as ChainObjectResponse).object,
    text: objectToString((response as ChainObjectResponse).object),
  })
}
```

geclos: Test test test test

PR #349, opened 2024-10-04
Feature/remove cyclical dependencies everywhere

geclos:


packages/core/src/events/events.d.ts
andresgutgon, line 1: Why a `.d.ts`?

geclos: why not it's just types 🤷🏼

andresgutgon: Don't know, what's the advantage? Does it have any tradeoff?

PR #350, opened 2024-10-04
Diff editor

csansoon:
Added an optional `diff` property to the Text Editor, which must contain: `newValue`, `description`, `onAccept` and `onReject`.
When this property is defined, the editor turns into a diff mode, where the regular `value` is shown as the original (red) and `diff.newValue` is shown as modified (green). Additionally, a new menu at the bottom of the editor appears with the description, and an accept and reject buttons that will just run the passed functions. The diff editor can be modified before accepting, although all changes are discarded if it's not accepted.

The diff editor currently has no limitations on read-only editors. The actor is responsible for controlling this state and running the actions triggered by ⁠onAccept and ⁠onReject.

image


packages/web-ui/src/ds/molecules/DocumentTextEditor/Editor/DiffEditor.tsx
```diff
@@ -0,0 +1,120 @@
...
+      }
+
+      editor.layout()
+    },
+    [],
+  )
+
+  return (
+    <div className='flex flex-grow relative'>
+      <div className='absolute top-0 left-0 right-0 bottom-0'>
```
andresgutgon, line 82: Do a `<EditorWrapper>` with these divs

geclos (approved):
look good 👌🏼

PR #355, opened 2024-10-06
feature: better error messages on provider api keys and commits actions

geclos:


packages/core/src/lib/errors.ts
```diff
@@ -50,3 +50,8 @@ export class UnauthorizedError extends LatitudeError {
   public statusCode = 401
   public name = 'UnauthorizedError'
 }
+
+export const databaseErrorCodes = {
+  foreignKeyViolation: '23503',
+  uniqueViolation: '23505',
+}
```
andresgutgon, line 57: This should be here https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/lib/Transaction.ts#L21

packages/core/src/services/providerApiKeys/destroy.ts
```diff
@@ -1,21 +1,42 @@
...
+        .returning()
+      const deleted = result[0]
 
-    return Result.ok(deleted)
+      return Result.ok(deleted)
+    } catch (error) {
+      if (
+        error instanceof DatabaseError &&
+        error.code === databaseErrorCodes.foreignKeyViolation
+      ) {
```
andresgutgon, line 31: This is not the right way of handle this. We don't have to wait to arrive to the DB to fail. This has to be validated in the action with a zod async validation. 

Example:
https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/actions/user/setupAction.ts#L22-L25

andresgutgon: This is like if do a catch in a Rails transaction. We we would do that having models and validations?

andresgutgon: Please change it to don't set a bad precent

geclos: i disagree why add an extra db roundtrip to all requests for a validation when you can just let the db validation fail and return a nicely formatted error 🤷🏼

packages/core/src/lib/errors.ts
geclos: will move even if this is just ocd

packages/core/src/services/providerApiKeys/destroy.ts
geclos: i absolutely do not think this is a bad precedent

andresgutgon: But this is exactly how Rails works. Is a validation check againts Db. Why matters the roundtrip in a POST operation? I don't think is the bottle neck.

Also services should receive valid data the same way they receive valid scopes.

Having to deal with try/catch feels nasty. Having zod which provides a nice layer between client and server

packages/core/src/lib/errors.ts
andresgutgon: How is this ocd? Is the same unique violation error in two different places. For that matter, put it in the destroy service as a local constant

packages/core/src/services/providerApiKeys/destroy.ts
andresgutgon: Any way. I said my part feel free to ignore

Issue #330, opened 2024-10-02
Clain reward popover doesn't close after claiming


andresgutgon:
<img width="527" alt="image" src="https://github.com/user-attachments/assets/f931af43-8fc6-4d2a-af5f-12e3998d2f92">

andresgutgon:
Also both popover are cropped. Compare to account
<img width="240" alt="image" src="https://github.com/user-attachments/assets/36241406-6c2b-4b49-8b00-a4e1ab89192b">

PR #356, opened 2024-10-06
chore: fix claim rewards service

geclos:
It was accepting the referral even if the user already existed. Also added tests.


andresgutgon:
Is this working for you?
https://github.com/latitude-dev/latitude-llm/issues/330

When I tried in production popover didn't closed

geclos:
didn't fix that bug, just this:

> It was accepting the referral even if the user already existed. Also added tests.

PR #359, opened 2024-10-07
Better serialized logs types

csansoon:
Improve types on Serialized Logs, as it is now going to be used in multiple places.


packages/web-ui/src/ds/tokens/skeleton.ts
```diff
@@ -0,0 +1,14 @@
+export const skeleton = {
+  height: {
+    h8: 'h-[8px]', // 8px
```
andresgutgon, line 3: Review always tailwind please. `h8` is `h-2` and `h7` is `h-2.5` 
https://tailwindcss.com/docs/height

packages/web-ui/src/ds/tokens/skeleton.ts
```diff
@@ -0,0 +1,14 @@
...
+    h7: 'h-[10px]', // 10px
+    h6: 'h-3', // 12px
+    h5: 'h-3.5', // 14px
+    h4: 'h-4', // 16px
+    h3: 'h-5', // 20px
+    h2: 'h-[26px]', // 26px
```
andresgutgon, line 9: Put this in tailwindConfig. The same way we have the font

packages/web-ui/src/ds/tokens/skeleton.ts
```diff
@@ -0,0 +1,14 @@
...
+    h1: 'h-9', // 36px
+  }
```
andresgutgon, line 11: Last, order from `h1` to `h8` as the font. Is breaking my brain having it in reverse order 😂

PR #361, opened 2024-10-07
Prompt refiner

csansoon:
Created the prompt refiner action and UI.
For now, the Refine button is only available from a draft.

https://github.com/user-attachments/assets/fba29e70-588e-439e-83f0-a444d05424db


apps/web/src/components/TablePaginationFooter/index.tsx
```diff
@@ -42,7 +42,13 @@ export function TablePaginationFooter({
         <div className='flex flex-row items-center gap-x-1'>
           <Text.H5M color='foregroundMuted'>Page</Text.H5M>
           <div className='max-w-14'>
-            <GoToPageInput key={pagination.page} pagination={pagination} />
+            <GoToPageInput
+              key={pagination.page}
+              page={pagination.page}
+              totalPages={pagination.totalPages}
+              pageSize={pagination.pageSize}
+              baseUrl={pagination.baseUrl}
```
andresgutgon, line 50: Why?

packages/web-ui/src/ds/tokens/skeleton.ts
```diff
@@ -0,0 +1,14 @@
+export const skeleton = {
+  height: {
+    h8: 'h-[8px]', // 8px
+    h7: 'h-[10px]', // 10px
+    h6: 'h-3', // 12px
+    h5: 'h-3.5', // 14px
+    h4: 'h-4', // 16px
+    h3: 'h-5', // 20px
+    h2: 'h-[26px]', // 26px
+    h1: 'h-9', // 36px
```
andresgutgon, line 10: All I comment in other place.

apps/web/src/actions/copilot/refinePrompt.ts
```diff
@@ -0,0 +1,87 @@
...
+  .handler(async ({ ctx, input }) => {
+    if (!env.COPILOT_WORKSPACE_APIKEY) {
+      throw new BadRequestError('COPILOT_WORKSPACE_APIKEY is not set')
+    }
+    if (!env.COPILOT_PROJECT_ID) {
+      throw new BadRequestError('COPILOT_PROJECT_ID is not set')
+    }
+    if (!env.COPILOT_REFINE_PROMPT_PATH) {
+      throw new BadRequestError('COPILOT_REFINE_PROMPT_PATH is not set')
+    }
```
andresgutgon, line 36: Not to change now. But is a pain to have to define all these checks. I think we should make all mandatory and let our env take care of it on runtime. By [setting it here](https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/env.ts#L14)

andresgutgon (approved):

packages/web-ui/src/ds/tokens/skeleton.ts
csansoon: talked offline

packages/core/src/repositories/repository.ts
```diff
@@ -59,6 +59,29 @@ export default abstract class Repository<
...
+      const missingIds = ids
+        // @ts-expect-error
+        .filter((id) => !result.find((r) => r.id === id))
+        .map((id) => "'" + id + "'")
+        .join(', ')
+
+      return Result.error(
+        new NotFoundError(`Records with ids ${missingIds} not found`),
+      )
+    }
```
geclos, line 80: hmm a bit surprised this is in base repository since it has some funky behaviour i dont expect a findmany to fail if it does not find all of the ids provided

csansoon: hmm, really? I'd expect it to return an array of N items when I entered an array of N ids

csansoon: Easy change though

PR #364, opened 2024-10-08
feature: GET document endpoint in JS/TS sdk

geclos:


packages/sdks/typescript/src/index.ts
```diff
@@ -74,6 +97,38 @@ export class Latitude {
...
+
+    if (!response.ok) {
+      throw new LatitudeApiError({
+        status: response.status,
+        statusText: response.statusText,
+        serverResponse: await response.text(),
+      })
+    }
+
+    return (await response.json()) as DocumentVersion & { config: Config }
```
andresgutgon, line 129: I realize we're not handling `.json()`. This can fail no?

andresgutgon (dismissed):

packages/sdks/typescript/src/index.ts
geclos: yes it can

csansoon (dismissed):

PR #365, opened 2024-10-08
Add UUID to evaluation results and populate the column with a data migration

andresgutgon:
# What?
Add UUID to evaluation results and populate the column with a data migration

We need to have a `errorableId` that can be created before the document log or the evaluation results are created. 

Creating the errors happens inside `runChain` and for that we pass a pre-generated UUID. 

Evaluation results don't have `uuid` as document logs. 

This PR add the field and also introduce a very basic version of data migrations to be run inside production web container agains PRODUCTION database. 

After this migration is run successfully, I'll set `evaluation_results.uuid` to NOT NULL


packages/core/src/schema/models/evaluationResults.ts
```diff
@@ -17,6 +17,7 @@ export const evaluationResults = latitudeSchema.table(
   'evaluation_results',
   {
     id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
+    uuid: uuid('uuid').unique(),
```
andresgutgon, line 20: For now until we implement proper uuid pasing from `chainRun` the uuid is auto generated when the evaluation result is created.

packages/core/src/tests/useTestDatabase.ts
```diff
@@ -1,20 +1,9 @@
...
-  beforeEach(startTransaction)
-  afterEach(rollbackTransaction)
-  afterAll(async () => {
-    unpatchPgForTransactions()
-    await close()
-  })
+  beforeAll(testTransaction.start)
+  beforeEach(testTransaction.start)
+  afterEach(testTransaction.rollback)
+  afterAll(testTransaction.close)
```
andresgutgon, line 8: I bumped the version of this package and they changed the way is setup. 
If doesn't work I'll move back to prev version

https://github.com/romeerez/pg-transactional-tests/commit/79c60a96b57676212d50b2dbc608ea83212fc63a

geclos (dismissed):

PR #367, opened 2024-10-08
feature: subscriptions data model

geclos:


packages/core/src/schema/models/workspaces.ts
```diff
@@ -1,14 +1,41 @@
...
   id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
   name: varchar('name', { length: 256 }).notNull(),
+  currentSubscriptionId: bigint('current_subscription_id', {
+    mode: 'number',
+  }).references(() => subscriptions.id),
   creatorId: text('creator_id').references(() => users.id, {
     onDelete: 'set null',
   }),
   ...timestamps(),
 })
```
csansoon, line 18: We should add the time when the subscription changed, so that we can calculate when credits are renewed based on it.

geclos: I don't understand this comment 😅

csansoon: The credit usage is calculated as the total of `documentLogs` and `evaluationResults` created in the workspace since a certain date.

This date is currently determined by the workspace creation date. For example, if a user created their workspace on July 6th, it will be renewed on the 6th of every month thereafter.

However, when a subscription plan is added, we need to calculate the date from the last time the user changed their subscription plan.

geclos: oh i see

packages/core/src/plans.ts
```diff
@@ -0,0 +1,28 @@
...
+    users: undefined,
+  },
+  [SubscriptionPlan.TeamV1]: {
+    name: 'Team',
+    credits: 100_000,
+    users: undefined,
+  },
+  [SubscriptionPlan.EnterpriseV1]: {
+    name: 'Enterprise',
+    credits: undefined,
```
andresgutgon, line 26: Missing users

packages/core/src/schema/models/workspaces.ts
```diff
@@ -12,3 +13,26 @@ export const workspaces = latitudeSchema.table('workspaces', {
...
+
+export const plansEnum = latitudeSchema.enum('subscription_plans', [
+  SubscriptionPlan.HobbyV1,
+  SubscriptionPlan.HobbyV2,
+  SubscriptionPlan.TeamV1,
+  SubscriptionPlan.EnterpriseV1,
+])
+
+export const subscriptions = latitudeSchema.table(
```
andresgutgon, line 24: Do a `models/subscriptions.ts` we always do one model by file

packages/core/src/schema/models/workspaces.ts
```diff
@@ -12,3 +13,26 @@ export const workspaces = latitudeSchema.table('workspaces', {
   }),
   ...timestamps(),
 })
+
+export const plansEnum = latitudeSchema.enum('subscription_plans', [
+  SubscriptionPlan.HobbyV1,
+  SubscriptionPlan.HobbyV2,
```
andresgutgon, line 19: Why v1 and v2? Why not different names?

packages/core/src/schema/models/workspaces.ts
geclos: That would be the subscription creation date

packages/core/src/schema/models/workspaces.ts
geclos: Not really important tbh but at least an incremental pattern gives some extra info

PR #368, opened 2024-10-08
feature: compute credits from current subscription plan

geclos:


packages/core/src/services/workspaces/usage.ts
```diff
@@ -36,9 +37,8 @@ export async function computeWorkspaceUsage(
     db,
   )
 
-  const createdAtDate = workspace.createdAt
+  const createdAtDate = workspace.currentSubscription.createdAt
```
geclos, line 40: this can't be merged atm because we would be effectively resetting the credit count for all users.

csansoon: Either create a migration to set the default subscription.created_at date for each workspace the same as the workspace.created_at,

or just merge it and some people will be happy about it xd

csansoon (approved):

packages/core/src/services/workspaces/usage.ts
andresgutgon: What about a data migration that sets subscription.created_at = workspace.created_at?

packages/core/src/plans.ts
```diff
@@ -14,16 +14,16 @@ export const SubscriptionPlans = {
...
     name: 'Team',
     credits: 100_000,
-    users: undefined,
+    users: 5,
   },
   [SubscriptionPlan.EnterpriseV1]: {
     name: 'Enterprise',
-    credits: undefined,
-    users: undefined,
+    credits: 1_000_000,
```
andresgutgon, line 26: Credits are until they finish or for each month?

andresgutgon (approved):

packages/core/src/plans.ts
geclos: monthly

PR #371, opened 2024-10-08
Populate evaluation result null UUIDs  with SQL migration

andresgutgon:
# What?
I tried to setup a javascript data migration system but I failed and for now we can go with pure SQL. Reverting prev atempt


packages/core/drizzle/0069_gifted_kulan_gath.sql
```diff
@@ -0,0 +1 @@
+UPDATE "latitude"."evaluation_results" SET uuid = gen_random_uuid() WHERE uuid IS NULL;
```
andresgutgon, line 1: The migration

PR #373, opened 2024-10-09
Do all data model changes to finish storing errors.

andresgutgon:
# What?
Schema changes needed for storing errors

- [x] Make not null `evaluation_result.uuid`
- [x] Add `finish_reason` to provider logs
- [x] Add column `errorable_uuid` to run_errors  

## Another PR
- [x] Fix evaluation results and document logs repositories by pointing to uuid instead of id. [here](https://github.com/latitude-dev/latitude-llm/pull/375)

### DIFF
--- a/packages/core/src/repositories/runErrors/evaluationResultsRepository/index.test.ts
+++ b/packages/core/src/repositories/runErrors/evaluationResultsRepository/index.test.ts
@@ -59,6 +59,7 @@ describe('EvaluationResultsRepository', () => {
       await createRunError({
         errorableType: ErrorableEntity.EvaluationResult,
         errorableId: evaluationResult.id,
+        errorableUuid: evaluationResult.uuid,
         code: RunErrorCodes.Unknown,
         message: 'Error message',
       })
@@ -105,6 +106,7 @@ describe('EvaluationResultsRepository', () => {
       await createRunError({
         errorableType: ErrorableEntity.EvaluationResult,
         errorableId: evaluationResult2.id,
+        errorableUuid: evaluationResult2.uuid,
         code: RunErrorCodes.Unknown,
         message: 'Error message',
       })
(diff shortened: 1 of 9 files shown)


packages/core/src/schema/models/evaluationResults.ts
```diff
@@ -17,7 +17,7 @@ export const evaluationResults = latitudeSchema.table(
   'evaluation_results',
   {
     id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
-    uuid: uuid('uuid').unique(),
+    uuid: uuid('uuid').unique().notNull(),
```
andresgutgon, line 20: This will be safe once this is fully deployed and migrated
https://github.com/latitude-dev/latitude-llm/pull/371

This should be 0 after data migration
image

packages/core/src/schema/models/providerLogs.ts
```diff
@@ -35,6 +35,7 @@ export const providerLogs = latitudeSchema.table('provider_logs', {
       onUpdate: 'cascade',
     }),
   model: varchar('model'),
+  finishReason: varchar('finish_reason').notNull().default('stop'),
```
andresgutgon, line 38: We want to save what was the AI reason to finish. By default we set `stop` because is the most common reason

andresgutgon:
Ready
image

packages/core/src/repositories/runErrors/evaluationResultsRepository/index.test.ts
```diff
@@ -105,6 +106,7 @@ describe('EvaluationResultsRepository', () => {
       await createRunError({
         errorableType: ErrorableEntity.EvaluationResult,
         errorableId: evaluationResult2.id,
+        errorableUuid: evaluationResult2.uuid,
```
geclos, line 109: :\ you are editing the tests but not the code?

geclos (approved):

packages/core/src/repositories/runErrors/evaluationResultsRepository/index.test.ts
andresgutgon: This interactor is not used, only here in these tests. I made NOT NULL `errorableUuid` so this was complaining

PR #375, opened 2024-10-09
Replace run_errors.errorable_id by run_errors.errorable_uuid

andresgutgon:
# What?
This PR comes after this [one is merged and deployed](https://github.com/latitude-dev/latitude-llm/pull/373)


packages/core/src/repositories/evaluationResultsRepository/index.ts
```diff
@@ -78,7 +78,7 @@ export class EvaluationResultsRepository extends Repository<
       .leftJoin(
         runErrors,
         and(
-          eq(runErrors.errorableId, evaluationResults.id),
+          eq(runErrors.errorableUuid, evaluationResults.uuid),
```
andresgutgon, line 81: This is critical change. But not so critical because there are 0 errors in DB now

geclos (approved):

PR #378, opened 2024-10-09
Extended and paginated results on refine modal

csansoon:
Now `EvaluationResultsWithMetadata` also contain a `documentContentHash` attribute, which can be used to determine whether an evaluation result was made from a log for an exact document content or not.

The list of evaluation results in the Refine modal menu is now fully paginated, and includes all results from logs made either from this commit, or any other commit that has the same contentHash.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,12 +5,7 @@
       "name": "Debug Web",
       "type": "node-terminal",
       "request": "launch",
-      "command": "cd apps/web && pnpm dev",
-      "serverReadyAction": {
-        "pattern": "- Local:.+(https?://.+)",
-        "uriFormat": "%s",
-        "action": "debugWithChrome"
-      }
+      "command": "cd apps/web && pnpm dev"
     }
   ]
 }
--- a/packages/core/src/services/evaluationResults/computeEvaluationResultsByDocumentContent.ts
+++ b/packages/core/src/services/evaluationResults/computeEvaluationResultsByDocumentContent.ts
@@ -1,21 +1,26 @@
+import { and, desc, eq, or } from 'drizzle-orm'
+
 import { Commit, Evaluation } from '../../browser'
 import { database } from '../../client'
-import { hashContent, Result } from '../../lib'
-import {
-  DocumentVersionsRepository,
-  EvaluationResultsRepository,
-} from '../../repositories'
+import { hashContent, paginateQuery, Result } from '../../lib'
+import { DocumentVersionsRepository } from '../../repositories'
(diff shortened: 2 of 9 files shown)


.vscode/launch.json
```diff
@@ -5,12 +5,7 @@
       "name": "Debug Web",
       "type": "node-terminal",
       "request": "launch",
-      "command": "cd apps/web && pnpm dev",
-      "serverReadyAction": {
-        "pattern": "- Local:.+(https?://.+)",
-        "uriFormat": "%s",
-        "action": "debugWithChrome"
-      }
```
csansoon, line 13: This tries to open chrome automatically when starting the server, which I don't want.
I am adding this to the repo because this is a standard can be used in any supported editor, not just mine.

packages/core/src/services/evaluationResults/computeEvaluationResultsByDocumentContent.ts
```diff
@@ -35,15 +36,21 @@ export async function computeEvaluationResultsByDocumentContent(
...
-    evaluationId: evaluation.id,
-    contentHash: hashContent(resolvedContent),
-  })
-  if (result.error) return result
 
-  return Result.ok(result.unwrap())
+  const { evaluationResultsScope, documentLogsScope, baseQuery } =
+    createEvaluationResultQuery(workspaceId, db)
+
+  const dynamicQuery = baseQuery
```
andresgutgon, line 43: Why is called dynamic? Is not `.$dynamic` in this sense no?

packages/core/src/services/evaluationResults/computeEvaluationResultsByDocumentContent.ts
```diff
@@ -35,15 +36,21 @@ export async function computeEvaluationResultsByDocumentContent(
...
+    .where(
+      and(
+        eq(evaluationResultsScope.evaluationId, evaluation.id),
+        or(
+          eq(documentLogsScope.contentHash, hashContent(resolvedContent)),
+          eq(commits.id, commit.id),
+        ),
+      ),
+    )
+    .orderBy(desc(evaluationResultsScope.createdAt))
```
andresgutgon, line 53: Do we have test for this new code?

geclos (dismissed):

packages/core/src/services/evaluationResults/computeEvaluationResultsByDocumentContent.ts
csansoon: yes

PR #380, opened 2024-10-09
Gradient Mesh Background

csansoon:
Actual Mesh Gradient shader (MeshGradient/Gradient.js file) made by Stripe and [Kevin Hufnagl](https://kevinhufnagl.com/).
[Here](https://whatamesh.vercel.app/)'s where I got it from.

Some things can be improved, and it may not need the lazy loader anymore. Can be easily revised in the future.


geclos:
closing for now

PR #381, opened 2024-10-09
Save errors when a chain run

andresgutgon:
## What?
This is the part where `run_errors` table is filled with all possible errors that can happen when a chain run is executed. 

## QA
- [x] QA run document successfully 
- [x] QA run document with error from-AI
- [x] QA chat in playground.
- [x] QA run document with error pre-AI
- [x] QA run evaluation successfully
- [x] QA run evaluation with error pre-AI
- [x] QA run evaluation with error from-AI
- [x] QA live evaluation job. Check it works.
- [x] QA run prompt in evaluations

## Left overs PR
- [x] Style progress bar indicator with new design [THIS PR](https://github.com/latitude-dev/latitude-llm/pull/465)
- [ ] Address [this comment](https://github.com/latitude-dev/latitude-llm/pull/381/files#r1806194348)

# Before merge this PR

### What's a valid evaluation?
A valid evaluation is one that:
1. Has not `run_errors` associated to their `uuid`
2. Has `provider_id` relation filled
3. Has `resultable` relatition filled.
With this description I think we're safe when generating evaluation results with errors. 
☝️ These checks [were done here](https://github.com/latitude-dev/latitude-llm/pull/401)

### Sentry errors 
With these changes, I did my best to contemplate all the places where an evaluation can fail. 
1. **On the job before calling runEvaluation**: Because document log or evaluation are not found. This throw an error and is not captured. I guess is fine because an evaluation result without evaluation doesn't makes sense. I think this will be catch for Sentry
2. **On runEvaluation Before calling `runChain`** this can fail because documentLog evaluated does not have provider logs (other cases already covered with tests)
3. **While calling runChain before calling AI service**: This can be errors like miss configured provider or a missconfigured AI provider. Or missing provider or compiler error.
4. **While running AI service**: If the AI service throw an error we catch it. If the AI service runs but ends with `finishReason` `error` or directly type `error` we catch it and return it in the `response`
5. **After `runChain` was called**: Here we catch when the response of an evaluation is not an object. Which happen if they configure the evaluation with Grok.
6. **In the job after runEvaluation is executed** When response.error from `runChain` is of type `unknown` this means we catch an error but we don't know what's. We still store as an error in the DB but we should `throw` the error in the Job so it can be catch by Sentry

☝️ This section is called **Sentry errors** but really only (1) and (6) throw an error that I think Sentry catch.


packages/core/src/services/chains/run.ts
```diff
@@ -121,7 +119,6 @@ async function iterate({
   const streamConsumer = new ChainStreamConsumer({
     controller,
     previousCount,
-    documentLogUuid,
```
andresgutgon, line 124: Review this. Bold move

packages/core/src/services/evaluationResults/create.ts
```diff
@@ -7,64 +7,77 @@ import {
...
       .values({
-        uuid: generateUUIDIdentifier(),
+        uuid,
         evaluationId: evaluation.id,
         documentLogId: documentLog.id,
-        providerLogId: providerLog.id,
         resultableType: evaluation.configuration.type,
-        resultableId: metadata[0]!.id,
         source: documentLog.source,
+        resultableId: metadataId,
```
andresgutgon, line 77: `resultableId` was already nilable. Any reason? Not complaining this is great for my use case

packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.ts
```diff
@@ -1,15 +1,51 @@
...
 
+function throwIfUnknownError(
+  error: unknown | undefined,
+): asserts error is Error {
+  const isAllGood =
+    !error || (isChainError(error) && error.errorCode !== RunErrorCodes.Unknown)
+
+  if (isAllGood) return
+
+  throw error
```
andresgutgon, line 23: This is important. We only throw an error in an evaluation run in 2 situations

1. Is a ChainError of type unknown. This is the catch in `runChain` not knowing what happened to this has to bubble to Sentry. Hopefully. 
2. is not a `ChainError` so it means something exploded.


Any other `ChainError` is known, saved and should be ignored by Sentry

packages/core/src/services/prompts/run.ts
```diff
@@ -29,6 +29,12 @@ export async function runPrompt({
   }
 
   const chain = createChain({ prompt: metadata.resolvedPrompt, parameters })
-  const run = await runChain({ workspace, chain, providersMap, source })
+  const run = await runChain({
+    workspace,
+    chain,
+    providersMap,
+    source,
+    persistErrors: false,
```
andresgutgon, line 37: This is an interesting case. We just run the prompt, we show the error in the evaluation playground but is not associated with a document or evaluation so no error to persist. 

Please confirm this is what we want

packages/core/src/events/handlers/notifyToClientEvaluationResultCreatedJob.ts
```diff
@@ -10,6 +10,8 @@ export const notifyToClientEvaluationResultCreatedJob = async ({
   data: EvaluationResultCreatedEvent
 }) => {
   const { evaluation, documentLog, evaluationResult } = event.data
+  // FIXME: Change this query to include the error
+  // This has to be done with care because it can break aggregations
```
andresgutgon, line 14: Once we have the errors saved into the DB and we see everything is working will show it in the UI, in evaluation results table and document logs table. 

This comment is to send in the realtime the error together with the evaluation result.

I'll do this in another PR.

packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.test.ts
```diff
@@ -61,84 +42,181 @@ vi.spyOn(progressTracker, 'ProgressTracker').mockImplementation(() => ({
...
-      data: {
-        batchId: 'batch-123',
-        evaluationId: 2,
-        documentUuid: 'doc-uuid',
-        completed: 1,
-        total: 1,
-      },
+      await expect(runEvaluationJob(jobData)).rejects.toThrowError(
+        new Error('Some error'),
+      )
```
andresgutgon, line 181: This is testing that we don't need to wait for the response. It was weird

```javascript
const { response } = runEvaluation()
await response
```

Now it will fail if 
1. There is an error before AI is called
2. There is an error in AI response

All we have to do is wait to finish and we get a Result
```javascript
const result = runEvaluation()
```

packages/core/src/services/chains/run.ts
```diff
@@ -147,26 +178,31 @@ async function iterate({
...
-      streamConsumedResult: aiStreamConsumedResult,
-      startTime: stepStartTime,
-    })
+    const response = await providerProcessor
+      .call({
+        aiResult,
+        finishReason: consumedStream.finishReason,
+        startTime: stepStartTime,
+      })
+      .then((r) => r.unwrap())
```
andresgutgon, line 191: Now if the AI respond with a chunks that are of `type: 'error'` or { type: 'finish', finishReason: 'error'}` we throw an error that is persisted in the errorable

packages/core/src/services/documentLogs/serialize.ts
```diff
@@ -1,32 +1,35 @@
...
 import { database } from '../../client'
-import { findWorkspaceFromDocumentLog } from '../../data-access'
 import { NotFoundError, PromisedResult, Result } from '../../lib'
 import { ProviderLogsRepository } from '../../repositories'
 import { serialize as serializeProviderLog } from '../providerLogs'
 
 export async function serialize(
-  documentLog: DocumentLog,
+  {
+    workspace,
```
andresgutgon, line 9: We already had the `workspace` in the 2 places where this is used. Removing this responsibility to this service

packages/core/src/services/evaluationResults/create.test.ts
```diff
@@ -199,8 +204,33 @@ describe('createEvaluationResult', () => {
     })
 
     expect(result.ok).toBe(false)
-    if (result.error) {
-      expect(result.error.message).toContain('Unsupported result type')
-    }
+    expect(result.error!.message).toContain('Unsupported result type')
+  })
+
+  it('saves unsupported type if result is not defined', async () => {
```
andresgutgon, line 210: If we have an evaluation_result without `result` it means something bad happend. Between the things that could happen is that we added for some reason a new evaluation type that's not valid in the DB. In that case we store the error and the evaluation result with `resultableType: null`

packages/core/src/services/evaluations/run.test.ts
andresgutgon, line 1: Evaluation run had 0 tests. Now it's covered I think

packages/core/src/services/evaluations/run.ts
```diff
@@ -102,50 +64,16 @@ export const runEvaluation = async (
...
 
-async function handleEvaluationResponse(
-  response: ChainStepResponse<'object'> | undefined,
-  documentUuid: string,
-  evaluation: EvaluationDto,
-  documentLog: DocumentLog,
-) {
-  if (!response) return
-  if (response.object === undefined) {
-    throw new LatitudeError(
```
andresgutgon, line 128: Evaluation run never throws errors now. If it throws an error is a bug on our code

packages/core/src/services/prompts/run.ts
```diff
@@ -29,6 +29,12 @@ export async function runPrompt({
...
+  })
```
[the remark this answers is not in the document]
andresgutgon: This needs to be confirmed but I think in the case of evaluation playground we don't have logs. Do we?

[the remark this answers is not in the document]
csansoon: I think we do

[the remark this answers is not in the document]
andresgutgon: I'm talking about this screen. If we do save logs for this playground where do we show those logs?
image

packages/core/src/services/evaluations/run.ts
csansoon: Why? Could an evaluation not contain errors in their syntax logic, or with the provider?

andresgutgon: We were checking here if the response from the AI had `object` key in the response. This is still checked but in another place.

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/RefineModal/steps/2_SelectEvaluationResults/SelectableEvaluationResultsTable.tsx
```diff
@@ -1,5 +1,3 @@
-'use client'
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/RefineModal/steps/2_SelectEvaluationResults/SelectableEvaluationResultsTable.tsx:1]: If a component is included in another component already declared as `use client` declaring the child component is not right.

Nextjs complains when passing functions as props. Everything works but you see the warnings
image

packages/core/src/services/chains/ProviderProcessor/saveOrPublishProviderLogs.ts
```diff
@@ -19,27 +17,30 @@ export async function saveOrPublishProviderLogs<T extends StreamType>({
...
   streamType: T
   data: LogData<T>
   saveSyncProviderLogs: boolean
-  // TODO: Use stream result check if contains an AI error to be
-  // persisted in the provider log
-  streamConsumedResult: TypedResult<
-    StreamChunk[],
-    ChainError<RunErrorCodes.AIRunError>
-  >
+  finishReason: StreamConsumeReturn['finishReason']
```
andresgutgon, line 25: ⚠️ Add this to table metadata

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/RefineModal/index.tsx
```diff
@@ -1,5 +1,3 @@
-'use client'
-
 import { ReactNode, useEffect, useMemo, useState } from 'react'
 
 import {
```
Comment (csansoon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/RefineModal/index.tsx:3]: How is it not a client component if it uses `useEffect`, `useMemo` and `useState`?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/RefineModal/steps/2_SelectEvaluationResults/SelectableEvaluationResultsTable.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/RefineModal/steps/2_SelectEvaluationResults/SelectableEvaluationResultsTable.tsx:1]: Hmm I don't see those warnings. Maybe the version of the linter?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/Messages.tsx
```diff
@@ -21,7 +21,13 @@ export function EvaluationResultMessages({
...
   }, [providerLog])
 
-  if (!providerLog) return null
+  if (!providerLog) {
+    return (
+      <Text.H5 color='foregroundMuted' centered>
+        There are no messages generated for this log
+      </Text.H5>
+    )
+  }
```
Comment (csansoon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/Messages.tsx:31]: Sometimes, the messages take some time to load, especially when a batch evaluation is currently being run. We should not display a message stating that a log does not exist when it is only still loading.

Not sure if this is the case here, but just something to take into account!

packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.test.ts
```diff
@@ -61,84 +42,181 @@ vi.spyOn(progressTracker, 'ProgressTracker').mockImplementation(() => ({
...
-    )
-    runEvaluationSpy.mockResolvedValueOnce({
-      unwrap: () =>
-        // @ts-ignore
-        Promise.resolve({
-          response: responsePromise,
+    it('increment successful counter', async () => {
+      runChainResponse = Result.ok({
+        streamType: 'object' as 'object',
+        object: { result: { result: '42', reason: 'Is always 42' } },
```
csansoon, line 119: 👌

packages/core/src/services/prompts/run.ts
csansoon: Is it not associated with the document though?

csansoon (approved):

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/RunErrorMessage/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+export function RunErrorMessage({ error }: { error: RunErrorField }) {
+  const resultError = getRunErrorFromErrorable(error)
+
+  if (!resultError) return null
+
+  let message = resultError.message
+  if (resultError.details) {
+    const keys = Object.keys(resultError.details)
+    for (const key of keys) {
+      // @ts-ignore
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/RunErrorMessage/index.tsx:14]: why this

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/MetadataItem/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+}: MetadataItemProps) {
+  const className = stacked
+    ? 'flex-flex-col gap-2'
+    : 'flex flex-row justify-between items-center gap-2'
+  return (
+    <div className={className}>
+      <Text.H5M color='foreground'>{label}</Text.H5M>
+      <div>
+        {loading ? (
+          <Skeleton className='w-12 h-4 bg-muted-foreground/10' />
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/MetadataItem/index.tsx:30]: the background class shouldn't be needed and you have the height prop for setting height

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/Messages.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/Messages.tsx:31]: People would have to click on the row and on message tab super quickly. I'll leave this way for now. We can always pass `isLoading` from store

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/RunErrorMessage/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/RunErrorMessage/index.tsx:14]: keys content is not typed

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx
```diff
@@ -21,12 +22,15 @@ function useFetchDocumentLog({ documentLogId }: { documentLogId: number }) {
     data: documentLog,
     isLoading,
     error,
-  } = useSWR<DocumentLogWithMetadata>(
+  } = useSWR<DocumentLogWithMetadataAndError>(
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx:25]: i find the naming is getting a bit crazy, could we keep it as it was and assume metadata includes error? 😅

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx
```diff
@@ -21,12 +22,15 @@ function useFetchDocumentLog({ documentLogId }: { documentLogId: number }) {
...
     ['documentLogs', documentLogId],
     useCallback(async () => {
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx:27]: hmm can we use useFetcher? not sure why it ain't used here, probably missed it somehow in my refactor

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DocumentLogInfo/Metadata.tsx
```diff
@@ -87,99 +55,120 @@ export function DocumentLogMetadata({
...
+}
+
+export function DocumentLogMetadata({
+  documentLog,
+  providerLogs: _providerLogs,
+}: {
+  documentLog: DocumentLogWithMetadataAndError
+  providerLogs?: ProviderLogDto[]
+}) {
+  const providerLogs = _providerLogs ?? []
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DocumentLogInfo/Metadata.tsx:142]: just add a default in the argument instead

apps/web/src/app/api/documentLogs/[id]/route.ts
```diff
@@ -14,28 +12,24 @@ export const GET = errorHandler(
...
       }: {
-        params: { id: number }
+        params: { id: string }
         workspace: Workspace
       },
     ) => {
       const id = params.id
+      const result = await fetchDocumentLogWithMetadata({
+        workspaceId: workspace.id,
+        documentLogId: +id,
```
Comment (geclos) [inline apps/web/src/app/api/documentLogs/[id]/route.ts:22]: i'm getting confused, is this ID or uuid? if it's a uuid why is the prop called documentlogid and we are coercing to number? if it's number... then the i don't understand why you are doing these changes?

apps/web/src/app/api/documentLogs/uuids/[uuid]/route.ts
```diff
@@ -19,24 +17,19 @@ export const GET = errorHandler(
       },
     ) => {
       const uuid = params.uuid
+      const result = await fetchDocumentLogWithMetadata({
+        workspaceId: workspace.id,
+        documentLogUuid: uuid,
```
Comment (geclos) [inline apps/web/src/app/api/documentLogs/uuids/[uuid]/route.ts:22]: here it's called documentLogUuid?

packages/core/src/jobs/job-definitions/batchEvaluations/runDocumentJob.ts
```diff
@@ -59,15 +60,11 @@ export const runDocumentForEvaluationJob = async (
       source: LogSources.Evaluation,
     }).then((r) => r.unwrap())
 
-    // TODO handle this error better
-    const res = await result.response
-    if (!res) throw new BadRequestError('Document run failed')
```
geclos, line 64: how is this handled now? runDocumentAtCommit has to always return before result.response resolves yet result.response might throw, so we need to wait for response to finish in this particular case.

packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.ts
geclos: nice maybe leave this comment in the code actually, it's helpful

packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.ts
```diff
@@ -1,15 +1,51 @@
...
+  documentLogUuid,
+}: {
+  workspaceId: number
+  evaluationId: number
+  documentLogUuid: string
+}) {
+  const documentLogsScope = new DocumentLogsRepository(workspaceId)
+  const evaluationsScope = new EvaluationsRepository(workspaceId)
+  const docLogResult = await documentLogsScope.findByUuid(documentLogUuid)
+
```
geclos, line 38: 👾

packages/core/src/repositories/runErrors/documentLogsRepository/index.ts
```diff
@@ -38,4 +42,19 @@ export class DocumentLogsWithErrorsRepository extends Repository<
       .where(eq(workspaces.id, this.workspaceId))
       .as('documentLogsWithErrorsScope')
   }
+
+  async findByUuid(uuid: string) {
```
geclos, line 46: check there's an index for this column in the model, i think there wasn't

packages/core/src/services/chains/ChainStreamConsumer/consumeStream.ts
```diff
@@ -59,5 +69,5 @@ export async function consumeStream({
     })
   }
 
-  return validate(chunks)
+  return parseChunks(chunks)
```
geclos, line 72: are we changing public API? stuff that we return in the public API? if so are the changes backwards compatible?

packages/core/src/services/chains/run.ts
```diff
@@ -88,29 +116,31 @@ export async function runChain({
     stream,
     response,
     resolvedContent: chain.rawText,
-    documentLogUuid,
```
geclos, line 91: veeery weird that we return errorableUuid, honestly I would return both props even if they are the same

packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -1,17 +1,70 @@
...
+  source: LogSources
+  errorableUuid: string
+  parameters: Record<string, unknown>
+  resolvedContent: string
+  customIdentifier?: string
+  duration?: number
+  response?: ChainResponse<StreamType>
+}) {
+  const durantionInMs = duration ?? 0
+  publisher.publishLater({
```
geclos, line 40: hmm you are publish the documentrun event even if the document run failed, is that ok? any event handlers that we should be worried about?

packages/core/src/services/documentLogs/_createDocumentLogQuery.ts
```diff
@@ -19,7 +22,7 @@ export function createDocumentLogQuery(workspaceId: number, db = database) {
         .as('cost_in_millicents'),
     })
     .from(scope)
-    .innerJoin(providerLogs, eq(providerLogs.documentLogUuid, scope.uuid))
+    .leftJoin(providerLogs, eq(providerLogs.documentLogUuid, scope.uuid))
```
geclos, line 25: this is matching now all records in scope, is that what you want? we are increasing the size of this query exponentially

packages/core/src/services/chains/ChainStreamConsumer/consumeStream.ts
andresgutgon: We're not I think. This file is responsible for sending the chunks in streaming. Which happen in the same way at the end of the file.

The return value is used internally inside `runChain`

packages/core/src/jobs/job-definitions/batchEvaluations/runDocumentJob.ts
andresgutgon: `result.response` never throws now. Please review `runChain` and tell me this is the case

packages/core/src/services/documentLogs/_createDocumentLogQuery.ts
andresgutgon: Other wise I don't catch logs without providerLogs. This can happen now that we capture errors before hitting the AI

geclos: buf, we need to fix pagination of document logs then, the list of document logs needs to be filtered at this point otherwise any query using this cte will be crazy slow with big numbers

apps/web/src/app/api/documentLogs/[id]/route.ts
Comment (andresgutgon) [reply apps/web/src/app/api/documentLogs/[id]/route.ts:22]: Is both. We have a little mess. In the UI we use `uuid` in the URL. But in the evaluationResult.documentLogId. So when we request the logs we need both apis. The DRY here is that `fetchDocumentLogWithMetadata` handle both cases

packages/core/src/repositories/runErrors/documentLogsRepository/index.ts
andresgutgon: image

andresgutgon: This has being like this since forever I think. Feel free to add it. I'll add a FIXME not doing here for sure 👍

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx:25]: We have `DocumentLogsRepository` and `DocumentLogsWithErrorRepository`

packages/core/src/services/commits/RunDocumentChecker/index.ts
```diff
@@ -0,0 +1,76 @@
...
+      data: {
+        errorableUuid: this.errorableUuid,
+        errorableType: ErrorableEntity.DocumentLog,
+        code: error.errorCode,
+        message: error.message,
+        details: error.details,
+      },
+    }).then((r) => r.unwrap())
+  }
+}
```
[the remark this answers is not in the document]
andresgutgon: I like this way to be consistent with other services

packages/core/src/services/commits/runDocumentAtCommit.ts
andresgutgon: Talked with Cesar. Not publishing if there is error

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/Messages.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/Messages.tsx:31]: Not critical. Not doing here sorry

packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.ts
andresgutgon: Doc added

packages/core/src/services/prompts/run.ts
andresgutgon: Talked offline. We don't have logs for evaluations in evaluations playground. So we can not associate errors with any entity in DB

packages/core/src/services/prompts/run.ts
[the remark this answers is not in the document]
andresgutgon: This comment 
https://github.com/latitude-dev/latitude-llm/pull/381#discussion_r1806629808

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/EvaluationResults/EvaluationResultInfo/index.tsx:27]: I'll do this in a smaller PR when we're sure these changes are working

apps/web/src/app/api/documentLogs/uuids/[uuid]/route.ts
```diff
@@ -19,24 +17,19 @@ export const GET = errorHandler(
...
       }
 
-      const documentLogsRepo = new DocumentLogsRepository(workspace.id)
-      const _documentLog = await documentLogsRepo
-        .findByUuid(uuid)
-        .then((res) => res.unwrap())
-      const documentLog = await fetchDocumentLogWithMetadata({
-        workspaceId: workspace.id,
-        documentLogId: _documentLog.id,
-      }).then((res) => res.unwrap())
```
Comment (andresgutgon) [inline apps/web/src/app/api/documentLogs/uuids/[uuid]/route.ts:37]: This was doing twice the work for fining the document log

packages/core/src/events/handlers/notifyToClientEvaluationResultCreatedJob.ts
```diff
@@ -10,9 +10,10 @@ export const notifyToClientEvaluationResultCreatedJob = async ({
   data: EvaluationResultCreatedEvent
 }) => {
   const { evaluation, documentLog, evaluationResult } = event.data
-  const { evaluationResultsScope, baseQuery } = createEvaluationResultQuery(
-    evaluation.workspaceId,
-  )
+  // FIXME: DRY. This should be using same query as
+  // computeEvaluationResultsWithMetadata
+  const { evaluationResultsScope, baseQuery } =
+    createEvaluationResultQueryWithErrors(evaluation.workspaceId)
```
andresgutgon, line 16: We have to review this event handlers. I tried to use new `computeEvaluationResultsWithMetadata` but it already includes the `where` clause. But if we solve that we can remove all these 
`createEvaluationResultQueryWithErrors` that are not used anywhere now I think.

packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.ts
```diff
@@ -1,15 +1,58 @@
...
+async function fetchData({
+  workspaceId,
+  evaluationId,
+  documentLogUuid,
+}: {
+  workspaceId: number
+  evaluationId: number
+  documentLogUuid: string
+}) {
+  const documentLogsScope = new DocumentLogsWithErrorsRepository(workspaceId)
```
andresgutgon, line 41: This is important. I changed the normal `DocumentLogsRepository` for the one with errors.

Some may run an evaluation with a failed document. This way, we can trace the error and show the failed evaluation result.

packages/core/src/services/ai/index.ts
```diff
@@ -57,159 +63,99 @@ export async function ai({
...
+    if (UNSUPPORTED_STREAM_MODELS.includes(model)) {
+      return runNoStreamingModels({
+        schema,
+        output,
+        commonOptions,
+      })
+    }
 
-  if (UNSUPPORTED_STREAM_MODELS.includes(model)) {
-    if (output && schema) {
```
andresgutgon, line 102: 2 if level nesting + the try/catch now. I moved non-streaming to `runNoStreamingModels` service

PR #383, opened 2024-10-09
Fix: Compiler was reading all attributes as strings.

csansoon:
The compiler treated all HTML-like tag attributes as strings. As a result, our config schema didn’t allow modifying the ⁠temperature in a ⁠response chain step, because it wouldn’t read it as a number.

### DIFF
--- a/packages/compiler/src/compiler/base/nodes/tag.ts
+++ b/packages/compiler/src/compiler/base/nodes/tag.ts
@@ -38,10 +38,12 @@ async function resolveTagAttributes({
       continue
     }
 
-    let totalValue: string = ''
+    const accumulatedValue: unknown[] = []
     for await (const node of value) {
       if (node.type === 'Text') {
-        totalValue += node.data
+        if (node.data) {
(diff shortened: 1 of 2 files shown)


geclos (approved):

packages/compiler/src/compiler/base/nodes/tag.ts
```diff
@@ -38,23 +38,30 @@ async function resolveTagAttributes({
...
       }
     }
 
-    attributes[name] = totalValue
+    const finalValue =
+      accumulatedValue.length > 1
+        ? accumulatedValue.map(String).join('')
+        : accumulatedValue[0]
+
+    attributes[name] = finalValue
```
andresgutgon, line 64: A little test no?

andresgutgon (approved):

packages/compiler/src/compiler/base/nodes/tag.ts
csansoon: veeeeeengaa

PR #394, opened 2024-10-10
Read Metadata on web workers

csansoon:
Can't seem to get this to work. I'm getting this error when instantiating the Worker:

image

[NextJS documentation that may be interesting.](https://nextjs.org/docs/app/building-your-application/optimizing/scripts#offloading-scripts-to-a-web-worker-experimental)

### DIFF
--- a/apps/web/src/components/EditorHeader/index.tsx
+++ b/apps/web/src/components/EditorHeader/index.tsx
@@ -23,6 +23,7 @@ import { envClient } from '$/envClient'
 import { ROUTES } from '$/services/routes'
 import useProviderApiKeys from '$/stores/providerApiKeys'
 import Link from 'next/link'
+import { parse, stringify } from 'yaml'
 
 function useModelsOptions({
   provider,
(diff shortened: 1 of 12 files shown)


apps/web/src/components/EditorHeader/index.tsx
```diff
@@ -70,6 +71,54 @@ function selectModel({
...
+      /((?:\/\*[\s\S]*?\*\/\s*)?---\n)[\s\S]*?\n---/,
+      `$1${newFrontMatter}---`,
+    )
+  } catch (error) {
+    // If parsing fails, create new frontmatter
+    const newFrontMatter = stringify(updates)
+    return `---\n${newFrontMatter}---\n\n${prompt.replace(/(?:\/\*[\s\S]*?\*\/\s*)?---\n[\s\S]*?\n---\n/, '')}`
+  }
+}
+
```
geclos, line 121: I'm sorry but I had to reimplement the config setting via regex. I've added tests to account for common use cases

apps/web/src/workers/readMetadata.ts
```diff
@@ -0,0 +1,76 @@
...
+  for (const segment of segments) {
+    if (segment === '..') {
+      resultSegments.pop()
+    } else if (segment !== '.') {
+      resultSegments.push(segment)
+    }
+  }
+
+  return resultSegments.join('/')
+}
```
geclos, line 76: I also had to implement the referencefn without using path

apps/web/src/workers/readMetadata.ts
```diff
@@ -0,0 +1,72 @@
...
+
+self.onmessage = async function (event: { data: ReadMetadataWorkerProps }) {
+  const { document, documents, prompt, ...rest } = event.data
+  const referenceFn = readDocument(document, documents, prompt)
+  const metadata = await readMetadata({
+    ...rest,
+    prompt,
+    referenceFn: referenceFn ?? undefined,
+  })
```
geclos, line 16: a tradeoff is that we don't have configSchema validation since it depends on zod which i can't manage to get it working on web workers

packages/compiler/src/compiler/base/nodes/tags/message.ts
```diff
@@ -93,7 +93,7 @@ function buildMessage<R extends MessageRole>(
   if (role === MessageRole.system) {
     return {
       role,
-      content: (content[0]! as TextContent).text,
+      content: (content[0] as TextContent)?.text ?? '',
```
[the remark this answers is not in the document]
andresgutgon: How I can reproduce this bug? Because we're changing this in my PR

apps/web/src/components/EditorHeader/index.tsx
csansoon: 😭

packages/compiler/src/compiler/base/nodes/tags/message.ts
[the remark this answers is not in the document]
geclos: write this in the playground:

```
<system>
</system>
```

you should not see any errors in the preview

PR #401, opened 2024-10-11
(IMPORTANT) Errors epic: Make provider log optional on Evaluation Results

andresgutgon:
# What?
This is key in enabling error tracing in our evaluation results and document logs. 

> [!WARNING]  
> Errors can happen. But we still want to track the story of that failure in the same way. So we want to create failed evaluation results records with an error attached to it.

## TODO
- [x] Make `provider_log_id` field optional on `evaluation_results`
- [x] Make sure evaluation results without provider logs are not used for agreggations
- [x] Make sure evaluation results without resultable relation are used for aggregations

### DIFF
--- a/packages/core/src/schema/models/providerLogs.ts
+++ b/packages/core/src/schema/models/providerLogs.ts
@@ -2,6 +2,7 @@ import { Message, ToolCall } from '@latitude-data/compiler'
 import {
   bigint,
   bigserial,
+  index,
   integer,
   json,
   jsonb,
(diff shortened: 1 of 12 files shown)


csansoon (approved):

packages/core/src/schema/models/providerLogs.ts
```diff
@@ -17,41 +18,56 @@ import { latitudeSchema } from '../db-schema'
...
+    id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
+    uuid: uuid('uuid').notNull().unique(),
+    documentLogUuid: uuid('document_log_uuid'),
+    providerId: bigint('provider_id', { mode: 'number' })
+      .notNull()
+      .references(() => providerApiKeys.id, {
+        onDelete: 'restrict',
+        onUpdate: 'cascade',
+      }),
+    errorId: bigint('error_id', { mode: 'number' }).references(
```
geclos, line 41: why does provider log have a pointer to error?

andresgutgon: An error can happen

1. Before AI is called. 
2. While AI is called. When chunks bring `type: error` or `finish` with `finishReason: 'error'`


This pointer is here to save the error + provider log at the same time with a pointer to the error from the provider log


I can not do this. But the question is should we throw an error when stream finish with `type: error` or `finishReason: error`?

geclos: the pointer should be from error to the provider log no? I honestly wouldn't add this pointer if we can avoid it it seems unnecessary

andresgutgon: Removed

PR #402, opened 2024-10-11
Fix LATITUDE-LLM-APP-3S swapped mode and mean evaluation aggr components

andresgutgon:
# What?
☝️ That


geclos (approved):
no need for PR approval go ahead please

PR #431, opened 2024-10-15
Fix - cannot create evaluations

csansoon:
Evaluations were taking too much time to load, which may be because of the POST action being queued. Additionally, there was no UI for loading state, which just looked like it was broken and not working as expected.

Moved the logic to an API fetch instead of a server action, and added UI clues to the loading state.


geclos (approved):

apps/web/src/app/api/documents/[projectId]/[commitUuid]/[documentUuid]/connected-evaluations/route.ts
```diff
@@ -20,11 +20,11 @@ export const GET = errorHandler(
     ) => {
       const { documentUuid } = params
       const scope = new ConnectedEvaluationsRepository(workspace.id)
-      const evaluations = await scope
+      const connectedEvaluations = await scope
         .filterByDocumentUuid(documentUuid)
         .then((r) => r.unwrap())
 
-      return NextResponse.json(evaluations, { status: 200 })
+      return NextResponse.json(connectedEvaluations, { status: 200 })
```
Comment (geclos) [inline apps/web/src/app/api/documents/[projectId]/[commitUuid]/[documentUuid]/connected-evaluations/route.ts:27]: you removed the filtering by document uuid that was present in the action

apps/web/src/stores/evaluations.ts
```diff
@@ -20,27 +21,36 @@ export default function useEvaluations(
...
     isLoading,
     error: swrError,
   } = useSWR<EvaluationDto[]>(
-    compact(['evaluations', ...flatten(Object.entries(opts?.params ?? {}))]),
-    async () => {
-      const [data, error] = await fetchEvaluationsAction(opts?.params)
+    compact(['evaluations', documentUuid]),
+    useCallback(async () => {
+      const route = documentUuid
+        ? _API_ROUTES.evaluations.detail({ documentUuid })
```
geclos, line 35: ah you did a separate route, ok 👌🏼

apps/web/src/stores/evaluations.ts
```diff
@@ -1,14 +1,15 @@
...
+import { compact } from 'lodash-es'
 
 import type { EvaluationDto } from '@latitude-data/core/browser'
 import { useToast } from '@latitude-data/web-ui'
 import { createEvaluationAction } from '$/actions/evaluations/create'
 import { destroyEvaluationAction } from '$/actions/evaluations/destroy'
-import { fetchEvaluationsAction } from '$/actions/evaluations/fetch'
 import { updateEvaluationContentAction } from '$/actions/evaluations/updateContent'
 import useLatitudeAction from '$/hooks/useLatitudeAction'
+import { _API_ROUTES } from '$/services/routes/api'
```
geclos, line 12: import ROUTES instead of _API_ROUTES

apps/web/src/stores/evaluations.ts
```diff
@@ -20,27 +21,36 @@ export default function useEvaluations(
...
+        const error = await response.json()
 
-      if (error) {
         toast({
           title: 'Error fetching evaluations',
           description: error.formErrors?.[0] || error.message,
           variant: 'destructive',
         })
+
         throw error
```
geclos, line 49: this is probably wrong we should not throw but instead return an empty array also you are throwing a json here

PR #438, opened 2024-10-16
Chore/fixes a bunch of minor bugs

geclos:
Fixes a set of bugs that probably crashed the app but weren't really issues


geclos:
🐮 👦🏼

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/DocumentsLayout/index.tsx
```diff
@@ -28,10 +31,22 @@ export default async function DocumentsLayout({
...
+      workspaceId: session.workspace.id,
+    })
+  } catch (error) {
+    console.warn((error as Error).message)
+
+    if (error instanceof NotFoundError) {
+      return redirect(ROUTES.dashboard.root)
+    }
+
+    throw error
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/DocumentsLayout/index.tsx:47]: We need an abstraction here. By abstraction undertsand how to handle 404 in nextjs propoerly without needing this try/catch

PR #442, opened 2024-10-16
Use node-fetch on javascript SDK

andresgutgon:
## What?
We realized that the sdk is run on client's server and we can't control the nodejs version or if the have `fetch` polyfill. Also one user saw our sdk failing with TLSSocket.onSocketEnd TypeError terminated other.side closed. The stack trace we saw was pointing to undinci https://github.com/nodejs/undici/issues/583
Which looks is the nodejs fetch implementatin. So we try to use node-fetch to see if this mitigates the error

### DIFF
--- a/packages/sdks/typescript/src/index.ts
+++ b/packages/sdks/typescript/src/index.ts
@@ -1,3 +1,5 @@
+import { Readable } from 'stream'
+
 import type { Config, Message } from '@latitude-data/compiler'
 import type {
   ChainCallResponseDto,
@@ -7,10 +9,14 @@ import type {
 } from '@latitude-data/core/browser'
 import env from '$sdk/env'
 import { GatewayApiConfig, RouteResolver } from '$sdk/utils'
+import { nodeFetchResponseToReadableStream } from '$sdk/utils/nodeFetchResponseToReadableStream'
 import { BodyParams, HandlerType, UrlParams } from '$sdk/utils/types'
 import { ParsedEvent, ReconnectInterval } from 'eventsource-parser'
 import { EventSourceParserStream } from 'eventsource-parser/stream'
+import nodeFetch, { Response } from 'node-fetch'
 
+const MAX_RETRIES = 2
+const WAIT_IN_MS_BEFORE_RETRY = 1000
 export type StreamChainResponse = {
   conversation: Message[]
   response: ChainCallResponseDto
@@ -166,7 +172,7 @@ export class Latitude {
       }
 
       return this.handleStream({
-        stream: response.body!,
+        body: response.body! as Readable,
         onEvent,
         onFinished,
         onError,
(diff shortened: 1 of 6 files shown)


packages/sdks/typescript/src/index.ts
```diff
@@ -195,26 +201,27 @@ export class Latitude {
...
   }: StreamResponseCallbacks & {
-    stream: ReadableStream
+    body: Readable
   }) {
     let conversation: Message[] = []
     let uuid: string | undefined
     let chainResponse: ChainCallResponseDto | undefined
 
     const parser = new EventSourceParserStream()
+    const stream = nodeFetchResponseToReadableStream(body)
```
andresgutgon, line 224: Node fetch doesn't play well directly with `ReadableStream` web streams. I did this helper function to convert node-fetch events to a readable stream. What do you think?

packages/sdks/typescript/src/index.ts
```diff
@@ -330,7 +339,7 @@ export class Latitude {
   }
 
   private bodyToString(body: object = {}) {
-    return JSON.stringify(body)
+    return JSON.stringify({ ...body, __internal: { source: this.source } })
```
geclos, line 342: whats this

packages/sdks/typescript/src/utils/nodeFetchResponseToReadableStream.ts
```diff
@@ -0,0 +1,19 @@
...
+      nodeStream.on('end', () => {
+        controller.close()
+      })
+
+      nodeStream.on('error', (err) => {
+        controller.error(err)
+      })
+    },
+  })
+}
```
geclos, line 19: this is very delicate indeed, are there any other events from nodeStream, or only data end and error?

geclos (approved):

csansoon (approved):

packages/sdks/typescript/src/utils/nodeFetchResponseToReadableStream.ts
andresgutgon: This is how body is constructed in `node-fetch` 
https://github.com/node-fetch/node-fetch/blob/main/src/body.js#L37-L89

I don't see any specific. 

This method is typed as `Readable` but is a bit of a lie. If you check in Typescript what node-fetch return is a `NodeJS.ReadableStream`. But that's not a real type a read that is equivalent to Readable
image


Here are the events for readable:
https://nodejs.org/api/stream.html#class-streamreadable

geclos: indeed it was delicate 😂

andresgutgon: It was very delicate. Still is, not sure I did well. We'll see

Issue #444, opened 2024-10-16
Switch email provider and redirect to app when hitting `/login` if user has a valid session


andresgutgon:
# What?
We are having some delivery issues with our current provider to see if this improve the situation.

andresgutgon:
PR: https://github.com/latitude-dev/latitude-llm/pull/445

PR #445, opened 2024-10-16
Switch email provider to Mailgun. Redirect to dashboard on login if user has a valid session

andresgutgon:
# What?
☝️ That

### TODO
- [x] Deploy infra environment

## Issue 
https://github.com/latitude-dev/latitude-llm/issues/444


packages/env/src/index.ts
```diff
@@ -70,6 +71,7 @@ export const env = createEnv({
     FROM_MAILER_EMAIL: z.string(),
     GATEWAY_HOSTNAME: z.string(),
     LATITUDE_DOMAIN: z.string(),
+    LATITUDE_EMAIL_DOMAIN: z.string(),
```
geclos, line 74: if this is not optional and you don't set them in web/gateway the production containers will fail

apps/infra/src/app/production/shared.ts
```diff
@@ -118,6 +120,7 @@ export const environment = pulumi
       { name: 'GATEWAY_HOSTNAME', value: 'gateway.latitude.so' }, // This is the host gateway is available at
       { name: 'GATEWAY_SSL', value: 'true' },
       { name: 'LATITUDE_DOMAIN', value: 'latitude.so' },
+      { name: 'LATITUDE_EMAIL_DOMAIN', value: 'mail.latitude.so' },
```
geclos, line 123: we are changing domains for emails?

andresgutgon: Yes, already configured in Mailgun

packages/env/src/index.ts
andresgutgon: Changed

packages/core/src/mailers/mailers/adapters/mailgun.ts
```diff
@@ -13,12 +13,15 @@ export type MailgunResponse = {
 export default function createMailgunTransport({
   transportOptions,
 }: MailerOptions) {
-  const domain = env.LATITUDE_DOMAIN
-  const apiKey = env.MAILER_API_KEY
+  const domain = env.LATITUDE_EMAIL_DOMAIN
+  const apiKey = env.MAILGUN_MAILER_API_KEY
 
   if (!domain || !apiKey) return null
```
geclos, line 19: should we raise instead of returning null if the domain env vars aren't defined?

PR #446, opened 2024-10-16
Fixes LATITUDE-LLM-APP-4D Invalid state: Controller is already closed

andresgutgon:
I had to choose between `end` and `close` events and think I picked the wrong one. Let's see if this fixes the issue. By the way, this does not happen all the time. I just tested production and worked for me.

## Sentry issue
https://latitude-l5.sentry.io/issues/5996210933/?alert_rule_id=15461351&alert_type=issue&environment=production&notification_uuid=5e4bbccb-4064-409f-9799-f92cd1868ca6&project=4507922531418112&referrer=slack

### DIFF
--- a/packages/sdks/typescript/src/utils/nodeFetchResponseToReadableStream.ts
+++ b/packages/sdks/typescript/src/utils/nodeFetchResponseToReadableStream.ts
@@ -15,16 +15,26 @@ export function nodeFetchResponseToReadableStream(nodeStream: Readable) {
       })
 
       /**
-       * The 'close' event is emitted when the stream and any of its underlying resources
-       * (like a file descriptor) have been closed.
+       * The 'end' event is emitted when there is no more data to be consumed from the stream.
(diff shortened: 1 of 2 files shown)


geclos:
why not both?

andresgutgon:
> why not both?

It fails. Even our simple test fails if you use both

packages/sdks/typescript/src/utils/nodeFetchResponseToReadableStream.ts
```diff
@@ -15,15 +15,9 @@ export function nodeFetchResponseToReadableStream(nodeStream: Readable) {
       })
 
       /**
-       * The 'close' event is emitted when the stream and any of its underlying resources
-       * (like a file descriptor) have been closed.
-       * The event indicates that no more events will be emitted, and no further computation will occur.
-       *
        * Other similar event is `end` event, which is emitted when there is no more data to be consumed from the stream.
```
geclos, line 18: Fix the sentence now that you've removed the other ones

PR #448, opened 2024-10-17
feature: support tool calling from compiler and gateway

geclos:


packages/compiler/src/compiler/base/nodes/tags/message.ts
```diff
@@ -120,7 +120,6 @@ function buildMessage<R extends MessageRole>(
 
     return {
       role,
-      id: String(attributes.id),
```
geclos, line 123: not sure what this id is tbh don't see it anywhere in ai sdk documentation

apps/gateway/src/common/messageSchema.ts
```diff
@@ -32,13 +32,22 @@ export const messageSchema = z
...
+            type: z.literal(ContentType.toolCall),
+            toolCallId: z.string(),
+            toolName: z.string(),
+            args: z.record(z.any()),
+          }),
+        ),
+      ),
       toolCalls: z.array(
         z.object({
           id: z.string(),
```
csansoon, line 53: `toolCalls` is not part of the `content`, is it? An assistant message could contain both a set of `toolCalls` and a `content` string.

packages/core/src/helpers.ts
```diff
@@ -9,6 +9,8 @@ export function objectToString(
   message = DEFAULT_OBJECT_TO_STRING_MESSAGE,
 ) {
   try {
+    if (!object) return ''
```
geclos, line 12: we were stryingifying stuff like null into 'null'

apps/gateway/src/common/messageSchema.ts
geclos: nein, it can also contain an array of tool calls within the content in fact

csansoon: At least according to openai schema. Is vercel's ai different?

apps/gateway/src/common/messageSchema.ts
```diff
@@ -51,7 +60,14 @@ export const messageSchema = z
...
-      id: z.string(),
+      content: z.array(
+        z.object({
+          type: z.literal(ContentType.toolResult),
+          toolCallId: z.string(),
+          toolName: z.string(),
+          result: z.string(),
+          isError: z.boolean().optional(),
+        }),
+      ),
```
csansoon, line 71: A `tool` message should only contain an `id` and `content`, at least according to the openai schema. Why do we need the rest?

packages/compiler/src/compiler/base/nodes/tags/message.ts
```diff
@@ -120,7 +120,6 @@ function buildMessage<R extends MessageRole>(
...
       content,
     } as ToolMessage
   }
```
csansoon, line 125: Why? A `tool` message requires an `id`, which identifies the `tool_call` that requested it

geclos: the schema is just wrong this is the valid schemas both for tool result and tool calling requested https://sdk.vercel.ai/docs/reference/ai-sdk-core/stream-text#messages-role not sure where this id thing came from

apps/gateway/src/common/messageSchema.ts
geclos: not true according to ai sdk documentation (and it wasn't working either)

apps/gateway/src/common/messageSchema.ts
geclos: yes it's different

packages/compiler/src/compiler/base/nodes/tags/message.ts
csansoon: It is required in the openai schema. I don't know how vercel would manage this if it is not included when openai requires it...

packages/compiler/src/compiler/base/nodes/tags/message.ts
csansoon: This differs from openai's schema, which is what the compiler is based on
https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages

Do we want to change the compiler's schema to fit vercel's instead?

packages/compiler/src/compiler/base/nodes/tags/message.ts
geclos: talked offline

packages/compiler/src/compiler/base/nodes/tags/message.ts
geclos: talked offline

PR #452, opened 2024-10-17
Fix scrolls and container limits

csansoon:
Removed some hardcoded height limitations to containers, and fixed the actual responsiveness and container sizes through CSS attributes (tailwind classes, don't worry). Man, flex containers are so useful and and easy to do wrong :)

Fixed by using `flex-grow` and `flex-shrink` attributes when needed.

Also, now Playground buttons stay always visible at the bottom without having to scroll

https://github.com/user-attachments/assets/f3eb16df-7e4f-4da4-80e6-07b44bc860b5

### DIFF
--- a/packages/web-ui/src/ds/atoms/SplitPane/index.tsx
+++ b/packages/web-ui/src/ds/atoms/SplitPane/index.tsx
@@ -18,7 +18,6 @@ function Pane({ children }: { children: ReactNode }) {
 }
 
 const PaneWrapper = ({
-  width = 'auto',
   children,
   isResizable = false,
   className,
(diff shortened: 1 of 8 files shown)


packages/web-ui/src/ds/atoms/SplitPane/index.tsx
```diff
@@ -114,16 +116,15 @@ function HorizontalSplit({
...
   return (
-    <div className='min-h-full w-full grid grid-cols-[auto,1fr]'>
+    <div className='flex flex-row w-full relative h-full max-h-full min-h-full'>
       <ResizablePane
         minWidth={minWidth}
         paneWidth={paneWidth}
         onResizePane={setPaneWidth}
         onResizeStop={onResizeStop}
       >
-        <PaneWrapper isResizable width={width} className={cssPanelHeight}>
```
andresgutgon, line 126: Make sure resize sidebar works

csansoon: It does! :D

PR #454, opened 2024-10-17
feature: fix jobs circular dependencies

geclos:
It's breaking production builds in certain situations


packages/core/src/jobs/constants.ts
```diff
@@ -24,30 +14,37 @@ export enum Jobs {
...
   [Queues.eventHandlersQueue]: {
     name: Queues.eventHandlersQueue,
     jobs: Object.values(EventHandlers).flat(),
   },
   [Queues.liveEvaluationsQueue]: {
     name: Queues.liveEvaluationsQueue,
-    jobs: [runLiveEvaluationJob],
+    jobs: ['runLiveEvaluationJob'],
   },
 } as const
```
csansoon, line 50: Why? Isn't this the same?

csansoon (dismissed):

packages/core/src/jobs/constants.ts
geclos: no because now we don't import those jobs here anymore, which previously caused circular dependency shenanigans in nextjs build

PR #457, opened 2024-10-18
Feature/improve eval results query perf

geclos:


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/page.tsx
```diff
@@ -37,15 +40,28 @@ export default async function ConnectedEvaluationPage({
     projectId: Number(params.projectId),
     uuid: params.commitUuid,
   })
-  const { rows, pagination } = await paginateQuery({
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/page.tsx:40]: i've ejected from paginatequery because it's based on the assumption that pagination is the last step of the process whereas i want it to be the first. If this works we will explore a way to DRY a bit everything.

packages/core/src/services/evaluationResults/computeEvaluationResultsWithMetadata.ts
```diff
@@ -1,35 +1,162 @@
...
+      eq(aggregatedFieldsSubQuery.id, evaluationResultsScope.id),
+    )
+    .innerJoin(
+      documentLogsScope,
+      eq(documentLogsScope.id, evaluationResultsScope.documentLogId),
+    )
+    .innerJoin(commits, eq(commits.id, documentLogsScope.commitId))
+}
+
+export async function computeEvaluationResultsWithMetadataCount(
```
geclos, line 124: the count computation is actually quite expensive and should be done separately and as few times as possible

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/page.tsx
```diff
@@ -37,15 +40,28 @@ export default async function ConnectedEvaluationPage({
...
-    }).$dynamic(),
+  const rows = await computeEvaluationResultsWithMetadata({
+    workspaceId: evaluation.workspaceId,
+    evaluation,
+    documentUuid: params.documentUuid,
+    draft: commit,
+    page: searchParams.page as string | undefined,
+    pageSize: searchParams.pageSize as string | undefined,
+  })
+  const countResult = await computeEvaluationResultsWithMetadataCount({
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/page.tsx:51]: Maybe we could even do a separate api call. This is bellow in the page. No need to make the table wait for it. Also that will fix the count refresh issue when websockets have new results

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/page.tsx:51]: 💯

Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/page.tsx:51]: Yes, but that's in the current UI. Return a 0 here if you don't want to change anything and then do a count store where you do in frontend this computation

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/page.tsx:51]: yes will do something like this

PR #464, opened 2024-10-18
Remove redundant apk update calls in Dockerfiles

PeterDaveHello:
Removed unnecessary `apk update` commands in all Dockerfiles. The `apk add` commands already utilize the `--no-cache` option, making the update step superfluous and ensuring the latest packages are used without maintaining a local cache. An additional `apk update` in another Docker image layer will just make the image larger with no benefits.


PeterDaveHello:

geclos:
thanks!

PR #465, opened 2024-10-18
Apply new progress evaluation bar design

andresgutgon:
# What?

## TODO
- [x] `useFetcher` in [here](https://github.com/latitude-dev/latitude-llm/pull/381/files#r1806194348)

### DIFF
--- a/packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.test.ts
+++ b/packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.test.ts
@@ -179,6 +179,39 @@ describe('runEvaluationJob', () => {
       })
     })
 
+    it('increment error counter when response has an error', async () => {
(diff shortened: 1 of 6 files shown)


packages/core/src/jobs/job-definitions/batchEvaluations/runEvaluationJob.test.ts
```diff
@@ -179,6 +179,39 @@ describe('runEvaluationJob', () => {
       })
     })
 
+    it('increment error counter when response has an error', async () => {
```
andresgutgon, line 182: We were missing updating the counter of errors on an evaluation when the error happens in the response. We were checking only the result of the run. 

Errors can happen:
1. Before `runChain` occours. Errors on `runEvaluation`
2. Errors on `runChain` which are inside the `run.response`

geclos (dismissed):

PR #466, opened 2024-10-18
Add support for openai o1 models

Ashad-h:
- Since o1 models doesn't support streaming, I simulated a stream, as it's much easier because all the latitude pipeline seems to be stream only.
- Adapted the playground to display correct loading animation.

### DIFF
--- a/packages/core/src/services/ai/index.ts
+++ b/packages/core/src/services/ai/index.ts
@@ -106,13 +106,24 @@ export async function ai({
         output: output as any,
       })
 
+      const chunks: StreamChunk[] = [
+        {
+          type: 'object',
+          object: result.object,
+        },
+        {
+          type: 'finish',
+          finishReason: result.finishReason,
+          usage: result.usage,
+          response: result.response,
+        },
+      ]
+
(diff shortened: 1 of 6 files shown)


geclos:
Run pnpm prettier to fix any formatting issues

packages/core/src/services/ai/index.ts
```diff
@@ -90,6 +92,67 @@ export async function ai({
...
+    return Result.ok({
+      type: 'text',
+      data: {
+        fullStream: fullStream as any,
+        text: Promise.resolve(result.text),
+        usage: Promise.resolve(result.usage),
+        toolCalls: Promise.resolve(toolCalls),
+      },
+    })
+  }
```
geclos, line 154: If you could add some unit tests for this in the accompanying test file it would be awesome 🙏🏼

geclos (commented):
fantastic initiative

packages/core/src/services/ai/index.ts
Ashad-h: Done :)

geclos (approved):

PR #467, opened 2024-10-19
Added rate limit in the gateway

### LINKED ISSUE #450: Implement rate limiting in gateway
Quick win

Ashad-h:
Fix #450 

- Allow to set custom headers for Errors
- Use hono app.onError instead of a plain middleware to handle error. This allow to throw error inside middlewares.
- Using https://github.com/animir/node-rate-limiter-flexible
  - Added a rate limiter for all routes in a middleware
  - It use redis in background to save rate limit

## TODO
- [x] Create different redis host for rate limit (I'm using QUEUE_HOST for now)

## Question
- Do you have a way to benchmark the gateway latency to see the impact of this ?


geclos:
this is awesome! Now that you are at it might as well include the compression middleware? https://hono.dev/docs/middleware/builtin/compress

apps/gateway/src/middlewares/rateLimit.ts
```diff
@@ -0,0 +1,52 @@
...
+        enableOfflineQueue: true,
+        maxRetriesPerRequest: null,
+    }),
+    points: RATE_LIMIT_POINTS,
+    duration: RATE_LIMIT_DURATION,
+});
+
+const rateLimitMiddleware = () =>
+    createMiddleware(async (c, next) => {
+        const workspaceId: number = c.get('workspace').id
```
geclos, line 33: hmm instead of doing it by workspace id we could do it by api key and put this middleware before the auth middleware so that we avoid unnecessary db rountrips to get the workspace before immediately rate limiting the user

geclos:
> Do you have a way to benchmark the gateway latency to see the impact of this ?

Not for local development, no. I usually recommend using [drill](https://github.com/fcsonline/drill) for this. I might write add a benchmark for the api to the repo at some point.

apps/gateway/src/routes/app.ts
```diff
@@ -30,6 +32,6 @@ app.route(jetPaths(ROUTES).Api.V2.Documents.Base, documentsRouterV2)
 app.route(jetPaths(ROUTES).Api.V2.Conversations.Base, chatsRouterV2)
 
 // Must be the last one!
-app.use(errorHandlerMiddleware())
+app.onError(errorHandlerMiddleware)
```
geclos, line 35: much better 👌🏼

apps/gateway/package.json
```diff
@@ -28,7 +28,8 @@
     "hono": "^4.5.3",
     "jet-paths": "^1.0.6",
     "lodash-es": "^4.17.21",
-    "zod": "^3.23.8"
+    "zod": "^3.23.8",
+    "rate-limiter-flexible": "^5.0.3"
```
geclos, line 32: hmm not sure how you installed these deps but you are missing the changes to pnpm lock file. Make sure to install any deps with `pnpm install`

geclos:
> Create different redis host for rate limit (I'm using QUEUE_HOST for now)

Use de cache host instead (core exports a cache method to instantiate the redis client to the cache).

There is no need to set different rate limits per subscription for now

Ashad-h:
> This is awesome! Now that you are at it might as well include the compression middleware? https://hono.dev/docs/middleware/builtin/compress

The compress middleware is breaking the token streaming in the playground, I don't understand why

geclos:
> The compress middleware is breaking the token streaming in the playground, I don't understand why

:/ no wrorries then, i will take at look at it at some point

geclos (approved):

PR #471, opened 2024-10-21
Soft delete support

csansoon:
PR is about 300-400 LOC, not over 3k... You know the deal with Drizzle.

Added soft delete on drafts, evaluations and providerApiKeys. Soft-deleted rows, or data related to soft-deleted rows (like document logs from removed commits, or evaluation results from removed evaluations) will not be available from the respective repositories. Deleted logs and results are taken into account when calculating workspace usage though.

Additionally, "deleting" a `providerApiKey` will censor the token forever. This means that data can still be aggregated with this provider, it cannot be used to generate new logs.

### DIFF
--- a/apps/web/src/components/layouts/AppLayout/Header/Breadcrumb/Evaluations/index.tsx
+++ b/apps/web/src/components/layouts/AppLayout/Header/Breadcrumb/Evaluations/index.tsx
@@ -16,6 +16,7 @@ export function EvaluationBreadcrumbItems({
   segments: string[]
 }) {
   const evaluationUuid = segments[0]!
+  const evaluationSegment = segments[1] as EvaluationRoutes | undefined
 
   const { data: evaluations, isLoading } = useEvaluations()
   const currentEvaluation = useMemo(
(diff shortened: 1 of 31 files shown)


apps/web/src/components/layouts/AppLayout/Header/Breadcrumb/Evaluations/index.tsx
```diff
@@ -25,14 +26,17 @@ export function EvaluationBreadcrumbItems({
...
+      const baseRoute = ROUTES.evaluations.detail({ uuid: p.uuid })
+      const href = evaluationSegment
+        ? (baseRoute[evaluationSegment]?.root ?? baseRoute.root)
+        : baseRoute.root
+
+      return {
+        label: p.name,
+        href,
+      }
+    })
```
csansoon, line 39: Fixed a bugfix where the Breadcrumb component failed on `/evaluations/<uuid>/destroy`

andresgutgon (commented):
Missing test for checking in the repository only not soft deleted models are return.

csansoon:
@andresgutgon That's tested in the `destroy` service for each model

andresgutgon (dismissed):
Oh ok got it

geclos (approved):
beautiful

PR #472, opened 2024-10-21
Improve document logs table perfomance

andresgutgon:
# What?
Doing something similar to what we did with evaluation results, we filter by page and object the rows to be shown in the aggregated table of logs. This should improve performance

## TODO
- [x] Fix test
- [x] Use it in next Page route for document logs
- [x] Use it in next API route for document logs
- [x] Use it in websockets notify document log created
- [x] QA


packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts
```diff
@@ -1,27 +1,113 @@
-import { and, desc, eq } from 'drizzle-orm'
+import { and, desc, eq, getTableColumns, sql, sum } from 'drizzle-orm'
 
 import { Commit } from '../../browser'
 import { database } from '../../client'
 import {
-  createDocumentLogQuery,
-  getCommitFilter,
-} from './_createDocumentLogQuery'
```
andresgutgon, line 8: I 🔥 this archive. Now all the "with metadata" things are in this file. I left a FIXME in evaluations to do the same

packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts
```diff
@@ -1,27 +1,145 @@
...
+  allowAnyDraft,
+}: {
+  scope: DocumentLogWithErrorScope
+  documentUuid?: string
+  allowAnyDraft?: boolean
+  draft?: Commit
+}) {
+  const byDocumentUuid = documentUuid
+    ? eq(scope.documentUuid, documentUuid)
+    : sql`1 = 1`
```
andresgutgon, line 45: If `documentUuid` is not passed we don't care and use `1 = 1`. This case is covering `/evaluations` playground where we get a `documentLog.resolvedContent` for the evaluation variables

We don't have the scope of `documentUuid`

packages/core/src/tests/factories/documentLogs.ts
```diff
@@ -86,6 +88,35 @@ export async function createDocumentLog({
...
+  const workspace = (await findWorkspaceFromCommit(commit))!
+  const documentContent = await getResolvedContent({
+    workspaceId: workspace.id,
+    document,
+    commit,
+  }).then((r) => r.unwrap())
+  const documentLogUuid = uuid()
+  let providerLogs: ProviderLog[] = []
+
+  if (!skipProviderLogs) {
```
andresgutgon, line 111: Make provider logs creation optional on this factory

packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts
geclos: hmm don't get it is this gonna match all document logs a pelo? wouldn't allow it

geclos (approved):
lgtm

packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts
andresgutgon: Here
image

andresgutgon: In the playground we can import any document log from any project and any document

andresgutgon: But we filter by `documentLogUuid` https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/services/documentLogs/fetchDocumentLogWithMetadata.ts#L40-L41

csansoon (approved):

PR #481, opened 2024-10-22
Custom provider rules

csansoon:
Some providers have their custom rules on how a conversation should look like. This breaks the prompt and makes it throw an error. The current rules (that we know of) are:

- **Anthropic**: They do not have support for system messages.
- **Google**: They only allow system messages to be at the beginning of the prompt.

To mitigate this, we will automatically adapt prompts to fit these rules. When this happens, a warning message will appear in the prompt's Playground.

Note: We can only determine if a prompt will trigger this rule when it’s compiled. In a chain, we’ll only display a message if this rule is broken from the first step. Although we adapt the prompt if the rule is not followed in subsequent steps, no warning message will be displayed in the Playground.


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Preview.tsx
```diff
@@ -62,15 +74,32 @@ export default function Preview({
...
+    if (!provider) return
+    if (!conversation) return
+
+    const { messages, ruleMessage } = applyCustomRules({
+      providerType: provider.provider,
+      messages: conversation.messages,
+    })
+    setFixedMessages(messages)
+    setWarningMessage(ruleMessage)
+  }, [provider, conversation])
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Preview.tsx:87]: Why not merge this useEffect with the other

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Preview.tsx
```diff
@@ -62,15 +74,32 @@ export default function Preview({
...
+
   return (
     <div className='flex flex-col flex-1 gap-2 h-full overflow-hidden'>
+      {warningMessage && (
+        <div className='w-full px-4 py-2 rounded-md bg-yellow'>
+          <Text.H6 color='warningMutedForeground'>{warningMessage}</Text.H6>
+        </div>
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Preview.tsx:94]: You could add this variant `warning` to the existing Alert
https://github.com/latitude-dev/latitude-llm/blob/main/packages/web-ui/src/ds/atoms/Alert/Primitives/index.tsx#L9-L17

Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Preview.tsx:94]: Let's try not to make add-hoc components that are part of the design system.

andresgutgon (dismissed):

PR #490, opened 2024-10-22
chore: fix performance of evaluation results table #3

geclos:
The total count is just too much for the db. We have to remove it and resort to dumber pagination in favour of better performance.


packages/core/src/services/evaluationResults/computeEvaluationResultsByDocumentContent.ts
```diff
@@ -98,98 +75,34 @@ export async function computeEvaluationResultsByDocumentContent(
...
+      desc(evaluationResultsScope.createdAt),
+      desc(evaluationResultsScope.id),
     )
+    .limit(pageSize)
+    .offset(offset)
 
   return Result.ok(rows)
 }
-
-export async function computeEvaluationResultsByDocumentContentCount(
```
andresgutgon, line 146: Should we also remove from logs?

geclos: hmm not really if it's just a count of document logs filtered by documentuuid and commit it should be fast, what's hurting this particular count is the multiples inner joins required

PR #494, opened 2024-10-22
Avoid throwing not found error when not login in nextjs pages

andresgutgon:
# What?
This issue 
https://latitude-l5.sentry.io/issues/5964814238/?project=4507922531418112&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D%20error.unhandled%3Atrue&referrer=issue-stream&statsPeriod=14d&stream_index=0


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx
```diff
@@ -13,7 +15,13 @@ export default async function DocumentPage({
   params: { projectId: string; commitUuid: string; documentUuid: string }
   searchParams: QueryParams
 }) {
-  const { workspace } = await getCurrentUser()
+  const session = await getSafeCurrentUser()
+
+  if (!session) {
+    redirect(ROUTES.auth.login)
+  }
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx:22]: This fix the issue. The thing is we want to do this in all authenticated pages. What would be the best abstraction. 

I'm trying HOC to get the session or redirect but with async components typescript is complaining. 

What other ideas we could use? Or is fine doing this same code in all the authenticated pages?

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx:22]: i think the appropiate way for doing this is middlewares

Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx:22]: The thing is that we have this check on private layout
https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/app/(private)/layout.tsx#L25-L28

And also this on fetcher 
https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/hooks/useFetcher.ts#L35-L42

Not sure we need to do a check with error by page inside `(private)` routes.

apps/web/src/app/(private)/_lib/createSdk.ts
```diff
@@ -23,16 +21,18 @@ async function getLatitudeApiKey() {
...
 }: {
+  workspace: Workspace
   projectId?: number
   apiKey?: string
   __internal?: { source: LogSources }
-} = {}) {
+}) {
   if (!apiKey) {
-    const result = await getLatitudeApiKey()
+    const result = await getLatitudeApiKey(workspace)
```
andresgutgon, line 35: No need to fetch again workspace when is already fetched by the users of this function. Passing it by argument

apps/web/src/app/(private)/dashboard/layout.tsx
```diff
@@ -20,12 +18,8 @@ export default async function DashboardLayout({
 }: Readonly<{
   children: ReactNode
 }>) {
-  const data = await getSession()
-  if (!data.session) return redirect(ROUTES.auth.login)
```
andresgutgon, line 24: This is already happening in the main layout. Is confusing having it here

apps/web/src/app/(private)/layout.tsx
```diff
@@ -26,6 +26,8 @@ export default async function PrivateLayout({
   if (!data.session) return redirect(ROUTES.auth.login)
 
   const { workspace, user } = await getCurrentUser()
+  if (!user) return redirect(ROUTES.auth.login)
```
andresgutgon, line 29: Without session there is no user. But doesn't hurt this check

csansoon (dismissed):

apps/web/src/app/(private)/layout.tsx
geclos: this is why I mentioned to do the check in a middleware as redirects in pages always throw. Nextjs docs actually recommend doing any pre-render redirects in middleware (for the future)

apps/web/src/app/(private)/settings/_components/WorkspaceApiKeys/index.tsx
```diff
@@ -44,6 +44,7 @@ export default function WorkspaceApiKeys() {
                     </TableCell>
                     <TableCell>
                       <Tooltip
+                        asChild
```
geclos, line 47: isn't that the default behaviour?

apps/web/src/services/auth/getCurrentUser.ts
```diff
@@ -11,25 +14,33 @@ export type SessionData = {
...
+/**
+ * This method is used in places where session has been already checked
+ * An example of it are inner `app/(private)` routes. Session was check in main
+ * layout. We don't want to throw a not found error if session is not present
+ */
 export const getCurrentUser = cache(async () => {
   const sessionData = await getSession()
-  const result = await getCurrentUserFromDB({ userId: sessionData?.user?.id })
-  const { user, workspace } = result.unwrap()
+  const { user, workspace } = await getUnsafelyCurrentUserFromDb({
```
geclos, line 25: nitpick: unsafelyGet instead of getUnsafely to match it with all the rest

geclos (approved):

apps/web/src/app/(private)/settings/_components/WorkspaceApiKeys/index.tsx
andresgutgon: I don't think. I had to add it. Not sure if default behaviour makes sense. I don't see recommended on radix docs 
https://www.radix-ui.com/primitives/docs/guides/composition#composing-multiple-primitives

PR #500, opened 2024-10-23
Move Typescript SDK to use our API v2

andresgutgon:
# What?
This change is incremental and everyone in v1 should be able to work without any changes. If they opt-in to use API v2, they need to pass apiVersion: 'v2'. And if they want to run a prompt without streaming they need to pass `stream: false`

## TODO
- [x] ~Maybe test somehow v2 and `streaming: false`~ I did an `./examples` playground in the monorepo to use our sdk
- [x] Fix gateway tests v2 with errors
- [x] Fix gateway tests v2 with valid response. I changed the output to match the stream version that the SDK is already returning
- [x] Implement v2 run method
- [x] Add new tests for non-streaming
- [x] Undo refactor of different sdk versions
- [x] QA playground keeps working with `v1`
- [x] Document `stream: boolean` in docs

### IMPORTANT
- [x] Test production build. I introduced a new package `@latitude-data/constants`. In tired of not being able to have a unified source of truth for enums and types. This is the first stone


packages/sdks/typescript/src/tests/common/chat.test.ts
```diff
@@ -0,0 +1,177 @@
...
+describe('/chat', () => {
+  beforeAll(() => server.listen())
+  afterEach(() => server.resetHandlers())
+  afterAll(() => server.close())
+
+  describe('v1', () => {
+    beforeAll(() => {
+      sdk = new Latitude(FAKE_LATITUDE_SDK_KEY, {
+        __internal: {
+          retryMs: 10,
```
andresgutgon, line 36: Faster test. We were spending 3 seconds in the retry test. Now I do twice one by api version

packages/sdks/typescript/src/versions/LatitudeSdkV1.ts
```diff
@@ -0,0 +1,3 @@
+import { LatitudeSdk } from '$sdk/versions/LatitudeSdk'
+
+export class LatitudeSdkV1 extends LatitudeSdk { }
```
andresgutgon, line 3: The point of all these changes is to try to do a patch version in the SDK because current sdk users should not see any changes.

examples/nodejs-sdk-usage/src/run.ts
```diff
@@ -0,0 +1,29 @@
+import { Latitude } from '@latitude-data/sdk'
+
+async function makeSdkRequest() {
```
andresgutgon, line 3: A playground to test in the development of the SDK. This makes our job easier

packages/constants/src/errors.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export type ApiErrorJsonResponse = {
+  name: string
+  message: string
+  details: object
+  errorCode: ApiResponseCode
+  dbErrorRef?: DbErrorRef
+}
+export type ApiResponseCode = RunErrorCodes | ApiErrorCodes | LatitudeErrorCodes
```
andresgutgon, line 52: Unified list of codes of all the things that can go wrong in an api call

packages/core/package.json
```diff
@@ -52,6 +52,7 @@
     "@faker-js/faker": "^8.4.1",
     "@latitude-data/compiler": "workspace:^",
     "@latitude-data/env": "workspace:*",
+    "@latitude-data/constants": "workspace:*",
```
andresgutgon, line 55: Next step in another moment is to move all constants that are from
1. Vercel AI
2. Compiler 

Into constants so sharing this types between `gateway` and `sdk` and `core` is perfect and no one depends on the others

packages/sdks/typescript/tsconfig.json
```diff
@@ -3,11 +3,12 @@
   "compilerOptions": {
     "moduleResolution": "Bundler",
     "outDir": "./dist",
-    "rootDir": ".",
+    "rootDir": "../../",
```
andresgutgon, line 6: This makes `.d.ts` generated file not complain. It can find constants

packages/web-ui/src/example-file.ts
andresgutgon, line 1: Not used, please confirm @geclos I think this was created by you

apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/documents/handlers/run.ts
```diff
@@ -82,18 +83,8 @@ export const runHandler = factory.createHandlers(
...
-      const error = new LatitudeError('Streaming did not return any events')
-      captureException(error)
-      throw error
-    }
-
-    const data = chainEventPresenter(lastEvent)
-    return c.json(data)
+    const response = await result.response.then((r) => r.unwrap())
+    const body = documentRunPresenter(response).unwrap()
+    return c.json(body)
```
Comment (andresgutgon) [inline apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/documents/handlers/run.ts:88]: This is an important change. The meet is in `documentRunPresenter`

apps/gateway/tsup.config.ts
```diff
@@ -23,5 +23,9 @@ export default defineConfig({
     '.html': 'empty',
   },
   external: [...rootDependencies, ...dependencies],
-  noExternal: ['@latitude-data/env', '@latitude-data/core'],
+  noExternal: [
+    '@latitude-data/env',
+    '@latitude-data/core',
+    '@latitude-data/constants',
```
andresgutgon, line 29: I added `@latitude-data/constants` to all our apps. This package would be the common package without other dependencies where we put all the shared constants and types.

apps/web/src/actions/copilot/requestSuggestion.ts
```diff
@@ -62,6 +62,7 @@ export const requestSuggestionAction = authProcedure
     }).then((r) => r.unwrap())
 
     const result = await sdk.run(env.COPILOT_CODE_SUGGESTION_PROMPT_PATH, {
+      stream: true,
```
andresgutgon, line 65: I moved all our internal use of SDK to work with new code. So all cases were using streaming and should keep using streaming. 

Please let me know if I forgot any

packages/sdks/typescript/src/tests/run.test.ts
```diff
@@ -0,0 +1,421 @@
...
+            errorCode: 'LatitudeError',
+          }),
+          message: 'Something bad happened',
+          errorCode: ApiErrorCodes.InternalServerError,
+          dbErrorRef: undefined,
+        }),
+      )
+    })
+  })
+})
```
andresgutgon, line 421: I think I covered all the new code here

packages/sdks/typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@latitude-data/sdk",
-  "version": "0.0.18",
+  "version": "1.0.0-beta.0",
```
andresgutgon, line 3: changed with this: `pnpm version premajor --preid beta`

geclos (dismissed):
LGTM

apps/web/src/actions/copilot/refinePrompt.ts
```diff
@@ -76,6 +76,7 @@ export const refinePromptAction = authProcedure
     }).then((r) => r.unwrap())
 
     const result = await sdk.run(env.COPILOT_REFINE_PROMPT_PATH, {
+      stream: true,
       parameters: {
         prompt: document.content,
         evaluation: evaluation.metadata.prompt,
```
csansoon, line 82: This should not be required, right? It's awaiting for the result anyways

csansoon (approved):

apps/web/src/actions/copilot/refinePrompt.ts
andresgutgon: What should be required? the `stream` argument?

apps/web/src/actions/copilot/requestSuggestion.ts
andresgutgon: Ok, setting to false

Issue #502, opened 2024-10-23
Fix auth checking with NextJS middleware 


andresgutgon:
# What?
Now we check auth by page but this is a bit lengthy and also we have some problems. The idea is to use [NextJS middleware redirect](https://nextjs.org/docs/app/building-your-application/routing/redirecting#nextresponseredirect-in-middleware) to have auth flow centralized

PR #503, opened 2024-10-24
Implement nextjs middleware to check auth

andresgutgon:
# What?
Solves this issue: 
https://github.com/latitude-dev/latitude-llm/issues/502

## Explanation
image

In our specific case we're using `fileURLToPath` which is NodeJS api not supported on edge runtime. We're [using it here](https://github.com/latitude-dev/latitude-llm/blob/main/packages/env/src/index.ts#L9-L11)

[Link to issue](https://github.com/vercel/next.js/discussions/46722#discussioncomment-11030344)

### DIFF
--- a/apps/web/src/services/routes.ts
+++ b/apps/web/src/services/routes.ts
@@ -18,6 +18,24 @@ export enum BackofficeRoutes {
   rewards = 'rewards',
 }
 
+const PUBLIC_ROOT_PATHS = {
+  setup: '/setup',
+  login: '/login',
+  magicLinks: '/magic-links',
+  invitations: '/invitations',
+}
+
+export function isPublicPath(pathname: string) {
+  const publicPaths = [
+    PUBLIC_ROOT_PATHS.setup,
+    PUBLIC_ROOT_PATHS.login,
+    PUBLIC_ROOT_PATHS.magicLinks,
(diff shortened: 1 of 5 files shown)


geclos:
> We're [using it here](https://github.com/latitude-dev/latitude-llm/blob/main/packages/env/src/index.ts#L9-L11)

We can probably stop using that given that this is only for development we can just hardcode a path and move on no?

andresgutgon:
This also fails. Basically any nodejs runtime api. Let's check together. Maybe you're right and I'm missing something
image

apps/web/src/services/routes.ts
```diff
@@ -18,6 +18,24 @@ export enum BackofficeRoutes {
   rewards = 'rewards',
 }
 
+const PUBLIC_ROOT_PATHS = {
+  setup: '/setup',
+  login: '/login',
+  magicLinks: '/magic-links',
+  invitations: '/invitations',
+}
```
andresgutgon, line 26: I prefer this instead of using the regex in middleware. I think we have more control. I left the middleware regex for nextjs urls stuff

geclos: agree

geclos (approved):

PR #506, opened 2024-10-24
chore: do not allow 4o models in default provider (except mini)

geclos:


packages/core/src/constants.ts
```diff
@@ -83,6 +83,7 @@ export enum LogSources {
 export enum RunErrorCodes {
   Unknown = 'unknown_error',
   DefaultProviderExceededQuota = 'default_provider_exceeded_quota_error',
+  DefaultProviderInvalidModel = 'default_provider_invalid_model_error',
```
andresgutgon, line 86: Missing generate migration for this one. Also add it to the schema enum

andresgutgon (dismissed):

PR #507, opened 2024-10-24
refactor(repositories): update repository inheritance to use RepositoryLegacy and RepositoryV2

geclos:
This commit refactors the repository classes to inherit from either RepositoryLegacy or RepositoryV2 instead of the original Repository class. The changes are made to improve the baseline performance of the repository classes. Legacy repositories make use of subqueries which can often be slow. New repository uses drizzle's dynamic queries which are much faster. This change should improve the performance of the document logs section.

- Updated various repository classes to extend RepositoryLegacy.
- Introduced RepositoryV2 for new repository implementations.
- Adjusted method calls and query scopes to align with the new repository structure.
- Ensured that all references to the old Repository class are replaced with the appropriate new classes.

This refactor aims to streamline the transition to the new repository structure while maintaining backward compatibility with legacy code.

### DIFF
--- a/packages/core/src/repositories/evaluationResultsRepository/index.ts
+++ b/packages/core/src/repositories/evaluationResultsRepository/index.ts
@@ -26,7 +26,7 @@ import {
   evaluations,
   runErrors,
 } from '../../schema'
-import Repository from '../repository'
+import Repository from '../repositoryV2'
 
 export const evaluationResultDto = {
   ...getTableColumns(evaluationResults),
@@ -50,10 +50,7 @@ export type EvaluationResultWithMetadata = EvaluationResultDto & {
   documentContentHash: string
 }
 
-export class EvaluationResultsRepository extends Repository<
-  typeof evaluationResultDto,
-  EvaluationResultDto
-> {
+export class EvaluationResultsRepository extends Repository<EvaluationResultDto> {
   get scope() {
     return this.db
       .select(evaluationResultDto)
@@ -104,7 +101,7 @@ export class EvaluationResultsRepository extends Repository<
           eq(evaluations.workspaceId, this.workspaceId),
         ),
       )
-      .as('evaluationResultsBaseQuery')
+      .$dynamic()
   }
 
   async findByContentHash({
@@ -114,17 +111,18 @@ export class EvaluationResultsRepository extends Repository<
     evaluationId: number
     contentHash: string
   }) {
-    const result = await this.db
-      .select(this.scope._.selectedFields)
-      .from(this.scope)
-      .innerJoin(documentLogs, eq(documentLogs.id, this.scope.documentLogId))
+    const result = await this.scope
+      .innerJoin(
+        documentLogs,
+        eq(documentLogs.id, evaluationResults.documentLogId),
+      )
       .where(
         and(
-          eq(this.scope.evaluationId, evaluationId),
+          eq(evaluationResults.evaluationId, evaluationId),
           eq(documentLogs.contentHash, contentHash),
         ),
       )
-      .orderBy(desc(this.scope.createdAt))
+      .orderBy(desc(evaluationResults.createdAt))
 
     return Result.ok(result.map(this.parseResult))
   }
--- a/packages/core/src/repositories/providerLogsRepository.ts
+++ b/packages/core/src/repositories/providerLogsRepository.ts
@@ -8,11 +8,12 @@ import {
   providerLogs,
   workspaces,
 } from '../schema'
-import Repository, { QueryOptions } from './repository'
+import { QueryOptions } from './repository'
+import Repository from './repositoryV2'
 
 const tt = getTableColumns(providerLogs)
 
-export class ProviderLogsRepository extends Repository<typeof tt, ProviderLog> {
+export class ProviderLogsRepository extends Repository<ProviderLog> {
   get scope() {
     return this.db
       .select(tt)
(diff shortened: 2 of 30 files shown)


packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts
```diff
@@ -36,7 +39,7 @@ function getCommonQueryConditions({
   draft,
   allowAnyDraft,
 }: {
-  scope: DocumentLogWithErrorScope
+  scope: any
```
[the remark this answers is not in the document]
geclos: honestly #peresa debugging the type for this local method tbh

packages/core/src/services/documentLogs/fetchDocumentLogWithMetadata.ts
```diff
@@ -44,9 +45,11 @@ export async function fetchDocumentLogWithMetadata(
   let logs: DocumentLogWithMetadataAndError[] = []
 
   if (documentLogUuid) {
-    logs = await baseQuery.where(eq(scope.uuid, documentLogUuid)).limit(1)
+    logs = await baseQuery
+      .where(eq(documentLogs.uuid, documentLogUuid))
+      .limit(1)
   } else if (documentLogId) {
-    logs = await baseQuery.where(eq(scope.id, documentLogId)).limit(1)
+    logs = await baseQuery.where(eq(documentLogs.id, documentLogId)).limit(1)
```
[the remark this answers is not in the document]
geclos: you could do `scope._selectedFields.id`, but I find documentLogs shorter 🤷🏼

packages/core/src/repositories/evaluationResultsRepository/index.ts
```diff
@@ -104,7 +101,7 @@ export class EvaluationResultsRepository extends Repository<
           eq(evaluations.workspaceId, this.workspaceId),
         ),
       )
-      .as('evaluationResultsBaseQuery')
+      .$dynamic()
```
geclos, line 104: this is the key. This forces this query to be part of whatever other query it's embedded into, preventing the use of subqueries which hinder performance.

csansoon (dismissed):
looks good

packages/core/src/repositories/providerLogsRepository.ts
```diff
@@ -67,11 +62,9 @@ export class ProviderLogsRepository extends Repository<typeof tt, ProviderLog> {
...
     }
 
-    const result = await this.db
-      .select()
-      .from(this.scope)
-      .where(eq(this.scope.documentLogUuid, documentLogUuid))
-      .orderBy(desc(this.scope.generatedAt))
+    const result = await this.scope
+      .where(eq(providerLogs.documentLogUuid, documentLogUuid))
+      .orderBy(desc(providerLogs.generatedAt))
```
andresgutgon, line 67: So no `select` and no `from` because is the select in the scope no?

packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts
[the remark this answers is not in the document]
andresgutgon: So all `DocumentLogWithErrorScope` was not working?

packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts
```diff
@@ -70,53 +73,55 @@ export function computeDocumentLogsWithMetadataQuery(
...
+      .from(documentLogs)
+      .innerJoin(filteredSubQuery, eq(filteredSubQuery.id, documentLogs.id))
+      .innerJoin(commits, eq(commits.id, filteredSubQuery.commitId))
       .innerJoin(
         aggregatedFieldsSubQuery,
-        eq(aggregatedFieldsSubQuery.id, scope.id),
+        eq(aggregatedFieldsSubQuery.id, filteredSubQuery.id),
       )
-      .orderBy(desc(scope.createdAt)),
+      .orderBy(desc(documentLogs.createdAt)),
```
andresgutgon, line 124: So `scope` is not used now here? only is exported. Maybe is not used outside and we can remove it

packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts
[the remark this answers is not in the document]
geclos: no because now it's a dynamic query yada yada, honestly i'm pretty sure i could fix this type but since this is a private method i chose to move on with my life #sorry

packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts
```diff
@@ -70,53 +73,55 @@ export function computeDocumentLogsWithMetadataQuery(
...
-    .select({
-      id: scope.id,
-      error: scope.error,
-    })
-    .from(scope)
-    .innerJoin(commits, eq(commits.id, scope.commitId))
+  const filteredSubQuery = scope
     .where(
-      getCommonQueryConditions({ scope, documentUuid, draft, allowAnyDraft }),
+      getCommonQueryConditions({
```
andresgutgon, line 78: How is tenancy enforced now?

packages/core/src/repositories/providerLogsRepository.ts
andresgutgon: Ok. Maybe a bit limiting and not able to do our own selects. But let's see

PR #508, opened 2024-10-25
feat: add custom rule handling for Anthropic and Google providers

geclos:
- Introduced `AppliedRules` type to handle custom rules for different providers.
- Updated `applyCustomRules` function to return `AppliedRules` or `undefined` based on provider type.
- Modified `applyAnthropicRules` and `applyGoogleRules` to return specific rule messages and handle system messages accordingly.
- Replaced `warningMessage` state with `warningRule` state in `Preview` component.
- Added `WarningMessage` component to display appropriate alerts based on the applied rule.
- Updated `Alert` component to support dynamic icon colors based on variant.
- Added documentation for custom rules in `docs/guides/prompt-manager/custom-rules.mdx`.
- Included new route for documentation in `routes.ts`.

These changes improve the handling of provider-specific rules, ensuring that users are informed about any limitations or modifications applied to their messages.

image


andresgutgon (approved):
I wonder if warnings like this should also be persisted as error (warning) logs

PR #510, opened 2024-10-28
feature: user generated logs

geclos:
This commit implements the ability to upload user generated logs to a project. It adds a new route for uploading logs and a new page for the user to upload logs. Logs are stored in the same way as regular document logs but they lack some optional fields such as duration, cost, token count or finish reason. The only required field is the messages array.


packages/core/drizzle/0082_sparkling_redwing.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE "latitude"."provider_logs" ALTER COLUMN "provider_id" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "latitude"."provider_logs" ALTER COLUMN "finish_reason" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "latitude"."provider_logs" ALTER COLUMN "config" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "latitude"."provider_logs" ALTER COLUMN "messages" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "latitude"."provider_logs" ALTER COLUMN "duration" DROP NOT NULL;--> statement-breakpoint
+ALTER TABLE "latitude"."provider_logs" ALTER COLUMN "generated_at" DROP NOT NULL;
```
andresgutgon, line 6: Why we're removing all this not null. I guess is because you want to handle the case of manually uploaded logs. But somehow looks like is making this model more weak no?

packages/core/src/schema/messages.ts
andresgutgon, line 1: Epa

packages/core/drizzle/0082_sparkling_redwing.sql
geclos: yup but there's no other way that isn't more painful

csansoon: ufff ya

apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/logs/handlers/post.ts
```diff
@@ -0,0 +1,50 @@
...
+        providerLog: {
+          messages,
+          responseText:
+            response ??
+            (messages[messages.length - 1].content?.text ||
+              messages[messages.length - 1].content),
+        },
+      },
+      commit,
+    }).then((r) => r.unwrap())
```
Comment (csansoon) [inline apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/logs/handlers/post.ts:46]: What will happen when a user runs the API's `chat` endpoint on this log? We need to manage it gracefully at least.

Comment (geclos) [reply apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/logs/handlers/post.ts:46]: hmm although technically possible i don't think this will ever happen as the uuid of this document uuid will never be returned in the sdk

Comment (geclos) [reply apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/logs/handlers/post.ts:46]: we will add a graceful handling anyways

apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/logs/handlers/post.ts
Comment (andresgutgon) [inline apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/logs/handlers/post.ts:1]: Small thing: A bit weird call it post. Better create

apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/logs/handlers/post.ts
```diff
@@ -0,0 +1,50 @@
...
+import { getData } from '$/routes/api/v1/projects/[projectId]/versions/[versionUuid]/documents/handlers/_shared'
+import { Factory } from 'hono/factory'
+import { z } from 'zod'
+
+const factory = new Factory()
+
+const schema = z.object({
+  path: z.string(),
+  messages: messagesSchema,
+  response: z.string().optional(),
```
Comment (andresgutgon) [inline apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/logs/handlers/post.ts:14]: What can they put here? Anything? I guess we just pass it to the evaluation as string no?

Comment (geclos) [reply apps/gateway/src/routes/api/v2/projects/[projectId]/versions/[versionUuid]/logs/handlers/post.ts:14]: the response from the assistant

apps/gateway/src/routes/api/v1/conversations/[conversationUuid]/index.ts
```diff
@@ -1,8 +1,7 @@
-import ROUTES from '$/common/routes'
 import { Hono } from 'hono'
 
 import { chatHandler } from './handlers/chat'
 
 export const chatsRouter = new Hono()
 
-chatsRouter.post(ROUTES.Api.V1.Conversations.Chat, ...chatHandler)
```
Comment (geclos) [inline apps/gateway/src/routes/api/v1/conversations/[conversationUuid]/index.ts:8]: I'm removing this crap, it's worse than having regular string literals even if it causes some minor duplicity

packages/sdks/typescript/package.json
```diff
@@ -19,16 +19,24 @@
...
+      },
+      "require": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.cjs"
+      }
     }
   },
+  "types": {
+    "index.d.ts": "./dist/index.d.ts"
+  },
```
geclos, line 39: our exports were all wrong and we weren't exporting types correctly, working on the example quickly surfaced these issues

PR #511, opened 2024-10-28
feature: add workspace tenancy to provider logs

geclos:
This commit adds a workspace tenancy to provider logs. Since provider logs won't have a mandatory providerId field, we lose the tenancy so we have to add it back with a new workspace_id direct reference.


packages/core/drizzle/0083_empty_albert_cleary.sql
```diff
@@ -0,0 +1,6 @@
+-- Custom SQL migration file, put you code below! --
+UPDATE provider_logs
+SET workspace_id = provider_api_keys.workspace_id
+FROM provider_api_keys
+WHERE provider_logs.provider_id = provider_api_keys.id
+AND provider_logs.workspace_id IS NULL;
```
geclos, line 6: will update 80k rows which is bordering the limits of what a would do with an sql migration tbh

csansoon (approved):
Nice. But provider logs will fail to be created from the moment it

packages/core/drizzle/0083_empty_albert_cleary.sql
andresgutgon: Fucking scary

packages/core/src/services/chains/ProviderProcessor/saveOrPublishProviderLogs.ts
```diff
@@ -29,8 +31,14 @@ export async function saveOrPublishProviderLogs<T extends StreamType>({
     data: { ...data, streamType } as AIProviderCallCompletedData<T>,
   })
 
+  const workspace = await unsafelyFindWorkspace(data.workspaceId)
+  if (!workspace) {
+    throw new NotFoundError('Workspace not found')
+  }
```
andresgutgon, line 37: Wait. No need for this. Pass `workspace` from `run.ts` This code is only used there:
https://github.com/latitude-dev/latitude-llm/blob/c8f02d0da03376a1fc5adc7adc662964dc2f8a6e/packages/core/src/services/chains/run.ts#L137

packages/core/drizzle/0083_empty_albert_cleary.sql
andresgutgon: Ping us if you want double check

PR #512, opened 2024-10-28
feat(migrations): update provider_logs with workspace_id from provider_api_keys

geclos:
This commit introduces a new SQL migration that updates the `workspace_id` field in the `provider_logs` table by joining it with the `provider_api_keys` table. The update is performed only for records where `workspace_id` is currently NULL. This ensures that all provider logs are correctly associated with their respective workspaces, improving data consistency and integrity.

Additionally, the corresponding snapshot and journal files have been updated to reflect this migration.


packages/core/drizzle/meta/_journal.json
```diff
@@ -1 +1,601 @@
...
\ No newline at end of file
+{
```
andresgutgon, line 1: Something weird happened in the previous PR

andresgutgon (approved):

packages/core/drizzle/meta/_journal.json
geclos: prettier formatting jsons, nothing to worry about

PR #514, opened 2024-10-28
Add Loops client to capture contact when a signup is created

andresgutgon:
# What?
Add loops client to capture contact information

## TODO
- [x] Event `userCreated` when signup occur 
- [x] When user created is created call loops API
- [x] Add test for user created event
- [x] Infra: deploy core variable for loops api


geclos (commented):
👌🏼

Issue #420, opened 2024-10-14
Evaluations – Remove conflicts in evaluation objective between configuration and prompt


samufyi:
What?

Right now, there is a config when creating an evaluation that sets the result (numeric between 1 and 5), but we don't pass this to the prompt, so the user can set a range between 9 and 20, and it would be the one that takes into account.

In summary, there are 2 sources of truth.

https://www.figma.com/design/ODioXiqX8aeDMonsh0HBui/Latitude-Cloud?node-id=2738-34189&t=C31y3Hbykh3pzF2x-4

PR #515, opened 2024-10-28
Simpler evaluations part 1.3

csansoon:
[Part 1.3](https://github.com/latitude-dev/latitude-llm/issues/420)
Perform the migration to move all configuration data from evaluations to evaluationMetadataLlmAsJudgeLegacy and remove the attribute from the evaluations table.


andresgutgon (approved):

packages/core/drizzle/0087_legacy_evaluations_part_2.sql
```diff
@@ -0,0 +1,11 @@
...
+UPDATE "latitude"."llm_as_judge_evaluation_metadatas"
+SET "configuration" = "latitude"."evaluations"."configuration"
+FROM "latitude"."evaluations"
+WHERE "latitude"."evaluations"."metadata_id" = "latitude"."llm_as_judge_evaluation_metadatas"."id";--> statement-breakpoint
+
+--- Remove evaluation metadatas with NULL configuratios (these are llm_as_judge_evaluation_metadatas rows that were left when removing the original evaluation)
+DELETE FROM "latitude"."llm_as_judge_evaluation_metadatas" WHERE "latitude"."llm_as_judge_evaluation_metadatas"."configuration" IS NULL--> statement-breakpoint
+
+--- Set configuration column as NOT NULL
+ALTER TABLE "latitude"."llm_as_judge_evaluation_metadatas" ALTER COLUMN "configuration" SET NOT NULL;
```
csansoon, line 11: Moving configuration from `evaluations` to `llm_as_judge_evaluation_metadatas`, and then removing the `llm_as_judge_evaluation_metadatas` that were left over after removing their original `evaluations`

packages/core/src/schema/types.ts
```diff
@@ -58,7 +58,7 @@ export type LlmAsJudgeEvaluationMetadata = InferSelectModel<
 export type Subscription = InferSelectModel<typeof subscriptions>
 
 export type EvaluationDto = Evaluation & {
-  configuration?: LlmAsJudgeEvaluationMetadata['configuration']
+  configuration: LlmAsJudgeEvaluationMetadata['configuration'] | null
```
geclos, line 61: why the null? isn't it supposed to be not null after your migration ☝🏼 ?

geclos (approved):

packages/core/src/schema/types.ts
csansoon: This is the original `configuration` attribute from the `evaluations`, not the new one in the `LegacyMetadata`, so technically every value should be Null here now. This is just for compatibility while I'm doing changes and will be removed in the future.

PR #522, opened 2024-10-29
feat(documentEditor): minor improvements to performance when computing metadata

geclos:
We were computing metadata several times per document edit. This change makes the code more imperative and it's clearer what's happening.


packages/core/src/repositories/providerLogsRepository.ts
```diff
@@ -29,6 +29,10 @@ export class ProviderLogsRepository extends Repository<ProviderLog> {
 
   async findByDocumentUuid(documentUuid: string, opts: QueryOptions = {}) {
     const query = this.scope
+      .innerJoin(
+        documentLogs,
+        eq(providerLogs.documentLogUuid, documentLogs.uuid),
+      )
```
geclos, line 35: bug

PR #524, opened 2024-10-30
Allow caching system, user and assistant messages for Anthropic

andresgutgon:
# What?
We want to allow with our syntax for users to tag messages with an Anthropic cache control tag.

## Extract Anthropic system messages as configuration
Also as part of this PR we extract first system messages in an anthropic prompt as part of the configuration. This way we can remove the warning

## Allow system messages to have not only text. 
And last change in this PR is that we allow system messages to have more than one message. This is something that's not valid in OpenAI but is valid in Anthropic so we allow this at compiler level but we restric for OpenAI at Rules level

## TODO
- [x] Fix existing tests
- [x] Double check assistant with string is fine for Vercel SDK
- [x] Double check `system.content` is transformed into string before sending Vercel SDk
- [x] Update docs


packages/compiler/src/compiler/compile.test.ts
```diff
@@ -563,15 +381,41 @@ describe('conditional expressions', async () => {
...
+        name: undefined,
+        content: [
+          {
+            type: 'text',
+            text: 'Bar!',
+          },
+        ],
+      },
+    ])
+    /* expect(result2).toBe('Bar!') */
```
geclos, line 418: 👀

packages/compiler/src/types/message.ts
```diff
@@ -42,8 +43,14 @@ export type ToolRequestContent = {
...
-  | TextContent
-  | ImageContent
+  | {
+      type: ContentType.text
+      text: string
+    }
+  | {
+      type: ContentType.image
+      image: string | Uint8Array | Buffer | ArrayBuffer | URL
+    }
```
geclos, line 53: why?

geclos (approved):

packages/compiler/src/compiler/compile.test.ts
andresgutgon: Is a left over. Check now we do an `toEqual` on `result2`

packages/compiler/src/compiler/base/nodes/tags/message.ts
```diff
@@ -107,9 +109,10 @@ function buildMessage<R extends MessageRole>(
 
   if (role === MessageRole.assistant) {
     return {
+      ...attributes,
       role,
       toolCalls: toolCalls.map(({ value }) => value),
-      content: (content[0]! as TextContent).text,
+      content,
```
geclos, line 115: so has content been transformed before? I don't see where we are getting the content from the array now

packages/compiler/src/types/message.ts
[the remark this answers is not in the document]
andresgutgon: This is wrong. I have a changes to be commited.

packages/compiler/src/compiler/chain.test.ts
```diff
@@ -101,7 +101,12 @@ describe('chain', async () => {
 
     const systemMessage = conversation.messages[0]!
     expect(systemMessage.role).toBe('system')
-    expect(systemMessage.content).toBe('System message')
+    expect(systemMessage.content).toEqual([
+      {
+        type: 'text',
+        text: 'System message',
+      },
```
andresgutgon, line 108: This is critical. Before we discarded system messages because from compiler we send only the first `content` item in a message as a string. Now we send an array of `content` items like in this text.

packages/compiler/src/compiler/compile.ts
```diff
@@ -157,14 +156,14 @@ export class Compile {
       this.baseNodeError(errors.invalidToolCallPlacement, toolNode)
     })
 
-    if (content.length > 0) {
-      const message = {
-        role: MessageRole.system,
-        content: (content[0] as TextContent).text,
```
andresgutgon, line 163: We were losing text that the user was writing

packages/compiler/src/types/message.ts
```diff
@@ -56,11 +57,11 @@ export type ToolCall = {
 interface IMessage {
   role: MessageRole
   content: MessageContent[]
+  [key: string]: unknown
```
andresgutgon, line 60: We acknowledge people can put whatever they want as an attribute

```javascript
<user something={{ foo: 'bar' }} something_else="hi_nice_to_meet_you" />
```

We do the same with `<text>` blocks. This is not documented but you can do this:
```markdown
<system>
  Read this book and ask user questins
  <text cache_control={{ type: 'ephimeral' }}>...THE_ENTIRE_BOOK</text>

  Something else no cached
</system>
```

This will result in 3 system messages the one with the book cached in Anthropic in this case.

packages/compiler/src/compiler/base/nodes/tags/content.ts
```diff
@@ -29,17 +30,19 @@ export async function compile(
       isInsideContentTag: true,
     })
   }
-  const textContent = popStrayText()
+  const textContent = removeCommonIndent(popStrayText())
```
csansoon, line 33: Thankss

packages/core/src/services/ai/providers/rules/vercel.test.ts
```diff
@@ -0,0 +1,338 @@
...
+    ] as Message[]
+
+    const rules = vercelSdkRules(
+      { rules: [], messages, config },
+      Providers.Anthropic,
+    )
+    expect(rules.messages).toEqual([
+      {
+        role: 'user',
+        experimental_providerMetadata: {
```
geclos, line 140: i assume this tranformation only gets applied when the provider is anthropic no?

packages/core/src/services/ai/providers/rules/vercel.test.ts
```diff
@@ -0,0 +1,338 @@
...
+                some_attribute: 'some_assistant_value',
+                another_attribute: { another_assistant: 'value' },
+              },
+            },
+          },
+        ],
+      },
+    ])
+  })
+})
```
geclos, line 338: lovely

packages/core/src/services/ai/providers/rules/vercel.ts
```diff
@@ -0,0 +1,129 @@
...
+  message,
+  provider,
+}: {
+  message: SystemMessage
+  provider: Providers
+}): Message[] {
+  const content = message.content as TextContent[] | string
+
+  // NOTO: `applyCustomRules` can be invoked multiple times
+  // during a chaing. if system message.content is already
```
geclos, line 28: chaing* typo

geclos (commented):
what happens with users that have added a sytem prompt via front matter?

```
---
system: this is my system prompt
---

<response system='another system prompt' />
```

examples ☝🏼

packages/core/src/services/ai/providers/rules/vercel.test.ts
andresgutgon: Only when there is metadata. And it ends like this:
``` javascript
{
  experimental_providerMetadata: {
    ['provider_name']: {//....custom tag attributes }
  }
}
```

andresgutgon:
> what happens with users that have added a system prompt via front matter or an attribute in response tag?

I have to check. Honestly didn't check this case

What should happen? This is production
image

packages/core/src/services/providerLogs/serialize.ts
```diff
@@ -52,7 +52,8 @@ export function formatContext(
       Array.isArray(message.content) &&
       message.content &&
       message.content[0] &&
-      'text' in message.content[0]
+      'text' in message.content[0] &&
+      message.content[0].type === ContentType.text
```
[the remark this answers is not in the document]
andresgutgon: I think is not necessary. Removing... Let's typescript what says

packages/core/src/services/ai/providers/rules/vercel.test.ts
```diff
@@ -0,0 +1,338 @@
...
+    const rules = vercelSdkRules(
+      { rules: [], messages, config },
+      Providers.Anthropic,
+    )
+
+    expect(rules.messages).toEqual([
+      ...(messages.slice(0, 1) as Message[]),
+      {
+        role: 'assistant',
+        content: [{ type: 'text', text: 'Hello! How are you?' }],
```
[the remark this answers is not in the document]
andresgutgon: For consistency. Now always returns an array of parts.

packages/core/src/services/providerLogs/serialize.ts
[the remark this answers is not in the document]
geclos: don't know where that code is but i know we also handle tool calls correctly in playground

geclos:
> What should happen? This is production

Right now users using anthropic are using this system prompt in front matter and <response /> to set up the system message in their anthropic prompts. Ideally we should not break this case

packages/core/src/services/providerLogs/serialize.ts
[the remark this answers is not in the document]
andresgutgon: That code is this file we're commenting

andresgutgon:
> Right now users using anthropic are using this system prompt in front matter and to set up the system message in their anthropic prompts. Ideally we should not break this case.

Ok I have to check. I don't see any system message in the playground when using `system` in the front matter. Check the screenshots

docs/guides/prompt-manager/provider-rules/anthropic.mdx
```diff
@@ -0,0 +1,70 @@
+1. [Not only system messages](#rule-1-not-only-system-messages)
+2. [No system messages after assistant or user messages](#rule-2-no-system-messages-after-assistant-or-user-messages)
+
+### Rule 1: Not only system messages
+System messages has to be followed by at least one user or assistant message. Anthropic does not consider system messages as part of the messages. They split in the following way:
```
geclos, line 5: have* to

run the docs through an llm to fix grammar/typos please

csansoon (approved):

PR #525, opened 2024-10-30
Simpler evaluations part 2.2

csansoon:
[Part 2.2](https://github.com/latitude-dev/latitude-llm/issues/420)

Mainly, two changes:

- Update the EvaluationDto type and EvaluationRepository to fetch data from all new polymorphic relations (`metadata` and `resultConfiguration`).
- Adapt current create function to include the result configuration in both old `evaluation.metadata.configuration` and new `evaluation.resultConfiguration`. The old version will be set as legacy in the future.


packages/core/src/repositories/evaluationsRepository.ts
```diff
@@ -1,29 +1,41 @@
...
+    }),
+    ['id', 'createdAt', 'updatedAt'],
+  ),
+  resultConfiguration: omit(
+    getSharedTableColumns(evaluations.resultType, {
+      [EvaluationResultableType.Boolean]: evaluationConfigurationBoolean,
+      [EvaluationResultableType.Number]: evaluationConfigurationNumerical,
+      [EvaluationResultableType.Text]: evaluationConfigurationText,
+    }),
+    ['id', 'createdAt', 'updatedAt'],
```
csansoon, line 37: Here, the `metadata` attribute contains either the attributes from `evaluationMetadataLlmAsJudgeAdvanced` or `evaluationMetadataLlmAsJudgeSimple`, depending on the value of `evaluations.metadataType`. The `getSharedTableColumns` returns the correct SQL query to return the attributes from either one or the other depending on the actual value!

The same happens with `resultConfiguration`, which can be the attributes from `evaluationConfigurationBoolean`, `evaluationConfigurationNumerical` or `evaluationConfigurationText`, depending on the value of `evaluations.resultType`.

apps/web/src/actions/evaluations/create.ts
```diff
@@ -34,13 +27,12 @@ export const createEvaluationAction = authProcedure
     { type: 'json' },
   )
   .handler(async ({ input, ctx }) => {
-    const result = await createEvaluation({
+    const result = await createAdvancedEvaluation({
```
geclos, line 30: what's an advanced evaluation

packages/core/drizzle/0091_bigint_evaluations_result_config.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "latitude"."evaluation_configuration_numerical" ALTER COLUMN "min_value" SET DATA TYPE bigint;--> statement-breakpoint
+ALTER TABLE "latitude"."evaluation_configuration_numerical" ALTER COLUMN "max_value" SET DATA TYPE bigint;
```
geclos, line 2: what types were they before?

packages/core/src/lib/compactObject.ts
```diff
@@ -1,9 +1,7 @@
...
-    if (value !== undefined) newObj[key] = value
-  })
-
-  return newObj
+export const compactObject = (
+  obj: Record<string, unknown | undefined>,
+): Record<string, unknown> => {
+  return Object.fromEntries(
+    Object.entries(obj).filter(([_, value]) => value !== undefined),
+  )
```
geclos, line 6: is this changing behaviour? this is used in many places

packages/core/src/schema/models/evaluationConfigurationNumerical.ts
```diff
@@ -7,8 +7,8 @@ export const evaluationConfigurationNumerical = latitudeSchema.table(
   'evaluation_configuration_numerical',
   {
     id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
-    minValue: numeric('min_value').notNull(),
-    maxValue: numeric('max_value').notNull(),
+    minValue: bigint('min_value', { mode: 'number' }).notNull(),
+    maxValue: bigint('max_value', { mode: 'number' }).notNull(),
```
geclos, line 11: this type change does not cause data removal no?

apps/web/src/actions/evaluations/create.ts
csansoon: An evaluation where the user defines the prompt (the ones we have now)

packages/core/src/services/evaluations/create.ts
```diff
@@ -194,3 +225,68 @@ async function findProvider(workspace: Workspace, db: Database) {
...
+        prompt: promptWithProvider,
+        configuration,
+        templateId: metadata.templateId ?? null,
+      } as Omit<EvaluationMetadataLlmAsJudgeAdvanced, 'id'>,
+      resultType: configuration.type,
+      resultConfiguration,
+    },
+    db,
+  )
+}
```
geclos, line 292: there's no way you are merging this without adding tests for this service 😅

packages/core/drizzle/0091_bigint_evaluations_result_config.sql
csansoon: "numeric", which for some reason is stored as a string. Anyways, this has to match with the value of `EvaluationResultableNumber`, which is a `bigint`

geclos (approved):
look good but add tests this is touching sensible code

packages/core/src/lib/compactObject.ts
csansoon: Nope, same behaviour. Could just undo the change, just felt this was more elegant :)

packages/core/src/schema/models/evaluationConfigurationNumerical.ts
csansoon: No, there's still no rows whatsoever in this table. There will be from now on tho

packages/core/src/services/evaluations/create.ts
csansoon: There are tests for this service! They check that everything is still working as expected.

PR #530, opened 2024-11-01
feat(commits): add caching for document responses

geclos:
[feat(commits): Implement cached response for chain responses](https://github.com/latitude-dev/latitude-llm/pull/530/commits/9b8a969715bdf0513a4041af4f52918821d3600f)
- Added `DocumentCache` class to handle caching of chain step responses based on workspace, config, and conversation.
- Introduced `getCachedResponse` and `setCachedResponse` functions to retrieve and store cached responses.
- Updated `runChain` function to utilize cached responses if available, reducing redundant calls to the AI module.
- Added tests to verify the caching behavior and ensure correct functionality with various configurations and scenarios.

This change improves performance by avoiding repeated AI calls for the same conversation and configuration, leveraging cached responses when appropriate.

### DIFF
--- a/packages/core/src/services/commits/promptCache.ts
+++ b/packages/core/src/services/commits/promptCache.ts
@@ -1,0 +1,80 @@
+import { hash } from 'crypto'
+
+import { Config, Conversation } from '@latitude-data/compiler'
+
+import { ChainStepResponse, StreamType, type Workspace } from '../../browser'
+import { cache } from '../../cache'
+
+function generateCacheKey(
+  workspace: Workspace,
+  config: Config,
+  conversation: Conversation,
+): string {
+  const k = hash(
+    'sha256',
+    `${JSON.stringify(conversation)}:${JSON.stringify(config)}`,
+  )
+
+  return `workspace:${workspace.id}:prompt:${k}`
+}
+
+function shouldCache(config: Config): boolean {
+  const temp = parseFloat(config.temperature as any)
+  return isNaN(temp) || temp === 0
+}
+
+async function getFromCache(
+  key: string,
+): Promise<ChainStepResponse<StreamType> | undefined> {
+  try {
+    const c = await cache()
+    const cachedResponseStr = await c.get(key)
+    return cachedResponseStr ? JSON.parse(cachedResponseStr) : undefined
+  } catch (e) {
+    return undefined
+  }
+}
+
+async function setToCache(
+  key: string,
+  response: ChainStepResponse<StreamType>,
+): Promise<void> {
+  try {
+    const c = await cache()
+    await c.set(key, JSON.stringify(response))
+  } catch (e) {
+    // Silently fail cache writes
+  }
+}
+
+export async function getCachedResponse({
+  workspace,
+  config,
+  conversation,
+}: {
+  workspace: Workspace
+  config: Config
+  conversation: Conversation
+}) {
+  if (!shouldCache(config)) return undefined
+
+  const key = generateCacheKey(workspace, config, conversation)
+  return await getFromCache(key)
+}
+
+export async function setCachedResponse({
+  workspace,
+  config,
+  conversation,
+  response,
+}: {
+  workspace: Workspace
+  config: Config
+  conversation: Conversation
+  response: ChainStepResponse<StreamType>
+}) {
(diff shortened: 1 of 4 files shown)


packages/core/src/services/documents/getDocumentMetadata.ts
```diff
@@ -1,47 +1,37 @@
...
   if (!docInCommit) {
-    return Result.error(
-      new UnprocessableEntityError('Document not found in commit', {}),
-    )
+    return Result.error(new NotFoundError('Document not found in commit'))
   }
-
-  if (commit.mergedAt != null && document.resolvedContent != null) {
-    return Result.ok(document.resolvedContent!)
-  }
```
geclos, line 43: I removed this early return which is a small optimization to avoid computing the metadata of a document, we always want to compute the metadata in order to know if the response is cached or not

packages/core/src/services/commits/RunDocumentChecker/index.ts
```diff
@@ -8,66 +8,48 @@ import { createRunError } from '../../runErrors/create'
 
 type RunDocumentErrorCodes = RunErrorCodes.ChainCompileError
 
-export class RunDocumentChecker {
```
andresgutgon, line 11: Is this pure ocd? 😂

andresgutgon (dismissed):

packages/core/src/services/documents/getDocumentMetadata.ts
geclos: yeah I think it's worth it when we trade it off with the cache

geclos:
Cannot be merged because of chains

packages/core/src/services/commits/promptCache.ts
```diff
@@ -0,0 +1,98 @@
...
+  config,
+  conversation,
+  response,
+}: {
+  workspace: Workspace
+  config: Config
+  conversation: Conversation
+  response: ChainStepResponse<StreamType>
+}) {
+  const documentCache = new DocumentCache({ workspace, config, conversation })
```
andresgutgon, line 96: Why a class in this case?

csansoon (approved):
looks good!

packages/core/src/services/commits/promptCache.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    return `workspace:${this.workspace.id}:prompt:${k}`
+  }
+
+  private shouldCache(): boolean {
+    const temp = parseFloat(this.config.temperature as any)
+    return isNaN(temp) || temp === 0
+  }
+
+  async get(): Promise<ChainStepResponse<StreamType> | undefined> {
```
andresgutgon, line 45: Micro ocd. But put private functions under public methods

packages/core/src/services/commits/promptCache.ts
geclos: yeah no real reason to be honest it's what the ai generated lol, I will change it to pure functions

packages/core/src/services/commits/promptCache.ts
```diff
@@ -0,0 +1,98 @@
...
+      'sha256',
+      `${JSON.stringify(this.conversation)}:${JSON.stringify(this.config)}`,
+    )
+
+    return `workspace:${this.workspace.id}:prompt:${k}`
+  }
+
+  private shouldCache(): boolean {
+    const temp = parseFloat(this.config.temperature as any)
+    return isNaN(temp) || temp === 0
```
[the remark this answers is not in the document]
geclos: because temperature changes the llm output even if with the same input, so it's a cache-breaking configuratiaon

packages/core/src/services/commits/promptCache.ts
geclos: I'm removing the class

andresgutgon (approved):

packages/core/src/services/commits/promptCache.ts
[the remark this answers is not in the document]
andresgutgon: So prompts with temperature are never cached? This should be explained in the docs of the cache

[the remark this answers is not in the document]
geclos: yup

PR #532, opened 2024-11-04
Add Sign Up button to Log In page

neoxelox:
Added Sign Up button to Log In Page.

### DIFF
--- a/apps/web/src/components/layouts/FocusLayout.tsx
+++ b/apps/web/src/components/layouts/FocusLayout.tsx
@@ -6,7 +6,7 @@ export default function FocusLayout({
   children,
 }: {
   header: ReactNode
-  footer: ReactNode
+  footer?: ReactNode
   children?: ReactNode
 }) {
   return (
@@ -14,7 +14,11 @@ export default function FocusLayout({
       <div className='flex flex-col gap-y-6 max-w-[22rem]'>
         <div>{header}</div>
         {children}
-        <div>{footer}</div>
+        {footer && (
+          <div className='flex flex-col items-center justify-center gap-y-6'>
+            {footer}
+          </div>
+        )}
       </div>
     </div>
   )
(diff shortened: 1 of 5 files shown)


apps/web/src/components/layouts/FocusLayout.tsx
```diff
@@ -2,16 +2,19 @@ import { ReactNode } from 'react'
 
 export default function FocusLayout({
   header,
+  footer,
   children,
 }: {
   header: ReactNode
+  footer: ReactNode
   children?: ReactNode
 }) {
```
csansoon, line 11: `FocusLayout` is being used in more places, we should make `footer` optional.

packages/web-ui/src/ds/molecules/FocusFooter/index.tsx
```diff
@@ -0,0 +1,9 @@
+import { ReactNode } from 'react'
+
+export default function FocusFooter({ content }: { content: ReactNode }) {
```
andresgutgon, line 3: This is super specific for the signup. Doesn't look like something part of design system.

apps/web/src/components/layouts/FocusLayout.tsx
```diff
@@ -2,16 +2,19 @@ import { ReactNode } from 'react'
...
   return (
     <div className='flex flex-col items-center justify-center h-screen'>
       <div className='flex flex-col gap-y-6 max-w-[22rem]'>
         <div>{header}</div>
         {children}
+        <div>{footer}</div>
```
andresgutgon, line 17: Do an if here and put that focus footer div styles
```javascript
{footer ? (<div classNmame="...">{footer}</div>) : null}
```

andresgutgon (approved):

apps/web/src/components/layouts/FocusLayout.tsx
csansoon: Why the styling there? The header is not doing that

andresgutgon: Good point. Because people change opinions. I did the header and now I changed my opinion.

PR #533, opened 2024-11-04
Simpler evaluations — part 2.4

csansoon:
[Part 2.4](https://github.com/latitude-dev/latitude-llm/issues/420)

Although legacy evaluation configuration is still available from the `metadata` (the `configuration` jsonb column), all the code has been adapted to use the new `configuration` attribute instead (the respective `EvaluationConfiguration` table for each result type).

Additionally, added the prompt that will generate from the configuration of the new "simple evaluations"


packages/core/src/services/evaluations/prompt/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+{{ messages.all }}
+
+Also, here is some aditional metadata about the LLM response. It may or may not be relevant to your objective.
+ - Cost: {{ cost }} cents.
+ - Duration: {{ duration }} milliseconds.
+
+You must respond with a JSON object with the following properties:
+ - value: ${valueInformation(evaluation)}
+ - reason: A string explaining your evaluation decision.
```
csansoon, line 104: The evaluation that will be generated from the new "simple evaluations" configuration!

apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/dashboard/_components/EvaluationStats.tsx
```diff
@@ -34,10 +37,14 @@ export default function EvaluationStats({
     useConnectedDocuments({ evaluation })
 
   useEffect(() => {
-    readMetadata({ prompt: evaluation.metadata.prompt }).then((metadata) => {
-      const metadataModel = (metadata.config['model'] as string) ?? 'Unknown'
-      setModel(metadataModel)
-    })
+    if (evaluation.metadataType === EvaluationMetadataType.LlmAsJudgeAdvanced) {
+      readMetadata({ prompt: evaluation.metadata.prompt }).then((metadata) => {
```
Comment (geclos) [inline apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/dashboard/_components/EvaluationStats.tsx:41]: can this promise fail? what should we do if it fails?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/MetricsSummary/index.tsx
```diff
@@ -27,7 +25,6 @@ export function MetricsSummary<T extends boolean>({
           commit={commit}
           evaluation={evaluation}
           documentUuid={documentUuid}
-          isNumeric={isNumeric}
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationId]/_components/MetricsSummary/index.tsx:30]: thanks

packages/core/src/services/evaluations/compiler/index.ts
```diff
@@ -1,44 +1 @@
...
-  const metadata = await readMetadata({
-    prompt: evaluation.metadata.prompt,
-    withParameters: Object.keys(PARAMETERS_FROM_LOG),
-  })
-
-  return Result.ok(metadata)
-}
-
-export { PARAMETERS_FROM_LOG }
+export { PARAMETERS_FROM_LOG } from './constants'
```
geclos, line 1: hombre might as well remove this file if it only exports this 😅

packages/core/src/services/evaluations/prompt/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+You’re an expert LLM evaluator. Your objective is to evaluate the response from an LLM model based on this goal:
+
+${evaluation.metadata.objective}
+
+${evaluation.metadata.additionalInstructions ? 'Additionally, you should follow these instructions:\n\n' + evaluation.metadata.additionalInstructions : ''}
+
+Now, evaluate the assistant response for the following conversation, based on your instructions:
+
+{{ messages.all }}
```
geclos, line 96: hmm this is a json array of the whole convo might be better to concat context and response instead

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/RefineModal/steps/2_SelectEvaluationResults/SelectableEvaluationResultsTable.tsx
```diff
@@ -29,21 +29,17 @@ export const ResultCellContent = ({
...
   }
 
-  if (
-    evaluation.metadata.configuration.type === EvaluationResultableType.Number
-  ) {
-    const minValue = evaluation.metadata.configuration.detail?.range.from ?? 0
-    const maxValue = evaluation.metadata.configuration.detail?.range.to ?? 10
+  if (evaluation.resultType === EvaluationResultableType.Number) {
+    const minValue = evaluation.resultConfiguration.minValue
+    const maxValue = evaluation.resultConfiguration.maxValue
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/RefineModal/steps/2_SelectEvaluationResults/SelectableEvaluationResultsTable.tsx:42]: Super nice

packages/core/src/services/evaluations/prompt/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+Also, here is some aditional metadata about the LLM response. It may or may not be relevant to your objective.
+ - Cost: {{ cost }} cents.
+ - Duration: {{ duration }} milliseconds.
```
geclos, line 100: cost and duration can be null or 0, I wouldn't include this section of the prompt if that is the case

packages/core/src/services/evaluations/prompt/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+You must respond with a JSON object with the following properties:
+ - value: ${valueInformation(evaluation)}
+ - reason: A string explaining your evaluation decision.
+`.trim(),
+  )
+}
```
geclos, line 107: what a beautiful file to add some tests to

packages/core/src/services/evaluations/prompt/index.ts
geclos, line 104: is that even necessary? we are already adding a json schema and structure output in the ai request

packages/core/src/services/evaluations/prompt/index.ts
```diff
@@ -0,0 +1,107 @@
...
+  const resultSchema = {
+    [EvaluationResultableType.Boolean]: 'boolean',
+    [EvaluationResultableType.Number]: 'number',
+    [EvaluationResultableType.Text]: 'string',
+  } as const
+
+  return Result.ok(
+    `
+---
+provider: ${providerResult.value!.name}
```
geclos, line 73: might as well unwrap here instead of asserting value

packages/core/src/services/evaluations/prompt/index.ts
```diff
@@ -0,0 +1,107 @@
...
+model: ${evaluation.metadata.model}
+temperature: 0.2
```
geclos, line 75: ? are we sure about this? I'd much rather evaluations be deterministic

andresgutgon, line 75: Why this temperature?

andresgutgon (approved):

packages/core/src/services/evaluations/prompt/index.ts
csansoon: We cannot predict the user's use case, we do not know if there's something in the conversation worth adding to the evaluation.

I think we should add a way for the user to select what should we add to the evaluation as context, but it has been discarded for now. Can be added in the future though

apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/dashboard/_components/EvaluationStats.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/dashboard/_components/EvaluationStats.tsx:41]: Is should not fail. A fail here would mean something in the db is wrong, which should not happen and is covered by the create service.

I think in this case it is better to just let it fail if something happens, and get a Sentry notification

packages/core/src/services/evaluations/prompt/index.ts
csansoon: This also adds the descriptions for these values, which are defined for the user. Also, OpenAI recommends to explain the schema in the context in addition to defining it in the configuration.

packages/core/src/services/evaluations/prompt/index.ts
csansoon: Hhmm yeah, I'll change it to `0`

packages/core/src/services/evaluations/prompt/index.ts
geclos: description can also be added in the json schema then!

packages/core/src/services/evaluations/prompt/index.ts
csansoon: Added some tests to assert the generated prompt does not fail.

Issue #527, opened 2024-10-30
Bug: check original log always displays latest log instead of the log that was evaluated


geclos:
1. Create an evaluation and activate live evaluation
2. Run the prompt
3. Chat with the prompt
4. Go to the evaluation result and click on "check original log"
5. Bug: it displays the full conversation rather than the prompt as it was evaluated in step 2

geclos:
https://github.com/latitude-dev/latitude-llm/pull/560

PR #536, opened 2024-11-04
feat(schema): migrate documentLogId pointer from evaluationResults to providerLogs

geclos:
Added new fields `evaluatedProviderLogId` and `evaluationProviderLogId` to the `evaluationResults` schema. These fields are intended to replace the `documentLogId` and providerLogId fields. The reason is that we want to track the precise provider log that was evaluated, not the document log. Because an evaluation can be run at different points in time over a document and the document might have different providerlogs associated with it at each point.

Related issue https://github.com/latitude-dev/latitude-llm/issues/527

### DIFF
--- a/packages/core/src/schema/models/evaluationResults.ts
+++ b/packages/core/src/schema/models/evaluationResults.ts
@@ -21,12 +21,20 @@ export const evaluationResults = latitudeSchema.table(
     evaluationId: bigint('evaluation_id', { mode: 'number' })
       .notNull()
       .references(() => evaluations.id, { onDelete: 'cascade' }),
+    // TODO: remove
     documentLogId: bigint('document_log_id', { mode: 'number' })
       .notNull()
       .references(() => documentLogs.id),
+    // TODO: remove
     providerLogId: bigint('provider_log_id', { mode: 'number' }).references(
       () => providerLogs.id,
     ),
+    evaluatedProviderLogId: bigint('evaluated_provider_log_id', {
+      mode: 'number',
+    }).references(() => providerLogs.id),
+    evaluationProviderLogId: bigint('evaluation_provider_log_id', {
+      mode: 'number',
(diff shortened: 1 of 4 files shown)


packages/core/src/schema/models/evaluationResults.ts
```diff
@@ -21,22 +21,38 @@ export const evaluationResults = latitudeSchema.table(
     evaluationId: bigint('evaluation_id', { mode: 'number' })
       .notNull()
       .references(() => evaluations.id, { onDelete: 'cascade' }),
+    // TODO: remove
     documentLogId: bigint('document_log_id', { mode: 'number' })
```
andresgutgon, line 25: Are we losing something by not having the reference to the log? I guess we always can get it through provider log

andresgutgon (approved):

packages/core/src/schema/models/evaluationResults.ts
```diff
@@ -21,22 +21,38 @@ export const evaluationResults = latitudeSchema.table(
...
+    // TODO: remove
     providerLogId: bigint('provider_log_id', { mode: 'number' }).references(
       () => providerLogs.id,
     ),
+    evaluatedProviderLogId: bigint('evaluated_provider_log_id', {
+      mode: 'number',
+    }).references(() => providerLogs.id),
+    evaluationProviderLogId: bigint('evaluation_provider_log_id', {
+      mode: 'number',
+    }).references(() => providerLogs.id),
```
csansoon, line 37: Is this no the same as `providerLogId`? We could just replace the attribute name in drizzle, no need for a migration

geclos: not backwards compatible

packages/core/src/schema/models/evaluationResults.ts
geclos: n

PR #542, opened 2024-11-04
Evaluations: remove configuration from db schema

csansoon:
Last step of the migration: The configuration is nowhere found in the code, all that’s left is to remove it from the database.

### DIFF
--- a/packages/core/drizzle/0096_curvy_scrambler.sql
+++ b/packages/core/drizzle/0096_curvy_scrambler.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "latitude"."llm_as_judge_evaluation_metadatas" DROP COLUMN IF EXISTS "configuration";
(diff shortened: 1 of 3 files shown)


packages/core/drizzle/0095_handy_skreet.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "latitude"."llm_as_judge_evaluation_metadatas" DROP COLUMN IF EXISTS "configuration";
```
andresgutgon, line 1: I'll do the stupid question here. But the data migration already put the context of this in the new place right?

csansoon: Yes! The code in production is not even aware this attribute exists, everything is being added and read from the new attributes/tables/relations

andresgutgon (approved):

PR #543, opened 2024-11-04
New evaluations UI — part 1

csansoon:


apps/web/src/actions/evaluations/create.ts
```diff
@@ -12,28 +15,51 @@ export const createEvaluationAction = authProcedure
...
       description: z.string(),
-      configuration: z.object({
-        type: z.nativeEnum(EvaluationResultableType),
-        detail: z
-          .object({ range: z.object({ from: z.number(), to: z.number() }) })
-          .optional(),
-      }),
-      metadata: z
-        .object({
+      resultConfiguration: z.discriminatedUnion('type', [
```
geclos, line 18: subtle bug always use discriminatedUnions when typing api endpoints

csansoon (commented):
Can't approve my own PRs! But lgtm 👍

apps/web/src/actions/evaluations/updateContent.ts
```diff
@@ -11,19 +15,53 @@ export const updateEvaluationContentAction = authProcedure
   .input(
     z.object({
       id: z.number(),
-      metadata: z.object({
-        prompt: z.string(),
-      }),
+      metadata: z
+        .discriminatedUnion('type', [
```
geclos, line 19: ditto

apps/web/src/actions/evaluations/create.ts
csansoon: :o What's the difference?

apps/web/src/actions/evaluations/updateContent.ts
```diff
@@ -11,19 +15,53 @@ export const updateEvaluationContentAction = authProcedure
...
+            minValue: z.number().optional(),
+            maxValue: z.number().optional(),
+            minValueDescription: z.string().optional(),
+            maxValueDescription: z.string().optional(),
+          }),
+          z.object({
+            type: z.literal(EvaluationResultableType.Text),
+            valueDescription: z.string(),
+          }),
+        ])
```
andresgutgon, line 51: This can be dry `configuration: ConfigurationType,` and use in both actions

apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Editor/Simple/index.tsx
```diff
@@ -0,0 +1,5 @@
+'use client'
+
+export default function SimpleEvaluationEditor() {
+  return null
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Editor/Simple/index.tsx:4]: So in simple we show nothing?

apps/web/src/actions/evaluations/create.ts
geclos: union type checks against the first matching element whereas discriminatedUnion type checks against the element that matches the "type" prop. The former can cause unexpected issues such as always matching the first element in the union if it contains all optional elements

apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Editor/Simple/index.tsx
Comment (geclos) [reply apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Editor/Simple/index.tsx:4]: yeah there's no simple evals yet

apps/web/src/actions/evaluations/updateContent.ts
geclos: i don't get the comment sorry

andresgutgon (dismissed):

apps/web/src/actions/evaluations/updateContent.ts
andresgutgon: The enums are the same when creating and updating. Extract it to be reused in both.

PR #551, opened 2024-11-06
PromptL compiler v1 - Chapter 1: ifs and fors

csansoon:
This is the first from a series of PRs to implement [the new spec for the PromptL language](https://latitudedata.notion.site/PromptL-v1-12fa92110b9780d7824ef3c540e4bed2?pvs=4).

**What's in this PR?**

Here, I have cloned the current `compiler` package to a `prompt_compiler`. We want to keep both compilers to support the legacy syntax on all documents that are already using it.

In addition to this, I have implemented the `{{ if condition }} … {{ else }} … {{ endif }}` and `{{ for item, index in element }} … {{ endfor }}` syntax! (And adapted + added corresponding tests)

Most of the other code is just from the copy.
These are the actual files where I added the main stuff:
 - `parser/state/mustache.ts` (and the brand new `parser/state/mustache.test.ts`)
 - `compiler/base/nodes/if.ts` (and `compiler/base/nodes/if.test.ts`
 - `compiler/base/nodes/for.ts` (and `compiler/base/nodes/for.test.ts`
 -


packages/promptl_compiler/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@latitude-data/promptl_compiler",
```
andresgutgon, line 2: I bit weird this name imo. Maybe something shorter without underscore. Don't know what tho

packages/promptl_compiler/src/compiler/base/index.ts
andresgutgon, line 1: Empty

packages/promptl_compiler/package.json
csansoon: Changed to just `promptl` 👍

andresgutgon (approved):

geclos (approved):

PR #552, opened 2024-11-06
feature: migrate provider log id to evaluation provider log id

geclos:
Also fixes bug where provider logs weren't being saved for cached prompts and add tests for this case.

### DIFF
--- a/packages/core/src/services/chains/ChainValidator/index.ts
+++ b/packages/core/src/services/chains/ChainValidator/index.ts
@@ -23,169 +23,156 @@ export type ValidatedStep = {
 type JSONOverride = { schema: JSONSchema7; output: 'object' | 'array' }
 export type ConfigOverrides = JSONOverride | { output: 'no-schema' }
 
-export class ChainValidator {
-  private workspace: Workspace
-  private prevText: string | undefined
-  private chain: Chain
-  private providersMap: CachedApiKeys
-  private configOverrides?: ConfigOverrides
+type ValidatorContext = {
+  workspace: Workspace
+  prevText: string | undefined
+  chain: Chain
+  providersMap: CachedApiKeys
+  configOverrides?: ConfigOverrides
+}
 
-  constructor({
-    workspace,
-    prevText,
-    chain,
-    providersMap,
-    configOverrides,
-  }: {
-    workspace: Workspace
-    prevText: string | undefined
-    chain: Chain
-    providersMap: CachedApiKeys
-    configOverrides?: ConfigOverrides
-  }) {
-    this.workspace = workspace
-    this.prevText = prevText
-    this.chain = chain
-    this.providersMap = providersMap
-    this.configOverrides = configOverrides
-  }
+const getInputSchema = (
+  chainCompleted: boolean,
+  config: Config,
+  configOverrides?: ConfigOverrides,
+): JSONSchema7 | undefined => {
+  if (!chainCompleted) return undefined
+  const overrideSchema =
+    configOverrides && 'schema' in configOverrides
+      ? configOverrides.schema
+      : undefined
+  return overrideSchema || config.schema
+}
 
-  async call() {
-    const chainResult = await this.safeChain()
-    if (chainResult.error) return chainResult
+const getOutputType = (
+  chainCompleted: boolean,
+  config: Config,
+  configOverrides?: ConfigOverrides,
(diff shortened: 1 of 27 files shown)


packages/core/src/services/chains/ChainValidator/index.ts
```diff
@@ -23,169 +23,156 @@ export type ValidatedStep = {
...
+    config: rule.config as Config,
+    chainCompleted,
+    conversation: {
+      ...conversation,
+      messages: rule?.messages ?? conversation.messages,
+    },
+    schema: getInputSchema(chainCompleted, config, configOverrides),
+    output: getOutputType(chainCompleted, config, configOverrides),
+  })
 }
```
geclos, line 178: removed class implementation

packages/core/src/services/chains/ProviderProcessor/index.ts
```diff
@@ -1,148 +1,97 @@
...
-
-    return Result.error(
-      new ChainError({
-        code: RunErrorCodes.UnsupportedProviderResponseTypeError,
-        message: `Invalid stream type ${type} result is not a textStream or objectStream`,
-      }),
-    )
-  }
+  return processStreamObject({ aiResult, commonData })
 }
```
geclos, line 97: moved to functional implementation and removed the code to create provider log, it was coupled to this service

packages/core/src/services/chains/ProviderProcessor/processStreamText.ts
```diff
@@ -21,37 +20,3 @@ async function processResponse({
...
-  aiResult: Awaited<AIReturn<'text'>>
-}) {
-  const response = await processResponse({ aiResult, commonData })
-  const providerLogsData = await createProviderLogData({
-    response,
-    aiResult,
-    commonData,
-  })
-  return { response, providerLogsData }
-}
```
geclos, line 57: simplified these two files in a single method that build the provider log dto

packages/core/src/services/chains/ChainValidator/index.ts
andresgutgon: Ya lo veo

packages/core/src/services/chains/run.ts
```diff
@@ -210,20 +226,42 @@ async function runStep({
...
+        workspace,
+        source,
+        provider: step.provider,
+        conversation: step.conversation,
+        stepStartTime,
+        errorableUuid,
+        response,
+      }),
+      saveSyncProviderLogs: step.chainCompleted,
+    })
```
andresgutgon, line 263: The thing I don't like about not using classes is that force to use more functions. And make this entry point more complex. `run.ts` should be super easy to read and understand. Is a key piece. 

That was my reasoning when I encapsulated some pieces in classes.

andresgutgon (approved):

packages/core/src/services/chains/run.ts
geclos: hmm the file should be exactly the same length or less since changing the implementation from classes to functional does not change the amount of calls. In fact it should be simpler since this:

```
validateChain(...)
```

is simpler to parse than:
```
const chainValidator = new ChainValidator(...)

chainValidator.call()
```

If you are seeing more function calls is because I've decoupled some code that was previously coupled (because I need it to not be coupled to fix the bug), mainly ProviderProcessor which was doing more than I needed it to do

packages/core/src/services/chains/run.ts
```diff
@@ -256,3 +297,16 @@ async function runStep({
...
+  const invalidType = type !== 'text' && type !== 'object'
+  if (!invalidType) return Result.nil()
+
+  return Result.error(
+    new ChainError({
+      code: RunErrorCodes.UnsupportedProviderResponseTypeError,
+      message: `Invalid stream type ${type} result is not a textStream or objectStream`,
+    }),
+  )
+}
```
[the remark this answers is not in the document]
geclos: We can put it in another file if it helps. But this is actually better, because it splits responsibilities and makes the code easier to parse. 

This:
```
// chains/run.ts

checkValidType(...)
parseResponse(...)
createProviderLog(...)
```

Is better than this:
```
// chains/run.ts
providerProcessor(...)

// providerProcessor.ts
checkValidType(...)
parseResponse(...)
createProviderLog(...)
```

PR #557, opened 2024-11-06
feat(core): add SQL migration for updating evaluated provider log IDs

geclos:
This commit introduces a new SQL migration file `0097_mushy_xavin.sql` that updates the `evaluated_provider_log_id` in the `evaluation_results` table. The migration ensures that the latest provider log is associated with each evaluation result, provided the `evaluated_provider_log_id` is currently null. This allows us to point to the actual provider log that was evaluated, instead of the document log which does not represent a particular conversation in time.

### DIFF
--- a/packages/core/drizzle/0097_mushy_xavin.sql
+++ b/packages/core/drizzle/0097_mushy_xavin.sql
@@ -1,0 +1,17 @@
+-- Custom SQL migration file, put you code below! --
+
+WITH latest_provider_logs AS (
+  SELECT 
+    pl.id as provider_log_id,
+    pl.document_log_uuid,
+    ROW_NUMBER() OVER (PARTITION BY pl.document_log_uuid ORDER BY pl.generated_at DESC) as rn
+  FROM provider_logs pl
+  WHERE pl.document_log_uuid IS NOT NULL
+)
+UPDATE evaluation_results er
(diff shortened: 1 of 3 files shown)


packages/core/drizzle/0097_mushy_xavin.sql
```diff
@@ -0,0 +1,17 @@
...
+  FROM provider_logs pl
+  WHERE pl.document_log_uuid IS NOT NULL
+)
+UPDATE evaluation_results er
+SET evaluated_provider_log_id = lpl.provider_log_id
+FROM document_logs dl
+JOIN latest_provider_logs lpl ON dl.uuid = lpl.document_log_uuid
+WHERE er.document_log_id = dl.id
+AND lpl.rn = 1
+AND er.evaluated_provider_log_id IS NULL;
```
geclos, line 17: 35k rows getting migrated here

csansoon, line 17: Maybe filter by `rn = 1` in the subquery instead? Although performance is not really important here...

csansoon (approved):
lgtm

packages/core/drizzle/0097_mushy_xavin.sql
andresgutgon: Scary

PR #565, opened 2024-11-07
PromptL compiler v1 - Chapter 2: Content tags and tool calls

csansoon:
This is the second from a series of PRs to implement [the new spec for the PromptL language](https://latitudedata.notion.site/PromptL-v1-12fa92110b9780d7824ef3c540e4bed2?pvs=4).

**What's in this PR?**

Here, I have changed the `<text>` and `<image>` tags to `<content-text>` and `<content-image>`, and added a generic `<content>` tag which requires a "type" attribute.

In addition to this, now tool calls are treated as content too. This is because it is defined in the same way as other contents and has the same requirements.

Node: Some providers, such as Anthropic, do consider tool calls as message content! Although others, like OpenAI, do not. We’ve decided to treat them as content by default. However, we’ll introduce an automatic adapter in a future PR. This adapter will format the returned object according to the selected API. As a result, the compiler’s internal types must hold all necessary information. This will enable transformation to any API schema, regardless of the internal schema in use internally by the compiler's logic.

Other minor changes:
- We now allow an optional argument `defaultRole` to select which role will be used when grouping stray text into a message.
- All message content will be defined by an array of Contents (text, images, or tool calls – the latter one only being allowed inside assistant messages), instead of a single string. This allows compatibility with all providers.


packages/promptl/src/parser/state/mustache.test.ts
```diff
@@ -1,50 +1,38 @@
...
-  errorClass: new () => T,
-): T => {
-  try {
-    action()
-  } catch (err) {
-    expect(err).toBeInstanceOf(errorClass)
-    return err as T
-  }
-  throw new Error('Expected an error to be thrown')
-}
```
andresgutgon, line 18: So much dry. Nice

pnpm-lock.yaml
andresgutgon, line 1: Why so many changes? I don't see any change in `package.json`?

andresgutgon (changes requested):
Review why `pnpm-lock.yaml` has so many changes and 👍

pnpm-lock.yaml
csansoon: My vscode prettified it lol

geclos:
> However, we’ll introduce an automatic adapter in a future PR

hmm thinking about this it might be worth making these adapters as external packages and have the compiler accept custom adapters, so that we can leverage the community to build their own adapters for other providers since I'm sure we won't be able to support all edge cases for all providers out of the box. The adapters should be pure functions that receive a particular input from the compiler that is always the same, and output whatever the provider they target requires. 

It's also important that these adapters are as decoupled from the compiler internals as possible, in order to reduce the barrier of entry for developers.

Issue #559, opened 2024-11-06
Weird coloring in some UI


geclos:
![image](https://github.com/user-attachments/assets/8b19aff0-9364-491d-ba4c-0a7df539cbc3)
![image](https://github.com/user-attachments/assets/06f3a1a8-2cc4-4f1e-a433-0df428600c77)
![image](https://github.com/user-attachments/assets/28438cc6-62e9-437a-ad4b-592657bd37ed)

andresgutgon:
PR: https://github.com/latitude-dev/latitude-llm/pull/568

geclos:
careful this PR does not fix this one which is in the "deploy this prompt" modal

<img width="384" alt="image" src="https://github.com/user-attachments/assets/eabe3332-637d-4a48-b2f5-0c2420ecdff1">

andresgutgon:
Fixed: https://github.com/latitude-dev/latitude-llm/commit/4fdc8a05c4d76d2dc973a26b7ebfc8dd42199554

PR #568, opened 2024-11-07
Improve tooltip coloring by defining defaults based on tooltip variant.

andresgutgon:
# What?
This change still permits custom tooltip content with HTML inside. But for simple plain text tooltips color is handled inside it based on the tooltip variant

### Issue
https://github.com/latitude-dev/latitude-llm/issues/559


apps/web/src/app/(private)/datasets/_components/DatasetsTable/index.tsx
```diff
@@ -94,9 +94,7 @@ export function DatasetsTable({
                       />
                     }
                   >
-                    <Text.H6B color='white'>
-                      Show file preview (first 100 rows)
-                    </Text.H6B>
```
andresgutgon, line 99: Don't use `<Text>` components for simple content. Tooltip takes care of handling it based on variant and dark mode

geclos (approved):

PR #569, opened 2024-11-07
Improve LLM Chunk error message

andresgutgon:
# What?
Until know we were telling "Unknown AI error". Now we put the actual error returned from the llm provider

### Before
image

### After
image


packages/core/src/services/chains/ChainStreamConsumer/consumeStream.ts
```diff
@@ -34,8 +41,21 @@ function createAIError(message: string): ChainError<RunErrorCodes.AIRunError> {
...
-function getErrorMessage(error: unknown): string {
-  return error instanceof Error ? error.message : 'Unknown AI chunk error'
+function getErrorMessage({
+  error,
+  providerName,
+}: {
+  error: unknown
+  providerName: Providers
+}): string {
+  const intro = `${capitalize(providerName)} had this error: `
```
geclos, line 51: returned* this error

geclos (dismissed):

PR #570, opened 2024-11-08
Link a dataset to a document version

andresgutgon:
# What?
These changes are part of [this task](https://github.com/orgs/latitude-dev/projects/2/views/4?pane=issue&itemId=86378792)

We want to allow users to pick parameters for a document from the rows of a dataset they have in their workspace. We link that dataset_id with the document version.
The `dataset_id` is an attribute that count as content change to be consired when a new version of the document is merged. We do this to allow users iterate the prompt through time using different datasets. So they can back to the pass and see the old prompt working with the old dataset.

image

### TODO
- [x] Add `dataset_id` nullable relation column to document version
- [x] Detect dataset id change as part of the changes when publishing a new version
- [x] Update dataset ID when merging the new version.
- [x] Fix tests and add a couple more

### DIFF
--- a/packages/core/src/schema/models/documentVersions.ts
+++ b/packages/core/src/schema/models/documentVersions.ts
@@ -12,6 +12,7 @@ import {
 import { latitudeSchema } from '../db-schema'
 import { timestamps } from '../schemaHelpers'
 import { commits } from './commits'
+import { datasets } from './datasets'
 
 export const documentVersions = latitudeSchema.table(
   'document_versions',
@@ -25,6 +26,10 @@ export const documentVersions = latitudeSchema.table(
     commitId: bigint('commit_id', { mode: 'number' })
       .notNull()
       .references(() => commits.id, { onDelete: 'cascade' }),
+    datasetId: bigint('dataset_id', { mode: 'number' }).references(
+      () => datasets.id,
+      { onDelete: 'set null' },
+    ),
     deletedAt: timestamp('deleted_at'),
     ...timestamps(),
   },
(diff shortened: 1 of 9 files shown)


packages/core/src/schema/models/documentVersions.ts
```diff
@@ -25,6 +26,10 @@ export const documentVersions = latitudeSchema.table(
     commitId: bigint('commit_id', { mode: 'number' })
       .notNull()
       .references(() => commits.id, { onDelete: 'cascade' }),
+    datasetId: bigint('dataset_id', { mode: 'number' }).references(
+      () => datasets.id,
+      { onDelete: 'set null' },
```
andresgutgon, line 31: If the dataset is deleted we set the pointer to null

packages/core/src/services/documents/update.test.ts
```diff
@@ -370,4 +370,48 @@ describe('updateDocument', () => {
...
+    }).then((r) => r.unwrap())
+
+    await recomputeChanges({ draft: commit, workspaceId: workspace.id })
+
+    const commitDocs = await docsScope
+      .getDocumentsAtCommit(commit)
+      .then((r) => r.unwrap())
+
+    const newDoc = commitDocs.find((d) => d.path === 'doc1')
+    expect(newDoc?.datasetId).toBe(dataset.id)
```
csansoon, line 415: I don't think assigining a dataset to a document should trigger an update. I think it should even be doable in merged commits. This property does not "define" the document, just how it's used in the ui

andresgutgon: But what if I create a new draft that changes the document a lot? I could change the dataset used in that playground. 

Now production version of that document will be pointing to that dataset too

geclos (dismissed):

packages/core/src/services/documents/update.test.ts
csansoon: yesyes I don't say the same dataset should apply to all versions. But I don't think changing a dataset should be treated as "a change to the document". If it does, I would not be able to change the dataset in a merged commit without having to create a new draft just for that.

andresgutgon: Talked offline. Carlos was right. Changed

csansoon (approved):

Issue #540, opened 2024-11-04
Implement default provider configurations


geclos:
Users want to set up a default provider and model so they don't have to repeat the same changes everytime they create a new prompt.

https://www.figma.com/design/ODioXiqX8aeDMonsh0HBui/Latitude-Cloud?node-id=3051-20078&node-type=frame&t=6y2Vpbhct0CCCLHx-0

https://www.figma.com/design/ODioXiqX8aeDMonsh0HBui/Latitude-Cloud?node-id=857-38101&node-type=frame&t=q5FFEcDozM9C1iH7-0

PR #577, opened 2024-11-08
Feature/540 default provider and model

neoxelox:
Closes https://github.com/latitude-dev/latitude-llm/issues/540

- Allows to create a provider with an optional default model
- Allows to set and unset the default provider
- Unsets the default provider if that provider is deleted
- Uses the default provider and model when creating a new document or simple/advanced evaluation
- Fixes problem with `Select` element not updating its value when using it as a managed input; and options when updated dynamically
- Fixes problem with current workspace not being updated because it was cached within the Session context


andresgutgon:
I know merge squash. But still "esta feo" don't left wips. If you change put all under one commit with a clear message
image

apps/web/src/actions/workspaces/update.ts
```diff
@@ -21,9 +22,9 @@ export const updateWorkspaceAction = authProcedure
       .find(input.workspaceId)
       .then((r) => r.unwrap())
 
-    const updatedWorkspace = await updateWorkspace({
-      workspace,
+    const updatedWorkspace = await updateWorkspace(workspace, {
       name: input.name,
+      defaultProviderId: input.defaultProviderId,
```
andresgutgon, line 27: These kind of changes would be better use a different action. Not generic workspace update. This way we can track an event and see how many users set default provider

andresgutgon: Not necessary to change unless we think is interesting to track this as event. Ask @cesr

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
```diff
@@ -51,44 +55,81 @@ export default function NewApiKey({
...
-              onChange={(newValue) => setProviderType(newValue)}
-              defaultValue={data?.provider}
-              options={PROVIDER_OPTIONS}
-            />
-          </FormField>
-          <FormField label='Name'>
+          <FormField
+            label='Provider'
+            description="This is the name you'll use in the prompt editor to refer to use this provider and model."
+          >
```
andresgutgon, line 61: No need for `<FormField>`. Remove it and pass `label` and `description` to input.

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
```diff
@@ -51,44 +55,81 @@ export default function NewApiKey({
...
             <Input
               required
               type='text'
               name='name'
-              defaultValue={data?.name}
-              placeholder='My API Key'
+              defaultValue={result?.name}
+              placeholder='My Provider'
+              // @ts-expect-error
+              errors={errors?.name}
```
andresgutgon, line 69: Why this? Let's find out

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
```diff
@@ -51,44 +55,81 @@ export default function NewApiKey({
...
+              required
+              name='provider'
+              onChange={(value) => setProvider(value as Providers)}
+              defaultValue={result?.provider}
+              options={PROVIDER_OPTIONS}
+              // @ts-expect-error
+              errors={errors?.provider}
+            />
+          </FormField>
+          <FormField label='API Key'>
```
andresgutgon, line 83: Same. You can remove all `FormField` components here

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
```diff
@@ -51,44 +55,81 @@ export default function NewApiKey({
...
+              description='URL to your OpenAI compatible API.'
+            >
               <Input
                 required
                 type='text'
                 name='url'
-                defaultValue={data?.url}
+                defaultValue={result?.url}
                 placeholder='http://localhost:11434/v1'
+                // @ts-expect-error
```
andresgutgon, line 105: This is not right. Let's understand why you put all these ts-expect-error(s)

apps/web/src/app/(private)/settings/apikeys/new/page.tsx
```diff
@@ -0,0 +1,16 @@
+'use client'
```
andresgutgon, line 1: Don't use client on pages. Pages are always SSR

apps/web/src/app/api/workspaces/current/route.ts
```diff
@@ -0,0 +1,19 @@
+import { WorkspaceDto } from '@latitude-data/core/browser'
+import { authHandler } from '$/middlewares/authHandler'
+import { errorHandler } from '$/middlewares/errorHandler'
+import { NextRequest, NextResponse } from 'next/server'
+
+export const GET = errorHandler(
```
andresgutgon, line 6: Why you need this? We already have `useSession` which gives the current workspace

packages/web-ui/src/providers/SessionProvider/index.tsx
```diff
@@ -7,10 +7,9 @@ export type SessionUser = {
   name: string | null | undefined
   email: string
 }
-export type SessionWorkspace = { id: number; name: string }
+
 interface ISessionContext {
   currentUser: SessionUser
-  workspace: SessionWorkspace
```
andresgutgon, line 13: Ok, this is related to my other comment. I would like to understand this change

apps/web/src/actions/workspaces/update.ts
neoxelox: Okay, will ask, although we could do something like `if (defaultProviderId) trackEvent()` in the action itself.

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
neoxelox: Okay, will change, although it was already like this, I only added another field.

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
neoxelox: I am assuming this `errors` prop allows to show an specific error underneath the field if there is any after submitting the form. I saw this pattern in others forms such as `apps/web/src/app/(public)/setup/SetupForm/index.ts`... so I though it was missing and just added this for consistency.

I see other forms that do `const errors = error?.fieldErrors as Record<string, string[]>` to avoid putting a ` // @ts-expect-error` repeatedly. Will change for that or I can just remove the errors field entirely if not necessary.

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
neoxelox: Ok

apps/web/src/app/(private)/settings/apikeys/new/page.tsx
neoxelox: Okay, I just based the page on the destroy provider one, which also has `use client`: `apps/web/src/app/(private)/settings/apikeys/[providerApiKeyId]/destroy/page.tsx`, maybe it is also wrong?

I see other also doing the same, ex: `apps/web/src/app/(private)/settings/users/[userId]/destroy/page.tsx`...

apps/web/src/app/api/workspaces/current/route.ts
neoxelox: I was told to do this change, because the workspace from the session was not updating after executing the update hook (store?), as it was always cached within the session. Notice that the already existing feature of updating the workspace name, had the same bug. It, however, showed the updated value in the UI because it was an input element.

apps/web/src/actions/workspaces/update.ts
andresgutgon: Yes, but it is better to have actions that do one thing: `create workspace`, `update workspace`, `add default provider`.

Yes, adding a default provider is a workspace update. But is more concrete. So let's make concrete from the action.

packages/web-ui/src/providers/SessionProvider/index.tsx
neoxelox: This is related to the other change I explained above. I deleted the workspace from the useSession, and replace it with the useCurrentWorkspace hook (store?), so that the workspace value actually changes when updated.

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
andresgutgon: I know. I just noticed this bad pattern in another place
https://github.com/latitude-dev/latitude-llm/pull/578

So the idea here is to be a good citizen and adapt the code now that you're touching it

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
neoxelox: See explanation above.

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
andresgutgon: The question is why `error.fieldErrors` is not typed like this. What type is expecting the `errors` prop? Don't worry we can check on Monday

apps/web/src/app/api/workspaces/current/route.ts
andresgutgon: So, the thing is to have the useSession hook synchronized with the workspace store. Even initialize the workspace store on the layout backend entry point. But again, I'm not sure so let's talk briefly on Monday : )

apps/web/src/app/(private)/settings/apikeys/new/page.tsx
andresgutgon: Is not super important. You can ignore this comment. But using `use client` on page makes me feel a bit dirty tbh 😂. Yes, pure OCD

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/index.tsx
```diff
@@ -73,8 +68,6 @@ export default function NewProviderApiKey({
             label='Provider'
             description="This is the name you'll use in the prompt editor to refer to use this provider and model."
             placeholder='My Provider'
-            // @ts-expect-error
-            errors={errors?.name}
```
andresgutgon: This is also an option instead of changing the action 😂

apps/web/src/actions/evaluations/create.ts
```diff
@@ -85,18 +89,32 @@ async function enrichWithProvider({
...
+    return {
+      ...rest,
+      prompt: `---
+provider: ${provider.name}
+model: ${model}
+---
+`.trim(),
+    } as EvaluationMetadataLlmAsJudgeAdvanced
+  }
+
```
geclos, line 117: unrelated changes in this pr, try to rebase instead of merge

packages/web-ui/src/ds/atoms/Select/index.tsx
```diff
@@ -60,6 +60,9 @@ export function Select({
   required = false,
 }: SelectProps) {
   const [selectedValue, setSelected] = useState(value ?? defaultValue)
+  useEffect(() => {
+    setSelected(value ?? defaultValue)
+  }, [value, defaultValue])
```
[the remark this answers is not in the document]
neoxelox: * self-applauds *

packages/core/src/services/workspaces/update.ts
```diff
@@ -1,23 +1,22 @@
...
-  workspace: Workspace
-  name: string
-}) {
-  return await Transaction.call(async (tx) => {
+export async function updateWorkspace(
+  workspace: Workspace,
+  values: Partial<Workspace>,
+  db = database,
+) {
+  return Transaction.call<Workspace>(async (tx) => {
```
csansoon, line 13: Make sure what `values` we're allowing to pass from the user. We shouldn't allow to change values other than name and the new default fields. Although it's okay if that restriction is somewhere else I guess

geclos (changes requested):
do a separate PR with only the contents from the drizzle folder and merge that first please

packages/core/src/services/workspaces/update.ts
neoxelox: Okay, will do, I based the changed on the `user update` service `packages/core/src/services/users/update.ts`, maybe in another PR we should restrict the fields too.

apps/web/src/actions/workspaces/setDefaultProvider.ts
```diff
@@ -0,0 +1,29 @@
...
+    const workspacesScope = new WorkspacesRepository(userId)
+    const workspace = await workspacesScope
+      .find(input.workspaceId)
+      .then((r) => r.unwrap())
+
+    const updatedWorkspace = await updateWorkspace(workspace, {
+      defaultProviderId: input.defaultProviderId,
+    }).then((r) => r.unwrap())
+
+    return updatedWorkspace!
```
geclos, line 28: you should not need this assertion, add it to the update workspace service instead

apps/web/src/app/(private)/settings/_components/ProviderApiKeys/index.tsx
```diff
@@ -15,38 +18,51 @@ import {
...
   return (
     <div className='flex flex-col gap-4'>
-      <NewApiKey open={open} setOpen={setOpen} />
       <div className='flex flex-row items-center justify-between'>
         <Text.H4B>Providers</Text.H4B>
-        <Button fancy variant='outline' onClick={() => setOpen(true)}>
+        <Button
+          fancy
+          variant='outline'
+          onClick={() => router.push(ROUTES.settings.providerApiKeys.new.root)}
```
geclos, line 44: nice 👌🏼

apps/web/src/app/api/workspaces/current/route.ts
```diff
@@ -0,0 +1,19 @@
...
+      {
+        workspace,
+      }: {
+        workspace: WorkspaceDto
+      },
+    ) => {
+      return NextResponse.json(workspace, { status: 200 })
+    },
+  ),
+)
```
geclos, line 19: lovely

apps/web/src/components/EditorHeader/index.tsx
```diff
@@ -187,19 +163,22 @@ export default function EditorHeader({
     }),
   )
 
+  const provider = selectedProvider
+    ? providersByName[selectedProvider]
+    : undefined
```
geclos, line 168: I feel we are doing this ternary in a dozen places, can we DRY a bit?

apps/web/src/hooks/useFormAction.ts
```diff
@@ -58,10 +62,14 @@ export function useFormAction<
 
   const action = useCallback(
     async (formData: FormData) => {
-      const json = formDataToJson(
+      let json = formDataToJson(
         formData,
       ) as inferServerActionInput<TServerAction>
 
+      if (onSubmit) {
+        json = onSubmit(json)
```
geclos, line 70: hmm this smells like @andresgutgon recommendation.. Are we sure about this? a callback that needs to return the payload it receives and inadvertendly changes the payload schema that gets sent to the api... leaky abstraction

If you need onSubmit from form you could just have used the server action directly, example:

```typescript
const { update } = useProviderApiKeys()
const onSubmit = (e) => {
    const formData = new FormData(e.currentTarget)
    const json = formDataToJson(formData)

    update(json)
}

return <form onSubmit={onSubmit) ... />
```

same number of lines, no leaky abstraction involved

apps/web/src/services/routes.ts
```diff
@@ -51,6 +51,9 @@ export const ROUTES = {
   settings: {
     root: '/settings',
     providerApiKeys: {
+      new: {
+        root: '/settings/apikeys/new',
+      },
```
geclos, line 56: niiiice

apps/web/src/stores/currentWorkspace.ts
```diff
@@ -42,8 +44,39 @@ export default function useCurrentWorkspace(opts?: SWRConfiguration) {
...
+      }
+
+      toast({
+        title: 'Default provider updated',
+      })
+
+      mutate(workspace)
+      return workspace
+    },
+    [mutate, data],
```
geclos, line 78: for next times you can use `useLatitudeAction` which encapsulates the onsuccess and onerror use cases, besides providing some nice defaults in terms of error reporting

apps/web/src/stores/datasetPreviews.ts
```diff
@@ -19,7 +17,7 @@ export default function useDatasetPreview(
   )
 
   const { data = [], ...rest } = useSWRImmutable(
-    ['workspace', workspace.id, 'datasets_preview', dataset?.id],
+    ['datasets_preview', dataset?.id],
```
geclos, line 20: lovely

docs/guides/getting-started/providers.mdx
```diff
@@ -14,22 +14,35 @@ The name you introduce for each provider will be the one you will have to use in
...
 
 ## Using Providers in Prompts
 
 Once you've set up a provider, you can use it in your prompts:
 
 1. Open the Prompt editor
-1. Create a new prompt or edit an existing one
-1. In the prompt configuration, select your desired provider from the dropdown menu or in the code
+2. Create a new prompt or edit an existing one
+3. In the prompt configuration, select your desired provider from the dropdown menu or in the code
```
geclos, line 45: fantastic

packages/core/src/services/ai/providers/models/index.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  it('returns first supported model for latitude provider', async () => {
+    provider.provider = Providers.OpenAI
+    provider.name = 'latitude'
+    const latitudeProvider = 'latitude'
+
+    const result = findFirstModelForProvider({ provider, latitudeProvider })
+
+    expect(result).toBe(Object.values(PROVIDER_MODELS[Providers.OpenAI]!)[0])
+  })
+})
```
geclos, line 80: nice

packages/web-ui/src/ds/atoms/Select/index.tsx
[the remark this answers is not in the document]
geclos: hmm why was this required the onchange callback should get triggered on mount if you provide a value/defaultValue, effectively setting the selected value already. I'm wrong?

geclos (commented):
some comments but overall good work!

packages/core/src/services/providerApiKeys/create.ts
```diff
@@ -17,17 +17,22 @@ export type Props = {
...
-  { workspace, provider, token, url, name, author }: Props,
+  { workspace, provider, token, url, name, defaultModel, author }: Props,
   db = database,
 ) {
   return Transaction.call(async (tx) => {
     if (provider === Providers.Custom && !url) {
       return Result.error(new BadRequestError('Custom provider requires a URL'))
     }
 
+    if (defaultModel === '') {
```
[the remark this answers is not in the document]
neoxelox: The default behaviour of the uncontrolled model input, of the create provider modal, was sending the model as an empty string when the input is empty. However, I also controlled this in the modal:

```typescript
onSubmit: (payload) => ({
      ...payload,
      defaultModel: payload.defaultModel || undefined, // Ensure defaultModel is either non-empty or undefined
    }),
````

I think it would be nice to let this check there to ensure no default model is saved as an empty string, because then we would have to check in all places not to use it as the default model if it is an empty string and not "null".

File: packages/web-ui/src/ds/atoms/Select/index.tsx (hunk 11 above)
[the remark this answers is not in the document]
Comment (csansoon) [reply packages/web-ui/src/ds/atoms/Select/index.tsx:65]: The `setSelected` was only triggered by the own component when the user selects an option, but there was no way to do this programmatically. Now, changing the `value` argument will select it too.

File: apps/web/src/actions/providerApiKeys/create.test.ts [hunk 24]
```diff
@@ -113,7 +125,7 @@ describe('createProviderApiKeyAction', () => {
       expect(error).toBeDefined()
     })
 
-    it('allows creating to providers with same name if one is deleted', async () => {
+    it('allows creating two providers with same name if one is deleted', async () => {
```
andresgutgon, line 128: I was thinking. This is one of my classic English mistakes. But it wasn't me ;0

apps/web/src/components/EditorHeader/index.tsx
```diff
@@ -156,106 +103,101 @@ export default function EditorHeader({
...
-    provider: selectedProvider
-      ? providersByName[selectedProvider]?.provider
-      : undefined,
-    isDefaultProvider,
+  const modelOptions = useModelOptions({
+    provider: provider ? providersByName[provider]?.provider : undefined,
+    name: provider ? providersByName[provider]?.name : undefined,
   })
 
+  // onPromptMetadataChange
```
andresgutgon, line 148: 🔥

apps/web/src/components/EditorHeader/index.tsx
```diff
@@ -156,106 +103,101 @@ export default function EditorHeader({
...
+      if (!selectedProvider) return
+      if (selectedProvider === provider) return
+
+      let firstModel
+      if (providersByName[selectedProvider]) {
+        firstModel = findFirstModelForProvider({
+          provider: providersByName[selectedProvider],
+          latitudeProvider: envClient.NEXT_PUBLIC_DEFAULT_PROJECT_ID,
+        })
+      }
```
andresgutgon, line 172: I have seen this same if check + firtModel find 3 times. Maybe you could pass the if inside the function.

apps/web/src/components/Providers/WebsocketsProvider/index.tsx
```diff
@@ -25,11 +25,12 @@ function useJoinWorkspace({ connection }: { connection: IWebsocketConfig }) {
   const { currentUser } = useSession()
   const { data: workspace } = useCurrentWorkspace()
   return useCallback(() => {
+    if (!workspace?.id) return
```
andresgutgon, line 28: How is possible not to have workspace in private routes?

apps/web/src/helpers/forms.ts
```diff
@@ -0,0 +1,9 @@
+import { inferServerActionInput, TAnyZodSafeFunctionHandler } from 'zsa'
+
+export function formDataToAction<
+  T extends (data: any) => Promise<[any, null] | [null, any]>,
+>(data: FormData): inferServerActionInput<TAnyZodSafeFunctionHandler & T> {
+  const obj: { [key: string]: unknown } = {}
+  data.forEach((value, key) => (obj[key] = value))
+  return obj as inferServerActionInput<TAnyZodSafeFunctionHandler & T>
+}
```
andresgutgon, line 9: We have like 3 abstractions for actions. I would use `useLatitudeAction` in your code instead of doing one more different way of working with actions.

apps/web/src/stores/currentWorkspace.ts
```diff
@@ -42,8 +44,39 @@ export default function useCurrentWorkspace(opts?: SWRConfiguration) {
       mutate(workspace)
       return workspace
     },
-    [mutate],
+    [mutate, data],
+  )
+
+  const { execute: setDefaultProvider } = useServerAction(
```
andresgutgon, line 50: Here if you use `useLatitudeAction` you don't need to deal with form data transformation. Also errors are typed correctly I think.

packages/core/src/services/providerApiKeys/create.test.ts
```diff
@@ -0,0 +1,287 @@
...
+    expect(result.error!.message).toEqual('This token is already in use')
+
+    const providersScope = new ProviderApiKeysRepository(workspace.id)
+    const providers = await providersScope.findAll().then((r) => r.unwrap())
+
+    expect(providers.map((p) => p.id)).toEqual([oldProvider.id])
+
+    expect(mocks.publisher).not.toHaveBeenCalled()
+  })
+})
```
andresgutgon, line 287: Awesome.

andresgutgon (dismissed):
Nice work. Some minor comments

apps/web/src/components/EditorHeader/index.tsx
neoxelox: There are 4 different components (4 editors) with similar functionality.... do you mean you want `findFirstModelForProvider` to accept an undefined provider and return no model if so? For me that it's a little strange but will do.

andresgutgon: I mean, is what you're doing outside no? you init `let model` as undefined and then assign it if the it exists

apps/web/src/helpers/forms.ts
neoxelox: But I need to get the data from the form before executing the actual action somehow...! Otherwise, the default behaviour of the uncontrolled model input, of the create provider modal, will send the model as an empty string when the input is empty.

neoxelox: I was using `useFormAction`, and modified it to be able to change the captured payload, but I was told not to modify that hook and make the code in the component. However, with the`formDataToJson` function of the `useFormAction` I was not getting any types, so I decided to move that function outside of the `useFormAction` file and enhance it to allow for type safety/inference....

apps/web/src/components/EditorHeader/index.tsx
neoxelox: Yes, but I am not calling the function with no provider because.... why call that function if you don't have one? For me it's strange that function may accept no provider.... but don't worry I will change it

apps/web/src/components/Providers/WebsocketsProvider/index.tsx
neoxelox: I didn't think of so, should I just assert workspace cannot be undefined?

apps/web/src/helpers/forms.ts
neoxelox: Just to clarify this function is just to transform the form data to a JS object. If I am not mistaken, even when using useLatitudeAction, I will still have to use it.... I use this function before calling the action:

```typescript
const { create } = useProviderApiKeys()
  const onSubmit = useCallback(
    async (event: FormEvent<HTMLFormElement>) => {
      event.preventDefault()

      const payload = formDataToAction<typeof create>(
        new FormData(event.currentTarget),
      )

      const [_, error] = await create({
        ...payload,
        defaultModel: payload.defaultModel || undefined, // Ensure defaultModel is either non-empty or undefined
      })
      if (!error) onOpenChange(false)
    },
    [create],
  )
  ```

apps/web/src/components/Providers/WebsocketsProvider/index.tsx
geclos: getting the workspace is an async operation so it's normal that it can be null while loading

apps/web/src/helpers/forms.ts
geclos: this is good 👌🏼

apps/web/src/stores/currentWorkspace.ts
geclos: not true, useLatitudeAction expects whatever the underlying server actions expects, and we always use json so it's json.

geclos (approved):

PR #580, opened 2024-11-11
feat(sdk/typescript): handle 5xx errors gracefully from sdk

geclos:


packages/sdks/typescript/src/utils/syncRun.ts
```diff
@@ -54,13 +55,19 @@ export async function syncRun(
   })
 
   if (!response.ok) {
```
andresgutgon, line 57: Why not check status code?

andresgutgon (approved):

packages/sdks/typescript/src/utils/syncRun.ts
geclos: it's another option 🤷🏼

PR #600, opened 2024-11-13
Fix webapp build avoiding lodash omit

neoxelox:
Fix webapp build avoiding lodash omit

### DIFF
--- a/packages/core/src/services/ai/providers/models/index.ts
+++ b/packages/core/src/services/ai/providers/models/index.ts
@@ -1,5 +1,3 @@
-import { omit } from 'lodash-es'
-
 import { ProviderApiKey } from '../../../../schema/types'
 
 export enum Providers {
@@ -111,7 +109,11 @@ export function listModelsForProvider({
   if (!models) return {}
 
   if (name && name === latitudeProvider) {
-    return omit(models, DEFAULT_PROVIDER_UNSUPPORTED_MODELS)
+    return Object.fromEntries(
+      Object.entries(models).filter(
+        ([key]) => !DEFAULT_PROVIDER_UNSUPPORTED_MODELS.includes(key),
+      ),
+    )
   }
 
   return models


packages/core/src/services/ai/providers/models/index.ts
```diff
@@ -1,5 +1,3 @@
-import { omit } from 'lodash-es'
```
andresgutgon, line 1: Why this? We're already using lodash in other places

neoxelox: I don't know why but it fails to build the webapp (that same omit and code was in a component). I think this is the simplest way to avoid locking the CD pipeline for now.

```text
../../packages/core/src/services/ai/providers/models/index.ts
Dynamic Code Evaluation (e. g. 'eval', 'new Function', 'WebAssembly.compile') not allowed in Edge Runtime 
Learn More: https://nextjs.org/docs/messages/edge-dynamic-code-evaluation

The error was caused by importing 'lodash-es/lodash.js' in '../../packages/core/src/services/ai/providers/models/index.ts'.
```

geclos (approved):

packages/core/src/services/ai/providers/models/index.ts
```diff
@@ -1,5 +1,3 @@
...
-
 import { ProviderApiKey } from '../../../../schema/types'
 
 export enum Providers {
```
geclos, line 3: the solution is you don't wanna remove this omit is to move any enum declared in this file to the browser export from core

neoxelox: mmmm, I don't understand how is that related. The `Providers` enum has always been in there and the build was indeed failing because of the lodash import.

My guess is that all the times lodash has been used in the webapp was in client side. However, I guess after moving some code (that included this lodash import) from a webapp component to the core, and exporting the function that uses lodash underneath through `browser.ts`, the server side part of Next was somewhere importing something from core through `browser.ts` and thus it had to resolve the lodash import indirectly.

PR #609, opened 2024-11-13
feat(evaluations): HITL evaluations MVP migrations

geclos:
Implements the necessary migrations to support the HITL evaluations.


packages/core/drizzle/0100_lonely_revanche.sql
```diff
@@ -0,0 +1 @@
+ALTER TYPE "latitude"."metadata_type" ADD VALUE 'default';
```
csansoon, line 1: `default`? What about `hitl` or `human_in_the_loop` to be more clear?

geclos: yeah we've talked about it and hitl is not good enough but will probably change it to "manual"

Issue #499, opened 2024-10-23
Automatic API documentation with Zod Open API


geclos:
https://hono.dev/examples/zod-openapi we already use zod so it's a no brainer

NewtTheWolf:
i take it

andresgutgon:
Un assigning. At the moment, no one is working on this one

PR #614, opened 2024-11-14
Gateway OpenAPI schema live documentation

neoxelox:
# What?
Implement OpenAPI + Swagger UI for Latitude gateway API endpoints

image

### Issue
https://github.com/latitude-dev/latitude-llm/issues/499

### TODO
- [x] Do prove of concept with `/run` api v2 ✅ 
- [x] Move evaluate tests to new folder 👈 I'm here
- [x] Consider if some schemas in core should go in gateway
- [x] Migrate all api routes to use OpenAPI interface
- [x] 🔥 `@hono/zod-validator`


packages/core/src/constants.ts
```diff
@@ -420,3 +422,102 @@ export type DiffValue = {
...
+      }),
+    ]),
+  }),
+])
+
+export const runSyncAPIResponseSchema = z.object({
+  uuid: z.string(),
+  conversation: z.array(messageSchema),
+  response: chainCallResponseDtoSchema,
+})
```
andresgutgon, line 523: Move all this to gateway openapi schema

apps/gateway/src/routes/v1/chat/chat.handler.ts
```diff
@@ -0,0 +1,45 @@
...
+import { addMessages } from '@latitude-data/core/services/documentLogs/index'
+import { captureException } from '@sentry/node'
+import { streamSSE } from 'hono/streaming'
+
+import { AppRouteHandler } from '$/openApi/types'
+import { ChatRoute } from '$/routes/v1/chat/chat.route'
+import { chainEventPresenter } from '$/common/documents/getData'
+
+// @ts-expect-error: streamSSE has type issues
+export const chatHandler: AppRouteHandler<ChatRoute> = async (c) => {
```
andresgutgon, line 12: For some reason types are not working at this level. cc @neoxelox

apps/gateway/src/routes/v2/documents/get/get.route.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const getRouteV1 = getRouteFactory({
+  path: ROUTES.v1.documents.get,
+  tags: ['V1_DEPRECATED'],
+})
+export const getRouteV2 = getRouteFactory({
+  path: ROUTES.v2.documents.get,
+  tags: ['Documents'],
+})
```
andresgutgon, line 40: With this abstraction in place reusing routes is as easy as changing the path if the implementation didn't change

apps/gateway/src/openApi/configureOpenAPI.ts
```diff
@@ -0,0 +1,53 @@
+import type { OpenAPIHono } from '@hono/zod-openapi'
+
+import { swaggerUI } from '@hono/swagger-ui'
+
+import packageJson from '../../package.json'
```
neoxelox, line 5: lol

apps/gateway/src/openApi/createApp.ts
```diff
@@ -0,0 +1,9 @@
+import { OpenAPIHono } from '@hono/zod-openapi'
+
+export function createRouter() {
+  return new OpenAPIHono({ strict: false })
```
neoxelox, line 4: what does `strict` mean in there?

apps/gateway/src/openApi/schemas/ai.ts
```diff
@@ -0,0 +1,98 @@
...
+})
+
+export const toolCallSchema = z.object({
+  id: z.string(),
+  name: z.string(),
+  arguments: z.record(z.any()),
+})
+
+export const configSchema = z.object({}).passthrough()
+export const providerLogSchema = z.object({}).passthrough()
```
neoxelox, line 18: no schema for provider log?

andresgutgon: It has not really been used in the API afaik

apps/gateway/src/openApi/createApp.ts
andresgutgon: Ni idea

andresgutgon: This https://hono.dev/docs/api/hono#strict-mode

apps/gateway/src/routes/v1/chat/chat.handler.ts
neoxelox: mmm that's probably why I did what I did, handlers + route def altogether... this types are magical... but if `c.req.valid` does complain when getting unknown parameters from the request schema... I guess I am okay with that

apps/gateway/src/routes/v2/conversations/createEvaluationResult/createEvaluationResult.handler.ts
```diff
@@ -122,3 +74,45 @@ async function validatedEvalConnectedToDocument({
     throw new BadRequestError('The evaluation is not connected to this prompt')
   }
 }
+
+// @ts-expect-error: streamSSE has type issues
```
neoxelox, line 78: is the type issue really because of streamSSE? I think it is because we are separating the handler from the magical createRoute, I would put that in the comment instead of streamSSE? (Also this handler does nothing about SSEs)

apps/gateway/src/routes/v2/documents/get/get.handler.test.ts
```diff
@@ -18,7 +18,7 @@ describe('GET documents', () => {
   describe('unauthorized', () => {
     it('fails', async () => {
       const res = await app.request(
-        '/api/v1/projects/1/versions/asldkfjhsadl/documents/path/to/document',
+        '/api/v2/projects/1/versions/asldkfjhsadl/documents/path/to/document',
```
neoxelox, line 21: why the change from v1 to v2?

andresgutgon: Before the implementation was in v1. So the test. Now I moved the implementation to v2, so the test

neoxelox (commented):
Buena currada refactorizando todo! The only thing I would say it to check out the comments saying "Types not working because of SSEs", I don't think its that. I think it is because we are separating the handler from the magical createRoute, I would put that in the comment instead?

apps/gateway/src/routes/app.ts
```diff
@@ -21,24 +22,16 @@ app.get('/health', (c) => {
...
-app.route(
-  '/api/v2/projects/:projectId/versions/:versionUuid/documents',
-  documentsRouterV2,
-)
-app.route('/api/v2/conversations', conversationsRouterV2)
-app.route('/api/v2/otlp', otlpTracesRouter)
+app.route('/', v1Routes)
+app.route('/', documents)
+app.route('/', conversations)
+app.route('/', telemetry)
```
geclos, line 34: hmm this is very painful for the reader no? now there's no easy way to know what routes we are exposing you'd have to check every router

apps/gateway/src/routes/v2/conversations/chat/chat.handler.test.ts
```diff
@@ -118,7 +118,12 @@ describe('POST /chat', () => {
         messages: [
           {
             role: MessageRole.user,
-            content: 'fake-user-content',
```
geclos, line 121: hmm this should be valid content

apps/gateway/src/routes/v2/conversations/createEvaluationResult/createEvaluationResult.handler.ts
geclos: yeah there's no sse herem, why are we escaping the type checking?

apps/gateway/src/routes/v2/documents/logs/create.handler.ts
```diff
@@ -0,0 +1,40 @@
+import { LogSources } from '@latitude-data/core/browser'
+import { generateUUIDIdentifier } from '@latitude-data/core/lib/generateUUID'
+import { createDocumentLog } from '@latitude-data/core/services/documentLogs/create'
+import { AppRouteHandler } from '$/openApi/types'
+import { CreateLogRoute } from '$/routes/v2/documents/logs/create.route'
+import { getData } from '$/common/documents/getData'
+
+// @ts-expect-error: broken types
```
geclos, line 8: i mean could we fix types 😅 it looks like this PR is breaking all types

apps/gateway/src/routes/app.ts
geclos: What I would have like is something similar to rails where routes.rb declares all the routes and routers are just the controllers with the handlers for each endpoint in that route

apps/gateway/src/routes/v2/conversations/chat/chat.handler.test.ts
andresgutgon: Not according to this schema
https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/constants.ts#L226-L243


This started failing now because before we did this "truquito"

```javascript
import { messageSchema } from '@latitude-data/core/browser'
const messagesSchema = z.array(z.any(messageSchema))
```
☝️ This is using any

What I did was remove the `any`
```javascript
const messagesSchema = z.array(messageSchema)
```

And now types works!

image

apps/gateway/src/routes/v2/documents/logs/create.handler.ts
andresgutgon: Before we were not using openapi so no old types broken. Just this
```
AppRouteHandler<CreateLogRoute>
```
And is because it does a complex generic black magic. But types inside the handlers all work nicely

apps/gateway/src/routes/app.ts
andresgutgon: You can see all the routes in one place in `routes.ts`. Here are the paths

image

PR #623, opened 2024-11-15
Order project list by edited at

neoxelox:
Closes:  https://github.com/latitude-dev/latitude-llm/issues/607

This feature orders the project list shown in the main dashboard by a computed project edited at descendingly. It also removes the prompt count, because (as discussed) it is not important, and the value shown is not consistent with the actual number of prompts the user sees in its draft (and it is difficult to compute because we would have to traverse all the commits to know).

### DIFF
--- a/packages/core/src/repositories/projectsRepository.ts
+++ b/packages/core/src/repositories/projectsRepository.ts
@@ -1,13 +1,4 @@
-import {
-  and,
-  count,
-  eq,
-  getTableColumns,
-  isNotNull,
-  isNull,
-  max,
-  sql,
-} from 'drizzle-orm'
+import { desc, eq, getTableColumns, isNull, max, sql } from 'drizzle-orm'
 
 import { Project } from '../browser'
 import { NotFoundError, Result } from '../lib'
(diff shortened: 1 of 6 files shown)


packages/core/src/repositories/projectsRepository.ts
```diff
@@ -58,54 +49,34 @@ export class ProjectsRepository extends RepositoryLegacy<typeof tt, Project> {
...
-          ),
-        lastCreatedAtDocument: aggredatedData.lastCreatedAtDocument,
+        lastEditedAt: aggredatedData.lastEditedAt,
       })
       .from(this.scope)
       .leftJoin(aggredatedData, eq(aggredatedData.id, this.scope.id))
       .where(isNull(this.scope.deletedAt))
+      .orderBy(
+        desc(
+          sql`COALESCE(${aggredatedData.lastEditedAt}, ${this.scope.createdAt})`,
```
neoxelox, line 77: This coalesce is to not have new projects with no documents at the top or at the bottom, but to follow the logical timeline.

geclos (approved):

andresgutgon (approved):

PR #625, opened 2024-11-17
Exclude logs with errors on document paramters UI

andresgutgon:
# What?
☝️ That

## TODO
- [x] Exclude logs with errors in the list of logs
- [x] Exclude logs with errors in the pagination of logs
- [x] Exclude logs with errors in the position location of one log
- [x] Add tests 💅


packages/core/src/services/documentLogs/computeDocumentLogs.ts
```diff
@@ -0,0 +1,88 @@
...
+    page?: string
+    pageSize?: string
+  },
+  db = database,
+) {
+  const repo = new DocumentLogsRepository(workspaceId, db)
+  const offset = calculateOffset(page, pageSize)
+  const scope = repo.scope.as('document_logs')
+  const fields = scope._.selectedFields
+  return db
```
andresgutgon, line 38: Question:
In [other places](https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/services/documentLogs/computeDocumentLogsWithMetadata.ts#L55-L67) we do `repo.scope.where(...)` 

But I tried here and it didn't work. You know why?

geclos: you should be able to do it since scope is a dynamic query in the case of document logs repository

packages/core/src/services/documentLogs/computeDocumentLogs.ts
```diff
@@ -0,0 +1,88 @@
...
+    .select(fields)
+    .from(scope)
```
geclos, line 40: nein, forbidden, this will fetch all document logs for that workspace before doing any other operation, you have to use the dynamic query or craft your own.

packages/core/src/services/documentLogs/computeDocumentLogs.ts
```diff
@@ -0,0 +1,88 @@
...
+  },
+  db = database,
+) {
+  const repo = new DocumentLogsRepository(workspaceId, db)
+  const scope = repo.scope.as('document_logs')
+  const countList = await db
+    .select({
+      count: sql`count(*)`.mapWith(Number).as('total_count'),
+    })
+    .from(scope)
```
geclos, line 78: nein, forbidden

apps/web/src/services/routes/api.ts
```diff
@@ -37,6 +37,11 @@ export const _API_ROUTES = {
                 documentLogs: {
                   root: `${documentRoot}/documentLogs`,
                   pagination: `${documentRoot}/documentLogs/pagination`,
+                  detail: (documentLogUuid: string) => {
+                    return {
+                      position: `${documentRoot}/documentLogs/${documentLogUuid}/position`,
+                    }
+                  },
```
csansoon, line 44: cool

packages/core/src/services/documentLogs/fetchDocumentLogWithPosition.ts
```diff
@@ -22,19 +46,32 @@ export async function fetchDocumentLogWithPosition(
...
+
+  const repo = getRepo({ workspaceId: workspace.id, db, excludeErrors })
+  const scope = repo.scope.as('document_logs')
+  const result = await db
     .select({
-      count: sql`COUNT(*)`.as('count'),
+      count: sql`COUNT(*)`.mapWith(Number).as('count'),
     })
-    .from(documentLogs)
+    .from(scope)
```
geclos, line 57: forbidden

andresgutgon: Shit

packages/core/src/services/documentLogs/computeDocumentLogs.ts
csansoon: I think you can actually do that with this repository. It's even used in some of the repository's methods themselves! Although maybe you can no longer do that after adding an `.as(…)` alias.

packages/core/src/services/documentLogs/computeDocumentLogs.ts
andresgutgon: Yeah, this sucks a bit no?

packages/core/src/services/documentLogs/computeDocumentLogs.ts
geclos: yeah no if you add an alias you can't do it, but you shouldn't add an alias 😅

packages/core/src/services/documentLogs/computeDocumentLogs.ts
csansoon: wdym

geclos: >  this will fetch all document logs for that workspace before doing any other operation, you have to use the dynamic query or craft your own.

packages/core/src/services/documentLogs/computeDocumentLogs.ts
```diff
@@ -0,0 +1,93 @@
...
+        eq(runErrors.errorableType, ErrorableEntity.DocumentLog),
+      ),
+    )
+    .where(
+      and(
+        isNull(runErrors.id),
+        eq(workspaces.id, workspaceId),
+        getCommonQueryConditions({ scope: documentLogs, documentUuid, draft }),
+      ),
+    )
```
geclos, line 90: yeah this will work

geclos (approved):

PR #626, opened 2024-11-17
Redirect to log correct page in paginated logs list

andresgutgon:
# What?
We were doing a shitty version of this in the client with a useEffect and I realize that we have a backend. So I just moved to it. Also: Make the playground resizable and avoid requesting a log when one is in flight

## TODO
- [x] Redirect to log detail and correct `page` if `?logUuid=some-uuid` is found
- [x] Enable resize of document editor and playground

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx
@@ -1,17 +1,38 @@
+import { Commit, Workspace } from '@latitude-data/core/browser'
 import { QueryParams } from '@latitude-data/core/lib/pagination/buildPaginatedUrl'
 import { computeDocumentLogsWithMetadataQuery } from '@latitude-data/core/services/documentLogs/computeDocumentLogsWithMetadata'
-import { Button, TableWithHeader, Text } from '@latitude-data/web-ui'
-import {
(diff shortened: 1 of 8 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx
```diff
@@ -23,71 +44,48 @@ export default async function DocumentPage({
...
+  })
+
+  if (currentLogPage && currentLogPage !== page) {
+    const route = ROUTES.projects
+      .detail({ id: projectId })
+      .commits.detail({ uuid: commit.uuid })
+      .documents.detail({ uuid: documentUuid }).logs.root
+    return redirect(
+      `${route}?page=${currentLogPage}&logUuid=${documentLogUuid}`,
+    )
```
Comment (csansoon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx:64]: Actually, should we add these parameters in the ROUTES instead? Something like `ROUTES.projects.[…].logs.with({ page, logUuid })`

csansoon (approved):

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/page.tsx:64]: Not today. I have a huge pile of changes to be merged

PR #627, opened 2024-11-17
Different input localstorage key for playground param sources

andresgutgon:
# What?
We want independent input values depending on the tab the user uses to populate the prompt parameters. The old version without a different source should keep working as before because I duplicated the hook that will be removed when enabling the new UI for everyone.

image

## TODO
- [x] Manual inputs source in local storage
- [x] Dataset inputs source in local storage 
- [x] History logs inputs source in local storage
- [x] QA without a feature flag keeps working

> [!NOTE]  
> This PR looks bigger than it is. I'm duplicating old input functionality to keep it working while we test all the changes. After all, a good big part of the code in this PR will be removed


apps/web/src/hooks/useDocumentParameters/index.ts
```diff
@@ -1,139 +1,292 @@
...
+      setDataset,
+    },
+    history: {
+      logUuid: inputs['history'].logUuid,
+      inputs: inputs['history'].inputs,
+      setInput: setHistoryInput,
+      setInputs: setHistoryInputs,
+      setHistoryLog,
+    },
+  }
```
andresgutgon, line 291: This hook is huge and it could have some tests. But I don't want to invest more time here. Also I feel being fully typed is easy to follow. Also all the methods are small

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/index.tsx
```diff
@@ -5,18 +5,44 @@ import { useCallback, useState } from 'react'
...
 import { useFeatureFlag } from '$/hooks/useFeatureFlag'
 
 import { Header } from '../Header'
 import Chat from './Chat'
 import { DocumentParams } from './DocumentParams'
-import { InputParams } from './DocumentParams/Input'
+import { OldInputs } from './OldInputs'
 import Preview from './Preview'
 
+// FIXME: Remove oldParameters when we remove feature flag
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/index.tsx:21]: why do we have old parameters? everyone is with the new ui no?

apps/web/src/hooks/useDocumentParameters/oldHook/index.ts
```diff
@@ -0,0 +1,142 @@
...
+        metadataParameters,
+        newParams,
+      })
+
+      setInputs(newInputs)
+    },
+    [inputs, newParams, setInputs],
+  )
+
+  return { parameters, inputs, setInputs, onMetadataProcessed }
```
csansoon, line 141: Why keeping the old hook though?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/index.tsx:21]: Not yet

apps/web/src/hooks/useDocumentParameters/oldHook/index.ts
andresgutgon: We're still under feature flag

Issue #606, opened 2024-11-13
Add relevant <title> tags to all pages/sections in the app


cesr:
Having multiple tabs open with Latitude is a nightmare because I never know what's in each of them. Pls fix.

PR #632, opened 2024-11-18
Add meta tags

neoxelox:
Closes: https://github.com/latitude-dev/latitude-llm/issues/606

Adds correspondent Metatags (Metadata) to every page.

> NOTE: This PR is in WIP, I need to check with the team when to stop or flatten the titles. As I did the system as a cascade tree I will just have to trim or flatten when necessary.

### DIFF
--- a/apps/web/src/hooks/useMetatags.ts
+++ b/apps/web/src/hooks/useMetatags.ts
@@ -1,0 +1,47 @@
+import { env } from '@latitude-data/env'
+import type { Metadata, ResolvedMetadata } from 'next'
+
+const DEFAULT_TITLE = 'The Open-Source LLM Development Platform'
+const DEFAULT_DESCRIPTION =
+  'Latitude is an end-to-end platform for prompt engineering where domain experts can collaborate with engineers to ship and maintain production-grade LLM features.'
+
+// This hook is necessary to define default metadata correctly, because
+// Next metadata merging would overwrite the nested objects totally.
+export function useMetatags({
+  title,
+  description,
+  parent,
+}: {
+  title?: string
+  description?: string
+  parent?: ResolvedMetadata
+}): Metadata {
+  let parentTitle = parent?.title?.absolute || ''
+  let metaTitle =
+    title && parentTitle
+      ? `${title} - ${parentTitle}`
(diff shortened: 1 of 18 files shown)


apps/web/src/hooks/useMetatags.ts
```diff
@@ -0,0 +1,47 @@
+import { env } from '@latitude-data/env'
+import type { Metadata, ResolvedMetadata } from 'next'
+
+const DEFAULT_TITLE = 'The Open-Source LLM Development Platform'
+const DEFAULT_DESCRIPTION =
+  'Latitude is an end-to-end platform for prompt engineering where domain experts can collaborate with engineers to ship and maintain production-grade LLM features.'
+
+// This hook is necessary to define default metadata correctly, because
+// Next metadata merging would overwrite the nested objects totally.
+export function useMetatags({
```
andresgutgon, line 10: Why not doing in the nextjs server?
https://nextjs.org/docs/app/building-your-application/optimizing/metadata#dynamic-metadata

neoxelox: I don't catch what do you exactly mean. The metadata is always generated in the server. I am already using generateMetadata wherever I need to generate dynamic metadata that is dependent of the path. For example, a title with name of a project...

I created useMetatags to have sane defaults and to allow correct merging of fields. This is actually advised in that same docs: https://nextjs.org/docs/app/building-your-application/optimizing/metadata#merging

useMetatags does not replace generateMetadata, because it is also used when exporting metadata directly (because the title is static).

csansoon: The `use` name is usually reserved to hooks, which cannot be used in server components. I think that's why @andresgutgon thought this was being used in client components.

Since this is not a hook, as it does not return any reactive data and is ran in server components, I think we should move this somewhere else to avoid confusion

andresgutgon (dismissed):

apps/web/src/hooks/useMetatags.ts
andresgutgon: Exactly, talked offline

neoxelox: Got it, thanks 👍

PR #644, opened 2024-11-19
Fix timestamp formatting

neoxelox:
Closes: https://github.com/latitude-dev/latitude-llm/issues/643

This PR fixes a bug with the `relativeTime` util function that misrepresented the timestamp when it was between 6 and 7 days old. Moreover it adds test and removes a duplicated implementation.

Screenshot 2024-11-19 at 15 58 16

### DIFF
--- a/apps/web/src/lib/relativeTime.test.ts
+++ b/apps/web/src/lib/relativeTime.test.ts
@@ -1,0 +1,52 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { relativeTime } from './relativeTime'
+
+describe('relativeTime', () => {
+  const SECONDS = 1000
+  const MINUTES = 60 * SECONDS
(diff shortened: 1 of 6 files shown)


csansoon (approved):

apps/web/src/lib/relativeTime.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    expect(relativeTime(new Date(NOW.getTime() - 1 * YEARS))).toBe(
+      'Aug 1, 1999, 12:00:00 PM',
+    )
+  })
+
+  it('returns "-" when no date is provided', () => {
+    expect(relativeTime(null)).toBe('-')
+    expect(relativeTime(undefined)).toBe('-')
+  })
+})
```
geclos, line 52: beautiful

PR #645, opened 2024-11-19
Revert "Web API fully typed with generics If this works as expected it improves our DX (#624)"

andresgutgon:
# What?
It "improves DX" but breaks the build 🦆

PR #646, opened 2024-11-19
Properly fix web api by respecting request/response argument on NextJS API handlers

andresgutgon:
# ⚠️ DO NOT MERGE
This is not the right approach 

---
# What?
This PR is open for [reverting the types in web api](https://github.com/latitude-dev/latitude-llm/pull/645) but is a shame. So I took a look and I think the error was that when building nextjs app it didn't recognize our NextJS API handlers as valid

What I did was to take the second argument of the handler to pass the context (user, workspace)
But this is wrong:

❌ Wrong:
```javascript
const GET =  (req: NextRequest, context: LatitudeAPIContext) => Promise<NextResponse>
```

What nextjs expect is req/res

✅ Ok:
```javascript
const GET =  (req: NextRequest, res: NextResponse, context: LatitudeAPIContext) => Promise<NextResponse>
```

## What was the problem?
NextJS's build when deploying was failing. 

Now I did this:
```bash
docker compose --profile building build web
```

And it finishes 🎉 Successfully 
image

### DIFF
--- a/apps/web/src/middlewares/authHandler.ts
+++ b/apps/web/src/middlewares/authHandler.ts
@@ -4,7 +4,7 @@ import { getCurrentUserOrError } from '$/services/auth/getCurrentUser'
 import { NextRequest, NextResponse } from 'next/server'
 
 export type DefaultParams = Record<string, unknown>
-export type AuthContext<IReq extends DefaultParams> = {
+export type Context<IReq extends DefaultParams> = {
   user: User
   workspace: WorkspaceDto
   params: IReq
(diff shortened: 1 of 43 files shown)


apps/web/src/middlewares/authHandler.ts
```diff
@@ -21,24 +21,28 @@ async function getUserOrUnauthorized() {
...
+    res: NextResponse<IResp>,
+    _context: Context<IReq>,
+  ) => {
     const { user, workspace } = await getUserOrUnauthorized()
+    // This is Typescript lie, we know that res is NextResponse<IResp>
+    // But because we want the return signature of the handler to be
+    // (req: NextRequest, res: NextResponse<IResp>)
+    // So in runtime we know params is a Promise<IReq>
+    const requestContext = res as unknown as { params: Promise<IReq> }
+    const params = await requestContext.params
```
andresgutgon, line 43: [This docs](https://nextjs.org/docs/app/building-your-application/routing/route-handlers#dynamic-route-segments)

andresgutgon:
I realize the types broke with nextjs upgrade 😭 . This is not the right fix. Better to revert. Another time

PR #640, opened 2024-11-19
Fix evals editor not updating after saving changes

andresgutgon:
# What?
Evals editor doesn't update the prompt in the playground even after saving changes

## Issue
https://github.com/latitude-dev/latitude-llm/issues/639

PR #651, opened 2024-11-21
Fix evaluation editor breaking when evaluation is not in the local cache

andresgutgon:
# What?
I fixed a [bug here](https://github.com/latitude-dev/latitude-llm/pull/640) where the chat was not updated after saving the evaluation. The problem is that evaluation state was not changing because we were not using to local SWR storage.

The problem is I introduced a bug if you navigate directly to `/editor` in evaluations it fails because the evaluation is not jet there the first time the page render. Now I fallback to server fetched evaluation in that case


apps/web/src/stores/evaluations.ts
```diff
@@ -88,8 +88,8 @@ export default function useEvaluations(
   })
 
   const findEvaluation = useMemo(
-    () => (evaluationUuid: string) =>
-      data.find((e) => e.uuid === evaluationUuid)!,
+    () => (evaluation: EvaluationDto) =>
+      data.find((e) => e.uuid === evaluation.uuid) ?? evaluation,
```
andresgutgon, line 92: Now evaluation will fallback to server evaluation first time `?? evaluation`. But after that `data` will contain a fresh version of it

geclos: super weird find method which returns its argument instead of null if it doesn't find what it is looking for.  I think you should keep the original implementation and either populate the swr state with the server state at render time or handle the case for findevaludation returning null, which is always a possibility

geclos (dismissed):

apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Playground/index.tsx
```diff
@@ -45,7 +45,7 @@ export default function Playground({
   evaluation: EvaluationDto
 }) {
   const { findEvaluation } = useEvaluations()
-  const evaluation = findEvaluation(serverEvaluation.uuid)!
+  const evaluation = findEvaluation(serverEvaluation)
```
Comment (csansoon) [inline apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Playground/index.tsx:48]: Why not using the original `serverEvaluation` though?

Comment (andresgutgon) [reply apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Playground/index.tsx:48]: Because if you use the `serverEvaluation` is not reacting to changes when the prompt is saved.

csansoon (approved):

PR #652, opened 2024-11-21
Latitude PromptL implementation

csansoon:
Latitude will now use the right compiler (legacy or ✨PromptL✨) depending on the version of the document and advanced evaluations, both to run prompts and to evaluate syntax.

Note: All LLM as Judge Simple evaluations will always be treated as PromptL evaluations, and evacuating them will create them accordingly. Until the UI is setup, this is the only way to create a promptl document for now.

### DIFF
--- a/packages/core/src/services/chains/ChainValidator/index.ts
+++ b/packages/core/src/services/chains/ChainValidator/index.ts
@@ -1,5 +1,11 @@
-import { Chain, CompileError, Conversation } from '@latitude-data/compiler'
+import {
+  CompileError,
+  Conversation,
+  Chain as LegacyChain,
+  Message,
+} from '@latitude-data/compiler'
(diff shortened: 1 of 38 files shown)


packages/core/src/services/chains/ChainValidator/index.ts
```diff
@@ -160,7 +175,7 @@ export const validateChain = async (
 
   const rule = applyCustomRules({
     providerType: provider.provider,
-    messages: conversation.messages,
+    messages: conversation.messages as Message[],
```
andresgutgon, line 178: Why this? Is not the same signature?

csansoon: Different types from different packages with different enums (although same values). Typescript does not like that.

andresgutgon (approved):

PR #662, opened 2024-11-25
Bugfix: Non-stream API run requests failing with PromptL documents

csansoon:
Bug: Running a promptl document through the API with `stream: false` returned a "Conversation messages not found in response" error.

Explanation:
When running a prompt through the gateway, we store the provider logs asyncronously through workers to not add any delay to our response. However, since we always need the latest provider log, this one WAS created syncronously.

We used the ⁠`completed` attribute from the `chain.step` response to determine which step’s provider log had to be created synchronously. This attribute indicated which step was the last one. However, with PromptL, the meaning of `⁠completed` has changed. Now, it is true when the last step has already been run and there are no more steps ahead.

### DIFF
--- a/packages/core/src/services/chains/run.ts
+++ b/packages/core/src/services/chains/run.ts
@@ -207,18 +207,20 @@ async function runStep({
           errorableUuid,
           response: cachedResponse,
         }),
-        saveSyncProviderLogs: step.chainCompleted,
+        saveSyncProviderLogs: true, // TODO: temp bugfix, shuold only save last one syncronously
       })
+
+      const response = { ...cachedResponse, providerLog }
 
       if (step.chainCompleted) {
         streamConsumer.chainCompleted({
           step,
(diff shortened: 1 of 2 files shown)


packages/core/src/services/chains/run.ts
```diff
@@ -207,18 +207,22 @@ async function runStep({
           errorableUuid,
           response: cachedResponse,
         }),
-        saveSyncProviderLogs: step.chainCompleted,
+        saveSyncProviderLogs: true, // TODO: temp bugfix, shuold only save last one syncronously
       })
 
+      const response = { ...cachedResponse, providerLog }
+
+      console.log('CACHED providerLog', providerLog)
```
andresgutgon, line 215: 🔥

andresgutgon:
Once this fix is working would be great to have a test to ensure this is not broken

Issue #496, opened 2024-10-23
Share this prompt


cesr:
It's a public website similar to Notion's with a form users fill out and they get the answer from the LLM
We show a checkbox to allow responses or not
The prompt is shared and people can duplicate it

## Requirements

- [x] Show preview URL in share prompt modal
- [x] Enable button to publish/unpublish a document
- [x] Only live documents can be shared (drafts will return 404)
- [x] Maybe add description to parameters
- [x] Share button with indicator
- [x] Share modal with tabs
- [x] Action to share. Find or create
- [x] Action to unshare
- [x] Forking means also copy all their linked prompts inside that prompt (FUN)
- [x] Public url `/share/d/:publishedDocumentUuid`
- [x] Maybe a nice 404 indicating that the prompt might be unpublished and contact with workspace owner
- [x] Sidebar with parameters 
- [x] Users can run the document. 
- [x] Event for runs
- [x] Event for visits
- [x] Event for fork prompt
- [x] Users can follow conversation with AI (If follow conversation is enable)
- [x] Parameters update the URL so it can be shared like that
- [x] Ability to FORK the prompt. Flow through login/signup. Fork the hidden uuid in the signup or login form
- [x] Collapse prompt messages and only show assistant response

PR #664, opened 2024-11-25
Data schema for sharing publicly prompts 

andresgutgon:
# What?
The idea is to allow users in prompt UI to publish the prompt so people on their organization without access to Latitude can interact with this prompt.

I'm thinking on doing a different table with published documents. I'm doing this way so we can share public UUIDs and not internal document uuid. Also because we want to add extra configuration like if the user want to allow to consumers of this public prompt to follow up the conversation with the AI.

## Issue
https://github.com/latitude-dev/latitude-llm/issues/496

### DIFF
--- a/packages/core/src/schema/models/publishedDocuments.ts
+++ b/packages/core/src/schema/models/publishedDocuments.ts
@@ -1,0 +1,48 @@
+import { sql } from 'drizzle-orm'
+import {
+  bigint,
+  boolean,
+  index,
+  text,
+  uniqueIndex,
+  uuid,
+  varchar,
+} from 'drizzle-orm/pg-core'
+
+import { latitudeSchema } from '../db-schema'
+import { timestamps } from '../schemaHelpers'
+import { projects } from './projects'
+import { workspaces } from './workspaces'
+
+export const publishedDocuments = latitudeSchema.table(
+  'published_documents',
+  {
+    uuid: uuid('uuid')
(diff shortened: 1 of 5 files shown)


packages/core/src/schema/models/publishedDocuments.ts
```diff
@@ -0,0 +1,36 @@
...
+      .references(() => projects.id, { onDelete: 'cascade' }),
+    allowFollowConversation: boolean('allow_follow_conversation').notNull().default(false),
+    ...timestamps(),
+  },
+  (table) => ({
+    projectWorkspaceIdx: index('workspace_idx').on(table.workspaceId),
+    uniqueProjectDocumentUuid: uniqueIndex(
+      'unique_project_document_uuid_idx',
+    ).on(table.projectId, table.documentUuid),
+  }),
```
andresgutgon, line 35: My idea here is to look these public documents by `uuid` and then pick 

`workspaceId`, `projectId` and `documentUuid` and use our methods
```javascript
const documentsScope = new DocumentVersionsRepository(ctx.workspace.id)
const commitUuid = // Get live version for this project
const document = await documentsScope
  .getDocumentAtCommit({
    projectId,
    commitUuid,
    documentUuid,
  })
  .then((r) => r.unwrap())
```

What do you think?

csansoon (dismissed):

PR #668, opened 2024-11-26
Fix: schema ignored in PromptL

csansoon:
**Problem:**
With the legacy compiler, the user could not define specific configuration only for the last step of a chan, since the last step was executed at the end of the document without a `<response />` tag like the rest of them. Since specifying a custom schema only for the last step was a common practice, we decided that the schema defined in the global configuration would be ignored in all steps but the last one.

PromptL, however, cannot know which step is the last one until it has been ran. This means that all steps ignored the schema.

**Solution:**
In PromptL, you can in fact define specific configuration to the last step, as it can (should) be wrapped with a `<step>` tag, which accepts custom config just like `response` did before.

So, for legacy documents, we will keep running the old logic, but PromptL documents will not ignore any configuration from the global config into any step. Instead, if you want a specific configuration only applied to the last step, you can just define it there.

### DIFF
--- a/packages/core/src/services/chains/ChainValidator/index.ts
+++ b/packages/core/src/services/chains/ChainValidator/index.ts
@@ -40,12 +40,16 @@ type ValidatorContext = {
   configOverrides?: ConfigOverrides
 }
 
-const getInputSchema = (
-  chainCompleted: boolean,
-  config: Config,
-  configOverrides?: ConfigOverrides,
-): JSONSchema7 | undefined => {
-  if (!chainCompleted) return undefined
+const getInputSchema = ({
+  config,
+  configOverrides,
+  ignoreSchema = false,
+}: {
+  config: Config
+  configOverrides?: ConfigOverrides
+  ignoreSchema?: boolean
+}): JSONSchema7 | undefined => {
+  if (ignoreSchema) return undefined
   const overrideSchema =
     configOverrides && 'schema' in configOverrides
       ? configOverrides.schema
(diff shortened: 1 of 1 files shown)


geclos (approved):

packages/core/src/services/chains/ChainValidator/index.ts
```diff
@@ -206,7 +224,7 @@ export const validateChain = async (
       ...conversation,
       messages: rule?.messages ?? conversation.messages,
     },
-    schema: getInputSchema(chainCompleted, config, configOverrides),
-    output: getOutputType(chainCompleted, config, configOverrides),
+    schema,
+    output,
   })
 }
```
geclos, line 230: not sure if there are tests for chain validator but if there are you know what to do 👌🏼

packages/core/src/services/chains/ChainValidator/index.ts
```diff
@@ -198,6 +205,17 @@ export const validateChain = async (
...
+  const schema = getInputSchema({
+    config,
+    configOverrides,
+    ignoreSchema: promptlVersion === 0 && !chainCompleted,
+  })
+  const output = getOutputType({
+    config,
+    configOverrides,
+    ignoreSchema: promptlVersion === 0 && !chainCompleted,
+  })
```
andresgutgon, line 217: What happen with non-promptl prompts?

packages/core/src/services/chains/ChainValidator/index.ts
csansoon: There is not :(

packages/core/src/services/chains/ChainValidator/index.ts
csansoon: old logic. Ignore schema in all steps but the last one

PR #670, opened 2024-11-27
Feat/overview

geclos:


packages/core/src/services/projects/computeProjectStats.ts
```diff
@@ -0,0 +1,178 @@
...
+    runsPerModel,
+    costPerModel,
+    rollingDocumentLogs,
+    totalEvaluations,
+    totalEvaluationRuns,
+    evaluationCosts,
+  }
+
+  return Result.ok(stats)
+}
```
andresgutgon, line 178: Nice file. No tests tho

geclos: fixored

PR #671, opened 2024-11-27
Fix Refine from evaluation results

andresgutgon:
# What?
In the first iteration of refine from evaluations I used `computeEvaluationResultsByDocumentContent` service which does a lot of things and pick a lot of evaluation results. This is wrong for the task of selecting only results comming from evaluation results table. The fix simplifies the query by just using evaluation results repository and checking the evaluation is connected to the document where the refinement happens.


geclos (approved):

csansoon (approved):
👏👏👏

PR #672, opened 2024-11-27
sdk: render prompts and chains locally

csansoon:
Added `client.prompts.render` and `client.prompt.renderChain` methods to the SDK's Latitude Client.

Examples:

### Render prompt
```ts
const prompt = await sdk.prompts.get(...)

const { config, messages } = await sdk.prompts.render({
  prompt,
  parameters: { name: 'John' },
})

const response = await openai.chat.completions({ ...config, messages })
```

### Render chain
```ts
const prompt = await sdk.prompts.get(...)

await sdk.prompts.renderChain({
  prompt,
  parameters: { name: 'John' },
  onStep: async ({ config, messages }) => {
    return await openai.chat.completions({ ...config, messages })
  }
})
```


andresgutgon (approved):

apps/gateway/package.json
```diff
@@ -22,7 +22,7 @@
     "@latitude-data/constants": "workspace:^",
     "@latitude-data/core": "workspace:^",
     "@latitude-data/env": "workspace:^",
-    "@latitude-data/promptl": "^0.1.2",
+    "@latitude-data/promptl": "^0.1.5",
```
geclos, line 25: didn't you just updated it to 0.2?

geclos (approved):

apps/gateway/package.json
csansoon: true

packages/core/src/services/documents/scan.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const docInCommit = docs.find((d) => d.documentUuid === document.documentUuid)
+
+  if (!docInCommit) {
+    return Result.error(
+      new UnprocessableEntityError('Document not found in commit', {}),
+    )
+  }
+
+  const referenceFn = async (
```
andresgutgon, line 42: I will use the same function. Please export so i can use it. Meantime I'll do my own

andresgutgon (approved):

PR #674, opened 2024-11-27
Fork document from one workspace to another

andresgutgon:
# What?
This PR only focuses on the actual service that does the forking. The authorization layer will be added later. I want to make this ready for use

image

## TODO
- [x] Add tests 
- [x] Implement fork
- [x] Choose a provider to use in copied documents
- [x] Fork to a new workspace (one after signup)
- [x] Fork to an existing workspace with default provider
- [x] Fork to an existing workspace without default provider
- [x] ~~Fork to an existing workspace without providers. It's possible? 💥~~ . Nothing happens it sets to nothing so they have to create one provider and set in the forked document
- [x] QA Signup flow

### DIFF
--- a/apps/web/src/services/user/setupService.ts
+++ b/apps/web/src/services/user/setupService.ts
@@ -1,21 +1,6 @@
-import { Providers, RewardType } from '@latitude-data/core/browser'
-import { SessionData } from '@latitude-data/core/data-access'
-import { publisher } from '@latitude-data/core/events/publisher'
-import { Result } from '@latitude-data/core/lib/Result'
-import Transaction, {
-  PromisedResult,
-} from '@latitude-data/core/lib/Transaction'
-import { createApiKey } from '@latitude-data/core/services/apiKeys/create'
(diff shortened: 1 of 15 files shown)


packages/core/src/services/documents/scan.ts
```diff
@@ -12,33 +12,13 @@ import {
...
-
-  const docInCommit = docs.find((d) => d.documentUuid === document.documentUuid)
-
-  if (!docInCommit) {
-    return Result.error(
-      new UnprocessableEntityError('Document not found in commit', {}),
-    )
-  }
-
+  getDocumentByPath: (path: string) => DocumentVersion | undefined
```
andresgutgon, line 20: Review this change @csansoon I wanted to reuse some part of this when I already have the docs

apps/web/src/services/user/setupService.ts
andresgutgon, line 1: I moved all this service to core so I can test a new created workspace can fork documents

csansoon (approved):

PR #678, opened 2024-11-29
Fix: Tool call names

csansoon:
Vercel does not eat the `tools` configuration right away, and instead requires an additional transformation. Within this transformation, we were not maintaining the tool name

SCR-20241129-nhaq

Now, the name is maintained:

SCR-20241129-ngua


geclos:
this is because you've represented the tool calls as an array of tools, if you pass an object it works perfectly, i wouldn't apply custom transformations if not needed, just document the right way of declaring tools:

```
provider: Real openai
model: gpt-4o-mini
tools:
  weather:
    description: Gets the weather for particular location
    parameters:
      type: object
      properties:
        location:
          type: string
          description: Location to get the weather for
      required:
        - location
      additionalProperties: false
```

csansoon:
> this is because you've represented the tool calls as an array of tools, if you pass an object it works perfectly, i wouldn't apply custom transformations if not needed, just document the right way of declaring tools:
> 
> ```
> provider: Real openai
> model: gpt-4o-mini
> tools:
>   weather:
>     description: Gets the weather for particular location
>     parameters:
>       type: object
>       properties:
>         location:
>           type: string
>           description: Location to get the weather for
>       required:
>         - location
>       additionalProperties: false
> ```

@geclos  Hmm but that differs from the OpenAI's API documentation. I guess we could just document our way to do it though

PR #682, opened 2024-12-01
Implement fork action for already logged in users

andresgutgon:
# What?
Implement fork action for logged users

### TODO
- [ ] Fix tests for fork document service. Review incremental name generation. It's not working with more than 2 forked projects with the same name

## Off-topic: Modals adjustments
I made some modals adjustments because current status footer and header are sticky but looks weird. Example:
Header
image

Footer
image

After fix
image


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/layout.tsx
```diff
@@ -36,18 +37,29 @@ export default async function CommitLayout({
...
-      commit = commits.find((c) => !!c.mergedAt)
+      commit = headCommit
     } else {
+      const commits = await findCommitsByProjectCached({
+        projectId: project.id,
+      })
       commit = commits.find((c) => c.uuid === commitUuid)
     }
 
+    isHead = commit?.id === headCommit?.id
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/layout.tsx:61]: Now `isHead` is true when `/live/...` in URL or when a head UUID is in the URL

PR #693, opened 2024-12-02
Shared prompts show last assistant message by default only

andresgutgon:
# What?
By default, all the previous prompts and steps are collapsed, but if the user wants to check them, they can be uncollapsed.


apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Editor/Advanced/PromptlNotificationModal/index.tsx
```diff
@@ -41,6 +41,15 @@ export function PromptlNotificationModal({
       onOpenChange={() => dismissNotification(true)}
       title='Syntax Update Available'
       description='We have updated the syntax for Latitude prompts!'
+      footer={
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Editor/Advanced/PromptlNotificationModal/index.tsx:44]: Always use `footer` in modals for buttons 🙏

apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Playground/Chat/index.tsx
```diff
@@ -110,9 +110,8 @@ export default function Chat({
 
       switch (event) {
         case StreamEventTypes.Latitude: {
-          if (data.type === ChainEventTypes.Step) {
-            if (data.isLastStep) setChainLength(messagesCount + 1)
-          } else if (data.type === ChainEventTypes.Complete) {
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/evaluations/(evaluation)/[evaluationUuid]/editor/_components/Playground/Chat/index.tsx:115]: This was a bug. We lost running time line because `promptl` does not know what `isLastStep` so we can't rely on that. We do this instead of chain complete.
image

PR #700, opened 2024-12-04
Do not store in cached AI response the data model attributes like `providerLog` or `documentLogUuid`

andresgutgon:
# What?
We were storing these attributes and we had a bug where follow up conversations were linked to the original document log UUID because that's what was returned from running a chain that was cached

### DIFF
--- a/packages/core/src/services/commits/promptCache.test.ts
+++ b/packages/core/src/services/commits/promptCache.test.ts
@@ -3,6 +3,7 @@ import { beforeEach, describe, expect, it, vi } from 'vitest'
 import * as cacheModule from '../../cache'
 import { createProject } from '../../tests/factories'
 import { getCachedResponse, setCachedResponse } from './promptCache'
+import { LogSources } from '../../constants'
 
 describe('promptCache', async () => {
   const mockCache = {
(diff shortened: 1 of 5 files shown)


packages/core/src/services/commits/promptCache.test.ts
```diff
@@ -20,14 +21,38 @@ describe('promptCache', async () => {
...
-    streamType: 'text',
+    streamType: 'text' as 'text',
     text: 'cached response',
     usage: {
       promptTokens: 0,
       completionTokens: 0,
       totalTokens: 0,
     },
     toolCalls: [],
+    // This should not be cached
```
geclos, line 32: dont' get it you just removed these from the cache no?

packages/core/src/services/commits/promptCache.ts
```diff
@@ -71,10 +85,29 @@ export async function setCachedResponse({
...
+      streamType: 'object',
+      text: response.text,
+      usage: response.usage,
+      object: response.object,
+    })
+  } else {
+    throw new Error(
+      'Invalid "streamType" response, it should be "text" or "object"',
+    )
+  }
```
geclos, line 112: hmm i'm a bit worried we might miss something instead of picking the parameters you wanna cache why not cache everything except documentLogUuid and providerLog? This way we ensure there's no incompatibility between the chain response that is cached and the chain response that isn't cached

andresgutgon: What can we miss? Typescript is not complaining according to response types

andresgutgon: This throw is here in case we start producing a different type of response than `text` or `object` and we forget to handle it in the cache

PR #712, opened 2024-12-11
Implement filter logs by creation date and refactor a bit

andresgutgon:
# What?
Implement the backend part of the logs filtered by `createdAt`

## TODO
- [x] Do not encode `versions`, `createdAt` and `logSources` filters. It's not necessary.
- [x] Centralize a bit more all related with log filters `#FEELS_GOOD`
- [x] Check API_ROUTES. Do not encoded there
- [x] Fix generate api route with params
- [x] Send `createdAt` from `DatePicker`. Handle when no `to` is present
- [x] Implement filter in SQL + tests
- [x] Fix select icon. Remove radix icons and use lucide version
- [x] On select end of range set time end of day. Otherwhise we lost all logs that day
- [x] Fix DatePicker text for TODAY in dark mode 🐛


packages/core/src/services/documentLogs/logsFilterUtils/parseLogFilterParams.ts
```diff
@@ -0,0 +1,50 @@
...
+  return logSources.filter((s) => LOG_SOURCES.includes(s as LogSources))
+}
+
+export function parseSafeCreatedAtRange(
+  createdAt: string[] | string | undefined,
+) {
+  const [from, to] = createdAt?.toString()?.split(',') ?? []
+  if (!from || !to) return undefined
+
+  return { from: new Date(from), to: new Date(to) }
```
andresgutgon, line 16: :bug Fix this

packages/web-ui/src/ds/atoms/Select/Primitives/index.tsx
```diff
@@ -9,16 +9,11 @@ import {
   useMemo,
   useState,
 } from 'react'
-import {
-  CaretSortIcon,
-  CheckIcon,
-  ChevronDownIcon,
-  ChevronUpIcon,
-} from '@radix-ui/react-icons'
```
andresgutgon, line 17: This icons are redundant. We use lucide

PR #713, opened 2024-12-12
feat(evaluations): rename some span columns and change tenancy to the workspace

geclos:


andresgutgon (approved):
Is fucking crazy drizzle migrations metadata

PR #721, opened 2024-12-13
Fix/feature 597 bucket to public part 2

neoxelox:


apps/infra/src/core/s3.ts
```diff
@@ -33,3 +33,25 @@ export const publicBucket = new aws.s3.BucketV2(
 )
 
 export const publicBucketName = publicBucket.bucket
+
+new aws.s3.BucketPolicy(
```
neoxelox, line 37: This will disappear after rebasing, when merging the first part of this fix

packages/core/src/services/files/upload.ts
```diff
@@ -38,10 +66,9 @@ export async function uploadFile(
 
   try {
     await disk.putFile(key, file).then((r) => r.unwrap())
-    const url = await disk.getSignedUrl(key, {
-      expiresIn: undefined,
-      contentDisposition: 'inline',
-    })
+    // TODO: Use temporal signed URLs, with a (micro)service
+    // acting as a reverse proxy refreshing the signed urls
```
andresgutgon, line 70: Don't put this TODO, I don't think we'll do in the short term no?

andresgutgon (approved):

packages/core/src/services/files/upload.ts
neoxelox: Just for future reference so we don't forget!

andresgutgon: Don't put too much todos or team morale will lower 😂 Just the necessary

PR #731, opened 2024-12-17
feat(telemetry): handle non-oai vercel calls in some scenarios

geclos:
Generating objects with non-oai models via vercel produced some unexpected spans. Adapted code to handle those. Also implemented upsert of spans in case they already exist.

### DIFF
--- a/examples/chat-app/instrumentation.ts
+++ b/examples/chat-app/instrumentation.ts
@@ -1,33 +1,14 @@
 import dotenv from 'dotenv'
 import { Latitude } from '@latitude-data/sdk'
-import { Anthropic } from '@anthropic-ai/sdk'
-import OpenAI from 'openai'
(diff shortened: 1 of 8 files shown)


examples/chat-app/instrumentation.ts
```diff
@@ -1,33 +1,14 @@
...
 import { Latitude } from '@latitude-data/sdk'
-import { Anthropic } from '@anthropic-ai/sdk'
-import OpenAI from 'openai'
+import { VercelSpanProcessor } from '@latitude-data/telemetry'
 
 // Load environment variables first
 dotenv.config()
 
-export const latitude = new Latitude(process.env.LATITUDE_API_KEY, {
+export const latitude = new Latitude('9d5a427b-f4db-42c4-ac03-41e30675bac2', {
```
geclos, line 8: dev key can be safely published

Issue #720, opened 2024-12-13
Move editor dataset mappings to DB


andresgutgon:
# What?
Now we persist dataset mapping of a prompt in Latitude Editor in local storage. 
![image](https://github.com/user-attachments/assets/b13061ca-ba8c-4478-b1ac-62d7e7dc0043)

We want to persist this mapping in DB. this way is shared between all the users in the workspace

## TODO

- [x] Move dataset mapping to the database as a JSONB in the document version table
- [x] Run in batch from editor does not persist the mapping 🐛 
- [x] Run in batch from evaluation does not persist the mapping 🐛 

### Run in batch from the editor
![image](https://github.com/user-attachments/assets/186e6672-c134-4764-b1b3-8ec6bde0a81d)

### Run in batch from evaluation
![image](https://github.com/user-attachments/assets/495df943-e98f-4cc3-b449-eacc373c6eb6)

andresgutgon:
PR: https://github.com/latitude-dev/latitude-llm/pull/734

PR #733, opened 2024-12-17
Persist dataset mapping in document version

andresgutgon:
# What?
This is explained in more detail [here](https://github.com/latitude-dev/latitude-llm/issues/720), but we want to store the dataset the users chose for the document version. 

Now we're storing on local storage. This has the problem that it can't be shared and all users have to do the same mapping

## TODO
- [ ] Persist datataset mapping and dataset inputs in document version table in a `JSONB` field. By `datasetId`. This means users don't lose mapping if they change selected dataset.
- [ ] Run in batch from editor does not persist the mapping 🐛
- [ ] Run in batch from evaluation does not persist the mapping 🐛


apps/web/src/hooks/useDocumentParameters/index.ts
```diff
@@ -1,59 +1,10 @@
 import { useCallback, useMemo } from 'react'
 
 import { recalculateInputs } from '$/hooks/useDocumentParameters/recalculateInputs'
-import { DocumentLog, ParameterType } from '@latitude-data/core/browser'
+import { DatasetSource, DocumentLog, INPUT_SOURCE, Inputs, InputSource, PlaygroundInput, PlaygroundInputs } from '@latitude-data/core/browser'
```
andresgutgon, line 4: I moved all the types and constants related to document parameters to the core. 
The reason is that now I want to type the jsonb field in the DB with those types. So, having it on web was not possible.

apps/web/src/hooks/useDocumentParameters/index.ts
```diff
@@ -339,11 +280,11 @@ export function useDocumentParameters({
       setInputs: setManualInputs,
     },
     dataset: {
+      // TODO: Fetch from DB
       datasetId: inputs['dataset'].datasetId,
       rowIndex: inputs['dataset'].rowIndex,
       inputs: inputs['dataset'].inputs,
       mappedInputs: inputs['dataset'].mappedInputs,
-      setInput: setDatasetInput,
```
andresgutgon, line 346: This was not used. Removing it

packages/core/src/lib/documentPersistedInputs.ts
```diff
@@ -0,0 +1,54 @@
...
+    inputs: Record<string, DatasetInput>
+    mappedInputs: Record<string, number>
+  }
+  history: {
+    logUuid: string | undefined
+    inputs: Record<string, HistoryInput>
+  }
+}
+
+export type LinkedDataset = Omit<PlaygroundInputs<'dataset'>['dataset'], 'datasetId'>
```
andresgutgon, line 49: The idea is to persist `rowIndex`, `inputs` and `mappedInputs` in the new `jsonb` column. This will be done in the next PR

PR #734, opened 2024-12-17
Read and Write dataset mapped inputs from document version

andresgutgon:
# WHAT?
We want to persist more permanently the mapping of a selected dataset for a document so anyone in the workspace can see that mapping

## TODO
- [x] Persist dataset mapping and dataset inputs in document version table in a ~~`JSONB`~~ `JSON` field. By `datasetId`. This means users don't lose mapping if they change selected dataset.
- [x] Run in batch from editor does not persist the mapping 🐛
- [x] Run in batch from evaluation does not persist the mapping 🐛
- [x] Check with existing local storage mapping configuration


apps/web/src/hooks/useDocumentParameters/index.ts
```diff
@@ -72,28 +80,70 @@ function mapLogParametersToInputs({
...
+  document: DocumentVersion
+  localInputs: PlaygroundInputs<'dataset'>['dataset']
+}) {
+  const datasetId = document.datasetId
+  if (!datasetId) return EMPTY_LINKED_DATASET
+
+  const all = document.linkedDataset ?? {}
+  const isEmpty = Object.keys(all).length === 0
+
+  if (isEmpty) {
```
andresgutgon, line 96: QA with an existing local storage with mapped fields. Switch to main, set the map and back to this branch

apps/web/src/hooks/useDocumentParameters/index.ts
```diff
@@ -281,28 +325,26 @@ export function useDocumentParameters({
...
       setDataset,
+      copyToManual: copyDatasetInputsToManual,
     },
     history: {
       logUuid: inputs['history'].logUuid,
       inputs: inputs['history'].inputs,
       setInput: setHistoryInput,
       setInputs: setHistoryInputs,
       setHistoryLog,
+      mapDocParametersToInputs,
```
andresgutgon, line 347: QA history

packages/core/src/services/documents/saveLinkedDataset.ts
```diff
@@ -0,0 +1,43 @@
...
+  const prevLinked = document.linkedDataset ?? {}
+  return await Transaction.call(async (tx) => {
+    const linkedDataset = {
+      ...prevLinked,
+      [dataset.id]: {
+        rowIndex: data.rowIndex,
+        inputs: data.inputs,
+        mappedInputs: data.mappedInputs,
+      },
+    }
```
andresgutgon, line 33: Do a test for this. We have to ensure old dataset mappings are preserved and only the dataset ID passed is overwritten

geclos (approved):

PR #735, opened 2024-12-17
feature(telemetry): search  box fixes

geclos:
- Load filters in the server component
- Load filters from search params
- Fix not_* filters
- Add trace id filter
- Show suggestions on focus


apps/web/src/app/(private)/traces/_components/TracesTable/SearchBox/constants.ts
```diff
@@ -0,0 +1,37 @@
+import { SearchColumn, Operator } from '../types'
+
+export const TIME_OPERATORS: Operator[] = [
+  { label: 'is', value: 'eq' },
+  { label: 'before', value: 'lt' },
+  { label: 'after', value: 'gt' },
+]
```
andresgutgon, line 7: Maybe you could use the new fancy date picker instead. For date picking, the date picker is nicer

andresgutgon (dismissed):

apps/web/src/app/(private)/traces/_components/TracesTable/SearchBox/constants.ts
geclos: yeah i will implement this pretty soon

PR #740, opened 2024-12-18
Revert document changes service

csansoon:
This PR adds two services to documentVersions:

**Reset to version:**
`resetToDocumentVersion` will create the necessary changes in the current draft to reset the state of a selected document to how it was in any previous commit.

**Compute revert changes:**
`computeDocumentRevertChanges` will compute what changes have been made to a document from a selected commit to another one, and then calculate what changes must be done in the current draft to revert them.
This one does not perform the changes automatically, as I will like to first display to the user what changes would be made to the document before actually applying them, as the document's content diff reversal may be tricky when documents are completely different.

Not being used yet, it will be added to the History UI in a future PR.


packages/core/src/services/documents/resetToVersion.ts
```diff
@@ -0,0 +1,75 @@
...
+  {
+    documentVersion,
+    draft,
+  }: {
+    documentVersion: DocumentVersion
+    draft: Commit
+  },
+  db = database,
+): Promise<TypedResult<DocumentVersion, Error>> {
+  const workspace = await findWorkspaceFromCommit(draft, db)
```
andresgutgon, line 26: Pass the workspace instead of making an extra request. You'll we have to do it any way when using in an action

packages/core/src/services/documents/computeRevertChanges.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  const changesToRevert: Partial<DocumentVersion> = omit(
+    getChangesToRevert({ originalDocument, changedDocument }),
+    [
+      'id',
+      'resolvedContent',
+      'contentHash',
+      'createdAt',
+      'updatedAt',
+      'mergedAt', // Why is this here (?)
```
andresgutgon, line 89: Who are you asking?

csansoon: just asking out loud waiting to see if someone magically responds

andresgutgon (approved):

PR #741, opened 2024-12-18
History tweaks

csansoon:
Some minor fixes for the History page:
- Draggable columns automatically resize when the window size changes
- Decreased minimum column size to account for smaller screens
- Some styles fixes


andresgutgon (approved):

packages/web-ui/src/ds/atoms/SplitPane/index.tsx
```diff
@@ -88,6 +89,7 @@ const SplitPane = ({
   className?: string
   visibleHandle?: boolean
   dragDisabled?: boolean
+  autoResize?: boolean
```
andresgutgon, line 92: Maybe we should also add this to the editor panels

PR #743, opened 2024-12-18
refactor(traces): simplify traces pagination and realtime updates

geclos:
- Remove pagination params from useRealtimeTraces hook
- Set default values for page size and filters
- Improve URL parameter handling in pagination component (was a bit broken)
- Made Typescript SDK render method not depend on latitude-specific prompt abstractions such as resolvedContent or provider

### DIFF
--- a/apps/web/src/app/(private)/traces/_components/TracesTable/index.tsx
+++ b/apps/web/src/app/(private)/traces/_components/TracesTable/index.tsx
@@ -55,12 +55,7 @@ export function TracesTable({ traces: serverTraces }: Props) {
     },
   )
 
-  useRealtimeTraces({
-    page: Number(page),
-    pageSize: Number(pageSize),
-    filters,
-    fallbackData: serverTraces,
-  })
+  useRealtimeTraces()
 
   const handleSearch = (completedSearches: CompletedSearch[]) => {
     const params = new URLSearchParams(searchParams)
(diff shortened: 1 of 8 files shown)


apps/web/src/app/(private)/traces/_components/TracesTable/index.tsx
```diff
@@ -55,12 +55,7 @@ export function TracesTable({ traces: serverTraces }: Props) {
     },
   )
 
-  useRealtimeTraces({
-    page: Number(page),
-    pageSize: Number(pageSize),
-    filters,
-    fallbackData: serverTraces,
-  })
+  useRealtimeTraces()
```
geclos, line 58: so that real time updates do not show up if i'm not on first page

apps/web/src/components/TablePaginationFooter/GoToPageInput/index.tsx
```diff
@@ -24,15 +24,18 @@ export function GoToPageInput({
...
 
     if (baseUrl) {
       router.push(
         buildPaginatedUrl({
           baseUrl,
           page: targetPage,
           pageSize,
-          queryParams,
+          queryParams: urlParams.toString(),
           encodeQueryParams: false,
```
geclos, line 39: this was broken it was adding 2 page search params instead of replacing the existing one

andresgutgon: This is used also in logs. Can you check it works ok when logs has filters and go to page 2 or 3...?

andresgutgon (dismissed):

PR #746, opened 2024-12-19
fix(telemetry): concurrency fixes

geclos:
In theory open telemetry spans are unique per trace but the reality is that nothing stops clients from sending spans with the same ID multiple times, and indeed they do. We handle this case by merging the spans on conflict where it is sensible to do so.

Also implemented some minor UI fixes and type fixes.

### DIFF
--- a/packages/core/drizzle/0113_quiet_wendigo.sql
+++ b/packages/core/drizzle/0113_quiet_wendigo.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "latitude"."spans" DROP CONSTRAINT "spans_span_id_unique";--> statement-breakpoint
+ALTER TABLE "latitude"."spans" ADD CONSTRAINT "spans_span_id_trace_id_unique" UNIQUE("span_id","trace_id");
--- a/packages/core/drizzle/0114_flippant_sunset_bain.sql
+++ b/packages/core/drizzle/0114_flippant_sunset_bain.sql
@@ -1,0 +1,2 @@
+ALTER TABLE "latitude"."spans" RENAME COLUMN "prompt_tokens" TO "input_tokens";--> statement-breakpoint
+ALTER TABLE "latitude"."spans" RENAME COLUMN "completion_tokens" TO "output_tokens";
--- a/packages/core/src/jobs/job-definitions/traces/processOtlpTracesJob.ts
+++ b/packages/core/src/jobs/job-definitions/traces/processOtlpTracesJob.ts
@@ -71,5 +71,5 @@ export const processOtlpTracesJob = async (
     traces: tracesToCreate,
     // @ts-expect-error - Fix when we fix types in compiler
     spans: processedSpans,
-  })
+  }).then((r) => r.unwrap())
 }
--- a/packages/core/src/services/traces/bulkCreateTracesAndSpans.ts
+++ b/packages/core/src/services/traces/bulkCreateTracesAndSpans.ts
@@ -24,9 +24,8 @@ export async function bulkCreateTracesAndSpans(
   db = database,
 ) {
   // Find existing traces and spans outside the transaction
-  const [existingTraceIds, existingSpanIds] = await Promise.all([
+  const [existingTraceIds] = await Promise.all([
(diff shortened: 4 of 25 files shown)


packages/core/drizzle/0113_quiet_wendigo.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "latitude"."spans" DROP CONSTRAINT "spans_span_id_unique";--> statement-breakpoint
+ALTER TABLE "latitude"."spans" ADD CONSTRAINT "spans_span_id_trace_id_unique" UNIQUE("span_id","trace_id");
```
geclos, line 2: span ids are only required to be unique per trace, not globally, so changing the index

packages/core/drizzle/0114_flippant_sunset_bain.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "latitude"."spans" RENAME COLUMN "prompt_tokens" TO "input_tokens";--> statement-breakpoint
+ALTER TABLE "latitude"."spans" RENAME COLUMN "completion_tokens" TO "output_tokens";
```
geclos, line 2: minor column renaming

packages/core/src/jobs/job-definitions/traces/processOtlpTracesJob.ts
```diff
@@ -71,5 +71,5 @@ export const processOtlpTracesJob = async (
     traces: tracesToCreate,
     // @ts-expect-error - Fix when we fix types in compiler
     spans: processedSpans,
-  })
+  }).then((r) => r.unwrap())
```
geclos, line 74: i want this job to retry if the service fails for whatever reason so we unwrap

packages/core/src/services/traces/bulkCreateTracesAndSpans.ts
```diff
@@ -215,60 +180,139 @@ async function createTraces(
...
+            WHEN ${spans.totalCostInMillicents} = 0 AND EXCLUDED.total_cost_in_millicents IS NOT NULL AND EXCLUDED.total_cost_in_millicents != 0
+            THEN EXCLUDED.total_cost_in_millicents
+            ELSE ${spans.totalCostInMillicents}
+          END
+        `,
+        documentUuid: sql`COALESCE(${spans.documentUuid}, EXCLUDED.document_uuid)`,
+        distinctId: sql`COALESCE(${spans.distinctId}, EXCLUDED.distinct_id)`,
+        commitUuid: sql`COALESCE(${spans.commitUuid}, EXCLUDED.commit_uuid)`,
+        updatedAt: new Date(),
+      },
```
geclos, line 315: the meat is here, basically merging duplicated spans via "on conflict" postgres handler and a bunch of sql rules

packages/web-ui/src/ds/molecules/Chat/Message/index.tsx
```diff
@@ -65,7 +65,7 @@ export function MessageItem({
           className='flex-shrink-0 bg-muted w-1 rounded-lg cursor-pointer hover:bg-primary transition-colors'
           onClick={() => setCollapseMessage(!collapsedMessage)}
         />
-        <div className='flex flex-1 flex-col gap-1'>
+        <div className='flex flex-grow flex-col gap-1 overflow-x-auto'>
```
andresgutgon, line 68: Be aware that this component is also used in shared prompts. Maybe check if it looks ok

packages/core/src/services/traces/bulkCreateTracesAndSpans.ts
andresgutgon: A lot of meat here 🤯

andresgutgon (dismissed):

packages/web-ui/src/ds/molecules/Chat/Message/index.tsx
geclos: reviewed and it looks good

PR #747, opened 2024-12-19
Revert document changes UI

csansoon:
The first of 4 actions for the History page:
- [x] Revert changes from a specific document
- [ ] Reset document to a specific version
- [ ] Revert all changes from a commit
- [ ] Reset project to a commit

All 4 of these actions will reuse the same Modal (`<HistoryActionModal />`). For this reason, the Modal exists in the root `<ProjectChanges>` component, and the rest handle the state for this modal through a Context and Provider (sorry @andresgutgon)

With this action you can revert changes from any commit (including the current draft). Removing a change will add the required changes to the current draft, if the user is in one, or a new one if it is not.

https://github.com/user-attachments/assets/45598d65-17ed-411b-976d-e7f611137e99


andresgutgon (approved):
Awesome!

PR #751, opened 2024-12-23
Projects lastEditedAt field

csansoon:
Because of how `documentVersions` work, recalculating changes in a draft will make the "last edited at" field in the project list display an update. To fix that, I added a new actual `projects.lastEditedAt` field, which gets updated when:
- document is created
- document is updated
- draft is created
- draft is merged
- draft is deleted
- evaluation is connected

https://github.com/user-attachments/assets/43d2578c-6b9c-4ed7-9f78-fbfd6ae76c8c


andresgutgon (approved):
Missing test to ensure ping is called. Other than that nice 👍

PR #753, opened 2024-12-27
Undo commits

csansoon:
My final two PRs regarding Project History Actions:
- Revert whole commits
- Reset project to the state it had in any previous commit

:D


andresgutgon (commented):
Looks good but there is too much code in the actions and no tests

Actions are like controllers in MVC. It should be thin, and the majority of the code should be in a service where tests are done. Actions are just gatekeepers to validate input data and permissions.

apps/web/src/actions/history/resetCommitVersion/resetCommitVersionAction.ts
```diff
@@ -0,0 +1,89 @@
...
+      targetDraftUuid: z.string().optional(),
+      commitUuid: z.string(),
+    }),
+  )
+  .handler(async ({ input, ctx }) => {
+    const { workspace, project } = ctx
+    const { targetDraftUuid, commitUuid } = input
+
+    const commitScope = new CommitsRepository(workspace.id)
+    const headCommit = await commitScope
```
geclos, line 28: remember a project can have no head commit (new projects start with no merged commits)

geclos (dismissed):
As @andresgutgon suggests move code from actions to services and add integration tests please

apps/web/src/actions/history/resetCommitVersion/resetCommitVersionAction.ts
csansoon: That's fine, you cannot reset to a draft. This restriction is applied in the UI, so calling it would require to break it (which users may expect broken logic too)

andresgutgon (approved):

PR #755, opened 2024-12-30
Add Azure and Google custom configs to docs

neoxelox:


packages/core/src/services/ai/index.ts
```diff
@@ -121,8 +121,8 @@ export async function ai({
       messages,
       provider,
       apiKey,
+      url: url ?? undefined,
```
andresgutgon, line 124: This is not the same as before. Make sure passing undefined for url is not having side effects

neoxelox: Why it is not the same? In the function that parameter is optional

andresgutgon: Now is always passing `url: undefined` before was passing nothing. Probably does not matter at all. It depneds on how is used url under the hood

neoxelox: I understand that but I am calling a function with an optional parameter, it should be undefined in both cases

Issue #684, opened 2024-12-02
Python SDK


geclos:
Port our typescript SDK to python (only v2)

PR #757, opened 2025-01-02
Add Python SDK tests

neoxelox:
Related to: https://github.com/latitude-dev/latitude-llm/issues/684

### DIFF
--- a/packages/sdks/python/tests/prompts/chat_test.py
+++ b/packages/sdks/python/tests/prompts/chat_test.py
@@ -1,0 +1,302 @@
+import json
+from typing import List, cast
+from unittest.mock import Mock
+
+import httpx
+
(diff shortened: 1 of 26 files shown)


packages/sdks/python/tests/prompts/chat_test.py
```diff
@@ -0,0 +1,302 @@
...
+
+        [
+            self.assert_requested(
+                request,
+                method="POST",
+                endpoint=endpoint,
+                body={
+                    "messages": [json.loads(message.model_dump_json()) for message in messages],
+                    "stream": options.stream,
+                    "__internal": {"source": "api"},
```
andresgutgon, line 76: Don't you have a constant for source?

neoxelox: Yeah, true, I could have used that. In the next PR to change the docs I will change these 👍

Issue #598, opened 2024-11-12
Tool calling epic


cesr:
- We need a way to paste tool responses in the playground chat to test this type of prompt
- When the LLM uses a tool, it should generate an input for the user to add the tool response (or mock it)

andresgutgon:
This has to work also with batch calls. When user run a prompt in batch the user should have the option to specify if the function call responses will be AI generated (mocking) or the user will answer manually each time the prompt call one of the functions

PR #758, opened 2025-01-02
Handle tool calls in playground

andresgutgon:
# What?
We have a task to allow users to give a response on one or more tool calls that the AI can request in our playground. The thing is that the work necessary to make this work is also essential for making tool calling work on our SDK and Gateway. So this PR is a spike to have a picture of what are the moving parts necessary to make this happen.

image

## Related issue
https://github.com/latitude-dev/latitude-llm/issues/598

### Possible affected clients
These are the places I think these changes affect
1. Web app playground
2. Running prompts in batch (from playground)
3. Users using our SDK clients
4. Users using our API

## TODO
- [x] Serialize chain AST and scope [Done here](https://github.com/latitude-dev/promptl/pull/11)
- [x] Store in a cache the serialized chain so it can be resumed
- [x] Generate a unique identifier of a SDK run so users can reference it when they call the tool. Can be this the `documentLogUuid`?
- [x] Present users with a UI in the playground to introduce tool call response
- [x] Send user's tool call response to gateway and resume previous playground session from cache. 
- [x] Do the resume conversation logic inside `addMessage` if a cache chain exists for that message and there are tool calls. 
- [x] Removed cached in redis chain when conversation is successfully resumed and completed.
- [x] ~~Shared prompts tool calling story 💀~~ Don't allow tool calling responses in shared prompts UI cc @cesr 
- [x] In the playground, send multiple tool calls at the same time when using chat form and some of the tools don't have a response.
- [x] INSTALL new promptl version
- [x] QA  ⚠️ 

### Tool calls when running prompt in Batch in Playground 🧠 🧠 🧠 🤯 
How we can do this?

### In SDK
Talked with @csansoon this is how it would look to have tool calls in SDK
```javascript
const response = sdk.run(
  'weather_experet', 
  { location: "Barcelona" }, 
  { 
    tools: [
      // Same name as in schema
      get_weather: (location: string) => {
         // Call some API or DB and get weather in "loc"
         return { temperature: 23 }
      }
    ]
  }
)
```

## Considerations
- [ ] AI can respond with multiple tool calls at the same time. So maybe we need to handle it at the SDK level as a `Promise.all` and wait for all the calls to be finished before continuing.
- [ ] What happens if the work to be done in one or more of the tool calls is heavy and needs to be put in a queue in the user's system? Do you think we should pass the reference to the cached partial response (serialized chain) in the tool calls?


apps/web/src/actions/sdk/runDocumentAction.ts
```diff
@@ -28,6 +32,7 @@ export async function runDocumentAction({
   parameters,
 }: RunDocumentActionProps) {
   const { workspace, user } = await getCurrentUserOrError()
+  let tools = []
```
andresgutgon, line 35: Requests to SDK in our playground are stateless. We need to implement a cache to remember where the user stopped because there were one o more tool call requested by the AI

packages/core/src/services/chains/run.ts
```diff
@@ -305,6 +313,8 @@ async function runStep({
     } else {
       streamConsumer.stepCompleted(response)
 
+      // TODO: Here instead of running the state if the step has
+      // toolCalls we serialize the chain up until this point
```
andresgutgon, line 317: When the AI is cached or when is fresh in both cases the step response can be `stepCompleted` or `chainCompleted` and we're doing the same. 

Now we also want to handle tool calling and chain serialization so I think we need to DRY this in to a new unified method that handle all the cases

apps/gateway/package.json
```diff
@@ -30,7 +30,7 @@
     "drizzle-orm": "^0.33.0",
     "hono": "^4.6.6",
     "lodash-es": "^4.17.21",
-    "promptl-ai": "^0.3.5",
+    "promptl-ai": "file:/Users/andres/code/latitude/promptl",
```
andresgutgon, line 33: ⚠️ PUBLISH the promptl package and use new version before merging this

apps/web/next.config.mjs
```diff
@@ -39,7 +39,7 @@ const nextConfig = {
       bodySizeLimit: '25mb',
     },
   },
-  webpack: (config) => {
+  turbo: (config) => {
```
andresgutgon, line 42: This remove a warning in the nextjs logs and looks it's working. We're no longer using Webpack 🎉

apps/web/package.json
```diff
@@ -51,7 +51,7 @@
     "oslo": "1.2.0",
     "pdfjs-dist": "^4.9.155",
     "posthog-js": "^1.161.6",
-    "promptl-ai": "^0.3.5",
+    "promptl-ai": "file:/Users/andres/code/latitude/promptl",
```
andresgutgon, line 54: ⚠️

geclos:
This API does not follow the pattern of other methods in the SDK:

```
const response = sdk.run(
  'weather_experet', 
  { location: "Barcelona" }, 
  { 
    tools: [
      // Same name as in schema
      get_weather: (location: string) => {
         // Call some API or DB and get weather in "loc"
         return { temperature: 23 }
      }
    ]
  }
)
```

Keep in mind sdk format follows this format: required arguments as positional + last _options_ argument. This means that all required arguments are positional and then any optional arguments go in a last _options_ arguments that is an object. In this case I imagine this would look something like:

```
const response = sdk.run(
  'weather_experet', 
  { 
   parameters: { location: 'barcelona'},
    tools: [
      get_weather: (location: string) => {
         return { temperature: 23 }
      }
    ]
  }
)
```

geclos:
> When stopping the chain because there are tool calls. Emit a new event or emit chainCompleted? ask @geclos and @csansoon for opinions. Alex voted for a new, event

Isn't this the same event than when a step is completed? Since a tool call request wouldbe the response from a step and providing the response would start a new step in the chain.

andresgutgon:
> Isn't this the same event than when a step is completed?

If we emit stepCompleted the stream is not stopped. I'm talking in our playground. For that I send `chainCompleted` to stop execution in the UI. 

This works but maybe is not a good idea

packages/core/src/services/chains/buildStep/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    finalResponse: ChainStepResponse<StreamType>
+  }): Promise<ChainStepResponse<StreamType>> {
+    const isPromptl = chain instanceof PromptlChain
+    const hasTools = isPromptl && hasToolCalls({ response: finalResponse })
+
+    if (hasTools) {
+      await cacheChain({ workspace, chain, documentLogUuid })
+    }
+
+    if (step.chainCompleted || hasTools) {
```
andresgutgon, line 75: I'm not sure about this. This is a breaking behaviour. 

To keep compatibility with existing code and also to ship this only for playground I think I'll change this to one of these 2 options:

1. Emit a new event `ChainStepWithToolCall` and stop the playground UI stream
2. Check on step completed event if there are toolCalls and stop the playground UI. 

All this file is a refactor to DRY code that was happening when:
1. Prompt run was cached
2. Prompt run was not cached.

But in terms of behaviour change I think this is the only change.

packages/core/src/services/chains/chainCache/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  workspace,
+  documentLogUuid,
+  chain,
+}: {
+  workspace: Workspace
+  chain: PromptlChain<Message>
+  documentLogUuid: string
+}) {
+  const key = generateCacheKey({ documentLogUuid, workspace })
+  const serialized = chain.serialize()
```
andresgutgon, line 64: [chain.serialize()](https://github.com/latitude-dev/promptl/pull/11) is a new method introduced in that PR

apps/web/src/actions/sdk/runDocumentAction.ts
```diff
@@ -50,6 +50,15 @@ export async function runDocumentAction({
...
   >()
+  // TODO: Maybe use this same action for 2 different cases:
+  // 1. Run a document
+  // 2. Continue running a document when the user provide one or more tool calls
+  //
+  // In both cases it generates the stream and the UI keeps the chat messages present in the stream
+  // and add more messages to the same stream session.
+  //
+  // For doing this we need a new gateway endpoint to send the tool calls to the server and run the chain
+  // also the SDK method to be called here
```
andresgutgon, line 61: I'll remove this comment. For now I opted for a different action. Let's see how it feels.

packages/core/src/helpers.ts
```diff
@@ -61,25 +63,53 @@ export function buildCsvFile(csvData: CsvData, name: string): File {
...
+    role: MessageRole.assistant,
+    content: [] as MessageContent[],
+    toolCalls: [],
+  }
+  if (!data) return undefined
+
+  const text = data.text
+  const object = type === 'object' ? data.object : undefined
+  const toolCalls = type === 'text' ? (data.toolCalls ?? []) : []
+  let content: MessageContent[] = []
```
neoxelox, line 95: nit: you can just avoid the content variable and push directly to the message.content

packages/core/src/services/chains/ChainValidator/index.ts
```diff
@@ -76,31 +76,56 @@ const getOutputType = ({
...
+      let prevText = getTextFromMessages(prevContent)
       const { completed, conversation } = await (chain as LegacyChain).step(
         prevText,
       )
       return Result.ok({ chainCompleted: completed, conversation })
     }
+
     const { completed, messages, config } = await (chain as PromptlChain).step(
-      prevText,
+      prevContent as PromptlMessage[] | undefined | string,
```
andresgutgon, line 120: We now can add to promptl an array of messages as the previous response.

packages/core/src/helpers.ts
```diff
@@ -88,18 +130,74 @@ export function buildConversation(providerLog: ProviderLogDto) {
...
+    const toolResponseContents = toolCallResponses.map((toolCallResponse) => {
+      return {
+        type: ContentType.toolResult,
+        toolCallId: toolCallResponse.id,
+        toolName: toolCallResponse.name,
+        result:
+          typeof toolCallResponse.result === 'string'
+            ? JSON.parse(toolCallResponse.result)
+            : toolCallResponse.result,
+        isError: toolCallResponse.isError || false,
```
andresgutgon, line 146: I don't like the fact that we're building response message with the types of old compiler. But I don't want to change this now.

We should do a PR killing all compiler types and start using prompt ASAP

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/useMessages.ts
```diff
@@ -0,0 +1,75 @@
...
+function getUnrespondedToolRequests<V extends PromptlVersion>({
+  version: _,
+  messages,
+}: {
+  version: V
+  messages: VersionedMessage<V>[]
+}) {
+  // FIXME: Kill compiler please. I made this module compatible with
+  // both old compiler and promptl. But because everything is typed with
+  // old compiler prompts in promptl version are also formatted as old compiler
```
Comment (csansoon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/useMessages.ts:24]: I second this.

packages/web-ui/src/ds/molecules/Chat/ChatTextArea/ToolCallForm/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+              responded
+            </>
+          ) : (
+            <>
+              <Badge variant='muted'>{currentToolRequest}</Badge> of{' '}
+              <Badge variant='muted'>{totalToolRequests}</Badge> tools to be
+              responded
+            </>
+          )}
+        </Text.H6>
```
csansoon, line 89: Now that this is the first thing you read on the Tool Response input, since it is the top text, I'd move the (?) Tooltip here instead, to explain what do "tools to be responded" mean.

packages/web-ui/src/ds/molecules/DocumentTextEditor/Editor/RegularEditor.tsx
```diff
@@ -34,7 +34,8 @@ export function RegularMonacoEditor({
   const { monacoRef, handleEditorWillMount } = useMonacoSetup({ errorFixFn })
 
   const [defaultValue, _] = useState(value)
-  const [isEditorMounted, setIsEditorMounted] = useState(false) // to avoid race conditions
+  // to avoid race conditions
+  const [isEditorMounted, setIsEditorMounted] = useState(false)
```
csansoon, line 38: xd

csansoon (dismissed):
lgtm!

PR #764, opened 2025-01-08
Fix tool calls in chat

neoxelox:
`chain-complete` was returning tool calls stringified as a content type `text` instead of content type `tool-call`

```json
{
  "type": "chain-complete",
  "config": {
    "provider": "OpenAI",
    "model": "gpt-4o-mini",
    "temperature": 0.1,
    "tools": {
      "get_the_weather": {
        "description": "Retrieves the current weather for a specified location.",
        "parameters": {
          "type": "object",
          "properties": {
            "location": { "type": "string", "description": "The city and country, e.g., 'Valencia, Spain'." }
          },
          "required": ["location"],
          "additionalProperties": false
        }
      }
    }
  },
  "response": {
    "streamType": "text",
    "text": "",
    "usage": { "promptTokens": 110, "completionTokens": 17, "totalTokens": 127 },
    "toolCalls": [
      {
        "id": "redacted",
        "name": "get_the_weather",
        "arguments": { "location": "Barcelona, Spain" }
      }
    ]
  },
  "messages": [
    {
      "role": "assistant",
      "toolCalls": [
        {
          "id": "redacted",
          "name": "get_the_weather",
          "arguments": { "location": "Barcelona, Spain" }
        }
      ],
      "content": "[\n  {\n    \"type\": \"tool-call\",\n    \"toolCallId\": \"redacted\",\n    \"toolName\": \"get_the_weather\",\n    \"args\": {\n      \"location\": \"Barcelona, Spain\"\n    }\n  }\n]"
    }
  ],
  "uuid": "redacted"
}
```

It also seems we were eating away a text message that came along with the tool call request:
image

There was also a bug that made us stuck when there was an error in the streaming chat:
Screenshot 2025-01-08 at 17 33 47

Aaaand also fixed this error that appeared because we have tool calls defined in various fields with different arguments: `toolCalls`, inside content but with `toolArguments` and inside content too but with `args`:

Screenshot 2025-01-08 at 17 31 56


packages/core/src/services/providerLogs/addMessages.ts
```diff
@@ -174,23 +171,24 @@ async function iterate({
...
       }),
       saveSyncProviderLogs: true,
     })
 
-    const text = parseResponseText(response)
+    const response = { ..._response, providerLog }
+
     ChainStreamConsumer.chainCompleted({
       controller,
-      response: { ...response, providerLog, text },
```
neoxelox, line 185: Here I'm just following what we do in `runChain`, this extra `parseResponseText(response)` is not needed, as it is already done in the `ChainStreamConsumer`

neoxelox: Update, now, no more `parseResponseText(response)` 😎

packages/core/src/services/providerLogs/addMessages.ts
```diff
@@ -23,12 +23,50 @@ import { ChainError } from '../chains/ChainErrors'
...
+          toolCallId: toolCall.id,
+          toolName: toolCall.name,
+          args: toolCall.arguments,
+        } as ToolRequestContent
+      }),
+      toolCalls: providerLog.toolCalls,
+    })
+  }
+
+  return messages
```
andresgutgon, line 67: This is similar to what we have in the other place no?

neoxelox: Yes it is similar

PR #769, opened 2025-01-14
Fix drizzle

csansoon:
Fixes Drizzle not able to import `@latitude-data/compiler`, which makes it unable to open the studio or generate migrations

### DIFF
--- a/packages/core/src/helpers.ts
+++ b/packages/core/src/helpers.ts
@@ -2,11 +2,9 @@ import { z } from 'zod'
 import { CsvData, ParameterType } from './constants'
 import { ProviderApiKey, ProviderLogDto } from './schema/types'
 
-import {
-  ContentType,
+import type {
   Message,
   MessageContent,
-  MessageRole,
   ToolRequestContent,
 } from '@latitude-data/compiler'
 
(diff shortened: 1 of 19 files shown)


packages/core/src/helpers.ts
```diff
@@ -2,11 +2,9 @@ import { z } from 'zod'
...
 import { ProviderApiKey, ProviderLogDto } from './schema/types'
 
-import {
-  ContentType,
+import type {
   Message,
   MessageContent,
-  MessageRole,
   ToolRequestContent,
 } from '@latitude-data/compiler'
```
csansoon, line 9: The actual issue

packages/core/src/helpers.ts
```diff
@@ -67,21 +65,21 @@ export function buildConversation(providerLog: ProviderLogDto) {
...
         {
-          type: ContentType.text,
+          type: 'text',
           text: providerLog.response,
         },
       ],
       toolCalls: [],
-    }
+    } as Message
   }
```
csansoon, line 77: I hate this.

packages/core/src/helpers.ts
neoxelox: lollipop

andresgutgon (approved):

neoxelox (approved):

PR #774, opened 2025-01-16
Autonomous Agents behaviour prototype

csansoon:
First prototype for Autonomous Agents.

In this PR, the following feature has been added:
- When a prompt has a `type: agent` configuration, a new `agent_finish_workflow` tool will be injected into the prompt, which will run an indefinite amount of steps until this tool has been called.
- The `schema` configuration from the prompt is repurposed in agents, where it only affects to the schema of the finish tool, instead of the response format for each intermediate step.
- If an Agent prompt already contains user-defined steps, all of those are executed without the return tool injected. The tool is only available for the autonomous steps executed by the agent.

https://github.com/user-attachments/assets/5d0bbec0-7ca0-4957-a7df-42045fe89b24

(Note on video: First attempt was blazing fast because of response caching. I removed it by setting a temperature)

### DIFF
--- a/packages/core/src/services/chains/agents/run.ts
+++ b/packages/core/src/services/chains/agents/run.ts
@@ -1,87 +1,53 @@
-import { Chain as LegacyChain } from '@latitude-data/compiler'
-import { RunErrorCodes } from '@latitude-data/constants/errors'
-import { Chain as PromptlChain } from 'promptl-ai'
-
-import { ProviderApiKey, Workspace } from '../../browser'
+import {
+  runChain,
+  SomeChain,
+  RunChainArgs,
+  ChainResponse,
+  CachedApiKeys,
+  createChainRunError,
+} from '../run'
+import { generateUUIDIdentifier } from '../../../lib/generateUUID'
+import {
+  LatitudeChainCompleteEventData,
+  LogSources,
+} from '@latitude-data/constants'
+import { ErrorableEntity } from '../../../constants'
 import {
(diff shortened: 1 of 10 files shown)


packages/core/src/services/chains/agents/run.ts
```diff
@@ -0,0 +1,331 @@
...
+      providersMap,
+      controller,
+      previousCount: messageCount,
+      previousResponse: response,
+    })
+  } catch (e: unknown) {
+    const error = streamConsumer.chainError(e)
+    throw error
+  }
+}
```
andresgutgon, line 331: In this PR or in the next this file need some testing

packages/core/src/services/chains/agents/run.ts
```diff
@@ -0,0 +1,331 @@
...
+        stepStartTime,
+        errorableUuid,
+        response: _response,
+      }),
+      saveSyncProviderLogs: true, // TODO: temp bugfix, shuold only save last one syncronously
+    })
+
+    const response = { ..._response, providerLog }
+
+    await setCachedResponse({
```
andresgutgon, line 301: The cache is the same for a chain and for an agent. We need to store the same, right?

andresgutgon (approved):

packages/core/src/services/chains/agents/run.ts
csansoon: Yeah, just input config, messages, and the response.

PR #779, opened 2025-01-17
maxSteps config for chains and agents

csansoon:
Defaults to 20 and has an absolute maximum of 150.

image


packages/core/src/schema/models/runErrors.ts
```diff
@@ -20,6 +20,7 @@ export const errorCodeEnum = latitudeSchema.enum('run_error_code_enum', [
   RunErrorCodes.EvaluationRunUnsupportedResultTypeError,
   RunErrorCodes.EvaluationRunResponseJsonFormatError,
   RunErrorCodes.DefaultProviderInvalidModel,
+  RunErrorCodes.MaxStepCountExceededError,
```
andresgutgon, line 23: Migration!

csansoon: done

andresgutgon (approved):

PR #783, opened 2025-01-20
Fix: error thrown when there are no changes to revert

csansoon:


packages/core/src/services/commits/computeRevertChanges.ts
```diff
@@ -191,9 +191,5 @@ export async function computeChangesToRevertCommit(
     targetDraftDocuments: documentsInDraft.value,
   })
 
-  if (changesToRevert.length === 0) {
-    return Result.error(new ConflictError('No changes to revert'))
```
andresgutgon, line 195: I think here you can throw the error and will arrive to `onError` Can you check?

csansoon: It does arrive to the `onError` handler, yes, but the sentry notification would already have been sent

geclos:
@csansoon should we close this?

geclos:
closing

PR #784, opened 2025-01-20
Run documents in batch when the prompt has tool calls

andresgutgon:
# What?
We implemented tool calling mocking in our playground. But we also need to handle tool calling when the running of the prompt is in batch

### TODO
- [x] Create a document for gen tool responses in development workspace
- [x] Implement AI service to auto-generate tool responses based on tool calls
- [x] Handle tool calling when running in batch from the playground
- [x] Handle tool calling when running an evaluation in batch
- [x] ~~Provision missing environment variables in production~~ I'll postpone from now. I did a default variable if the env is not there.
- [x] Create a document for gen tool responses in the production workspace. Copy from dev

### Next PR
- [ ] Tests for run batch service when tool calls


packages/core/src/jobs/job-definitions/documents/runDocumentAtCommit/generateToolResponseMessages.ts
```diff
@@ -0,0 +1,85 @@
...
+  const response = responseResult.value
+  if (response.streamType === 'text') {
+    return Result.error(
+      new Error(
+        'Invalid response. AI responed with text while generating a mock tool response',
+      ),
+    )
+  }
+
+  const toolCallResponses = response.object as ToolCallResponse[]
```
andresgutgon, line 73: Validate that the returned object is like an array of `ToolCallResponse.` AI could hallucinate

csansoon (approved):

PR #792, opened 2025-01-22
Fix old OpenAI old O1 family models not supporting system messages

andresgutgon:
# What?
Using `o1-mini` when we send `system` messages VercelSDK [remove those messages](https://github.com/vercel/ai/blob/main/packages/openai/src/openai-chat-language-model.ts#L957) so we convert system messages to `user` messages.

Otherwise OpenAI receives an empty array
image

## Remove non-streaming hack
This was needed before we updated to vercel SDK 1.1.0 which now has support for `o1` family models in streaming mode


docs/guides/prompt-manager/provider-rules/openai.mdx
```diff
@@ -0,0 +1,24 @@
...
+---
+
+1. [System messages are not supported in old O1 family](#system-messages-are-not-supported-in-old-o1-family)
+
+### System messages are not supported in old O1 family
+
+You can read more about the OpenAI message format in the [OpenAI documentation](https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages).
+But basically the idea in old reasoning models of the `o1` family like `o1-preview` or `o1-mini` System messages are not supported. Instead you use `<user>` or `<assistant>` messages. Latitude replace the system messages with user messages. You always can wrap a text with `<assistant>` tags to get assistant messages.
+
+<Info>Official "o1" model works fine</Info>
```
csansoon, line 13: Improve texts and 👌

csansoon (dismissed):

PR #794, opened 2025-01-22
Handle agent tool calls in SDK and playground

csansoon:
Implement chain's tool calling behaviour for Agents, both for SDK and Playground.

Additionally, modified messages UI to be more understandable at a glance. Both tool request and tool responses got their own container. The final Agent Response is shown as a variation of a Tool Request (which makes it the most consistent to how it actually works).

image


packages/core/src/services/chains/agents/AgentStepValidator/index.ts
```diff
@@ -130,15 +130,20 @@ export const validateAgentStep = async (
   })
   if (freeQuota.error) return freeQuota
 
+  const messages = conversation.messages
+  const assistantResponse = messages[messages.length - 1]
+
+  if (Array.isArray(prevContent)) {
+    const [_assMessage, ...toolResponses] = prevContent
```
andresgutgon, line 137: lol `assMessage` 🦋

packages/core/src/services/chains/agents/AgentStepValidator/index.ts
```diff
@@ -130,15 +130,20 @@ export const validateAgentStep = async (
...
+    messages.push(...toolResponses)
+  }
```
andresgutgon, line 139: What is doing this code? I would put in function with a name

packages/core/src/services/documentLogs/addMessages/index.ts
```diff
@@ -14,6 +25,33 @@ type CommonProps = {
...
+        type: ContentType.toolResult,
+        toolName: AGENT_RETURN_TOOL_NAME,
+        toolCallId: agentFinishToolCall.id,
+        isError: false,
+        result: {},
+      },
+    ],
+  }
+
+  return [agentFinishToolResult]
```
andresgutgon, line 52: Yep, a bit weird that this module knows about agents

packages/core/src/services/providerLogs/addMessages.ts
```diff
@@ -73,11 +71,6 @@ export async function addMessages({
     responseResolve = resolve
   })
 
-  messages = [
-    ...buildConversation(serializeProviderLog(providerLog)),
-    ...messages,
-  ]
-
```
andresgutgon, line 80: All of this, why is it disappearing?

packages/web-ui/src/ds/molecules/Chat/Message/index.tsx
```diff
@@ -224,91 +200,160 @@ const Content = ({
...
+                }
+                ellipsis
+              >
+                {value.toolCallId}
+              </Text.H6>
+            </TruncatedTooltip>
           </div>
+          <CodeBlock language='json'>
+            {JSON.stringify(value.result, null, 2)}
+          </CodeBlock>
```
andresgutgon, line 343: This switch case is super big. Extract both tool request and tool response in other components

packages/web-ui/src/ds/molecules/Chat/Message/index.tsx
```diff
@@ -224,91 +200,160 @@ const Content = ({
...
-    case ContentType.toolResult:
-      return (
-        <div key={`${index}`} className='py-2 w-full'>
-          <div className='overflow-hidden rounded-xl w-full'>
-            <CodeBlock language='json'>
-              {JSON.stringify(value as ToolContent, null, 2)}
-            </CodeBlock>
+  if (value.type === ContentType.toolResult) {
+    return (
+      <div key={`${index}`} className='py-2 w-full'>
```
andresgutgon, line 303: When extracting it don't pass `index` put `key` in the extracted component

andresgutgon (dismissed):

packages/core/src/services/chains/agents/AgentStepValidator/index.ts
csansoon: The `prevContent` contains both assistant and tool messages. The `messages` array also already contains the assistant message, so I separate them and push only the tool responses to that array.

packages/core/src/services/providerLogs/addMessages.ts
csansoon: I moved it to the parent service

packages/core/src/services/documentLogs/addMessages/index.ts
csansoon: Oh, it will know more in the next PR. The "addMessage" will be the one that will decide if adding a message should just run the AI once with the new message, or if it has to start a new agent workflow with the new message, depending on whether the prompt is an agent or not.

Same as `runDocumentAtCommit`, where it needs to know if it has to run a regular chain or an agent.

PR #798, opened 2025-01-23
Simplify chain streaming workflow

csansoon:
We have three (3) different workflows where AI generates a response based on an array of messages and a config, and send events back through a controller: RunChain, RunAgent and AddMessage. These methods implemented all of that logic independently, although they have lots of common code.

In this PR, I applied DRY to the common steps in all of those workflows in the backend, and how events are managed in the frontend's Chat component.

The main DRY'd method is a new function called `executeStep`. This method takes care of all the required steps and events that must happen each time a step is executed, independently of the reason. First, a `chain-step` event must be sent. Then, the ai provider is called and all of those events are redirected through the stream. Finally, a `chain-step-complete` event must be sent, and the providerLog created. This is now all managed by this new method.


packages/core/src/services/chains/runStep/index.ts
```diff
@@ -0,0 +1,232 @@
...
+        finishReason: previousResponse?.finishReason ?? 'stop',
+      })
+
+      previousResponse!.chainCompleted = true
+      return previousResponse!
+    }
+
+    const nextStepCount = getNextStepCount({ stepCount, chain, step }).unwrap()
+
+    const response = await executeStep({
```
andresgutgon, line 158: We have here `runStep` and `executeStep`. It's fine but for someone new or we after sometime checking this code again maybe is a bit confuse. 

`executeStep` is related to calling the AI and Streaming. So maybe instead of executeStep something like `streamAIResponse`. Also the return value is the `response`

packages/core/src/services/agents/runStep/index.ts
```diff
@@ -0,0 +1,154 @@
...
+      chainCompleted: step.chainCompleted,
+      schema: step.schema,
+      output: step.output,
+    })
+
+    const toolCalls = getToolCalls({ response })
+
+    // Stop the chain if there are tool calls
+    if (toolCalls.length) {
+      await cacheChain({
```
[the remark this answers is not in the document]
csansoon: yess :)

packages/core/src/services/chains/runStep/index.ts
csansoon: Ohhh I like that name

PR #802, opened 2025-01-24
Chat with Agents

csansoon:
This PR adds the feature to chat with agents, and fixed some other bugs with the chain streaming workflow.

https://github.com/user-attachments/assets/70fdfc6c-b0a4-46fe-9518-55e0a0fdc60b

Now, the addMessage endpoint runs the `documentLogs/addMessage` service, which depending on context will chose between the following:
- `addChatMessage`: Adds the message to a finished prompt (or chain) and returns a new response.
- `resumeAgent`: Adds a message to an agent conversation, either paused or finished, and continues the autonomous flow.
- `resumePausedPrompt`: Adds tool responses to a cached chain.

Pausing agents may or may not result in a cached chain. If the hardcoded chain is paused before it has finished, it will be cached just as a regular chain. If the chain already finished and the agent pauses in an autonomous step, it is not cached.

Bugfixes:
- Sometimes system messages do not appear in the playground when they are included after a step that gets paused.
- An error is raised when an agent is stopped by a tool-call (although the client behaviour is unaltered).
- A client error is raised when a chain/prompt/agent execution fails in the playground. This will be revised in the future.
- The response from the last step of a chain was sometimes repeated when starting the agent workflow.

### DIFF
--- a/packages/core/src/services/documentLogs/addMessages/findPausedChain.ts
+++ b/packages/core/src/services/documentLogs/addMessages/findPausedChain.ts
@@ -1,10 +1,4 @@
 import { Workspace } from '../../../browser'
-import { Result } from '../../../lib'
-import {
-  CommitsRepository,
-  DocumentLogsRepository,
-  DocumentVersionsRepository,
-} from '../../../repositories'
 import { getCachedChain } from '../../chains/chainCache'
 
 export async function findPausedChain({
@@ -17,27 +11,8 @@ export async function findPausedChain({
   const cachedData = await getCachedChain({ workspace, documentLogUuid })
   if (!cachedData) return undefined
 
-  const logsRepo = new DocumentLogsRepository(workspace.id)
-  const logResult = await logsRepo.findByUuid(documentLogUuid)
-  if (logResult.error) return logResult
-
-  const documentLog = logResult.value
-  const commitsRepo = new CommitsRepository(workspace.id)
-  const commitResult = await commitsRepo.find(documentLog.commitId)
-  if (commitResult.error) return commitResult
-
-  const commit = commitResult.value
-  const documentsRepo = new DocumentVersionsRepository(workspace.id)
-  const result = await documentsRepo.getDocumentAtCommit({
-    commitUuid: commit?.uuid,
-    documentUuid: documentLog.documentUuid,
-  })
-  if (result.error) return result
-
-  return Result.ok({
-    document: result.value,
-    commit,
+  return {
     pausedChain: cachedData.chain,
     previousResponse: cachedData.previousResponse,
-  })
+  }
 }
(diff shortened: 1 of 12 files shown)


packages/core/src/services/documentLogs/addMessages/findPausedChain.ts
```diff
@@ -17,26 +12,7 @@ export async function findPausedChain({
...
-  const documentsRepo = new DocumentVersionsRepository(workspace.id)
-  const result = await documentsRepo.getDocumentAtCommit({
-    commitUuid: commit?.uuid,
-    documentUuid: documentLog.documentUuid,
-  })
-  if (result.error) return result
-
   return Result.ok({
-    document: result.value,
-    commit,
```
andresgutgon, line 39: So now who fetch document and commit?

packages/core/src/services/documentLogs/addMessages/index.ts
```diff
@@ -89,30 +28,61 @@ export async function addMessages({
...
+    documentUuid: documentLog.documentUuid,
+  })
+  if (documentResult.error) return documentResult
+  const document = documentResult.value
+
+  const providerLogRepo = new ProviderLogsRepository(workspace.id)
+  const providerLogResult =
+    await providerLogRepo.findLastByDocumentLogUuid(documentLogUuid)
+  if (providerLogResult.error) return providerLogResult
+  const providerLog = providerLogResult.value
```
andresgutgon, line 53: Why all this is now inside addMessages

packages/core/src/services/documentLogs/addMessages/resumeChain/index.ts
```diff
@@ -62,22 +57,15 @@ export async function resumeConversation({
   })
   const errorableUuid = documentLogUuid
 
-  // These are all the messages that the client
-  // already have seen. So we don't want to send them again.
-  const previousCount =
-    pausedChain.globalMessagesCount + responseMessages.length
```
andresgutgon, line 68: Why this change?

packages/core/src/services/documentLogs/addMessages/findPausedChain.ts
csansoon: The parent, `addMessages` service. It is required to decide which workflow to follow

packages/core/src/services/documentLogs/addMessages/index.ts
csansoon: All of this is context required to decide which workflow to follow.

packages/core/src/services/documentLogs/addMessages/resumeChain/index.ts
csansoon: The length of extra messages are added inside of the `runStep` processes. This was being added twice, which caused a bug not showing messages from between steps.
Specifically, this is being added in the `buildPrevContent` method, used by both chains and agents.

packages/core/src/services/documentLogs/addMessages/index.ts
andresgutgon: Extract into a named function with a name like `getDataFor....`

PR #803, opened 2025-01-24
Tools in Python SDK

neoxelox:
Closes: https://github.com/latitude-dev/latitude-llm/issues/799
Related to: https://github.com/latitude-dev/latitude-llm/issues/684

### DIFF
--- a/packages/sdks/python/src/latitude_sdk/sdk/prompts.py
+++ b/packages/sdks/python/src/latitude_sdk/sdk/prompts.py
@@ -20,17 +20,15 @@ from latitude_sdk.sdk.types import (
     ChainEvents,
     ChainEventStep,
     ChainEventStepCompleted,
-    ContentType,
     FinishedEvent,
     Message,
-    MessageRole,
     OnToolCall,
     OnToolCallDetails,
     Prompt,
     SdkOptions,
     StreamCallbacks,
     StreamEvents,
-    ToolCall,
+    StreamTypes,
     ToolMessage,
     ToolResult,
     ToolResultContent,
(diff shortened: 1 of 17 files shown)


packages/sdks/python/src/latitude_sdk/sdk/prompts.py
```diff
@@ -160,6 +170,73 @@ async def _handle_stream(
...
+
+        return list(tool_calls.values())
+
+    async def _handle_tool_calls(
+        self,
+        result: FinishedEvent,
+        tool_calls: List[ToolCall],
+        options: Union[RunPromptOptions, ChatPromptOptions],
+    ) -> Optional[FinishedEvent]:
+        assert options.tools is not None
```
neoxelox, line 201: This should not happen as this function is only run if the users sets tools, otherwise we just return the result with the answered tool calls without results from the llm.

However, I could defend this case with a custom raise ApiError to have a better error

packages/sdks/python/src/latitude_sdk/sdk/prompts.py
```diff
@@ -216,21 +309,24 @@ async def run(self, path: str, options: RunPromptOptions) -> Optional[RunPromptR
...
 
-                if options.on_finished:
-                    options.on_finished(FinishedEvent(**dict(result)))
+            if options.tools:
+                tool_calls = self._search_unanswered_tool_calls(result.conversation)
+                if tool_calls:
+                    try:
+                        # NOTE: The last sdk.chat called will already call on_finished
+                        final_result = await self._handle_tool_calls(result, tool_calls, options)
+                        return RunPromptResult(**dict(final_result)) if final_result else None
```
andresgutgon, line 322: What if there are more tool calls in another step? How is the recursivity done?

neoxelox: See `_handle_tool_calls`, at the end I do `next_result = await self.chat(result.uuid, tool_messages, options)`

csansoon (dismissed):
lgtm

PR #805, opened 2025-01-29
Code BuiltIn tool

csansoon:
### Latitude Tools
Added support for Latitude tools: A series of built-in tools that will get ran in Latitude's backend!

Currently available: `code`.

Latitude tools are defined as such:
```
---
provider: openai
model: gpt-4o-mini
latitudeTools:
  - code
---
```

Latitude tools can be combined with regular tools!

---
### Code built-in tool

The `code` Latitude tool uses CodeSandbox to execute custom python or javascript code in a secure and isolated environment. It even allows installing dependencies before execution!

### DIFF
--- a/packages/core/src/lib/chainStreamManager/index.ts
+++ b/packages/core/src/lib/chainStreamManager/index.ts
@@ -1,4 +1,4 @@
-import { Message, ToolCall } from '@latitude-data/compiler'
+import { Message, ToolCall, ToolMessage } from '@latitude-data/compiler'
 import { StreamEventTypes } from '@latitude-data/constants'
 import {
   ChainEvent,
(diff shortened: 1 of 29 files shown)


packages/core/src/lib/chainStreamManager/index.ts
```diff
@@ -154,6 +159,48 @@ export class ChainStreamManager {
...
+          toolMessage = buildToolMessage({
+            toolName: toolCall.name,
+            toolId: toolCall.id,
+            result: result,
+          })
+        } catch (error) {
+          toolMessage = buildToolMessage({
+            toolName: toolCall.name,
+            toolId: toolCall.id,
+            result: Result.error(error as Error),
```
neoxelox, line 191: isError should be true here

packages/core/src/services/builtInTools/definitions.ts
```diff
@@ -0,0 +1,74 @@
...
+          type: 'string',
+          description: 'The code to run.',
+        },
+        dependencies: {
+          type: 'array',
+          items: {
+            type: 'string',
+          },
+          description:
+            'An optional list of all the required dependencies to run the script. Adding dependencies will severely increate the execution time, so do not include them unless required.',
```
neoxelox, line 68:
```suggestion
            'An optional list of all the required dependencies to run the script. Adding dependencies will severely increase the execution time, so do not include them unless required.',
```

neoxelox (approved):

packages/core/src/lib/chainStreamManager/index.ts
neoxelox: I saw that you check inside if it is an error and you set isError accordingly.

packages/core/src/services/builtInTools/runCode/withDependencies.ts
```diff
@@ -0,0 +1,62 @@
...
+function getRunCommand({
+  language,
+  filename,
+}: {
+  language: SupportedLanguage
+  filename: string
+}) {
+  if (language === 'python') return `python ${filename}`
+  if (language === 'javascript') return `node ${filename}`
+  throw new Error(`Unsupported language: ${language}`)
```
andresgutgon, line 32: What type of run error message generates this? Do we create a log with error result in this case?

andresgutgon (approved):

packages/core/src/services/builtInTools/runCode/withDependencies.ts
andresgutgon: Got it

Issue #808, opened 2025-01-29
Change Python SDK tools callback signature


andresgutgon:
# What?
We have now this:
```typescript
tools: {
  get_coordinates: async ({ id, arguments: { location } }) => {
    const data = fetch('https://api.example.com/coordinates?location=' + location)
    const { latitude, longitude } = await data.json()
    return {
      id,
      name: 'get_coordinates',
      result: { latitude, longitude },
    }
  },
}
```

Reviewing with the team we think is easier to simplify this way

```typescript
tools: {
  get_coordinates: async ({ location }, { toolId, toolName }) => {
    const data = fetch('https://api.example.com/coordinates?location=' + location)
    const { latitude, longitude } = await data.json()
    return { latitude, longitude }
  },
}
```

PR #809, opened 2025-01-30
Change typescript sdk tool call callback signature

andresgutgon:
# What?
This simplifies the way users can define tool call callback functions in our Typescript SDK

## Issue
https://github.com/latitude-dev/latitude-llm/issues/808

### DIFF
--- a/packages/sdks/typescript/package.json
+++ b/packages/sdks/typescript/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@latitude-data/sdk",
-  "version": "1.1.1",
+  "version": "2.0.0",
   "description": "Latitude SDK for Typescript",
   "author": "Latitude Data SL <hello@latitude.so>",
   "license": "LGPL-3.0",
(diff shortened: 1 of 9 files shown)


packages/sdks/typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@latitude-data/sdk",
-  "version": "1.1.1",
+  "version": "2.0.0",
```
andresgutgon, line 3: Although no one is using yet tool calling from our sdk this change is a breaking change that deserves a major version bump. Also I think is nice to introduce tool calling with a new version

packages/sdks/typescript/src/utils/types.ts
```diff
@@ -147,9 +148,9 @@ export type ToolCallDetails = {
 
 export type ToolSpec = Record<string, Record<string, unknown>>
 export type ToolHandler<T extends ToolSpec, K extends keyof T> = (
-  toolCall: { id: string; name: K; arguments: T[K] },
+  toolCall: T[K],
   details: ToolCallDetails,
-) => Promise<void | ToolCallResponse>
+) => Promise<unknown>
```
csansoon, line 153: Niceee! :D

csansoon (approved):

PR #814, opened 2025-01-30
Search logs by custom identifier

neoxelox:
This PR allows to search logs by custom identifier.

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/index.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/index.tsx
@@ -48,7 +48,7 @@ export function DocumentLogFilters({
       <div className='max-w-40'>
         <Input
           placeholder='Custom identifier'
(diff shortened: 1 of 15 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/index.tsx
```diff
@@ -41,6 +45,13 @@ export function DocumentLogFilters({
         isDefault={filters.isLogSourcesDefault}
         reset={() => filters.onSelectLogSources(LOG_SOURCES)}
       />
+      <div className='max-w-40'>
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/index.tsx:48]: Do you need this? Ideally space should be distributed among filters

Comment (neoxelox) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/index.tsx:48]: This is to set a maximum value for the custom identifier search box, otherwise it tries to fill all the space. I tried to set the class "max-w-40" directly to the input, but it seems inside the input there is a FormField where the classes cannot be modified easily.

andresgutgon (dismissed):

Issue #806, opened 2025-01-29
Paywalls


andresgutgon:
# TODO

- [x] Change button in run counter on nav, Stripe link pass by Cesar
- [x] Hide copilot features on Open Source. Because it does not work.
- [x] If limit members overdue show indication in runs counter (Figma pending)
- [x] Add `LATITUDE_CLOUD` and `LATITUDE_CLOUD_PAYMENT_URL` env vars to display over limits and more.
- [x] Backoffice workspaces run/members over limits
- [x] If team plan overlimits button with contact us

andresgutgon:
Companies with more than 1 member and more than 10k runs in total 
And view runs group by month

andresgutgon:
PR: https://github.com/latitude-dev/latitude-llm/pull/821

andresgutgon:
PR: https://github.com/latitude-dev/latitude-llm/pull/826

PR #817, opened 2025-01-31
Implement overview of runs by workspace in backoffice

andresgutgon:
# What?
Have an overview of runs executed by the workspace. Part of this issue https://github.com/latitude-dev/latitude-llm/issues/806

## TODO
- [x] Implement usage query + tests
- [x] Use query in backoffice table

### DIFF
--- a/packages/core/src/services/workspaces/usageOverview/getUsageOverview.ts
+++ b/packages/core/src/services/workspaces/usageOverview/getUsageOverview.ts
@@ -1,0 +1,91 @@
+import { desc, sql, eq, max } from 'drizzle-orm'
+import { database, Database } from '../../../client'
+import { buildDocumentLogsCountersQuery } from './documentLogsCountersQuery'
+import { buildEvaluationResultsCountersQuery } from './evaluationResultsCountersQuery'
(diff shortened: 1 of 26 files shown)


packages/core/src/services/workspaces/usageOverview/getUsageOverview.ts
```diff
@@ -0,0 +1,92 @@
...
+    page: number
+    targetDate?: Date
+  },
+  db: Database = database,
+) {
+  const renewalDates = buildSubscriptionWithRenewalDatesQuery(targetDate)
+  const documentLogsCountersQuery = buildDocumentLogsCountersQuery(renewalDates)
+  const evaluationResultsCountersQuery =
+    buildEvaluationResultsCountersQuery(renewalDates)
+  const query = db
```
andresgutgon, line 32: This query is not the more performance it could be. Or maybe yes. The problem it has is that I want aggregated counters for documentLogs and evaluationResults grouped by workspace->subscription calculated start period. 

I did 2 subqueries for these 2 aggregations and another subquery to do the calculated renewal date following the JS implementation. 

I would love to hear the best ideas. If any I would like to deploy this and see if runs with our current data volume.

csansoon (approved):

PR #820, opened 2025-02-03
Update Dockerfile in all the apps

andresgutgon:
The reason for doing this is documented in this PR: https://github.com/latitude-dev/latitude-llm/pull/818 I just forgot to do the same and the rest of the deployment dockerfiles


apps/web/docker/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN apk add --no-cache libc6-compat curl
 FROM alpine as base
 
 # Will be used to cache pnpm store
-RUN corepack enable && corepack prepare pnpm@10.0.0 --activate
+RUN npm install -g corepack@0.31.0 && corepack enable
```
andresgutgon, line 16: So, I misunderstood the issue. It's related with corepack not with pnpm. So we keep current version of `pnpm` and we bump `corepack`. 

Alpine NodeJS 20 comes with `corepack@29.4.1` which has the problem explained here https://github.com/nodejs/corepack/issues/612 

So what we do is update manually to a version that does not break the build. 

Let's see if deploy works now 🤞

PR #821, opened 2025-02-03
Add extra usage info to header and introduce cloud env variable

andresgutgon:
# What?
More info on this issue: https://github.com/latitude-dev/latitude-llm/issues/806#issue-2817990792

## TODO
- [x] Disable copilot features on non cloud installations
- [x] Display amount of members in the workspace. 
- [x] Setup in infra `LATITUDE_CLOUD` and `LATITUDE_CLOUD_PAYMENT_URL` env variables
- [x] Fix usage member realtime/client update


csansoon (dismissed):

apps/web/src/stores/workspaceUsage.ts
```diff
@@ -22,6 +22,7 @@ export default function useWorkspaceUsage(opts?: SWRConfiguration) {
         return {
           ...prevData,
           usage: prevData.usage + 1,
+          members: prevData.members + 1,
```
andresgutgon, line 25: ⚠️ This is wrong, fix it but syncing with members store and update this store when a member is added

PR #822, opened 2025-02-03
New event types

csansoon:
## **WHAT**  
This PR introduces a new, robust, and extendable Event Types system for chain streaming.

This PR is getting too long, so it will not contain the new API version. Instead, all internal logic has been updated, but the current API handlers will still return the legacy events thanks to a dedicated translator.

## **WHY**  
Currently, chain streaming relies on three main Event Types:  
- `Step` (step started)  
- `StepComplete` (step completed)  
- `Complete` (chain completed)  

As we introduce more complex features, such as built-in tools, these events no longer provide enough information to keep users informed at all times. To address this limitation, we are defining a new Event Types system that is both structured and future-proof.  

## **HOW**  
We have designed a comprehensive set of events where each represents a distinct step in the execution workflow. Every event contains `messages` and `documentLogUuid`, along with other relevant data specific to the step.  

#### **Chain Execution Workflow**  
A chain follows this structured sequence:  
1. The chain starts  
2. All chains steps are executed
3. The chain completes  

Each chain step execution follows this flow: 
- The step starts, returning all compiled messages  
- The provider generates a response  
- If built-in tools are included in the response, they are executed  
- If client-defined tools are required, the chain halts and requests them from the user  

### **New Event Types**  
Here’s the complete list of possible events during a chain execution:  
- **`ChainStarted`** – The chain execution begins  
- **`StepStarted`** – A step starts, providing all new compiled messages
- **`ProviderStarted`** – Provider is now generating a response, and contains the configuration sent to the provider
- **`ProviderCompleted`** – Includes the provider's response, finish reason, and token usage  
- **`ToolsStarted`** – Built-in tools start executing, along with tool details and arguments  
- **`ToolCompleted`** – One per built-in tool, containing the tool’s response  
- **`StepCompleted`** – Marks the step as completed  
- **`ChainCompleted`** – Marks the chain as completed  

Additionally, the chain execution may be interrupted by:  
- **`ToolsRequested`** – Sent after `ToolCompleted` if client-defined tools are requested, pausing the chain  
- **`ChainError`** – Sent at any point if an error occurs, stopping execution  

### **Managing Events Efficiently**  
To streamline event management across different chain workflows (running chains, agents, resuming paused chains, handling chat messages, etc.), we have introduced the `ChainStreamManager`.  

This class is responsible for:  
- Defining and managing the Stream
- Sending all required events  
- Managing event-related data  

By consolidating this logic, we simplify the previously complex workflows, making event handling more efficient and maintainable.  

### **Maintaining legacy API versions** 
The `v1` and `v2` versions of the API must be respected for users that may currently be using them. Instead of duplicating the code and handling events in a different way depending on the API version, a "translator" has been added to easily convert the new events workflow into the legacy one. This works thanks to the new event system, which not only can be expanded upon, but also contains all necessary information to reconstruct the original one.

In order to avoid growing this PR even larger than it already is, no `v3` version will be added here. Instead, although all internal logic has been modified, the practical usage will stay the same for this PR, and updated into a new API version in the next one.


apps/gateway/src/routes/v2/conversations/chat/chat.handler.ts
```diff
@@ -52,8 +56,7 @@ export const chatHandler: AppRouteHandler<ChatRoute> = async (c) => {
     )
   }
 
-  const response = (await result.response).unwrap()
-  const body = runPresenter(response).unwrap()
+  const body = runPresenter(await response).unwrap()
```
andresgutgon, line 59: Now response always return a value right?

packages/core/src/jobs/job-definitions/documents/runDocumentAtCommitWithAutoToolResponses/respondToToolCalls.test.ts
```diff
@@ -93,25 +93,24 @@ describe('respondToToolCalls', () => {
...
-              },
-            ],
-          },
-        }),
-      ),
+      error: Promise.resolve(undefined),
+      lastResponse: Promise.resolve({
+        providerLog: { uuid: 'log1' },
+        object: {
+          tool_responses: [
```
andresgutgon, line 100: use camelCase. Everywhere we use camelCase

packages/core/src/lib/chainStreamManager/index.ts
```diff
@@ -0,0 +1,229 @@
...
+   */
+  start(cb?: () => Promise<void>): {
+    stream: ReadableStream<ChainEvent>
+    messages: Promise<Message[]>
+    toolCalls: Promise<ToolCall[]>
+    error: Promise<ChainError<RunErrorCodes> | undefined>
+    lastResponse: Promise<ChainStepResponse<StreamType> | undefined>
+  } {
+    if (this.controller) throw new Error('Chain already started')
+    if (this.finished) throw new Error('Chain already finished')
```
andresgutgon, line 68: Who catch these errors?

packages/core/src/lib/chainStreamManager/index.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    const [messages, resolveMessages] = usePromise<Message[]>()
```
andresgutgon, line 70: Small nick pick. Don't call it `useSOMETHING` look like frontend. Call it `buildPromise` or something like that

packages/core/src/lib/chainStreamManager/step/streamAIResponse.ts
```diff
@@ -1,84 +1,88 @@
...
+        completionTokens: 0,
+        totalTokens: 0,
+      },
+    }
   }
 
   const aiResult = await ai({
-    messages,
-    config,
+    messages: [...conversation.messages], // TODO: vitest will cry when checking the parameters passed to this function when the object mutes afterwards. To fix this, we make a deep copy of the array so that it is immutable.
```
andresgutgon, line 84: Who is mutating the list?

packages/core/src/services/agents/run.ts
```diff
@@ -1,166 +1,105 @@
...
-                  .response as ChainStepResponse<StreamType>
-              }
-            }
-          }
-
-          if (!done) return readNextChainEvent()
-
-          // Chain has been completed. Remove the cached chain if there was any.
-          deleteCachedChain({ workspace, documentLogUuid: errorableUuid })
+  const streamResult = chainStreamManager.start(async () => {
```
andresgutgon, line 36: Here you are throwing errors inside `.start`. I would always return a Result and do the `.unwrap()` here. Let's not put handled errors inside services. 

The less try/catch we have the better

packages/core/src/services/agents/runStep/index.ts
```diff
@@ -7,138 +7,106 @@ import {
...
+    chainStreamManager.done()
+    return
+  }
 
-      return response
-    }
+  const validStepCount = assertValidStepCount({ stepCount, step })
+  if (validStepCount.error) {
+    chainStreamManager.error(validStepCount.error)
+    return
```
andresgutgon, line 79: Don't we want to throw here?

packages/core/src/services/chains/run.ts
```diff
@@ -53,68 +60,69 @@ export type RunChainArgs<
...
-        })
-        .catch(async (e: ChainError<RunErrorCodes>) => {
-          const error = await createChainRunError({
-            error: e,
-            errorableUuid,
-            errorableType,
-            persistErrors,
-          })
 
-          responseResolve(Result.error(error))
```
andresgutgon, line 108: Make sure errors are stored associated with the documentLog as they did before

packages/core/src/services/commits/runDocumentAtCommit.ts
```diff
@@ -159,12 +157,11 @@ export async function runDocumentAtCommit({
...
         source,
-        response: response.value,
-        duration: await run.duration,
-        publishEvent: !response.error,
+        duration: await runResult.duration,
+        publishEvent: true,
       })
 
-      return response
+      return r
```
andresgutgon, line 164: Never 1 letter variables 🙏

andresgutgon (approved):
Small comments, nice work +1

packages/core/src/lib/chainStreamManager/index.ts
csansoon: These are errors for wrongly using the `chainStreamManager`, not from unhandled failures. The idea here is they should never happen, and if they do that's our fault

packages/core/src/lib/chainStreamManager/index.ts
csansoon: thought it was really similar to `useState`, so that made sense to me xd

packages/core/src/services/agents/runStep/index.ts
csansoon: That works too, yep

apps/gateway/src/routes/v2/conversations/chat/chat.handler.ts
csansoon: Oh, this has been fixed. It returns two promised elements: `error` and `lastResponse`. `error` is undefined if everything is correct, and `lastResponse` contains the last generated response, which could be undefined if there was an error in the way (so `lastResponse` is always defined if `error` is undefined)

packages/core/src/jobs/job-definitions/documents/runDocumentAtCommitWithAutoToolResponses/respondToToolCalls.test.ts
csansoon: Yeah, I didn't change this. Leaving it this way since that's how the original test was

PR #824, opened 2025-02-04
Try to improve usage query. Previous version was too much for the db

andresgutgon:
# What?
☝️ That

### DIFF
--- a/packages/core/src/services/workspaces/usageOverview/getUsageOverview.ts
+++ b/packages/core/src/services/workspaces/usageOverview/getUsageOverview.ts
@@ -1,23 +1,21 @@
-import { desc, sql, eq, max } from 'drizzle-orm'
+import { sql, count, desc, eq, and, gte, isNull, max } from 'drizzle-orm'
 import { database, Database } from '../../../client'
-import { buildDocumentLogsCountersQuery } from './documentLogsCountersQuery'
-import { buildEvaluationResultsCountersQuery } from './evaluationResultsCountersQuery'
-import { workspaceUsageInfoQuery } from './workspaceUsageInfoQuery'
-import { buildSubscriptionWithRenewalDatesQuery } from '../utils/calculateRenewalDate'
+import {
+  commits,
+  documentLogs,
+  evaluationResults,
+  evaluations,
+  projects,
+  runErrors,
+  workspaces,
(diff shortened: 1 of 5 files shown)


packages/core/src/services/workspaces/usageOverview/getUsageOverview.ts
```diff
@@ -1,88 +1,117 @@
...
+  )
 
-      // Two months ago period
-      twoMonthsAgoPeriodAt: max(documentLogsCountersQuery.twoMonthsAgoPeriodAt),
-      twoMonthsAgoPeriodRuns: sql<number>`SUM(
-        ${documentLogsCountersQuery.twoMonthsAgoPeriodCount} +
-        ${evaluationResultsCountersQuery.twoMonthsAgoPeriodCount}
-      )`.as('two_months_ago_period_runs'),
+  return db
+    .with(logsCTE, evaluationResultsCTE)
```
andresgutgon, line 96: What I did was do 2 CTE. 
As far as I can see we could do 3 things

### A) All in the same query
Do all the joins for getting `documentLogs` and `evaluationResults` in the same query and do the counts and sum those counts. 

### B) Do two subqueries
Do 2 subqueries. One for `documentLogs` and other for `evaluationResults`

### C ) Do two CTEs
Do the same as with subqueries but with CTEs

### My preference
I think (A) is the more efficient because all the tables are filtered by limit/offset. The problem with this solution when I tried it in the tests is that it duplicates rows, and the counters are all wrong. I didn't know how to make it work.

Doing (B) makes the aggregations for logs and results run N times for each workspace. Which is O(N) issue. This will choke the DB almost certainly. 

Doing (C) with CTEs is doing the aggregations but only once and then is used in the final query. Still I feel it will consume a lot of memory. Will see.

neoxelox: If you want to debug/test performance of queries before deploying themn to production, I would get the resulting SQL query from the drizzle builder and then EXPLAIN ANALYZE the query directly in the PROD DB... Otherwise is deploying and redeploying blindly

andresgutgon: Done offline. Looks good in prod

Issue #832, opened 2025-02-05
Update provider models


andresgutgon:
# What?

- [x] Review new models in supported providers.
- [x] From Latitude provider leave only `gpt-4o` y `gpt-4o-mini` when used in evaluations.

PR #833, opened 2025-02-06
Update model list of providers and use only gpt4-o for evaluations

andresgutgon:
# What?
This PR update the list of models and their prices according to official documentation on provider's pages. It also unify the way we list model names. 
Lastly we want to use only gpt4o and gpt4o-mini for latitude provider. 

## Issue
https://github.com/latitude-dev/latitude-llm/issues/832

## TODO
- [x] Update Groq models
- [x] Update Google models
- [x] Update Anthropic models
- [x] Update Mistral models
- [x] Update OpenAI models
- [x] Make clear what's the mess we have with default provider 
- [x] Use only gpt4o-mini for latitude provider. 
- [x] Fix setupService tests
- [x] Implement range model token price calculation
- [x] Put N/A in providerLogs metadata and documentLogs metadata for `costs` for custom providers and Azure. We don't really calculate the price atm


apps/web/src/actions/documents/sharing/forkDocumentAction.ts
```diff
@@ -26,6 +26,6 @@ export const forkDocumentAction = authProcedure
         workspace: ctx.workspace,
         user: ctx.user,
       },
-      defaultProviderId: env.DEFAULT_PROVIDER_ID,
```
andresgutgon, line 29: We meant name not id here. This variable was wrong

apps/web/src/services/user/setupService.test.ts
```diff
@@ -73,16 +87,26 @@ describe('setupService', () => {
...
     expect(createdApiKey).toBeDefined()
+
+    // Check provider API key creation when ENV variables are present
+    const createdProviderApiKey =
+      await database.query.providerApiKeys.findFirst({
+        // @ts-expect-error - drizzle-orm types are not up to date
+        where: eq(providerApiKeys.workspaceId, workspace.id),
+      })
+    expect(createdProviderApiKey).toBeDefined()
+    expect(createdProviderApiKey?.authorId).toBe(user.id)
```
andresgutgon, line 98: This is back in the test because now we test with stub env variables

packages/core/src/services/ai/estimateCost/index.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    //   Total input cost = 0.16 + 0.18 = 0.34
+    //
+    // Output tokens:
+    //   - First 128,000 tokens at 5: (5 * 128,000) / 1,000,000 = 0.64
+    //   - Remaining 172,000 tokens at 10: (10 * 172,000) / 1,000,000 = 1.72
+    //   Total output cost = 0.64 + 1.72 = 2.36
+    //
+    // Overall cost = 0.34 + 2.36 = 2.70
+    expect(cost).toBeCloseTo(2.7)
+  })
```
csansoon, line 70: Cool!

packages/core/src/services/ai/estimateCost/index.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      model: 'gemini-1.5-flash',
+    })
+
+    // For tiered-model-flash:
+    // Both input and output tokens are below the threshold (128,000), so the first tier applies:
+    // Input: (0.075 * 100,000) / 1,000,000 = 0.0075
+    // Output: (0.3 * 50,000) / 1,000,000 = 0.015
+    // Total = 0.0075 + 0.015 = 0.0225
+    expect(cost).toBeCloseTo(0.0225)
+  })
```
csansoon, line 43: Maybe fake the model cost here. If we need to update this model's cost in the future, we'll have to update the tests too, although the tested logic did not change.

andresgutgon: The AI suggested that solution. I changed it for real data. I think it is not a big deal, and probably, the model will not change much of the price.

PR #835, opened 2025-02-06
API v3

csansoon:
Update the API version to the new v3.

In this v3, everything will still be the same, except for prompt execution of any kind (run / chat) with `stream: true`. In this case, all new events will be used!

Both SDKs are also updated to use the new v3 API version!

- [x] API v3
- [x] Typescript SDK
- [ ] Python SDK – Will do in a different PR


apps/gateway/src/routes.ts
```diff
@@ -47,4 +57,20 @@ export const ROUTES = {
...
+      run: `${V3_DOCUMENTS}/run`,
+      logs: `${V3_DOCUMENTS}/logs`,
+    },
+    conversations: {
+      chat: `${V3_CONVERSATION_DETAIL}/chat`,
+      evaluate: `${V3_CONVERSATION_DETAIL}/evaluate`,
+      createEvaluationResult: `${V3_EVALUATION_DETAIL}/evaluation-results`,
+    },
+    telemetry: {
+      traces: `${V3_TELEMETRY}/v1/traces`,
```
andresgutgon, line 73: This is weird. We have v1 here. I'm sure is a mistake i did when moving v2 to openapi

apps/gateway/src/routes/v2/index.ts
```diff
@@ -0,0 +1,9 @@
+import { createRouter } from '$/openApi/createApp'
+import { chatHandler, chatRoute } from './conversations/chat'
+import { runHandler, runRoute } from './documents/run'
+
+const router = createRouter()
+  .openapi(runRoute, runHandler)
+  .openapi(chatRoute, chatHandler)
+
+export default router
```
andresgutgon, line 9: I don't know why we don't do the same in v3. Just one index file to be imported. Can you check if can be done? again, this is the way i did in v2 move to openapu, so my bad

andresgutgon (dismissed):
Nice

apps/gateway/src/routes.ts
neoxelox: yea I think this was intended by gerard, like a subversioning

PR #840, opened 2025-02-11
Disable eval generator locally

neoxelox:


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/dashboard/_components/Layout.tsx
```diff
@@ -61,8 +83,10 @@ Don't rawdog your prompts!
 
 export default function EvaluationsLayoutClient({
   evaluations: fallbackData,
+  isEvaluationGeneratorEnabled,
 }: {
   evaluations: EvaluationDto[]
+  isEvaluationGeneratorEnabled?: boolean
 }) {
```
Comment (csansoon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/dashboard/_components/Layout.tsx:90]: Why is it an attribute? Why not read it straight from the env?

Comment (neoxelox) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/dashboard/_components/Layout.tsx:90]: Because is frontend code and the env var is not accesible from there. I could try to put it as a public envar, but last time we tried we were not able to do so...

andresgutgon (approved):

PR #842, opened 2025-02-12
Fix: latitude tools and client tools compatibility

csansoon:
### WAT
This PR fixes an error where both client tools and latitude tools request in the same assistant response are incompatible.
Before this PR, the Latitude tools were processed and executing before pausing the chain, but these tool responses were not stored neither in the paused chain or provider log.

https://github.com/user-attachments/assets/e6e7870c-379c-4812-a11a-2dabe00134d2

### HOW
Now, when the assistant responds with both client and latitude tools, the client tools will be addressed first: it pauses the chain and waits for the user to provide the client tools responses. Then, when resuming the chain, the rest of latitude tools will be processed and executed before continuing with the next step.
No need to store the Latitude Tools responses before pausing them. :)

https://github.com/user-attachments/assets/98490990-404e-494d-8977-f9c295219483

### DIFF
--- a/packages/core/src/services/chains/runStep/index.ts
+++ b/packages/core/src/services/chains/runStep/index.ts
@@ -1,5 +1,5 @@
 import { Chain as PromptlChain } from 'promptl-ai'
-import { type Message } from '@latitude-data/compiler'
+import { AssistantMessage, type Message } from '@latitude-data/compiler'
 import {
   ABSOLUTE_MAX_STEPS,
   DEFAULT_MAX_STEPS,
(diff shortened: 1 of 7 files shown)


andresgutgon (commented):
Add some tests

packages/core/src/services/chains/runStep/index.ts
```diff
@@ -90,6 +90,15 @@ export async function runStep({
...
   stepCount = 0,
 }: StepProps) {
+  if (newMessages?.length) {
+    const lastResponse = newMessages[0]! as AssistantMessage
+    const builtInToolCalls =
+      getBuiltInToolCallsFromAssistantMessage(lastResponse)
+    const latitudeToolResponses =
+      await chainStreamManager.executeLatitudeTools(builtInToolCalls)
+    newMessages.push(...latitudeToolResponses)
+  }
```
andresgutgon, line 100: This same logic in 2 places. is not possible to dry it?

csansoon: sure!

PR #852, opened 2025-02-14
Fix dataset preview generation

andresgutgon:
# What?
We were using our SDK in streaming mode to check outdated event types. But before fixing it, I decided to move this action to non-streaming. More easier to understand


apps/web/src/app/(private)/datasets/generate/GenerateDatasetContent.tsx
```diff
@@ -69,25 +65,18 @@ export function GenerateDatasetContent({
...
-    isLoading: previewIsLoading,
+    execute: runPreviewAction,
+    isPending: previewIsLoading,
     error: previewError,
-  } = useStreamableAction<typeof generateDatasetPreviewAction>(
-    generateDatasetPreviewAction,
-    async (event, data) => {
-      if (
-        event === StreamEventTypes.Latitude &&
-        data.type === LegacyChainEventTypes.Complete
```
andresgutgon, line 81: The problem is that the data returned by the SDK is different now, and this code was not changed.

PR #853, opened 2025-02-14
Search Latitude Tool (Again)

csansoon:
gateway had issues importing `@tavily/core` as it was not marked as an external dependency.

Just adding it as external does not work, since the gateway package itself does not have it as a dependency of its own, it's only added as a dependency of `@latitude-dev/core`.

The solution has been adding the package as a dependency to the gateway, as it automatically configures it as external and it does work this way.


apps/gateway/package.json
```diff
@@ -27,6 +27,7 @@
     "@sentry/cli": "^2.37.0",
     "@sentry/node": "^8.30.0",
     "@t3-oss/env-core": "^0.10.1",
+    "@tavily/core": "^0.3.1",
```
csansoon, line 30: Actual fix

andresgutgon (approved):

PR #855, opened 2025-02-14
Recover mailpit

andresgutgon:
# What?
☝️ We use this to see emails in developement

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,12 @@
 services:
+  mailpit:
+    image: axllent/mailpit
+    ports:
+      - 8025:8025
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,4 +1,10 @@
 services:
+  mailpit:
+    image: axllent/mailpit
+    ports:
+      - 8025:8025
+      - 1025:1025
+
```
geclos, line 7: add it to the tmuxinator config

geclos: also add a profile development and add this container to that profile please

geclos:
looks good merging

PR #863, opened 2025-02-19
Generate document suggestions

neoxelox:
Related to: https://github.com/latitude-dev/latitude-llm/issues/851


apps/web/src/actions/copilot/refinePrompt.ts
```diff
@@ -116,5 +118,10 @@ export const refinePromptAction = authProcedure
       },
     })
 
-    return result.response.text
+    return (
```
andresgutgon, line 121: Instead of lying here you can really throw an error if it's not `object`

```javascript
if (result.response.type !== 'object') {
  throw new Error('This should never happen')
}

// Now Typescript is happy and you can do `result.response.object`
```

packages/core/src/events/handlers/index.ts
```diff
@@ -28,7 +29,10 @@ export const EventHandlers: IEventsHandlers = {
   documentSuggestionDiscarded: [],
   documentRun: [],
   evaluationCreated: [],
-  evaluationResultCreated: [notifyToClientEvaluationResultCreatedJob],
+  evaluationResultCreated: [
+    requestDocumentSuggestionJob,
+    notifyToClientEvaluationResultCreatedJob,
+  ],
```
csansoon, line 35: Is a suggestion requested for any new evaluation result?

csansoon (dismissed):

apps/web/src/actions/documentSuggestions/generate.ts
```diff
@@ -1,35 +0,0 @@
...
-    }).then((r) => r.unwrap())
-
-    return {
-      suggestion: {
-        ...result.suggestion,
-        evaluationUuid: evaluation.uuid,
-        evaluationName: evaluation.name,
-      } as DocumentSuggestionWithDetails,
-    }
-  })
```
andresgutgon: So now who invokes `generate` service?

neoxelox: The backend when a new evaluation result from the playground is created

packages/core/src/events/handlers/requestDocumentSuggestionJob.ts
```diff
@@ -0,0 +1,36 @@
...
+    evaluationResult: { id: resultId },
+    documentLog: { commitId, documentUuid },
+  } = event.data
+
+  const workspace = await unsafelyFindWorkspace(workspaceId)
+  if (!workspace) throw new NotFoundError(`Workspace not found ${workspaceId}`)
+
+  const resultsRepository = new EvaluationResultsRepository(workspace.id)
+  const result = await resultsRepository.find(resultId).then((r) => r.unwrap())
+  if (result.source !== LogSources.Playground) return
```
andresgutgon, line 24: Nice

andresgutgon: Maybe do a test for this handler

PR #864, opened 2025-02-19
Agents as tools

csansoon:
New feature to run agents inside prompts!
Just add an `agents` config with a list of relative paths for all included agents:
```
---
provider: openai
model: gpt-4o-mini
agents:
 - ./agents/researcher
---
```

A new tool will be injected into the prompt, which will execute the agent and return the response!

https://github.com/user-attachments/assets/d79ebd4c-6502-412e-ae63-e0251d58c4ba

---

Note: still working on some tests.


apps/web/src/actions/prompts/run.ts
```diff
@@ -8,29 +8,41 @@ import { createStreamableValue } from 'ai/rsc'
...
-        workspace: ctx.workspace,
+        workspace,
         source: LogSources.Evaluation,
         prompt,
         promptlVersion,
         parameters,
         providersMap: await buildProvidersMap({
           workspaceId: ctx.workspace.id,
         }),
+        promptSource: evaluationResult.unwrap(),
```
andresgutgon, line 45: So now every run in the playground sends an evaluation Result to get the agent tools?

csansoon: nono, this is only the evaluations playground

andresgutgon (dismissed):

PR #866, opened 2025-02-20
Implement get all documents in python sdk

andresgutgon:
# What?
Implement `get_all` method to get all documents on a project from python SDK

### TODO
- [x] Implement method 
- [x] Add documentation


packages/sdks/python/src/latitude_sdk/sdk/prompts.py
```diff
@@ -16,6 +16,7 @@
     RunPromptRequestBody,
     RunPromptRequestParams,
 )
+from latitude_sdk.client.payloads import GetAllPromptRequestParams
```
neoxelox, line 19: import GetAllPromptRequestParams from .client (above)

packages/sdks/python/src/latitude_sdk/sdk/types.py
```diff
@@ -24,11 +26,22 @@ class Providers(StrEnum):
     Custom = "custom"
 
 
+class ParameterType(str, Enum):
```
neoxelox, line 29: use only StrEnum (as Providers do)

neoxelox: Also, I thought the convetion was camel case and not all uppercased

packages/sdks/python/tests/utils/fixtures.py
```diff
@@ -42,6 +42,7 @@
     ToolCall,
     ToolResult,
 )
+from latitude_sdk.sdk.types import ParameterType
```
neoxelox, line 45: import ParameterType from .sdk (like above)

examples/sdks/python/get-prompt/multiple/example.py
```diff
@@ -0,0 +1,29 @@
...
+VERSION_UUID = os.getenv("VERSION_UUID")
+
+
+async def main():
+    assert PROJECT_ID, "PROJECT_ID is not set"
+
+    sdk = create_sdk()
+
+    # Latest version
+    prompt = await sdk.prompts.get_all(GetAllPromptOptions(project_id=int(PROJECT_ID)))
```
neoxelox, line 21: why not create the sdk with the project id? just as you do in the single get example

andresgutgon: I don't know. To play with both cases

packages/sdks/python/tests/utils/fixtures.py
```diff
@@ -86,6 +88,7 @@
         "topP": 0.9,
     },
     "provider": "openai",
+    "parameters": {"question": {"type": ParameterType.Text}},
```
neoxelox, line 91: hardcode "text" as string to ensure deserialization is done correctly : )

neoxelox (approved):

PR #869, opened 2025-02-20
Schedule document suggestions + bug fixes

neoxelox:
Related to: https://github.com/latitude-dev/latitude-llm/issues/851


packages/core/src/jobs/index.ts
```diff
@@ -1,12 +1,24 @@
...
-export async function setupJobs() {
-  if (queues) return queues
-  queues = await setupQueues()
-
+export async function setupJobs(connection?: Redis) {
+  if (!queues) queues = await setupQueues(connection)
   return queues
 }
+
+export async function setupSchedules(connection?: Redis) {
```
andresgutgon, line 13: Why is this not inside `setupJobs`? Call it something generic and put all in one function

neoxelox: Because I want to call the setup schedules only from Workers initialization

andresgutgon: Got it +1

PR #880, opened 2025-02-24
chore: add bullmq dashboard to workers

geclos:


apps/workers/package.json
```diff
@@ -14,17 +14,21 @@
...
   "dependencies": {
+    "@bull-board/api": "^6.7.9",
+    "@bull-board/express": "^6.7.9",
     "@latitude-data/core": "workspace:^",
     "@latitude-data/env": "workspace:^",
     "@sentry/cli": "^2.37.0",
     "@sentry/node": "^8.30.0",
     "@t3-oss/env-core": "*",
     "@tavily/core": "^0.3.1",
+    "express": "^4.21.2",
```
andresgutgon, line 25: Would have being nice to use Hono as in the rest of the app
https://www.npmjs.com/package/@bull-board/hono

andresgutgon: Why is better with Express?

andresgutgon: Also why express 4? They released v5 I think

geclos: I think 5 is in beta as it defaults to 4 yet, not massively important since this won't grow to become a big server

geclos: > Why is better with Express?

They have better examples with expressjs integration with auth and other use cases

andresgutgon (dismissed):

PR #881, opened 2025-02-24
Cleaner recomputeChanges workflow

csansoon:
When reading what changes are in any draft, we were recomputing all changes, removing all old document versions, and recreating the recomputed ones. This made the document versions' `id` attribute really unstable. In addition to that, this will also make CASCADE relations depending on these rows really unsafe.

This commit improves this workflow. Now, only the reverted changes will be removed from document versions, and the ones that changed will be simply inserted or updated, keeping its old row and `id`.

### DIFF
--- a/packages/core/src/services/documents/recomputeChanges/index.ts
+++ b/packages/core/src/services/documents/recomputeChanges/index.ts
@@ -7,7 +7,7 @@ import {
   type CompileError,
 } from '@latitude-data/compiler'
 import { scan } from 'promptl-ai'
-import { eq } from 'drizzle-orm'
+import { and, eq, inArray, not } from 'drizzle-orm'
 
 import {
   Commit,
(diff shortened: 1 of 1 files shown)


andresgutgon (changes requested):
Tests

csansoon:
> Tests

All previous tests apply to make sure the workflow is unaltered. This change is done to enable some new features @neoxelox is working on.

packages/core/src/services/documents/recomputeChanges/index.ts
```diff
@@ -126,23 +126,75 @@ async function replaceCommitChanges(
...
+          .values(
+            docsToInsert.map((d) => ({
+              ...omit(d, ['id', 'commitId', 'updatedAt']),
+              commitId,
+            })),
+          )
+          .returning()
+      : []
+
+    const updatedDocs = docsToUpdate.length
```
neoxelox, line 175: Maybe you could do a single update query in bulk?

```sql
update test as t set
    column_a = c.column_a,
    column_c = c.column_c
from (values
    ('123', 1, '---'),
    ('345', 2, '+++')  
) as c(column_b, column_a, column_c) 
where c.column_b = t.column_b;
```

neoxelox (approved):

packages/core/src/services/documents/recomputeChanges/index.ts
csansoon: could be revised in the future though

PR #890, opened 2025-02-24
Implement new data model for editable datasets

andresgutgon:
# What?
We want to allow users to edit rows or add new rows in uploaded datasets. For that what we're going to do is store each row in the original CSV as new entry in `dataset_rows` table. That has a relation to `datasets_v2` table

## TODO
- [ ] Run migration
- [ ] Add types for new models


packages/core/src/schema/models/datasetsV2.ts
```diff
@@ -0,0 +1,44 @@
...
+    }),
+    columns: jsonb('columns').$type<Column[]>().notNull(),
+    ...timestamps(),
+  },
+  (table) => ({
+    datasetWorkspaceIdx: index('datasets_table_workspace_idx').on(table.workspaceId),
+    authorIdx: index('datasets_table_author_idx').on(table.authorId),
+    uniqueDatasetNameInWorkspace: uniqueIndex().on(
+      table.workspaceId,
+      table.name,
```
neoxelox, line 41: why?

andresgutgon: It was in the old table. We don't allow to have 2 datasets with the same name

neoxelox: makes no sense for me, lets talk about it on slack

packages/core/src/schema/models/datasetRows.ts
```diff
@@ -0,0 +1,16 @@
+import { pgTable, uuid, jsonb, bigserial } from 'drizzle-orm/pg-core'
+import { timestamps } from '../schemaHelpers'
+import { datasetsV2 } from './datasetsV2'
+
+type DatasetRowData = {
+  [key: string]: string | number | boolean | null | undefined
+}
+
+export const datasetRows = pgTable('dataset_rows', {
+  id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
```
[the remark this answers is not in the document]
andresgutgon: What's the use case? I think we're never going to want to fetch all rows of all datasets by workspace. Do we?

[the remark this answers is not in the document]
neoxelox: its for row level security and make scopes easier to implement

geclos (approved):

PR #891, opened 2025-02-24
Smarter agents

csansoon:
## WAT

Currently, creating a competent agent requires lots of prompting and the user knowing how the workflow actually works.

Without the required prompting, the LLM tends to make some of these errors:
- Responding right away, without performing a chain-of-thought and without using the agent tool.
- Entering an infinite loop asking the user for more input without realizing it needs to use the agent tool to stop the workflow.
- Not using the agent tool correctly.

This PR tries to fix these issues, giving context to the AI about the workflow and giving them a push to work in a structured chain-of-thought process, without requiring extra prompting from the user. In order to respect the users control over their own prompt, this must be done without injecting or modifying SYSTEM / USER messages from the prompt.

## HOW

Okay, so this is what I have done.
Now, when the AI before an autonomous workflow, I add a phantom Assistant message with the fake request to start an autonomous workflow. This tool does not exist, and will do nothing, but it tricks the AI into thinking they requested it and thus know how behave from this point forward. Now, since "they requested" entering an autonomous workflow, they understand that they must make an explicit request to exit this workflow and return a response. In addition to this, the tool response to the fake workflow start request also gives indications about the workflow they just started, how it works and how and when to end it. No additional instructions are added about the task or how to solve it.

--- 

Let's see an example:

Take a look at this prompt:
```markdown
---
provider: openai
model: gpt-4o-mini
type: agent
---
You are a helpful AI assistant designed to assist users with any questions they might have. Your goal is to guide the user towards scheduling an appointment. As soon as the user wants an appointment, help them with that.

Here you can find all the info the user might ask about:

<prompt path="./docs" />

<user>
  {{ message }}
</user>
```

With this prompt, the AI does not know anything about the agentic workflow. For this reason, it tends to just return a response to the user without using the agent tool, see that the user hasn't responded (because the AI is still in a loop) and ask for further input for the user. Then, since it has still not ended the workflow, it repeats the last message. And over, and over, and over again.

Now, with the fake message injection, this is how this conversation actually looks like to the AI:
```
[SYSTEM]
You are a helpful AI assistant designed to assist users with any questions they might have. Your goal is to guide the user towards scheduling an appointment. As soon as the user wants an appointment, help them with that.
Here you can find all the info the user might ask about:
<docs>

[USER]
<message>

[ASSISTANT] (fake workflow request)
{
  toolId: 'agent_start_1',
  toolName: 'start_autonomous_chain'
  arguments: {}
}

[TOOL] (fake tool response)
Autonomous workflow started.
All messages from now on are for internal use only and will not be presented to the user.
Use this workflow to perform a chain-of-thought about your task and execute required tools.
You must explain your thought process on each and every step of the chain.
Start by understanding your task and explaining the intended course of actions to yourself.
Once you have finished your current task, use the 'end_autonomous_chain' tool to finish it. The tool call must include the requested result.
```

Now, the AI is informed about the autonomous workflow, no system instruction has been added that could conflict with the original prompt, and it will start a chain-of-thought process.

### Before
image

### After
image

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptConfiguration/BehaviourSettings.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptConfiguration/BehaviourSettings.tsx
@@ -7,6 +7,7 @@ import {
 } from './utils'
 import { LatitudeTool } from '@latitude-data/constants'
 import { SubAgentSelector } from './_components/AgentSelector'
+import { useEffect } from 'react'
 
 export function BehaviourSettings({
   config,
@@ -21,6 +22,22 @@ export function BehaviourSettings({
     key: 'type',
     defaultValue: undefined,
   })
+
+  const {
+    value: disabledAgentOptimization,
+    setValue: setDisableAgentOptimization,
+  } = useConfigValue<boolean>({
+    config,
+    setConfig,
+    key: 'disableAgentOptimization',
+    defaultValue: false,
+  })
+
+  useEffect(() => {
+    if (agentValue !== 'agent') {
+      setDisableAgentOptimization(undefined)
+    }
+  }, [agentValue])
 
   const { tools, toggleTool } = useLatitudeToolsConfig({ config, setConfig })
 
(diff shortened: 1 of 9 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptConfiguration/BehaviourSettings.tsx
```diff
@@ -22,6 +23,22 @@ export function BehaviourSettings({
...
+  } = useConfigValue<boolean>({
+    config,
+    setConfig,
+    key: 'disableAgentOptimization',
+    defaultValue: false,
+  })
+
+  useEffect(() => {
+    if (agentValue !== 'agent') {
+      setDisableAgentOptimization(undefined)
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptConfiguration/BehaviourSettings.tsx:38]: Instead of this do this logic on the type callback function. Avoid `useEffect` like la peste 🙏

andresgutgon (dismissed):

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptConfiguration/BehaviourSettings.tsx
```diff
@@ -41,6 +59,28 @@ Unlike regular prompts or predefined Chains, Agents can adapt dynamically, respo
...
+            description={`When enabled, the agent will automatically know that they are in an autonomous workflow and how it works.
+                Otherwise, you will need to add custom prompting to avoid the AI falling into infinite loops and using the agent tools correctly.
+                This optimization is done without adding or modifying any SYSTEM or USER message from your prompt, it's all handled internally by Latitude.`}
+          >
+            <SwitchToogle
+              disabled={disabled}
+              checked={!disabledAgentOptimization}
+              onCheckedChange={() =>
+                setDisableAgentOptimization(
+                  disabledAgentOptimization ? undefined : true,
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptConfiguration/BehaviourSettings.tsx:77]: Lovely

PR #894, opened 2025-02-25
Improved Tool UI

csansoon:
Improved Tool UI. Now, instead of appearing as distinct messages, the tool responses appear right after their corresponding tool requests, improving the user experience. Additionally, a loading state will appear when the tool response has not been added to the conversation yet, indicating that the tool is still being executed.

image

### DIFF
--- a/packages/web-ui/src/ds/molecules/Chat/Message/ContentCard.tsx
+++ b/packages/web-ui/src/ds/molecules/Chat/Message/ContentCard.tsx
@@ -2,28 +2,32 @@ import { cn } from '../../../../lib/utils'
 import { Button, Icon, IconName, Text } from '../../../atoms'
 import { TruncatedTooltip } from '../../TruncatedTooltip'
 import { ReactNode } from 'react'
-import { TextColor } from '../../../tokens'
+import { BackgroundColor, colors, TextColor } from '../../../tokens'
 
 export function ContentCard({
   label,
   icon,
   bgColor = 'bg-yellow',
   fgColor = 'warningForeground',
-  children,
   info,
   infoColor,
+  children,
+  resultFooter,
+  separatorColor = 'mutedForeground',
 }: {
   label: string
   icon?: IconName
   bgColor: string
   fgColor: TextColor
-  children: ReactNode
   info?: string
   infoColor?: TextColor
+  children: ReactNode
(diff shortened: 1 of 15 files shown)


packages/web-ui/src/ds/molecules/Chat/Message/ContentCard.tsx
```diff
@@ -9,21 +9,25 @@ export function ContentCard({
...
   label: string
   icon?: IconName
   bgColor: string
   fgColor: TextColor
-  children: ReactNode
   info?: string
   infoColor?: TextColor
+  children: ReactNode
+  resultFooter?: ReactNode
+  separatorColor?: string
```
andresgutgon, line 26: This should be a token color

packages/web-ui/src/ds/molecules/Chat/Message/LatitudeTools/Extract.tsx
```diff
@@ -21,6 +25,21 @@ export function WebExtractLatitudeToolCallContent({
       bgColor='bg-success'
       fgColor='successForeground'
       info={toolCallId}
+      separatorColor={
+        toolResponse?.isError ? 'bg-destructive-muted-foreground' : undefined
```
andresgutgon, line 29: Yep, token color. Never raw css classes

packages/web-ui/src/lib/hooks/useToolContentMap.ts
```diff
@@ -0,0 +1,29 @@
...
+  MessageRole,
+  ToolContent,
+} from '@latitude-data/compiler'
+import { useMemo } from 'react'
+
+export function useToolContentMap(
+  messages: Message[],
+  toolContentMap?: Record<string, ToolContent>,
+) {
+  if (toolContentMap) return toolContentMap
```
andresgutgon, line 14: So this is never updated when messages change no?

andresgutgon (dismissed):

packages/web-ui/src/lib/hooks/useToolContentMap.ts
csansoon: If an optional default `toolContentMap` is given, it will use just that one instead. Otherwise, it iterates over the messages array each time it changes.

PR #897, opened 2025-02-25
New tools schema

csansoon:
Improves the tools definition to merge both custom and built-in tools into a single config section. Forget about `latitudeTools`, `customTools` or `clientTools`.

Before:
```yaml
---
provider: openai
model: gpt-4o
tools:
  get_weather:
    description: Returns the weather for a given location
    parameters:
      type: object
      properties:
        location:
          type: string
          description: A location name
latitudeTools:
  - code
  - search
  - extract
---
```

After:
```yaml
---
provider: openai
model: gpt-4o
tools:
 - latitude/*
 - get_weather:
      description: Returns the weather for a given location
      parameters:
        type: object
        properties:
          location:
            type: string
            description: A location name
---
```

---

Call now and get an extra real-time parsing!

image

image

### DIFF
--- a/packages/core/src/lib/chainStreamManager/step/promptInjection/tools.ts
+++ b/packages/core/src/lib/chainStreamManager/step/promptInjection/tools.ts
@@ -1,0 +1,136 @@
+import {
+  PromptConfig,
+  LatitudeTool,
+  ToolDefinition,
+  VercelConfig,
+} from '@latitude-data/constants'
+import { Result, TypedResult } from '../../../Result'
+import { BadRequestError, LatitudeError, NotFoundError } from '../../../errors'
+import {
+  getLatitudeToolDefinition,
+  getLatitudeToolInternalName,
+} from '../../../../services/latitudeTools/helpers'
+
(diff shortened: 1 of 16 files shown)


packages/core/src/lib/chainStreamManager/step/promptInjection/tools.ts
```diff
@@ -0,0 +1,136 @@
...
+    .map((s) => s.trim())
+  if (!toolSource || !toolName || other.length) {
+    return Result.error(new BadRequestError(`Invalid tool ID: '${toolId}'`))
+  }
+
+  if (toolSource === 'latitude') {
+    const toolName = toolId.slice('latitude:'.length) as LatitudeTool | '*'
+    if (toolName === '*') {
+      return Result.ok(
+        Object.fromEntries(
```
andresgutgon, line 28: All this work can be at module level. No need to do each time we call this method

csansoon: cool idea

PR #898, opened 2025-02-25
Integrations schema

csansoon:
Schema for new Integrations workspace configuration.

In the future, we'll have a selection of built-in integrations. For now, just custom MCP servers.


packages/core/src/schema/models/integrations.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export type IntegrationProviderConfig<P extends IntegrationTypes> =
+  P extends IntegrationTypes.CustomMCP ? CustomMCPConfiguration : never
+
+export const integrations = latitudeSchema.table(
+  'integrations',
+  {
+    id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
+    name: varchar('name').notNull(),
+    type: integrationTypesEnum('type').notNull(),
```
andresgutgon, line 31: You can type this. Also call the column `integration_type`. I try to avoid ultra generic column names because then you can have conflicts on joins for example

packages/core/src/schema/relations.ts
```diff
@@ -284,3 +286,14 @@ export const spansRelations = relations(spans, ({ one }) => ({
...
+export const integrationsRelations = relations(integrations, ({ one }) => ({
+  workspace: one(workspaces, {
+    fields: [integrations.workspaceId],
+    references: [workspaces.id],
+  }),
+  author: one(users, {
+    fields: [integrations.authorId],
+    references: [users.id],
+  }),
+}))
```
andresgutgon, line 299: We really don't use relations for nothing

csansoon: 🤷

packages/core/src/schema/models/integrations.ts
```diff
@@ -0,0 +1,58 @@
...
+      .$type<IntegrationType>()
+      .notNull(),
+    configuration:
+      jsonb('configuration').$type<
+        IntegrationProviderConfig<IntegrationType.CustomMCP>
+      >(),
+    workspaceId: bigint('workspace_id', { mode: 'number' })
+      .notNull()
+      .references(() => workspaces.id),
+    authorId: text('author_id')
```
andresgutgon, line 41: Use `varchar`

andresgutgon (approved):

PR #902, opened 2025-02-26
Integrations settings

csansoon:
UI and logic to create and remove "Custom MCP Server" integrations.

https://github.com/user-attachments/assets/35df3b61-0fac-4ef1-85ae-f813db0fbe87


apps/web/src/app/(private)/settings/_components/Integrations/New/buildIntegrationPayload.ts
```diff
@@ -0,0 +1,92 @@
+import { merge } from 'lodash-es'
+import { IntegrationType } from '@latitude-data/constants'
+import {
+  COMMON_INTEGRATION_INPUT_FIELDS_KEYS,
+  CommonIntegrationInputKey,
+  IntegrationInputSchema,
+} from '$/actions/integrations/inputSchema'
+
+const CONFIG_PARAM_REGX = /\[([^\]]+)\]/g
```
andresgutgon, line 9: This sounds familiar. Don't we do something like this in other settings?

packages/core/src/repositories/integrationsRepository.ts
```diff
@@ -0,0 +1,34 @@
+import { eq, getTableColumns } from 'drizzle-orm'
+
+import { Integration } from '../browser'
+import { NotFoundError, Result } from '../lib'
+import { integrations } from '../schema'
+import RepositoryLegacy from './repository'
+
+const tt = getTableColumns(integrations)
+
+export class IntegrationsRepository extends RepositoryLegacy<
```
andresgutgon, line 10: Ey! no legacy no?

csansoon: Oh, true!

apps/web/src/app/(private)/settings/_components/Integrations/New/buildIntegrationPayload.ts
csansoon: With ApiKeyProviders, yep.

andresgutgon: 😢

PR #923, opened 2025-03-03
Evaluations V2 models schemas

neoxelox:


packages/core/src/schema/models/evaluationResultsV2.ts
```diff
@@ -0,0 +1,44 @@
...
+  'evaluation_results_v2',
+  {
+    id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
+    workspaceId: bigint('workspace_id', { mode: 'number' })
+      .notNull()
+      .references(() => workspaces.id, { onDelete: 'cascade' }),
+    evaluationId: bigint('evaluation_id', { mode: 'number' })
+      .notNull()
+      .references(() => evaluationsV2.id, { onDelete: 'cascade' }),
+    experimentId: bigint('experiment_id', { mode: 'number' }),
```
andresgutgon, line 19: Couldn't be result a polymorphic table having reference to `evaluation` or `experiment`?

packages/core/src/schema/models/evaluationResultsV2.ts
```diff
@@ -0,0 +1,44 @@
...
+    // .references(() => experiments.id, { onDelete: 'set null' }), // TODO: Add this when experiment table is created
+    evaluatedLogId: bigint('evaluated_log_id', { mode: 'number' })
+      .notNull()
+      .references(() => providerLogs.id, { onDelete: 'cascade' }),
+    score: bigint('score', { mode: 'number' }).notNull(),
+    metadata: jsonb('metadata').notNull().$type<EvaluationResultMetadata>(),
+    // Denormalized metadata fields - create indexes if necessary
+    usedForSuggestion: boolean('used_for_suggestion'),
```
andresgutgon, line 27: This is nullable. Maybe set a default false?

neoxelox: Nullable is a falsy value, its okay not to set a default val. I prefer that because of consistency. If in the future we need to denormalize more common fields they will be nullable without default (it locks the row/table on the mig otherwise)...

packages/core/src/schema/models/evaluationResultsV2.ts
neoxelox: Over complicating things really

andresgutgon: So a result can have an experiment and a evaluation at the same time?

geclos: if we don't anticipate more than a couple pointers it's fine, if it's larger than that i would much rather have a polymorphic pointer

PR #952, opened 2025-03-06
Move Evaluations to Document tab

neoxelox:
Related to: https://github.com/latitude-dev/latitude-llm/issues/922


apps/web/src/stores/evaluations.ts
```diff
@@ -56,6 +56,14 @@ export default function useEvaluations(
       },
     },
   )
+  const create = useCallback(
+    async (args: Parameters<typeof executeCreate>[0]) => {
+      const [result, error] = await executeCreate({ ...args, documentUuid })
+      if (error) return
+      return result
```
andresgutgon, line 63: You don't need to do this to get the error

neoxelox: Its a wrapper to inject the store's documentUuid

andresgutgon (approved):

PR #962, opened 2025-03-07
Fix agent tool maps recomputing too much

andresgutgon:
# What?
Try to make more efficient the way we recompute metadata in the document editor. This [useEffect](https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/app/(private)/projects/%5BprojectId%5D/versions/%5BcommitUuid%5D/documents/%5BdocumentUuid%5D/_components/DocumentEditor/Editor/index.tsx#L243-L254) is forcing recompute metadata all the time.

### DIFF
--- a/apps/web/src/hooks/useMetadata.ts
+++ b/apps/web/src/hooks/useMetadata.ts
@@ -1,11 +1,16 @@
 'use client'
 
-import { useEffect, useRef, useState } from 'react'
+import { useEffect, useState } from 'react'
 
 import type { ConversationMetadata } from 'promptl-ai'
 import { useDebouncedCallback } from 'use-debounce'
 
 import type { ReadMetadataWorkerProps } from '../workers/readMetadata'
+
+const worker =
+  typeof window === 'undefined'
+    ? undefined
+    : new Worker(new URL('public/workers/readMetadata', import.meta.url))
 
 export function useMetadata({
   onMetadataProcessed,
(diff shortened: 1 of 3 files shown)


apps/web/src/hooks/useMetadata.ts
```diff
@@ -1,46 +1,36 @@
...
 
-import { useEffect, useRef, useState } from 'react'
+import { useEffect, useState } from 'react'
 
 import type { ConversationMetadata } from 'promptl-ai'
 import { useDebouncedCallback } from 'use-debounce'
 
 import type { ReadMetadataWorkerProps } from '../workers/readMetadata'
 
+const worker =
```
andresgutgon, line 10: I did worker at module level. Does it need to be recreated each time the hook is used?

andresgutgon: This was something I was checking before finding the other bug. This is not a bug per se but I think is better this way.

apps/web/src/hooks/useMetadata.ts
```diff
@@ -1,46 +1,36 @@
...
-    createWorker()
+    if (typeof window === 'undefined' || !worker) return
 
-    return () => {
-      workerRef.current?.terminate()
+    worker.onmessage = (event: { data: ConversationMetadata }) => {
+      setMetadata(event.data)
+      onMetadataProcessed?.(event.data)
     }
   }, [onMetadataProcessed])
```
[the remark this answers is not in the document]
andresgutgon: What If I want to have only one instance of the worker initialized at module level?

geclos (approved):

PR #982, opened 2025-03-12
Email trigger attachments

csansoon:
- Added support for PromptLFiles, which automatically include file parameters as message contents
- Added support for attachments in Email Triggers
- Removed container on Email Trigger's sent email content

### DIFF
--- a/packages/core/src/services/commits/RunDocumentChecker/index.ts
+++ b/packages/core/src/services/commits/RunDocumentChecker/index.ts
@@ -4,12 +4,26 @@ import {
   type ReferencePromptFn,
 } from '@latitude-data/compiler'
 import { RunErrorCodes } from '@latitude-data/constants/errors'
-import { Adapters, Chain as PromptlChain, scan } from 'promptl-ai'
+import {
+  Adapters,
+  isPromptLFile,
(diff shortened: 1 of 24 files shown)


packages/core/src/services/commits/RunDocumentChecker/index.ts
```diff
@@ -121,6 +146,90 @@ export class RunDocumentChecker {
...
+          url,
+          file: {
+            name: fileName,
+            type: contentType ?? '',
+            size: parseInt(contentLength ?? '0'),
+          } as File,
+        }),
+      )
+    } catch (e) {
+      return Result.error(new LatitudeError(`Error fetching file: ${url}`))
```
andresgutgon, line 189: Do we want to fail? All will fail even if only of the files is wrong no?

andresgutgon (approved):

packages/core/src/services/commits/RunDocumentChecker/index.ts
csansoon: If a file fails, it means the URL is not available... That would be unexpected behaviour for the user, right?

PR #953, opened 2025-03-06
Adapt document parameters to use datasets v2

andresgutgon:
# What?
Last part of datasets v2 in the app. 

## Issue
https://github.com/latitude-dev/latitude-llm/issues/918

## TODO
- [x] Change dataset parameter hook to have a uniq API between V1 and V2 Datasets
- [x] QA v1 datasets keep working after refactor
- [x] Read default parameters from metadata.
- [x] Make runtime value mapping of dataset values to avoid stale values when a dataset row cell is updated or dataset headers are re-ordered.
- [x] DO `withPosition` endpoint to get selected dataset row 
- [x] Change DatasetRow navigation to work as logs navigation (Both are rows in the DB now)
- [x] Syncronize parameters in datasets v2 with document parameters. Otherwhise preview fails
- [x] Add a Zustand global store for document parameters 
- [x] Implement copy to manual from datasets V2
- [x] Remove feature flag
- [x] QA document parameters with V1 ⏬ 

## Next 
- [ ] Datasets v1 migration
- [ ] Implement copy to manual parameters

PR #983, opened 2025-03-12
Add new dataset V2 document relations

andresgutgon:
# What?
Migrations for this PR
https://github.com/latitude-dev/latitude-llm/pull/953

### DIFF
--- a/packages/core/src/lib/documentPersistedInputs.ts
+++ b/packages/core/src/lib/documentPersistedInputs.ts
@@ -45,6 +45,11 @@ export type LinkedDataset = {
   mappedInputs: Record<string, number>
 }
 
+export type LinkedDatasetRow = {
+  datasetRowId: number | undefined
+  mappedInputs: Record<string, string>
+}
+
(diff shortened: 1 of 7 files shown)


packages/core/src/lib/documentPersistedInputs.ts
```diff
@@ -45,6 +45,11 @@ export type LinkedDataset = {
   mappedInputs: Record<string, number>
 }
 
+export type LinkedDatasetRow = {
+  datasetRowId: number | undefined
```
neoxelox, line 49: I dont know how this can be undefined but ok

neoxelox (approved):

packages/core/src/lib/documentPersistedInputs.ts
andresgutgon: I'll double check in this PR: https://github.com/latitude-dev/latitude-llm/pull/953 maybe it's not necessary to set as `undefined`

PR #984, opened 2025-03-12
feat: Implement scheduled document triggers with cron expressions

geclos:

PR #986, opened 2025-03-13
Implement dataset V1 to dataset V2 migration service

andresgutgon:
# What?
Migrate old datasets to datasets v2 include also the associated file. Also migrate linked datasets to documents to new datasets. 

The order is to migrate the dataset to V2 and the check if it's linked to any document. In case is linked transform old mapped inputs into new mapped inputs.

## TODO
- [x] Implement migration service for datasets for one workspace. 
- [x] Implement a wrapper service to run a list of workspaces or an `all: true` flag`
- [x] Discuss how to run this code in the server with @geclos 
- [x] Fix csv import wrapping cells with `""` quotes
- [x] Fix [this bug](https://latitude-l5.sentry.io/issues/6426511220)
- [x] Enable again Dataset V2 for lat


packages/core/src/client/index.ts
```diff
@@ -14,4 +15,5 @@ const pool = new Pool({
   statement_timeout: 30000, // 30 seconds
 })
 
+export const dbUtils = drizzleUtils
```
andresgutgon, line 18: 🔥 Do it in a more specific way. Avoid exporting all drizzle

packages/core/src/services/datasetRows/createRowsFromUploadedDataset.ts
```diff
@@ -129,13 +133,19 @@ export async function createRowsFromUploadedDataset(
...
+      // PARANOID MODE: I don't want to enable file deletion until we have migrated
+      // all datasets V1
+
+      /* const deleteResult = await disk.delete(fileKey) */
+      /**/
+      /* if (deleteResult.error) { */
+      /*   throw new Error( */
+      /*     `Error deleting file: ${deleteResult.error.message} datasetId: ${datasetId} workspaceId: ${workspaceId}`, */
+      /*   ) */
+      /* } */
```
andresgutgon, line 148: This service is used in new datasets V2 settings to generate rows in `dataset_rows` table out of the uploaded CSV by the user. 

I'm reusing all that logic for the migration from dataset V1 to V2. Because V1 datasets have stored in Disk/S3 the file and I pass that file to this service and everything is created in the new way as it should be. 

The only thing I don't want is to delete the file after the new data is created. Because of that I commented this code until we're fully migrated.

packages/core/src/services/datasetsV2/createFromFile.ts
```diff
@@ -38,34 +36,15 @@ export const createDatasetFromFile = async (
...
-    )
-  }
-
-  let columns: Column[] = []
-  const newColumns = Array.from(uniqueHeaders.keys()).map((columnName) => ({
-    name: columnName,
-  }))
-  columns = buildColumns({
-    newColumns,
-    prevColumns: columns,
```
andresgutgon, line 65: QA that uploading a CSV in new datasets works. I just moved this to a common function because I use in 2 places. But it should work as before.

neoxelox (dismissed):

packages/core/src/data-migrations/datasets-v2/utils/migrateWorkspaceDatasetsToV2.ts
```diff
@@ -0,0 +1,137 @@
...
+        datasetV2: {
+          model: dataset,
+          rowCount: createResult.migrationResult.datasetV2?.rowCount ?? 0,
+          documents: updatedDocuments,
+        },
+      },
+    }
+  })
+
+  const results = await Promise.all(allResult)
```
neoxelox, line 108: My only concern is that you are executing the migration in parallel for all datasets of the workspace. If there are multiple datasets linked to the same document, there could be a data race, because you don't do select for update when you get the linked datasets info.

andresgutgon: Only can be 1 dataset linked to a document https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/schema/models/documentVersions.ts#L45-L48

andresgutgon: Yes but linkedDataset and datasetId are related.

pnpm-lock.yaml
andresgutgon, line 1: HUGE change. But because we submitted [pnpm-lock.yaml installed with pnpm 6](https://github.com/latitude-dev/latitude-llm/pull/984) and now I'm using the one we have defined which is pnpm 9

apps/web/package.json
```diff
@@ -6,7 +6,7 @@
     "workers:build": "rollup -c rollup.config.workers.mjs",
     "build": "pnpm run workers:build && next build",
     "workers:watch": "rollup -c rollup.config.workers.mjs -w",
-    "dev": "NODE_DEBUG=latitude:debug next dev --turbo",
+    "dev": "WATCHPACK_POLLING=true NODE_DEBUG=latitude:debug next dev --turbo",
```
andresgutgon, line 9: This fixed my problem with max-watched files. In theory, `turbo` is not a webpack, but maybe they share some configurations. It looks like this had some effect. 

### The error I saw
```bash
> cd apps/web
> pnpm dev
Watchpack Error (watcher): Error: EMFILE: too many open files, watch
Watchpack Error (watcher): Error: EMFILE: too many open files, watch
Watchpack Error (watcher): Error: EMFILE: too many open files, watch
```

packages/core/src/client/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+  and,
+  not,
+  gte,
+  lte,
+  gt,
+  lt,
+  asc,
+  desc,
+  getTableColumns,
+} from 'drizzle-orm'
```
andresgutgon, line 14: I opted to put only a subset of tools from drizzle in the console. We always can expand it

packages/core/src/services/datasetsV2/createFromFile.ts
andresgutgon: done

packages/core/src/client/index.ts
```diff
@@ -12,6 +13,11 @@ const pool = new Pool({
   connectionString: env.DATABASE_URL,
   idle_in_transaction_session_timeout: 1800000, // 30 minutes
   statement_timeout: 30000, // 30 seconds
+  ssl: {
+    // FIXME: Do not merge with this here!!!!
```
andresgutgon, line 17: EPA ⚠️

packages/core/src/client/index.ts
```diff
@@ -12,6 +13,11 @@ const pool = new Pool({
   connectionString: env.DATABASE_URL,
   idle_in_transaction_session_timeout: 1800000, // 30 minutes
   statement_timeout: 30000, // 30 seconds
+  /* ssl: { */
+  /*   // FIXME: Do not merge with this here!!!! */
+  /*   rejectUnauthorized: false, */
+  /* }, */
```
andresgutgon, line 19: OJO ⚠️

apps/console/.eslintrc.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": ["./node_modules/@latitude-data/eslint-config/library.js"],
+  "env": {
+    "node": true
+  },
+  "rules": {
+    "no-constant-condition": "off"
+  }
+}
```
geclos, line 9: maybe move this new package to tools folder? I would like to keep the apps folder only for apps that are required for latitude to run.

geclos (approved):
lgtm

apps/web/src/hooks/useDocumentParameters/useAsyncDocumentParameters.ts
```diff
@@ -96,21 +98,30 @@ export function useAsyncDocumentParameters({
...
       asyncParameters.set(newLocalInputs)
     }
-  }, [localInputs, datasetVersion, asyncParameters.set, isMountedOnRoot])
+  }, [
+    source,
+    localInputs,
+    datasetVersion,
+    asyncParameters.set,
+    isMountedOnRoot,
+  ])
```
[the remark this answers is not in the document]
andresgutgon: I'm not happy with this part to be honest. Needs to be reviewed

apps/console/.eslintrc.json
andresgutgon: Noted, I'll move in another PR +1

PR #997, opened 2025-03-18
feat: remove environment-dependent hardcodings

geclos:
Some values were hardcoded in the codebase based on the environment. This commit removes those hardcoded values and replaces them with environment variables.

PR #1004, opened 2025-03-19
Fix missing environment variables and improve Traefik security in Docker Compose setup

Bald1nh0:
Hello, I don't often make PRs to open-source repositories, so please don't judge me too harshly if something is wrong. :)

I encountered an issue when trying to deploy the project using Docker Compose in a production environment. The first thing I noticed was the insecure Traefik setup. Additionally, several required environment variables were missing from .env.example:

```bash
Invalid environment variables: {
  BULL_ADMIN_USER: [ 'Required' ],
  BULL_ADMIN_PASS: [ 'Required' ],
  MCP_NODE_GROUP_NAME: [ 'Required' ],
  MCP_DOCKER_IMAGE: [ 'Required' ],
  MCP_SCHEME: [ 'Required' ]
}
```
To address these issues, I propose an improved Docker Compose configuration, which includes:

Updated Traefik version with SSL certificates and FQDN support.
A complete .env.example file with required variables.
Corrections to documentation links for self-hosted production deployment.
Let me know if any adjustments are needed!

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,13 +4,9 @@ NODE_ENV=production
 # Traefik
 TRAEFIK_HOST=traefik.${APP_DOMAIN}
 TRAEFIK_ACME_EMAIL=your@email.com
-TRAEFIK_ADMIN_PASS="your_secure_password" # Passwords must be hashed using MD5, SHA1, or BCrypt. Read more: https://doc.traefik.io/traefik/middlewares/http/basicauth/
-
-BULL_ADMIN_USER=admin
-BULL_ADMIN_PASS=password
-MCP_NODE_GROUP_NAME=default-group
-MCP_DOCKER_IMAGE=ghcr.io/latitude-dev/web:latest
-MCP_SCHEME=http
+# Passwords must be hashed using MD5, SHA1, or BCrypt. Read more: https://doc.traefik.io/traefik/middlewares/http/basicauth/
+# Example command to generate a password for user admin: echo $(htpasswd -nb admin password) | sed -e s/\\$/\\$\\$/g
+TRAEFIK_ADMIN_PASS="your_secure_password"
 
 # Database
 POSTGRES_USER=latitude
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,17 @@
 # Environment
 NODE_ENV=production
 
+# Traefik
+TRAEFIK_HOST=traefik.${APP_DOMAIN}
+TRAEFIK_ACME_EMAIL=your@email.com
+TRAEFIK_ADMIN_PASS="your_secure_password" # Passwords must be hashed using MD5, SHA1, or BCrypt. Read more: https://doc.traefik.io/traefik/middlewares/http/basicauth/
+
+BULL_ADMIN_USER=admin
+BULL_ADMIN_PASS=password
+MCP_NODE_GROUP_NAME=default-group
+MCP_DOCKER_IMAGE=ghcr.io/latitude-dev/web:latest
+MCP_SCHEME=http
+
 # Database
 POSTGRES_USER=latitude
 POSTGRES_PASSWORD=secret
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ For more detailed information on each step, explore our documentation or join ou
 
 ### Latitude Self-Hosted
 
-Follow the instructions in the [self-hosted guide](https://docs.latitude.so/self-hosted/production) to get started with Latitude Self-Hosted.
+Follow the instructions in the [self-hosted guide](https://docs.latitude.so/guides/self-hosted/production) to get started with Latitude Self-Hosted.
 
 After setting up Latitude Self-Hosted, you can follow the same steps as in the Latitude Cloud guide to create, test, evaluate, and deploy your prompts.
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,7 +33,7 @@ x-gateway-common: &gateway-common
   labels:
     - "traefik.enable=true"
     - "traefik.docker.network=web"
-    - "traefik.http.services.gateway-service.loadbalancer.server.port=8081"
+    - "traefik.http.services.gateway-service.loadbalancer.server.port=8080"
     - "traefik.http.routers.gateway-http.entrypoints=http"
     - "traefik.http.routers.gateway-http.service=gateway-service"
     - "traefik.http.routers.gateway-http.rule=Host(`gateway-latitude.${APP_DOMAIN}`)"
@@ -75,7 +75,7 @@ services:
       --providers.file.watch=true
       --api.dashboard=true
       --api.debug=false
-      --api.insecure=false
+      --api.insecure=${TRAEFIK_API_INSECURE:-false}
       --metrics.prometheus=true
       --metrics.prometheus.buckets=0.1,0.3,1.2,5.0
       --metrics.prometheus.addEntryPointsLabels=true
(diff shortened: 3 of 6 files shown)


.env.example
```diff
@@ -1,6 +1,17 @@
...
+# Traefik
+TRAEFIK_HOST=traefik.${APP_DOMAIN}
+TRAEFIK_ACME_EMAIL=your@email.com
+TRAEFIK_ADMIN_PASS="your_secure_password" # Passwords must be hashed using MD5, SHA1, or BCrypt. Read more: https://doc.traefik.io/traefik/middlewares/http/basicauth/
+
+BULL_ADMIN_USER=admin
+BULL_ADMIN_PASS=password
+MCP_NODE_GROUP_NAME=default-group
+MCP_DOCKER_IMAGE=ghcr.io/latitude-dev/web:latest
+MCP_SCHEME=http
```
geclos, line 13: i'm making these env vars optional [here](https://github.com/latitude-dev/latitude-llm/pull/997) so you can remove those 🙏🏼

README.md
```diff
@@ -88,7 +88,7 @@ For more detailed information on each step, explore our documentation or join ou
 
 ### Latitude Self-Hosted
 
-Follow the instructions in the [self-hosted guide](https://docs.latitude.so/self-hosted/production) to get started with Latitude Self-Hosted.
+Follow the instructions in the [self-hosted guide](https://docs.latitude.so/guides/self-hosted/production) to get started with Latitude Self-Hosted.
```
geclos, line 91: thanks

docker-compose.yml
```diff
@@ -19,53 +31,127 @@
...
+    restart: always
+    command:
+      --api=true
+      --ping=true
+      --providers.docker=true
+      --providers.file.directory=/etc/traefik/
+      --providers.file.watch=true
+      --api.dashboard=true
+      --api.debug=false
+      --api.insecure=false
```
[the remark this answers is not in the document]
geclos: ☝🏼 good point

docker-compose.yml
```diff
@@ -19,53 +31,127 @@ x-web-common: &web-common
...
+      --metrics.prometheus=true
+      --metrics.prometheus.buckets=0.1,0.3,1.2,5.0
+      --metrics.prometheus.addEntryPointsLabels=true
+      --entrypoints.http.address=:80
+      --entrypoints.https.address=:443
+      --entryPoints.http.forwardedHeaders.insecure
+      --entryPoints.https.forwardedHeaders.insecure
```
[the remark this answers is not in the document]
geclos: this goes back to the convo about making basic auth opt-in rather than default

docker-compose.yml
```diff
@@ -19,53 +31,127 @@ x-web-common: &web-common
...
+    restart: always
+    container_name: latitude-llm-db
     ports:
       - '5432:5432'
     env_file:
       - .env
     volumes:
       - ./docker/init-db.sh:/docker-entrypoint-initdb.d/init-db.sh
-      - ./docker/pgdata:/var/lib/postgresql/data
+      - pgdata:/var/lib/postgresql/data
```
geclos, line 143: you changed this volume path?

docker-compose.yml
```diff
@@ -143,4 +238,14 @@ services:
...
     <<: *websockets-common
     image: ghcr.io/latitude-dev/websockets:latest
+    restart: always
+    container_name: latitude-llm-websockets
     platform: linux/amd64
+
+# You need to create external network for Traefik to work (docker network create web)
+networks:
+  web:
+    external: true
```
geclos, line 248: do we need custom network config? couldn't we use default?

docs/guides/getting-started/quick-start.mdx
```diff
@@ -40,6 +40,6 @@ For more detailed information on each step, explore our documentation or join ou
 
 ## Latitude Self-Hosted
 
-Follow the instructions in the [self-hosted guide](https://docs.latitude.so/self-hosted/production) to get started with Latitude Self-Hosted.
+Follow the instructions in the [self-hosted guide](https://docs.latitude.so/guides/self-hosted/production) to get started with Latitude Self-Hosted.
```
geclos, line 43: thanks

docker-compose.yml
andresgutgon: I think this makes us miss our current local data.

geclos (commented):
Thanks for taking the time!

My only contention point is to make basic auth optional rather than opt-in since it adds an extra layer of configuration that many won't want out of the box.

.env.example
```diff
@@ -1,6 +1,17 @@
 # Environment
 NODE_ENV=production
 
+# Traefik
+TRAEFIK_HOST=traefik.${APP_DOMAIN}
+TRAEFIK_ACME_EMAIL=your@email.com
+TRAEFIK_ADMIN_PASS="your_secure_password" # Passwords must be hashed using MD5, SHA1, or BCrypt. Read more: https://doc.traefik.io/traefik/middlewares/http/basicauth/
```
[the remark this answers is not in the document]
Bald1nh0: You're absolutely right! My main concern is securing the Traefik dashboard in a production environment. Since this setup is meant for production deployment, I believe it's important to protect access to the dashboard by default.

However, I understand that not everyone may need basic auth. A good compromise would be to enable it by default but allow disabling it via an environment variable. Would that approach work for you?

Let me know your thoughts!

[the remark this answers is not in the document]
geclos: sounds good 👌🏼

docker-compose.yml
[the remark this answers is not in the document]
Bald1nh0: Good point about the email variable—it's necessary for issuing SSL certificates, so I’ll ensure it has a default value to prevent failures.

Regarding basic auth, in the current Traefik configuration, it’s not possible to disable authentication dynamically because it's directly set in the docker-compose.yml labels. Given that this setup is intended for production use, protecting the Traefik dashboard by default is a reasonable security measure.

If there’s a strong need for an opt-out mechanism, one approach could be modifying the docker-compose.override.yml file per deployment. But in general, I believe keeping authentication enabled by default is the safest approach.

docker-compose.yml
Bald1nh0: Yes, I intentionally added a specific path for the volume while also defining a named volume (pgdata). The goal was to make data storage more explicit and easier to manage across environments.

However, I understand that this change might cause data loss for existing setups because it effectively creates a new volume. If that's a concern, we could:

1. Keep the existing setup and document an alternative for those who prefer explicit local paths.
2. Provide migration instructions for users who want to move their existing data to the new path.

Would love to hear your thoughts - do you think we should revert this change, or should we document how to migrate?

docker-compose.yml
Bald1nh0: Yes, we need a custom network configuration because web is an external network. This is useful when integrating Traefik with multiple services running in different docker-compose setups, ensuring they can communicate properly.

If we use the default network, it might work in simpler cases, but for production deployments with multiple services, an explicit external network provides better control and flexibility.

docker-compose.yml
neoxelox: This is changing to a named volume, defined down in the docker compose. We should be able to keep the data by specifying a bind path:

```
pgdata:
    driver: local
    driver_opts:
      type: 'none'
      o: 'bind'
      device: '/srv/db-data'
```

docker-compose.yml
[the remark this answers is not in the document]
geclos: Ok I agree let's keep it on by default

geclos:
> I understand! If basic auth feels like an extra layer for some, we could also consider disabling the Traefik dashboard entirely by default, especially for users who don't need it exposed. This would simplify the configuration and ensure security without requiring additional setup.

It's fine after more thought I think it's reasonable to keep basic auth on by default.

Remember to [sign the CLA](https://github.com/latitude-dev/latitude-llm/pull/1004#issuecomment-2736838430)

Bald1nh0:

docker-compose.yml
Bald1nh0: I revert changes to prevent data loss on working systems.

geclos: thanks we can always document the alternative if needed

geclos (approved):

andresgutgon (approved):

geclos:
👏🏼 congrats

PR #1008, opened 2025-03-20
"fix": no message after tool call response

csansoon:
When a prompt returns a tool call, and it is executed in the SDK, no further response is generated.

Okay... This is not actually a bug, it's a feature —yes, really—. But it does not fit the user's expectations.

Here's why it is happening: Running a chain with two steps should only send 2 requests to your provider... right? And... a chain with 1 step should only send 1 request...? Well, that's what should happen in most cases, but there are some others where the expected behaviour is different!

Let's say I have a simple prompt with access to a tool `get_weather`, and the prompt is asked to tell me what the weather is. Well, this requires 2 steps! One for generating the tool request, and another one for generating the final response after getting the tool response. Weird, right? Welp, this PR fixes that.

### DIFF
--- a/packages/core/src/services/chains/runStep/index.ts
+++ b/packages/core/src/services/chains/runStep/index.ts
@@ -15,9 +15,11 @@ import { cacheChain } from '../chainCache'
 import { ChainStreamManager } from '../../../lib/chainStreamManager'
 import {
   ABSOLUTE_MAX_STEPS,
+  ChainStepResponse,
   DEFAULT_MAX_STEPS,
   MAX_STEPS_CONFIG_NAME,
   PromptConfig,
+  StreamType,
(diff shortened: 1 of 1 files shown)


packages/core/src/services/chains/runStep/index.ts
```diff
@@ -106,7 +110,11 @@ export async function runStep({
 
   // With PromptL, the chain complete is checked AFTER the step is executed.
   // If the chain is completed, no more steps must be ran.
-  if (chain instanceof PromptlChain && step.chainCompleted) {
+  if (
+    chain instanceof PromptlChain &&
+    step.chainCompleted &&
+    !!previousResponse // If no previous response has been generated, make an additional step
+  ) {
```
andresgutgon, line 117: Test me babe

PR #1024, opened 2025-03-24
Feature/webhooks

geclos:


.avanterules
andresgutgon, line 1: Nice, I really need to setup avante

packages/web-ui/src/ds/molecules/MultiSelect/index.tsx
```diff
@@ -0,0 +1,362 @@
...
+   * An array of option objects to be displayed in the multi-select component.
+   * Each option object has a label, value, and an optional icon.
+   */
+  options: {
+    /** The text to display for the option. */
+    label: string
+    /** The unique value associated with the option. */
+    value: string
+    /** Optional icon component to display alongside the option. */
+    icon?: React.ComponentType<{ className?: string }>
```
andresgutgon, line 39: Use Select type

packages/web-ui/src/ds/molecules/MultiSelect/index.tsx
```diff
@@ -0,0 +1,362 @@
...
+  }[]
+
+  /**
+   * Callback function triggered when the selected values change.
+   * Receives an array of the new selected values.
+   */
+  onChange?: (value: string[]) => void
```
andresgutgon, line 46: Make it as in select with a generic so the value here is what the options values are

packages/web-ui/src/ds/molecules/MultiSelect/index.tsx
```diff
@@ -0,0 +1,362 @@
...
+  /**
+   * Additional information to display in a tooltip over the multi-select component.
+   * Optional, can be used to provide additional information or instructions.
+   */
+  info?: string
+
+  /**
+   * Additional label to display above the multi-select component.
+   */
+  label?: string
```
andresgutgon, line 103: Do we need all these comments? too much no?

packages/web-ui/package.json
```diff
@@ -73,8 +75,8 @@
     "lucide-react": "^0.468.0",
     "monaco-editor": "^0.50.0",
     "postcss": "^8.4.39",
-    "promptl-ai": "^0.5.2",
     "postcss-import": "^16.1.0",
+    "promptl-ai": "^0.5.2",
     "react": "18.3.0",
     "react-dom": "18.3.0",
     "react-resizable": "^3.0.5",
```
andresgutgon, line 82: Uhmm, I don't see `Command` we were already using it?

packages/web-ui/src/ds/molecules/MultiSelect/index.tsx
geclos: i think it's better if it's always a string since this way it behaves like a regular select that coerces all its values to strings

packages/web-ui/src/ds/molecules/MultiSelect/index.tsx
geclos: yeah no

packages/web-ui/package.json
geclos: it's in atoms/Command

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Actions.tsx
```diff
@@ -1,4 +1,4 @@
-import { ClientOnly, SwitchToogle, Text } from '@latitude-data/web-ui'
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/Actions.tsx:1]: Do we need to do the git blame? :trollface:

packages/core/src/services/webhooks/listWebhooks.ts
```diff
@@ -0,0 +1,12 @@
...
+import { database } from '../../client'
+import { webhooks } from '../../schema/models/webhooks'
+import { type Webhook } from './types'
+
+export async function listWebhooks(
+  workspaceId: number,
+  db = database,
+): Promise<Webhook[]> {
+  return db.select().from(webhooks).where(eq(webhooks.workspaceId, workspaceId))
+}
```
geclos, line 12: i'm actually choosing not to use repositories because all the troubles we've experienced so far. My counter proposal are basic services to do gets.

packages/web-ui/src/ds/atoms/Switch/index.tsx
```diff
@@ -85,7 +100,21 @@ function SwitchInput({
           formDescriptionId={formDescriptionId}
           formMessageId={formMessageId}
         >
-          <SwitchToogle {...rest} />
+          <div>
+            <input
+              type='hidden'
+              name={name}
+              value={isChecked ? 'true' : 'false'}
```
geclos, line 107: this makes switch toggle compatible with regular form actions

packages/web-ui/src/ds/molecules/MultiSelect/index.tsx
andresgutgon: Check Select I don't think it is necessary. It's working fine

PR #1032, opened 2025-03-26
Run email triggers in jobs

csansoon:
Mailgun is re-triggering the webhook because it is not receiving a successful response in time (the request probably times out).

I moved the execution logic to a dedicated job. When the webhook is received, we just validate the info required to run the prompt, enqueue the job, and then return a successful response. This way the document is executed in the background and Mailgun is happy.

### DIFF
--- a/packages/core/src/services/documentTriggers/handlers/email/index.ts
+++ b/packages/core/src/services/documentTriggers/handlers/email/index.ts
@@ -10,11 +10,14 @@ import {
   EMAIL_TRIGGER_DOMAIN,
 } from '@latitude-data/constants'
 import { database } from '../../../../client'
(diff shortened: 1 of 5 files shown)


packages/core/src/services/documentTriggers/handlers/email/index.ts
```diff
@@ -110,56 +134,37 @@ export async function handleEmailTrigger(
...
+    senderEmail: senderEmail,
+    senderName: senderName,
+    messageId: messageId,
+    parentMessageIds: parentMessageIds,
+    subject: subject,
+    body: body,
+    attachments: uploadedFiles,
+  }
+
+  await jobQueues.defaultQueue.jobs.enqueueRunDocumentJob(runJobData)
```
csansoon, line 167: Job eunque logic

PR #1036, opened 2025-03-26
Allow mark datasets as golden

andresgutgon:
# What?
image


andresgutgon:
At the moment we're not going to implement it

PR #1039, opened 2025-03-27
Add semantic similarity metric

neoxelox:


apps/web/src/components/evaluations/rule/SemanticSimilarity.tsx
```diff
@@ -0,0 +1,159 @@
...
+}
+
+function ResultPanelMetadata(
+  _props: ResultPanelProps<
+    EvaluationType.Rule,
+    RuleEvaluationMetric.SemanticSimilarity
+  >,
+) {
+  return <></>
+}
```
andresgutgon, line 127: Instead of this I would do a fallback in the place where these components are used. And in the spec for this one set it to `null`

Issue #1028, opened 2025-03-25
Fix datasets V2 playground parameters race condition


andresgutgon:
# What?
We need to improve the way dataset parameters are loaded in playground. New datasets load the dataset but also the linked row. In all datasets this data was stored in local storage but now is fetched dynamically when the datasets parameters are loaded. 

## Solution
Maybe we need to load from local storage as a fallback while the fresh data come from the DB

### TODO
- [x] Fix raise condition loading playground on dataset tab selected
- [x] When doing batch run the mapping of the selected row is not there 🐛 
- [x] When navigating between rows in playground the values are not reflected on prompt preview.

### Frontend failed parsing of numbers as dates

<img width="304" alt="Image" src="https://github.com/user-attachments/assets/699f787f-d328-4ab0-9ce7-08a62b44b18e" />

PR #1040, opened 2025-03-27
Fix race conditions in the way we load dataset parameters on playground

andresgutgon:
# What?
I did this in a way that data is not persisted also in local storage, and this makes all the UI behave jumpy. Now, we also persist in localStorage, as with old datasets. I also tried to do an implementation where datasetV2 and dataset rows are not fetched until the user clicks on dataset parameters, but this provokes a race condition document emitting metadata parameters and datasets not being set correctly.

## Issue
https://github.com/latitude-dev/latitude-llm/issues/1028

## TODO
- [x] Store in local storage mapped inputs and resolved inputs for dataset parameters.
- [x] 🐛 Navigating between rows provoke infinite loop
- [x] 🐛 **SELECTING DATASET** should work when was previously selected or is the first time. Keep previous inputs and mappedInputs.
- [x] `useMetadata` is remounted many times and that provokes being `undefined` some times.
- [x] How to keep previous inputs when user comment those inputs in the prompt? 🧠 
- [x] 🐛 Changing parameter type erase data from preview
- [x] 🐛 Not matching columns on run document batch from dataset selected in parameters
- [x] 🐛 Not matching columns on run batch evaluation V1
- [x] 🐛 Not matching columns on run batch evaluation V2
- [x] Avoid wrapping in quotes when running a batch evaluation or document 
- [x] Review performance of writing on the playground editor

### Next PR
- [x] ~🔥 🔥 MOVE `dataset_rows` table to latitude schema 🔥 🔥~ All good by default is latitude schema. uff
- [ ] Review creating a dataset from a CSV 
- [ ] Review / fix generate dataset with AI with JSON columns
- [ ] 🐛 Make full dataset row table linkable
- [ ] 🐛 fix download logs. [Not working](https://app.latitude.so/projects/22/versions/live/documents/77035bac-871c-4cae-86e5-bfd6847fbb6e/logs)
- [ ] Review creating a dataset with AI.
- [ ] 🔥 Datasets V1 code + feature flags
- [ ] Rename dataset v2 to dataset
- [ ] [Review datasets documentation](https://docs.latitude.so/guides/datasets/overview#datasets)


geclos:
s/raise/race (updated the pr header)

apps/web/src/hooks/useDocumentParameters/index.ts
```diff
@@ -418,65 +420,70 @@ export function useDocumentParameters<
...
-        'manual',
-        recalculateInputs({
-          inputs: inputs.manual.inputs,
-          metadata,
-        }),
-      )
+      setValue((oldState) => {
+        const { doc } = getDocState(oldState, key)
+        console.log('MANUAL_INPUTS', doc.manual.inputs)
+        return updateInputsState({
```
andresgutgon, line 433: Here is where prev state is lost when someone comments a parameter in the prompt. Have to find a way of persisting prevState to recover and they restore that parameter

apps/web/src/hooks/useDocumentParameters/index.ts
```diff
@@ -418,65 +420,70 @@ export function useDocumentParameters<
...
-    ],
+    [key, document.datasetId, setValue],
   )
 
+  useEffect(() => {
+    if (!metadata) return
+
+    setParameters(Array.from(metadata.parameters))
+    onMetadataChange(metadata)
+  }, [metadata, setParameters])
```
andresgutgon, line 482: This is major change. Before this hook was exporting `onMetadataChange` and this function had a lot of dependencies that was making `useMetadata` being re-rendered and `metadata` was lost. I hate using `useEffect` but I think is better this way.

apps/web/src/stores/apiKeys.ts
```diff
@@ -10,7 +10,7 @@ import useSWR, { SWRConfiguration } from 'swr'
 export default function useApiKeys(opts?: SWRConfiguration) {
   const { toast } = useToast()
   const key = 'api/apiKeys'
-  const fetcher = useFetcher(ROUTES.api.apiKeys.root)
+  const fetcher = useFetcher<ApiKey[]>(ROUTES.api.apiKeys.root)
```
andresgutgon, line 13: Lot of changes in this PR are related with improving types in `useFecher`. I wanted to extract what is inside and also being typed independent of the store.

apps/web/src/stores/evaluationResultsV2/byDocumentLogs.ts
```diff
@@ -38,11 +38,10 @@ export default function useEvaluationResultsV2ByDocumentLogs(
...
   }, [documentLogUuids])
-  const fetcher = useFetcher(`${route}?${query}`)
-
-  const { data = {}, ...rest } = useSWR<
-    Record<string, ResultWithEvaluationV2[]>
-  >(
+  const fetcher = useFetcher<Record<string, ResultWithEvaluationV2[]>>(
+    `${route}?${query}`,
+  )
+  const { data = {}, ...rest } = useSWR(
```
neoxelox, line 44: In the other stores you did not remove the type from "useSWR", why remove it here?

apps/web/src/stores/evaluationResultsV2/index.ts
```diff
@@ -46,9 +46,9 @@ export function useEvaluationResultsV2<
     () => (search ? evaluationResultsV2SearchToQueryParams(search) : ''),
     [search],
   )
-  const fetcher = useFetcher(`${route}?${query}`)
+  const fetcher = useFetcher<EvaluationResultV2<T, M>[]>(`${route}?${query}`)
 
-  const { data = [], ...rest } = useSWR<EvaluationResultV2<T, M>[]>(
+  const { data = [], ...rest } = useSWR(
```
neoxelox, line 51: Same as above

packages/core/src/repositories/datasetRowsRepository.ts
```diff
@@ -78,7 +76,7 @@ export class DatasetRowsRepository extends Repository<DatasetRow> {
       .from(datasetRows)
       .where(and(this.scopeFilter, eq(datasetRows.datasetId, datasetId)))
       .limit(limit)
-      .orderBy(desc(datasetRows.createdAt))
+      .orderBy(desc(datasetRows.createdAt), desc(datasetRows.id))
```
neoxelox, line 79: Cannot comment on that line but `findManyByDataset` is missing this orderBy

packages/core/src/repositories/datasetRowsRepository.ts
```diff
@@ -91,25 +89,13 @@ export class DatasetRowsRepository extends Repository<DatasetRow> {
...
-
-    const row = rows[0]
+    const rowResult = await this.find(datasetRowId)
 
-    if (!row) {
-      return Result.error(
-        new Error(`Dataset row not found with id ${datasetRowId}`),
-      )
+    if (rowResult.error) {
+      return { position: 0, page: 1 }
```
neoxelox, line 95: This should be position 1 as the query is greater or EQUAL

packages/core/src/repositories/datasetRowsRepository.ts
```diff
@@ -119,15 +105,14 @@ export class DatasetRowsRepository extends Repository<DatasetRow> {
         and(
           this.scopeFilter,
           eq(datasetRows.datasetId, datasetId),
-          lt(datasetRows.id, row.id),
+          sql`(${datasetRows.id}, ${datasetRows.createdAt}) >= (${row.id}, ${new Date(row.createdAt).toISOString()})`,
```
neoxelox, line 108: This MUST be (createdAt, id) >= NOT the other way around. You are ordering by created at and then by id (if the created at are the same)!!!

packages/web-ui/src/ds/molecules/DocumentTextEditor/Editor/index.tsx
```diff
@@ -20,14 +20,15 @@ import { RegularMonacoEditor } from './RegularEditor'
 export function DocumentTextEditor({
   value,
   path,
-  metadata,
```
andresgutgon, line 23: I was trying to find out why Monaco is slow in dev and I didn't but I saw we're just using errors from metadata so I changed this prop

PR #1048, opened 2025-03-31
Fix dataset rows schema and add dataset info to eval results

neoxelox:


andresgutgon (approved):
I think this is ok 🤞

geclos (approved):

PR #1068, opened 2025-04-01
feature: improve package management and docker image sizes

geclos:
We've always had a bad setup regarding dependencies in which we installed all dependencies in the root package.json which caused many unnecessary dependencies to be bundled with apps that did not need them. This commit fixes that by moving all dependencies out of the workspace package.json and into the individual apps/packages package.json.

Additionally, I've also split all the exports from web-ui package instead of having single index.ts files that bundled all the exports. This results in much faster nextjs compile times and smaller build sizes.

If you care about details. The changes in this commit in terms of docker image sizes are:

Web: 4.5gb -> 1.4gb
Gateway: 3.5gb -> 1.09gb
Workers: no change (1.4gb aprox)


apps/web/scripts/update-web-ui-imports.ts
```diff
@@ -0,0 +1,361 @@
...
+      console.error(`Error processing file ${file}:`, error)
+      continue // Skip this file and continue with the next one
+    }
+  }
+}
+
+updateImports().catch((error) => {
+  console.error('Error updating imports:', error)
+  process.exit(1)
+})
```
geclos, line 361: these are some recast scripts i've used to migrate the imports not sure if it's valuable to include them in the repo for future reference

apps/web/src/app/(admin)/backoffice/_components/BackofficeTabs/index.tsx
```diff
@@ -2,7 +2,7 @@
 
 import { ReactNode } from 'react'
 
-import { TabSelector } from '@latitude-data/web-ui'
+import { TabSelector } from '@latitude-data/web-ui/molecules/TabSelector'
```
geclos, line 5: instead of exporting a index.ts file we export each component individually which allows nextjs to tree shake the chunks and results also in faster dev times

apps/web/scripts/update-web-ui-imports.ts
andresgutgon: I wouldn't

apps/web/src/app/(admin)/backoffice/_components/BackofficeTabs/index.tsx
andresgutgon: Nice

andresgutgon: So in the end we keep not compiling packages separately no?

geclos: yup, no compilation required since we don't have workspace packages anymore

PR #1073, opened 2025-04-02
Feature/remove compiler

geclos:
Compiler is one of the packages that takes the longest to compile and we
are only keeping it around for support of old prompt syntax. It's time
we deprecate it.


packages/constants/src/events/compiler/index.ts
```diff
@@ -0,0 +1,213 @@
...
+
+export type ElseBlock = BaseNode & {
+  type: 'ElseBlock'
+}
+
+export type IfBlock = BaseNode & {
+  type: 'IfBlock'
+  expression: unknown
+  else: ElseBlock | null
+}
```
geclos, line 213: i've moved the legacy compiler types to constants

PR #1077, opened 2025-04-02
Fix generating a dataset from a CSV o text coming from AI that contains JSON columns

andresgutgon:
# What?
Before, it was not possible to use our generate dataset feature for generating datasets with JSON columns.

## What?
- [x] ~~Review creating a dataset from a CSV~~ Not done. Is not CSV generation's fault
- [x] Review / fix generate dataset with AI with JSON columns
- [x] Review creating a dataset with AI.
- [x] 🐛 Make full dataset row table linkable

## Sentry issue
[Invalid Closing Quote: found non trimable byte after quote at line 2](https://latitude-l5.sentry.io/issues/6500286520)


packages/core/src/lib/readCsv/castCells.ts
```diff
@@ -52,37 +53,73 @@ function tryParseDate({
   return null
 }
 
+export function sanitizeJsonInCsv(csv: string): string {
+  return csv.replace(/({[^}]*})/g, (match) => `"${match.replace(/"/g, '""')}"`)
```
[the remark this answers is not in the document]
andresgutgon: Oh well

packages/core/src/lib/documentPersistedInputs.ts
```diff
@@ -67,7 +67,7 @@ export type PlaygroundInputs<S extends InputSource> = {
   }
   // DEPRECATED: Remove after a while
   dataset: LinkedDataset & { datasetId: number | undefined }
-  datasetV2: Record<number, LinkedDatasetRow>
+  datasetV2: Record<number, LinkedDatasetRow> | undefined
```
andresgutgon, line 70: The fix for the bug on local storage is not to lie here and assume users can have `undefined` this. After changing this I fixed all the places where this key was used.

packages/core/src/lib/readCsv/index.test.ts
```diff
@@ -0,0 +1,116 @@
...
+              email: 'alicej@example.com',
+              age: 25,
+              gender: 'female',
+            },
+          },
+          info: expectedCols,
+        },
+      ],
+    })
+  })
```
geclos, line 57: i would add some extra test cases maybe you can tell ai to generate some complex jsons with strings that have html escaping and what not, just to test it's robust.

packages/core/src/lib/readCsv/castCells.ts
```diff
@@ -52,37 +53,129 @@ function tryParseDate({
...
+      if (
+        (trimmed.startsWith('{') && trimmed.endsWith('}')) ||
+        (trimmed.startsWith('[') && trimmed.endsWith(']'))
+      ) {
+        return `"${trimmed.replace(/"/g, '""')}"`
+      }
+      return trimmed
+    })
+    .join(',')
+}
```
geclos, line 103: surprised there's no library for this

geclos (approved):
ok but let's add a bit more tests

neoxelox (changes requested):
I wouldn't touch the client CSV as they are responsible to generate a correct CSV with the JSON column scaped, or any other string scaped if it contains newlines or commas or whatever, look at what excel generates when puting a JSON in a column:

```csv
name,payload
Andres,"{""username"":""johndoe"",""email"":""johndoe@example.com"",""age"":28,""gender"":""male""}"
```

excel (the client) is responsible to generate a correct CSV

packages/core/src/lib/readCsv/castCells.ts
neoxelox: we shouldnt do this, its gonna break 101%

andresgutgon:
Agree, bad approach. I will change

packages/core/src/lib/readCsv/castCells.ts
andresgutgon: Not doing it

marfal:
Hey, any chance you'll be able to also export evaluations? I saw the cloud version already greatly improved export capabilities, but being able to export connected evaluations would be fantastic, as you could also later feed this to another system and do a more through qualitative and quantitative analysis of your prompts.

neoxelox:
> Hey, any chance you'll be able to also export evaluations? I saw the cloud version already greatly improved export capabilities, but being able to export connected evaluations would be fantastic, as you could also later feed this to another system and do a more through qualitative and quantitative analysis of your prompts.

Hey @marfal , sorry for the late response. I'm not sure what do you mean by "exporting connected evaluations". Could you create an issue explaining it a bit more? Thanks 👍

packages/core/src/services/datasetsV2/createFromJson.ts
```diff
@@ -0,0 +1,70 @@
...
+function generateRowsFromJson({
+  columns,
+  rows,
+}: {
+  columns: Column[]
+  rows: Record<string, unknown>[]
+}) {
+  return rows.map((row) =>
+    columns.reduce((acc, col) => {
+      const cellValue = row[col.name] ?? ''
```
neoxelox, line 19: In `packages/core/src/services/datasetRows/generatePreviewRowsFromJson.ts` you used `parseRowCell`, should you do that here too?

neoxelox (approved):

packages/core/src/services/datasetsV2/createFromJson.ts
andresgutgon: No. Preview goes straight to visualize the content in the table. This goes to DB to the JSONB column in the rows and we want to store original content: number, string, object,...

PR #1081, opened 2025-04-03
fix: errors from deleted prompts

csansoon:
If a prompt is deleted in a draft and it contained errors, the draft cannot be published because "a prompt contains errors" (but this prompt has been deleted...)

Steps to reproduce:
 - Create a prompt "a"
 - Create a prompt "b", importing "a".
 - Publish the version
 - Create a new draft
 - Rename the prompt "a" to something else. Now the prompt "b" contains errors because the import is wrong.
 - Delete the prompt "b"
 - Try to publish


andresgutgon (changes requested):
Little test & +1

PR #1095, opened 2025-04-04
Experiments schema

csansoon:
## EXPERIMENTS DATA SCHEMA

New Experiments table! Properties:
- `name`: Optional string to name and identify experiments
- `commitId`
- `evaluationUuids`
- `documentUuid`
- `datasetId`
- `metadata`: Configuration about the experiment execution, but not used for indexing:
  - `resolvedPrompt`
  - `promptHash`
  - `count`: Total number of logs to be generated
- `startedAt`: Time of start. Can be undefined if experiment is still pending
- `finishedAt`: Time of finish. Can be undefined if experiment is still pending or running

@neoxelox I did not add the `status`, this info can be extracted from `startedAt`, `finishedAt` and `results` :)

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/LogSourceFilter/index.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/LogSourceFilter/index.tsx
@@ -16,6 +16,7 @@ const LogSourceLabel: { [key in LogSources]: string } = {
   [LogSources.AgentAsTool]: 'Sub agents',
   [LogSources.EmailTrigger]: 'Email trigger',
   [LogSources.ScheduledTrigger]: 'Scheduled trigger',
+  [LogSources.Experiment]: 'Experiment',
 }
 
 function LogSourceCheckbox({
--- a/packages/constants/src/index.ts
+++ b/packages/constants/src/index.ts
@@ -1,15 +1,19 @@
 export enum LogSources {
   API = 'api',
   Playground = 'playground',
-  Evaluation = 'evaluation',
+  Evaluation = 'evaluation', // This is actually for "batch" evaluations (now Experiments)
+  Experiment = 'experiment',
   User = 'user',
   SharedPrompt = 'shared_prompt',
   AgentAsTool = 'agent_as_tool',
   EmailTrigger = 'email_trigger',
   ScheduledTrigger = 'scheduled_trigger',
 }
+export const NON_LIVE_EVALUABLE_LOG_SOURCES = [
+  LogSources.Evaluation,
+  LogSources.Experiment,
+]
 
-export const NON_LIVE_EVALUABLE_LOG_SOURCES = [LogSources.Evaluation]
 export enum Providers {
   OpenAI = 'openai',
   Anthropic = 'anthropic',
@@ -37,6 +41,12 @@ export enum DocumentTriggerParameters {
   Subject = 'subject',
   Body = 'body',
   Attachments = 'attachments',
+}
+
+export type ExperimentMetadata = {
+  resolvedPrompt: string
+  promptHash: string
+  count: number // Total number of to generate logs in the experiment
 }
 
 export * from './models'
--- a/packages/core/src/schema/models/documentLogs.ts
+++ b/packages/core/src/schema/models/documentLogs.ts
@@ -12,6 +12,7 @@ import { latitudeSchema } from '../db-schema'
 import { timestamps } from '../schemaHelpers'
 import { commits } from './commits'
 import { logSourcesEnum } from './providerLogs'
+import { LogSources } from '@latitude-data/constants'
 
 export const documentLogs = latitudeSchema.table(
   'document_logs',
@@ -30,7 +31,7 @@ export const documentLogs = latitudeSchema.table(
     parameters: jsonb('parameters').$type<Record<string, unknown>>().notNull(),
     customIdentifier: text('custom_identifier'),
     duration: bigint('duration', { mode: 'number' }),
-    source: logSourcesEnum('source'),
+    source: logSourcesEnum('source').$type<LogSources>(),
     ...timestamps(),
   },
   (table) => ({
--- a/packages/core/src/schema/models/evaluationResultsV2.ts
+++ b/packages/core/src/schema/models/evaluationResultsV2.ts
@@ -17,6 +17,7 @@ import { datasetRows } from './datasetRows'
 import { datasetsV2 } from './datasetsV2'
 import { providerLogs } from './providerLogs'
 import { workspaces } from './workspaces'
+import { experiments } from './experiments'
 
 export const evaluationResultsV2 = latitudeSchema.table(
   'evaluation_results_v2',
@@ -30,8 +31,10 @@ export const evaluationResultsV2 = latitudeSchema.table(
       .notNull()
       .references(() => commits.id, { onDelete: 'restrict' }),
     evaluationUuid: uuid('evaluation_uuid').notNull(),
-    experimentId: bigint('experiment_id', { mode: 'number' }),
-    // .references(() => experiments.id, { onDelete: 'restrict' }), // TODO: Add this when experiment table is created
+    experimentId: bigint('experiment_id', { mode: 'number' }).references(
+      () => experiments.id,
+      { onDelete: 'restrict', onUpdate: 'cascade' },
+    ),
     datasetId: bigint('dataset_id', { mode: 'number' }).references(
       () => datasetsV2.id,
       { onDelete: 'set null' },
--- a/packages/core/src/schema/models/experiments.ts
+++ b/packages/core/src/schema/models/experiments.ts
@@ -1,0 +1,53 @@
+import {
+  bigint,
+  bigserial,
+  index,
+  jsonb,
+  timestamp,
+  uuid,
+  varchar,
+} from 'drizzle-orm/pg-core'
+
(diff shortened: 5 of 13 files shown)


packages/core/src/schema/models/providerLogs.ts
```diff
@@ -18,17 +18,12 @@ import { latitudeSchema } from '../db-schema'
...
-  LogSources.User,
-  LogSources.SharedPrompt,
-  LogSources.AgentAsTool,
-  LogSources.EmailTrigger,
-  LogSources.ScheduledTrigger,
-])
+export const logSourcesEnum = latitudeSchema.enum(
+  'log_source',
+  Object.values(LogSources) as [string, ...string[]],
+)
```
csansoon, line 26: Automatically update the sql enum with the typescript enum values

packages/constants/src/index.ts
```diff
@@ -39,6 +40,18 @@ export enum DocumentTriggerParameters {
   Attachments = 'attachments',
 }
 
+export type ExperimentMetadata = {
+  resolvedPrompt: string
+  evaluationIds: number[]
```
neoxelox, line 45: not evaluation ids no?

packages/constants/src/index.ts
```diff
@@ -39,6 +40,18 @@ export enum DocumentTriggerParameters {
...
+  threshold: number // Percentage of logs with "passed" evaluations to consider the experiment passed
```
neoxelox, line 46: this should be in the model

packages/core/src/schema/models/experiments.ts
```diff
@@ -0,0 +1,44 @@
...
+import { timestamps } from '../schemaHelpers'
+import { commits } from './commits'
+import { ExperimentMetadata, ExperimentResults } from '@latitude-data/constants'
+import { datasetsV2 } from './datasetsV2'
+
+export const experiments = latitudeSchema.table(
+  'experiments',
+  {
+    id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
+    name: varchar('name', { length: 256 }),
```
andresgutgon, line 21: Why not mandatory and we autogenerate the name if not pased by the user?

packages/constants/src/index.ts
neoxelox: should be uuids for evals 2, anyway, this should be in the model not in metadata

packages/constants/src/index.ts
```diff
@@ -39,6 +40,18 @@ export enum DocumentTriggerParameters {
...
+  count: number // Total number of to generate logs in the experiment
+}
+
+export type ExperimentResults = {
```
neoxelox, line 50: where are u planing on using this?

packages/core/src/schema/models/experiments.ts
```diff
@@ -0,0 +1,44 @@
...
+    documentUuid: uuid('document_uuid').notNull(),
+    datasetId: bigint('dataset_id', { mode: 'number' }).references(
+      () => datasetsV2.id,
+      {
+        onDelete: 'restrict',
+        onUpdate: 'cascade',
+      },
+    ),
+    metadata: jsonb('metadata').$type<ExperimentMetadata>().notNull(),
+    results: jsonb('results').$type<ExperimentResults>(),
```
neoxelox, line 34: updating this could create a bottleneck for big experiments, because every result must select for update the same row in db... causing starvation

neoxelox (changes requested):
comments

packages/constants/src/index.ts
csansoon: Why in the model? It will not be used as a filter anyways, we would only need this value when performing an actual SELECT.

I can move it though, no issue. Just to be in the same page

packages/core/src/schema/models/experiments.ts
csansoon: What names would we use?

packages/core/src/schema/models/experiments.ts
csansoon: nono, it will only be generated when the experiments finish.

packages/constants/src/index.ts
neoxelox: semantics, threshold is not metadata, is part of the experiment configuration

packages/core/src/schema/models/experiments.ts
neoxelox: maybe with only counting results should be enough ? is fast aggregation

neoxelox: what i would add is a hasPassed boolean column, so we know in an instant that experiment has passed

packages/constants/src/index.ts
csansoon: Added it as a separate table (for *-* relations), but still used IDs instead of UUIDs, this is only for internal use

packages/core/src/schema/models/experiments.ts
andresgutgon: `Experiment #1` and being `1` the next number of experiments. Something like that?

packages/core/src/schema/models/documentLogs.ts
```diff
@@ -29,8 +31,15 @@ export const documentLogs = latitudeSchema.table(
     contentHash: text('content_hash').notNull(),
     parameters: jsonb('parameters').$type<Record<string, unknown>>().notNull(),
     customIdentifier: text('custom_identifier'),
+    experimentId: bigint('experiment_id', { mode: 'number' }).references(
```
andresgutgon, line 34: Should we add an index here?

packages/core/src/schema/models/evaluationResultsV2.ts
```diff
@@ -30,8 +31,10 @@ export const evaluationResultsV2 = latitudeSchema.table(
       .notNull()
       .references(() => commits.id, { onDelete: 'restrict' }),
     evaluationUuid: uuid('evaluation_uuid').notNull(),
-    experimentId: bigint('experiment_id', { mode: 'number' }),
-    // .references(() => experiments.id, { onDelete: 'restrict' }), // TODO: Add this when experiment table is created
+    experimentId: bigint('experiment_id', { mode: 'number' }).references(
```
andresgutgon, line 34: Same, makes sense an index?

packages/core/src/schema/models/experiments.ts
andresgutgon: I would make it a timestamp maybe `passed_at` this way we can know when it passed. Maybe not but could be handy

neoxelox: i mean, when it passes or not when it finishes

packages/core/src/schema/models/providerLogs.ts
```diff
@@ -37,6 +32,13 @@ export const providerLogs = latitudeSchema.table(
     workspaceId: bigint('workspace_id', { mode: 'number' }),
     uuid: uuid('uuid').notNull().unique(),
     documentLogUuid: uuid('document_log_uuid'),
+    experimentId: bigint('experiment_id', { mode: 'number' }).references(
```
neoxelox, line 35: Add an index here, where are gonna join with this table probably?

packages/core/src/schema/models/evaluationResultsV2.ts
neoxelox: index was already there

packages/core/src/schema/models/documentLogs.ts
neoxelox: ye, if we are gonna join or filter

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/LogSourceFilter/index.tsx
```diff
@@ -16,6 +16,7 @@ const LogSourceLabel: { [key in LogSources]: string } = {
   [LogSources.AgentAsTool]: 'Sub agents',
   [LogSources.EmailTrigger]: 'Email trigger',
   [LogSources.ScheduledTrigger]: 'Scheduled trigger',
+  [LogSources.Experiment]: 'Experiment',
```
Comment (neoxelox) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/LogSourceFilter/index.tsx:19]: Add this to "NON_LIVE_EVALUABLE_LOG_SOURCES" (and if you can move that constant to `packages/constants/src/evaluations/index.ts`, I dont know why is not in there)

packages/constants/src/index.ts
csansoon: changed my mind, added it as an array instead 👍

packages/constants/src/index.ts
csansoon: welp, no threshold for now

packages/core/src/schema/models/providerLogs.ts
csansoon: removed the attribute for now

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/LogSourceFilter/index.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/Filters/LogSourceFilter/index.tsx:19]: Added 👌 

Left the constant where it is. I like that it's right under the enum definition, and the enum should not be on "evaluations" either 'cause it's about logs instead

packages/core/src/schema/models/documentLogs.ts
csansoon: removed the attribute for now

neoxelox (approved):

PR #1099, opened 2025-04-06
Send provider options camel cased to Vercel SDK

andresgutgon:
# What?
Vercel SDK expect provider options camelCase. Then they snake case the same options before passing to the LLMs. For example Anthropic.

So the way this works is this:
1. User reads Anthropic's documentation and sees `"thinking: { "budget_tokens": 33 }"`
2. Latitude sends to Vercel AI SDK `"thinking: { "budgetTokens": 33 }"`
3. Vercel SDK sends to Anthropic `"thinking: { "budget_tokens": 33 }"`

🤡


packages/core/src/services/ai/providers/rules/vercel.test.ts
```diff
@@ -252,9 +252,9 @@ describe('applyVercelSdkRules', () => {
         content: 'I am good',
         providerOptions: {
           anthropic: {
-            some_attribute: 'some_system_value',
+            someAttribute: 'some_system_value',
             cacheControl: { type: 'ephemeral' },
-            another_attribute: { another_system: 'value' },
+            anotherAttribute: { anotherSystem: 'value' },
```
andresgutgon, line 257: ☝️ Before we were only camel casing `cacheControl` on messages. Now we camel case everthing that goes inside `providerOptions.anthropic`. 

One open question I have is if we should do the same for the rest of providers.

I'm only fixing this for now. No one complained about other providers afaik.

andresgutgon: I see this docs that implies camel case 
https://sdk.vercel.ai/providers/openai-compatible-providers?utm_source=chatgpt.com#provider-specific-options

But not sure. Sending everything in camel case for messages is an easy change. 

Let me know

packages/core/src/services/ai/index.ts
```diff
@@ -153,7 +153,7 @@ export async function ai({
         providerAdapter(model, {
           cacheControl: config.cacheControl ?? false,
           // Propagate provider config options for this language model
-          ...config,
+          ...omit(config, ['providerOptions']),
         })
```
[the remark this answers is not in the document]
geclos: good point

packages/core/src/services/ai/providers/rules/index.test.ts
```diff
@@ -34,4 +34,36 @@ describe('rules', () => {
...
+        providerOptions: {
+          anthropic: {
+            model: 'claude-3-7-sonnet-latest',
+            thinking: { type: 'enabled', budgetTokens: 1024 },
+          },
+        },
+      },
+    })
+  })
 })
```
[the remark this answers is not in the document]
geclos: good point

packages/core/src/services/ai/providers/rules/vercel.test.ts
csansoon: Yeah I'd do that for any `providerOptions`, that's what vercel apparently expects

csansoon (approved):

packages/core/src/services/ai/providers/rules/vercel.test.ts
andresgutgon: Done

PR #1121, opened 2025-04-09
feat: SSO with Google

geclos:
The only SSO you really need


apps/web/src/app/(public)/login/LoginForm/index.tsx
```diff
@@ -40,9 +44,27 @@ export default function LoginForm({
...
+          <div className='relative'>
+            <Separator />
+            <div className='absolute left-1/2 top-1/2 -translate-x-1/2 -translate-y-1/2'>
+              <div className='bg-background px-2'>
+                <Text.H6 color='foregroundMuted'>Or</Text.H6>
+              </div>
+            </div>
+          </div>
+
+          <Button variant='outline' fullWidth asChild>
```
andresgutgon, line 61: You have `iconProps` no need to use `<Icon />` inside buttons

apps/web/src/app/(public)/setup/SetupForm/index.tsx
```diff
@@ -70,9 +74,27 @@ export default function SetupForm({
...
+            <div className='absolute left-1/2 top-1/2 -translate-x-1/2 -translate-y-1/2'>
+              <div className='bg-background px-2'>
+                <Text.H6 color='foregroundMuted'>Or</Text.H6>
+              </div>
+            </div>
+          </div>
+
+          <Button variant='outline' fullWidth asChild>
+            <Link href='/api/auth/google/start' className='flex items-center'>
+              <Icon name='googleWorkspace' className='mr-2 h-4 w-4' />
```
andresgutgon, line 93: Same

apps/web/src/app/api/auth/google/start/route.ts
```diff
@@ -0,0 +1,39 @@
...
+    maxAge: 60 * 10, // 10 minutes
+    sameSite: 'lax',
+  })
+  cookieStore.set('google_code_verifier', codeVerifier, {
+    path: '/',
+    httpOnly: true,
+    secure: env.NODE_ENV === 'production',
+    maxAge: 60 * 10, // 10 minutes
+    sameSite: 'lax',
+  })
```
andresgutgon, line 31: All this cookie management put in the same place we have the rest

packages/core/src/services/auth/findOrCreateUserFromOAuth.ts
```diff
@@ -0,0 +1,138 @@
...
+ * Always ensures the OAuth account is linked.
+ * Returns the full User and Workspace objects.
+ */
+export function findOrCreateUserFromOAuth(
+  { providerId, providerUserId, email, name }: FindOrCreateUserFromOAuthInput,
+  db = database,
+) {
+  // Let TypeScript infer the return type of the transaction
+  return Transaction.call(async (tx) => {
+    // Helper function to get user and their first workspace
```
andresgutgon, line 28: Lot of comments from AI I guess

packages/core/src/services/auth/findOrCreateUserFromOAuth.ts
```diff
@@ -0,0 +1,138 @@
...
+    })
+
+    return Result.ok({
+      user: newUser,
+      // Explicitly cast or map properties if needed
+      workspace: newWorkspaceDto as Workspace, // Cast Dto to base Workspace type
+      isNewUser: true,
+    })
+  }, db)
+}
```
andresgutgon, line 138: Tests

apps/web/src/app/api/auth/google/start/route.ts
geclos: why? no i think thios is fine here it's just a temporary cookie for the google auth dance it does not make sense to abstract it to a service as it will never be reused

packages/core/src/services/auth/findOrCreateUserFromOAuth.ts
geclos: yeah been trying gemini, it's a bit pesado with the comments

apps/web/src/app/(public)/login/LoginForm/index.tsx
geclos: nah because it's full width it does not do it the way i want it

packages/core/src/services/auth/findOrCreateUserFromOAuth.ts
geclos: done

PR #1123, opened 2025-04-09
🔥 Remove dataset v1 leftovers

andresgutgon:
# What?
- [x] Remove old code 🔥
- [x] Rename datasetsV2 to just `datasets`


packages/core/src/schema/index.ts
```diff
@@ -1,5 +1,8 @@
 export * from './db-schema'
 
+// Legacy models
+export * from './legacyModels/datasetsV1'
```
andresgutgon, line 4: I kept old datasets table alive so we can access data for now

geclos: good

PR #1132, opened 2025-04-10
feat: remove lib index.ts

geclos:
it compiles too much stuff into a single import


packages/core/package.json
```diff
@@ -130,8 +130,10 @@
     "pg-transactional-tests": "^1.0.9",
     "supertest": "^7.0.0",
     "svelte": "^4.2.19",
+    "ts-morph": "^25.0.1",
+    "ts-node": "^10.9.2",
```
andresgutgon, line 134: why this? what does ts-morph. also we have svelte here. Do we need it?

geclos: it's like recast used for large refactors

apps/web/src/actions/admin/documentTriggers/manualTrigger/email.ts
```diff
@@ -4,7 +4,7 @@ import { z } from 'zod'
 
 import { withAdmin } from '../../../procedures'
 import { handleEmailTrigger } from '../../../../../../../packages/core/src/services/documentTriggers/handlers/email'
-import { env } from 'process'
+import { env } from '@latitude-data/env'
```
geclos, line 7: @latitude-dev/developers careful with these pleaase

packages/core/package.json
andresgutgon: How did you use in this case? Maybe a note in the README that we have this now at our disposal

PR #1156, opened 2025-04-16
Feature/onboarding

geclos:


packages/core/drizzle/0159_easy_madripoor.sql
andresgutgon, line 1: Redo the migrations too much shit in each migration 🙏

geclos:
will fix tests in next commit

PR #1161, opened 2025-04-21
feat: create evaluation in onboarding document

geclos:


apps/web/src/services/user/setupService.test.ts
```diff
@@ -96,6 +98,15 @@ describe('setupService', () => {
       })
     expect(createdProviderApiKey).toBeDefined()
     expect(createdProviderApiKey?.authorId).toBe(user.id)
+
+    // Check if onboarding evaluation was created
+    const createdEvaluation = await database.query.evaluationVersions.findFirst(
+      {
+        // @ts-expect-error - drizzle-orm types are not up to date
+        where: eq(evaluationVersions.workspaceId, workspace.id),
```
andresgutgon, line 106: You can avoid this by importing `eq` from `packages/core` from here:
https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/client/utils.ts

apps/web/src/services/user/setupService.test.ts
```diff
@@ -166,5 +175,14 @@ describe('setupService', () => {
       .where(eq(commits.projectId, importedProject!.id))
     expect(importedDocuments.length).toBe(1)
     expect(importedDocuments[0]!.document_versions.content).toEqual(prompt)
+
+    // Check if onboarding evaluation was created
+    const createdEvaluation = await database.query.evaluationVersions.findFirst(
+      {
+        // @ts-expect-error - drizzle-orm types are not up to date
+        where: eq(evaluationVersions.workspaceId, workspace.id),
```
andresgutgon, line 183: ditto

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DocumentLogsTable.tsx
```diff
@@ -260,11 +259,6 @@ export const DocumentLogsTable = forwardRef<HTMLTableElement, Props>(
                     {capitalize(documentLog.source || '-')}
                   </Text.H5>
                 </TableCell>
-                <TableCell>
-                  <Text.H5 color={cellColor}>
-                    {documentLog.customIdentifier || '-'}
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DocumentLogsTable.tsx:265]: Why?

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DocumentLogsTable.tsx:265]: it's useless 99% of the time

Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DocumentLogsTable.tsx:265]: I think Carlos was using it when working with agents. I don't think is useless. The only improvement is adding ellipsis and noWrap. But we can open the discussion on Slack.

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DocumentLogsTable.tsx:265]: i'm not removing the info just moving it to metadata, it definitely does not need to be in the table considering 99% of users don't use this value

PR #1165, opened 2025-04-22
feat: allow toggle evals v2 in head commit

geclos:


packages/core/src/services/evaluationsV2/toggleLiveMode.test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    expect(result.ok).toBeTruthy()
+    const updatedEvaluation = result.unwrap().evaluation
+    expect(updatedEvaluation.name).toBe(evaluation.name)
+    expect(updatedEvaluation.description).toBe(evaluation.description)
+    expect(updatedEvaluation.type).toBe(evaluation.type)
+    expect(updatedEvaluation.metric).toBe(evaluation.metric)
+    expect(updatedEvaluation.configuration).toEqual(evaluation.configuration)
+  })
+})
```
geclos, line 125: I see many services without tests guys please

packages/core/src/services/evaluationsV2/update.ts
```diff
@@ -93,9 +93,13 @@ export async function updateEvaluationV2<
       versionId: result.id,
     } as unknown as EvaluationV2<T, M>
 
-    await pingProjectUpdate({ projectId: commit.projectId }, tx).then((r) =>
```
geclos, line 96: do not write side effects in services instead publish an event and let handlers handle side effects

PR #1183, opened 2025-04-24
[WIP] Custom llm evaluation editor

andresgutgon:
# What?
Implement a kind of playground for running custom evaluation prompts


geclos:
> Copilot action pass prompt instead of documentUuid this way suggestions can be applied over evaluation prompts.
I wouldn't worry about this for the mvp

> Make evaluation run action with streaming for the playground
Don't we have this already implemented in the evaluation v1 editor/playground?

> Think about useDocumentParameters. It's coupled with documents in the sense that locaStorage is stored by document uuid
Again, not sure we need local storage persistence for parameteres in evaluation editor (in the mvp).

andresgutgon:
> Again, not sure we need local storage persistence for parameteres in evaluation editor

Maybe not. I'll check

PR #1191, opened 2025-04-25
Prettier Experiments

csansoon:
Fixed stylings and added some metadata from document logs.
Next PR will add real-time for that metadata too.


packages/core/src/services/experiments/create.ts
```diff
@@ -113,7 +113,7 @@ export async function createExperiment(
           datasetLabels,
           fromRow: fromRow,
           toRow: toRow ?? rowCount,
-          count: Math.min(rowCount, toRow ?? rowCount) - fromRow,
```
andresgutgon, line 116: Magic number. What's this?

packages/core/src/repositories/experimentsRepository.ts
```diff
@@ -221,6 +222,42 @@ export class ExperimentsRepository extends Repository<Experiment> {
...
+      )
+    }
+
+    return Result.ok({
+      count: result[0]!.count,
+      totalCost: result[0]!.totalCost,
+      totalDuration: result[0]!.totalDuration,
+    })
+  }
+
```
andresgutgon, line 260: No tests

PR #1194, opened 2025-04-28
Split LLM Custom Evaluation V2 into actual and expected output types

### LINKED ISSUE #1192: Split Custom LLM evaluation in 2 types
# Why?

Custom llm evaluations can be interesting for doing experiments with `expectedOutput`. But also for using as `live` evaluations. The problem is that a custom llm evaluation using the parameter `expectedOutput` will fails if it's used as live evaluation.

# Proposed solution.

### When creating from scratch
When creating a custom evaluation, show a checkbox explaining if the evaluation can or can not have expected output

Image

### When cloning from managed llm like (binary, rating or comparison)
1. When cloning from `binary` or `rating` create a custom llm eval without `expectedOutput`. This can be used as `live`
2. When cloning from `comparison` create a llm eval **WITH** `expectedOutput`

neoxelox:
Closes #1192

### DIFF
--- a/apps/web/src/components/evaluations/llm/Custom.tsx
+++ b/apps/web/src/components/evaluations/llm/Custom.tsx
@@ -11,6 +11,7 @@ import { FormField } from '@latitude-data/web-ui/atoms/FormField'
 import { FormFieldGroup } from '@latitude-data/web-ui/atoms/FormFieldGroup'
 import { IconName } from '@latitude-data/web-ui/atoms/Icons'
 import { NumberInput } from '@latitude-data/web-ui/atoms/NumberInput'
+import { SwitchInput } from '@latitude-data/web-ui/atoms/Switch'
 import {
   useCurrentCommit,
   useCurrentProject,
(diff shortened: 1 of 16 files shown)


apps/web/src/components/evaluations/llm/Custom.tsx
```diff
@@ -161,6 +164,29 @@ You're an expert LLM-as-a-judge evaluator. Your task is to judge whether the res
...
+          // both custom and custom labeled evaluations
+          settings.metric === LlmEvaluationMetric.CustomLabeled
+        }
+        label='Use expected output'
+        description='Use the expected output in the evaluation prompt'
+        onCheckedChange={(checked) => {
+          if (checked) {
+            setSettings({
+              ...settings,
+              // @ts-expect-error this component is reused for
```
andresgutgon, line 179: Why not pass in props both
```js
type AnyCustomLlm = LlmEvaluationMetric.Custom | LlmEvaluationMetric.CustomLabeled
//...
}: ConfigurationFormProps<EvaluationType.Llm, AnyCustomLlm>)
```

PR #1198, opened 2025-04-28
fix: paramaeters keeping state between input type changes

geclos:


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/DocumentParams/ManualParams/index.tsx
```diff
@@ -23,26 +22,19 @@ type ManualParameterProps = {
 }
 
 function LocalParameterInput({ param, input, setInput }: ManualParameterProps) {
-  const [localInput, setLocalInput] = useState(input.value ?? '')
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/DocumentParams/ManualParams/index.tsx:26]: unless you need to edit the contents of an input from outside the input itself, there's never a reason for deriving the state of the input into a JS local state

apps/web/src/components/ParameterInput/index.tsx
```diff
@@ -21,31 +21,93 @@ export function ParameterInput({
...
 
-      onChange('')
-    },
-    [uploadFile, onChange],
+    return value
+  }, [value])
+
+  return (
+    <TextArea
+      defaultValue={textValue}
```
geclos, line 90: this is the only change in this file (the rest is just splitting components), changed value to defaultValue

andresgutgon (approved):

PR #1199, opened 2025-04-28
Improve run prompts with tool calls in small screens

andresgutgon:
# What?
On small screens run button and reset button are not visible. Now users can collapse tool call parameters in the form

https://github.com/user-attachments/assets/51ae0ab7-668b-4d16-8dff-f7ae1c5fb6e4


apps/web/src/stores/documentVersions.ts
```diff
@@ -254,68 +263,111 @@ export default function useDocumentVersions(
...
+      isUpdatingContent,
+      assignDataset,
+      saveLinkedDataset,
+      mutate,
+      isAssigningDataset,
+      isLinkingDataset,
+      isDestroyingFile,
+      isDestroyingFolder,
+    ],
+  )
```
andresgutgon, line 372: This store is used in many places. I reviewed it and tried to memoize all possible functions and the return so it's stable

geclos: thanks!

PR #1202, opened 2025-04-29
fix: include tool calls and agent response to documentLogCreated event

csansoon:


packages/core/src/services/providerLogs/buildResponse.ts
```diff
@@ -1,7 +1,10 @@
...
-import { ProviderLog } from '../../browser'
+import { getAgentResponseFromToolCalls, ProviderLog } from '../../browser'
 
 export function buildProviderLogResponse(providerLog: ProviderLog) {
+  const agentResponse = getAgentResponseFromToolCalls(providerLog.toolCalls)
+  if (agentResponse) return objectToString(agentResponse)
+
   return (
     providerLog.responseText ||
     (providerLog.responseObject
```
csansoon, line 10: This is used in several provider log serializers all over the app. I think it is the right call to include the agent response as the `response` object in all of them.

csansoon: okay, nope, I’m reverting this change. Serializing it this way will cause many parts of the app to misinterpret it as a regular text response, in addition to the agentic response.

This has to be done with love, not a quickfix

PR #1207, opened 2025-04-30
feat: handle rate limits in batch evaluations

geclos:


packages/core/src/lib/chainStreamManager/ChainStreamConsumer/consumeStream.ts
```diff
@@ -29,20 +29,79 @@ interface ConsumeStreamParams {
...
+      data: chunk,
+    })
+  }
+
+  return { error, finishReason }
+}
+
+function getErrorCode(error: unknown) {
+  if (error instanceof APICallError) {
+    if (error.statusCode === 420) {
```
andresgutgon, line 82: LOL. 
image


Is this what vercel SDK returns?

neoxelox (approved):

packages/core/src/tests/factories/experiments.ts
```diff
@@ -0,0 +1,68 @@
...
+    toRow: args.toRow,
+    workspace: args.workspace,
+  }).then((r) => r.unwrap())
+
+  return {
+    experiment,
+    dataset: datasetResult.dataset,
+    workspace: args.workspace,
+  }
+}
```
geclos, line 68: @csansoon please add tests to experiments

andresgutgon: Why did you add the factory in this PR? Is it not used?

geclos: it is used, i added a test that uses it

PR #1216, opened 2025-05-05
Display evaluations v2 on the project sidebar

andresgutgon:
# What?
When editing an evaluation, the playground is super similar to the one used in the edition of the document. By displaying the current selected evaluation, we think it will make it clear

## TODO
- [x] Style evaluation item
- [x] Show what the current evaluation is

## Next PR
- [ ] Use evaluation icons in the header to make it short
- [ ] Search for `TODO(evalsv2)` and `Go to LLM evaluation editor when LLM V2 evaluations are available` and add the right link.
- [ ] Add to evals editor the schema to only allow the right parameters
- [ ] Enforce structure outputs on evaluations editor
- [ ] Review evals editor lagging

### DIFF
--- a/apps/web/src/components/Sidebar/Files/index.tsx
+++ b/apps/web/src/components/Sidebar/Files/index.tsx
@@ -74,6 +74,8 @@ export type FileNodeProps = {
   onRenameFile: (args: { node: Node; path: string }) => Promise<void>
 }
 
+const EMPTY_TMP_NODES: Node[] = []
+
 function FileNode({
   isMerged,
   node,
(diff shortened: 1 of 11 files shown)


apps/web/src/components/Sidebar/Files/index.tsx
```diff
@@ -251,10 +258,6 @@ export function FilesTree({
       isMount.current = true
       return
     }
-
-    if (currentPath) {
-      togglePath(currentPath)
-    }
```
andresgutgon, line 257: This was a bug because folder was not open when selected

geclos (commented):
dale 👌🏼

PR #1217, opened 2025-05-05
Experiments - link to logs

csansoon:
- Added Link to logs in Experiments
- Fixed a bug where experiment variants were not applied and it just ran the original prompt lmao
- Fixed the Range Selector usability


packages/core/src/services/documentLogs/logsFilterUtils/parseLogFilterParams.ts
```diff
@@ -44,6 +44,28 @@ export function parseSafeCustomIdentifier(
...
+  }
+
+  try {
+    experimentId = decodeURIComponent(experimentId as string).trim()
+    const experimentIdNumber = Number(experimentId)
+    if (!isNaN(experimentIdNumber)) {
+      return experimentIdNumber
+    }
+  } catch (error) {
+    return undefined
```
geclos, line 63: you can leave this empty and let the last return handle the undefined case

geclos (dismissed):
looks good

PR #1219, opened 2025-05-05
Scale prompt verdict automatically when cloning a managed LLM Eval V2

neoxelox:


geclos (commented):
tests, also wait for reviews please

PR #1225, opened 2025-05-06
fix: experiments with rate limiting

geclos:


packages/core/src/jobs/job-definitions/experiments/runDocumentForExperimentJob.test.ts
```diff
@@ -0,0 +1,94 @@
...
+      }),
+    )
+
+    await expect(runDocumentForExperimentJob(mockJob)).rejects.toThrow(
+      'rate limited!',
+    )
+
+    expect(mocks.updateExperimentStatus).not.toHaveBeenCalled()
+  })
+})
```
geclos, line 94: teeeeeests

andresgutgon (approved):

PR #1235, opened 2025-05-08
feat: stripe webhooks

geclos:
Updates subscription on new subscription update Stripe event. Very simple approach that only ensures user is in paid team plan if the stripe subsription is in that state.


apps/web/src/app/api/webhooks/stripe/route.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  let event: Stripe.Event
+  try {
+    event = stripe.webhooks.constructEvent(await req.text(), sig, webhookSecret)
+  } catch (error: any) {
+    throw new UnprocessableEntityError(`Webhook Error: ${error.message}`)
+  }
+
+  if (event.type === 'customer.subscription.updated') {
+    const subscription = event.data.object as Stripe.Subscription
```
andresgutgon, line 50: All this signature validation could also be moved to a service and be tested

geclos: i've added tests to the route

PR #1236, opened 2025-05-08
Implement analytics service

andresgutgon:
# What?
We want a telemetry/analytics job that tracks events done in all Latitude installations. That can be, at this moment, a `cloud` or an `oss` deployment. 
The `oss` version has an opt-out environment variable, so these events are not sent to Latitude analytics.

## TODO
- [x] Tests
- [x] Implement Out-Out environment variable
- [x] Implement data filtering
- [x] Add a line in deployment docs about opt-out analytics
- [x] Add Posthog config in the code
- [x] Check if I need to split the data migration in different parts. Nullable, fill-data, non-nullable with default...
- [x] QA cloud + development
- [x] QA cloud + production
- [x] QA OSS + production
- [x] QA OSS + development


packages/core/drizzle/0164_purple_warbound.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "latitude"."workspaces" ADD COLUMN "uuid" uuid DEFAULT gen_random_uuid() NOT NULL;
```
geclos, line 1: this is backfilling existing workspaces?

packages/core/src/lib/analytics/providers/Posthog.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  async capture(data: CollectorOutput<ProductEdition>): Promise<void> {
+    const { distinctId, event, properties } = data
+
+    this.posthog.capture({
+      distinctId,
+      event,
+      properties,
+    })
+    await this.posthog.shutdown()
```
geclos, line 27: why? this is gonna make the next time you call captur ein this instance fail no?

andresgutgon: This is what [we have now](https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/jobs/job-definitions/events/publishToAnalyticsJob.ts#L25-L34) 
image

How has this changed in my PR?

packages/core/drizzle/0164_purple_warbound.sql
andresgutgon: It did when I ran in my dev env

PR #1237, opened 2025-05-08
feat: warn users about old syntax deprecation

geclos:
It's time


packages/web-ui/src/ds/atoms/Alert/index.tsx
```diff
@@ -41,7 +41,7 @@ export function Alert({
         />
       )}
       <div className='flex flex-row items-center gap-4 lg:gap-8 justify-between'>
-        <div className='flex flex-col gap-2 whitespace-pre-wrap break-all'>
+        <div className='flex flex-col gap-2 whitespace-pre-wrap'>
```
neoxelox, line 44: I think the break-all was introduced for correctly displaying the error frames when there is a syntax error in promptl or something like that. Take a look to see if that is broken.

geclos: image

PR #1249, opened 2025-05-09
feat: make additionalProperties mandatory in integration/latitude tools

geclos:
Since it's optional in some providers but required for othersd, we make it mandatory and defeault to false as the most sensible option.

### DIFF
--- a/packages/constants/src/ai.ts
+++ b/packages/constants/src/ai.ts
@@ -83,7 +83,7 @@ export type ToolDefinition = JSONSchema7 & {
     type: 'object'
     properties: Record<string, JSONSchema7>
     required?: string[]
-    additionalProperties?: boolean
+    additionalProperties: boolean
   }
 }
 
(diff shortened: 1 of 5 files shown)


packages/constants/src/ai.ts
```diff
@@ -83,7 +83,7 @@ export type ToolDefinition = JSONSchema7 & {
     type: 'object'
     properties: Record<string, JSONSchema7>
     required?: string[]
-    additionalProperties?: boolean
+    additionalProperties: boolean
```
geclos, line 86: cc @csansoon made this change which works ok in local testing but not sure if I might be unadvertendly breaking something

csansoon: That's fine, it's just to represent the type internally to help develop

andresgutgon (approved):

csansoon:
Change the schema definition in `/packages/constants/src/helpers.ts`. That's where we generate the schema than will then be used to validate the configuration in real-time from the client editor

geclos:
> Change the schema definition in /packages/constants/src/helpers.ts. That's where we generate the schema than will then be used to validate the configuration in real-time from the client editor

But i don't want to force this prop in user-written schemas, this is only for internal tools

csansoon (approved):

Issue #1245, opened 2025-05-08
Tweak onboarding to create two experiments with different models


cesr:
Cositas:

- [x] Onboarding prompt should have an accuracy evaluation connected by default
- [x] Remove gap from preview vertical space between paragraphs (spans)
- [x] Remove skip onboarding
- [x] Onboarding should generate two experiments, with different models (4o and 4o-mini)
- [x] Add posthog to onboarding and all the others

PR #1251, opened 2025-05-09
Add experiments with 2 different models to onboarding

andresgutgon:
# What?
During onboarding introduce new users to experiments by creating one with 2 models comparison

## Issue
https://github.com/latitude-dev/latitude-llm/issues/1245

## TODO
- [ ] Onboarding prompt should have an accuracy evaluation connected by default
- [ ] Remove gap from preview vertical space between paragraphs (spans)
- [ ] Remove skip onboarding
- [ ] Onboarding should generate two experiments, with different models (4o and 4o-mini)
- [ ] Add posthog to onboarding and all the others


apps/web/src/app/providers.tsx
```diff
@@ -33,19 +33,22 @@ export function IdentifyUser({
...
+
     try {
-      if (user && !user.email.match(/@latitude\.so/)) {
-        posthog?.identify(user.email, {
-          email: user.email,
-        })
-        posthog?.group('workspace', String(workspace.id))
-        posthog?.startSessionRecording()
-      }
+      const isStaff = !!email.match(/@latitude\.so/)
```
neoxelox, line 41:
```suggestion
      const isStaff = !!email.match(/@latitude\.so$/)
```

this avoids matching some malicious emails such as `@latitude.so.attacker.domain`

apps/web/src/components/Sidebar/Files/DocumentHeader/index.tsx
```diff
@@ -67,6 +67,8 @@ export default function DocumentHeader({
     if (selected && !currentEvaluationUuid) return undefined
```
neoxelox, line 67: this will still make that the document is not clickable in "evaluations/experiments/logs" tab, do this instead:
```
if (!documentUuid) return undefined
if (!node.isPersisted) return undefined
if (selected && pathname.endsWith(documentUuid)) return undefined
```

apps/web/src/components/Sidebar/Files/DocumentHeader/index.tsx
```diff
@@ -64,9 +64,15 @@ export default function DocumentHeader({
   )
   const documentUuid = node.doc!.documentUuid
   const url = useMemo(() => {
-    if (selected && !currentEvaluationUuid) return undefined
-    if (!node.isPersisted) return undefined
     if (!documentUuid) return undefined
+    if (!node.isPersisted) return undefined
+    if (
+      selected &&
+      !currentEvaluationUuid &&
```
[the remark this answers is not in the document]
andresgutgon: Doesn't hurt

packages/core/src/services/evaluationsV2/validate.ts
```diff
@@ -49,28 +50,42 @@ export async function validateEvaluationV2<
...
-    .validate(
-      {
-        metric: settings.metric,
-        configuration: settings.configuration,
-        document: document,
-        commit: commit,
-        workspace: workspace,
-      },
-      db,
+  if (parseResult.error) {
```
neoxelox, line 57: dont do this and return the zod error as is please. The frontend needs the original zod error to show the error for each input in the form

apps/web/src/components/Sidebar/Files/DocumentHeader/index.tsx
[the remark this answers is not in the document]
neoxelox: but it doesn't make sense, you only dont want to navigate if the document is selected and you are already in the document editor. In all other cases you want to navigate on click

packages/web-ui/src/ds/molecules/ChatWrapper/Message/index.tsx
```diff
@@ -381,7 +381,7 @@ const ContentText = memo(
     return (
```
neoxelox, line 381: Can't comment on that line but add `wordBreak='breakAll'` to the `TextComponent` above

PR #1256, opened 2025-05-09
feature: stop experiments

csansoon:
Stop Experiments manually


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/experiments/_components/ExperimentsTable/ExperimentStatus.tsx
```diff
@@ -0,0 +1,76 @@
...
+    })
+  }, [experiment.uuid, execute, projectId, commitUuid, documentUuid])
+
+  if (isPending) {
+    return <Icon name='clock' color='foregroundMuted' />
+  }
+
+  if (isRunning) {
+    return (
+      <Tooltip
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/experiments/_components/ExperimentsTable/ExperimentStatus.tsx:41]: `asChild` my boy

packages/core/src/services/experiments/complete.ts
```diff
@@ -30,5 +31,10 @@ export async function completeExperiment(
     return Result.error(updateResult.error as LatitudeError)
   }
 
+  updateExperimentStatus({
+    workspaceId: experiment.workspaceId,
+    experiment: updateResult.unwrap(),
```
andresgutgon, line 36: Don't unwrap in services. Return the error :pray:

neoxelox: its already returned above

andresgutgon: Ah is true. So in this case to avoid confusion is better to do `updateResult.value`

PR #1264, opened 2025-05-12
Websockets Admin UI

andresgutgon:
# What?
With this we can have a bit more visibility on websockets connections

This is the official Admin UI from socket-io
https://socket.io/docs/v4/admin-ui/

### TODO
- [ ] Provision credentials for the admin

### DIFF
--- a/apps/websockets/src/server.ts
+++ b/apps/websockets/src/server.ts
@@ -16,6 +16,7 @@ import { env } from '@latitude-data/env'
 import cookieParser from 'cookie-parser'
 import express from 'express'
 import { Namespace, Server, Socket } from 'socket.io'
+import { instrument } from '@socket.io/admin-ui'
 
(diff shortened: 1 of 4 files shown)


apps/websockets/src/server.ts
```diff
@@ -40,12 +41,25 @@ const server = http.createServer(app)
...
+    type: 'basic',
+    // TODO: Provision these env variables
+    username: '',
+    // Encrypted password
+    password: '',
+  },
+  readonly: true,
+})
+
+io.of('/admin').use((_socket, next) => next())
```
neoxelox, line 61:
```suggestion
io.on('/admin').use((_socket, next) => next())
```

andresgutgon: Is `of` for defining a namespace

PR #1266, opened 2025-05-13
Add Experiment Variants created.

andresgutgon:
# What?
Create an event with the variants used when a list of variants are submited to create N experiments


csansoon (dismissed):

packages/core/src/services/experiments/assertRequirements.ts
```diff
@@ -9,12 +8,12 @@ export function assertEvaluationRequirements({
...
   evaluations: EvaluationV2[]
   datasetLabels: Record<string, string>
-}): TypedResult<undefined, LatitudeError> {
+}) {
   for (const evaluation of evaluations) {
     const spec = getEvaluationMetricSpecification(evaluation)
     if (!spec.supportsBatchEvaluation) {
       return Result.error(
-        new BadRequestError(
+        new Error(
```
geclos, line 16: LatitudeError 🙏🏼

packages/core/src/services/experiments/assertRequirements.ts
```diff
@@ -23,7 +22,7 @@ export function assertEvaluationRequirements({
     const datasetLabel = datasetLabels[evaluation.uuid]
     if (spec.requiresExpectedOutput && !datasetLabel) {
       return Result.error(
-        new BadRequestError(
+        new Error(
```
geclos, line 25: ditto

packages/core/src/services/experiments/createVariants.ts
```diff
@@ -84,14 +81,21 @@ export async function createExperimentVariants(
...
+      data: {
+        workspaceId: workspace.id,
+        documentUuid: document.documentUuid,
+        commitUuid: commit.uuid,
+        variants: inputVariants,
+      },
+    })
+
+    return Result.ok(experiments as Experiment[])
+  }, db)
```
geclos, line 100: hmm no test for this?

geclos (approved):

packages/core/src/services/experiments/createVariants.ts
andresgutgon: There are tests. The ones I changed

packages/core/src/services/experiments/assertRequirements.ts
andresgutgon: The transaction was complaining like crazy here

packages/core/src/services/experiments/assertRequirements.ts
andresgutgon: This is used inside a transaction. And transactions only understand `Error`

PR #1270, opened 2025-05-14
feat: move apiKeys touches to workers

geclos:
Also commented some circular dependency code


packages/core/src/lib/Transaction.ts
```diff
@@ -38,7 +38,12 @@ export default class Transaction {
   ): PromisedResult<ResultType> {
     try {
       let result: TypedResult<ResultType, Error>
-      await db.transaction(async (trx) => (result = await callback(trx)))
+
+      await db.transaction(async (trx) => {
+        result = await callback(trx)
+
+        if (result.error) throw result.error
```
geclos, line 45: so that you can return Result.error from transactions since i see you are all doing it

packages/core/src/services/experiments/complete.ts
```diff
@@ -31,10 +30,11 @@ export async function completeExperiment(
...
 
-  updateExperimentStatus({
-    workspaceId: experiment.workspaceId,
-    experiment: updateResult.value,
-  })
+  // TODO: uncomment when we've removed the circular dependency
+  // await updateExperimentStatus({
+  //   workspaceId: experiment.workspaceId,
+  //   experiment: updateResult.value,
+  // })
```
csansoon, line 37: Just replace it with
```ts
const websockets = await WebsocketClient.getSocket()
websockets.emit('experimentStatus', {
  workspaceId: experiment.workspaceId,
  data: {
    experiment: updateResult.value
  },
})
```

This way it still updates the front-end without entering a loop

PR #1285, opened 2025-05-16
Tracing span model

neoxelox:
Related to: #1273


packages/core/src/plans.ts
```diff
@@ -10,21 +10,25 @@ export const SubscriptionPlans = {
...
     name: 'Team',
     credits: 100_000,
     users: 5,
+    retention_period: 1825,
   },
   [SubscriptionPlan.EnterpriseV1]: {
     name: 'Enterprise',
     credits: 1_000_000,
     users: 1000,
+    retention_period: 3650,
```
andresgutgon, line 31: What's this?

neoxelox: The log (span) retention period mentioned in https://latitude.so/pricing

geclos (approved):

PR #1287, opened 2025-05-18
feat: improve count logs query performance

geclos:
This query was unnecessarily slow by joining to multiple tables when the filtering returned high cardinality (>600k results!). 

Removing some of the joins that weren't even necessary reduces the query time by 5x.

### DIFF
--- a/packages/core/src/repositories/documentLogsWithMetadataAndErrorsRepository/index.ts
+++ b/packages/core/src/repositories/documentLogsWithMetadataAndErrorsRepository/index.ts
@@ -14,6 +14,7 @@ import { DocumentLogWithMetadataAndError } from '../runErrors/documentLogsReposi
 import { NotFoundError } from './../../lib/errors'
 import { Result } from './../../lib/Result'
 
+// TODO: remove
 export class DocumentLogsWithMetadataAndErrorsRepository extends Repository<DocumentLogWithMetadataAndError> {
   get scopeFilter() {
     return eq(workspaces.id, this.workspaceId)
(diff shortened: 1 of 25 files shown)


packages/core/src/services/documentLogs/countDocumentLogs.ts
```diff
@@ -0,0 +1,19 @@
...
+    .from(documentLogs)
+    .innerJoin(commits, eq(documentLogs.commitId, commits.id))
+    .where(
+      and(
+        eq(documentLogs.documentUuid, documentUuid),
+        isNull(commits.deletedAt),
+      ),
+    )
+    .then((r) => r[0]?.count ?? 0)
+}
```
geclos, line 19: removed the run errors left join and moved the is null equality to the where filter rather than the join condition. This reduces query time by 5x.

packages/core/src/repositories/documentLogsWithMetadataAndErrorsRepository/index.ts
```diff
@@ -24,13 +25,15 @@ export class DocumentLogsWithMetadataAndErrorsRepository extends Repository<Docu
...
         commit: getTableColumns(commits),
+        // TODO: Denormalize these aggregations and persist them at write time
         tokens: sum(providerLogs.tokens).mapWith(Number).as('tokens'),
         duration: sum(providerLogs.duration)
           .mapWith(Number)
           .as('duration_in_ms'),
         costInMillicents: sum(providerLogs.costInMillicents)
           .mapWith(Number)
           .as('cost_in_millicents'),
+        // TODO: Denormalize the errors and persist them at write time
```
andresgutgon, line 36: The plan is to get rid of the errors table and put there in the logs and eval results?

geclos: We will remove the table yes but we will also migrate for doc/provider logs to traces and spans

PR #1296, opened 2025-05-20
Documentation sidebar

csansoon:
https://github.com/user-attachments/assets/8742278d-9db3-4c34-8b15-56ad4dee50df


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/index.tsx
```diff
@@ -184,10 +187,11 @@ export default function DocumentEditor({
       <SplitPane
         className='pt-6'
         direction='horizontal'
+        reversed
         gap={4}
-        initialPercentage={52}
         initialWidthClass='min-w-1/2'
-        minSize={300}
+        minSize={350}
+        initialPercentage={40}
```
Comment (csansoon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/index.tsx:194]: I reversed the Draggable Pane for the Playground. Now, when resizing the window, the playground stays with the same size and it is the editor part which grows and shrinks accordingly. Also, the playground now has a min size too, so it is no longer possible to lose it when making the window smaller (happened to the best of us)

apps/web/src/app/(private)/settings/_components/Webhooks/index.tsx
```diff
@@ -141,7 +143,12 @@ export default function Webhooks() {
   return (
     <>
       <TableWithHeader
-        title='Webhooks'
+        title={
+          <div className='flex flex-row items-center gap-2'>
+            <Text.H4B>Webhooks</Text.H4B>
+            <OpenInDocsButton route={DocsRoute.Webhooks} />
+          </div>
```
andresgutgon, line 150: Over and over same pattern. This could be `icon` prop inside `TableWithTitle`

andresgutgon (dismissed):

PR #1298, opened 2025-05-21
Telemetry SDKs V2

neoxelox:
Related to: https://github.com/latitude-dev/latitude-llm/issues/1273


packages/telemetry/typescript/tsconfig.json
```diff
@@ -1,18 +1,13 @@
...
+  "exclude": ["node_modules", "dist"],
   "compilerOptions": {
     "outDir": "dist",
     "paths": {
-      "$telemetry/*": [
-        "./src/*"
-      ]
+      "$telemetry/*": ["./src/*"],
+      "@latitude-data/constants/*": ["../../constants/src/*"],
+      "@latitude-data/sdk/*": ["../../sdks/typescript/src/*"]
```
andresgutgon, line 10: This should not be needed

packages/constants/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "@latitude-data/compiler": "workspace:^",
+    "@opentelemetry/semantic-conventions": "^1.22.0",
```
andresgutgon, line 29: How big is this thing?

constants should be as clean and lean as possible. Don't put here big dependencies 🙏

PR #1313, opened 2025-05-26
feat: export logs to datasets and file downloads in batch

geclos:
Allows exporting logs to datasets and file downloads in big numbers.


packages/core/src/services/exports/findByUuid.ts
```diff
@@ -0,0 +1,24 @@
...
+  uuid: string,
+  workspaceId: number,
+): Promise<Export | undefined> {
+  return await database.query.latitudeExports.findFirst({
+    where: and(
+      eq(latitudeExports.uuid, uuid),
+      eq(latitudeExports.workspaceId, workspaceId),
+    ),
+  })
+}
```
andresgutgon, line 24: Why not a collection? I wouldn't start another way of doing things while we have already one working

andresgutgon: Also a service is weird only for getting data

2a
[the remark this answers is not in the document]
geclos: woopsie

andresgutgon (dismissed):

packages/core/src/services/exports/findByUuid.ts
andresgutgon: ok

andresgutgon: Another tradeoff I see in putting all in `services/**/*.ts` is that making tests for each data access file will fill the folder with files. I like services to be only things that do one thing, writing data. 
At least put all data access in `data-access/**/*.ts`. Although I think should investigate again if Drizzle now supports a kind of tenancy that could work for us

PR #1326, opened 2025-05-28
feat: improve performance of project stats

geclos:


packages/core/src/services/projects/computeProjectStats.ts
```diff
@@ -144,95 +144,25 @@ export async function computeProjectStats({
...
 
-  // Get document UUIDs for evaluation queries
-  const documentUuids = await db
-    .select({ documentUuid: documentVersions.documentUuid })
-    .from(documentVersions)
-    .innerJoin(commits, eq(documentVersions.commitId, commits.id))
-    .where(and(eq(commits.projectId, project.id), isNull(commits.deletedAt)))
-    .then((result) => result.map((row) => row.documentUuid))
-
-  // NOTE(evalsv2): Do not delete this as it's crucial to compute historical
```
geclos, line 155: I think this is no longer true? We migrated all evaluation results to v2

PR #1327, opened 2025-05-28
feat: improve logs export

daavidrgz:
https://github.com/user-attachments/assets/7c55526a-c1ac-47e0-baea-eee8ded6224e

### New developed features
- When downloading logs, the user can see a preview (similar at how it's done when generating datasets)
- In that preview, the users can select the columns they want to include in the CSV
- Additionally, to avoid showing too many logs to the user in that preview, we retrieve only a set of representative logs based on their parameters. This is, for multiple logs with the exact same parameters, only one of them will be shown to the user.

### Frontend code improvements
- Move the `usePreviewLogs` hook to `/stores` in order to reuse it from both modals
- Generalize the PreviewTable component
- Modify the condition for downloading logs synchronously or not. Now it relies only on the number of logs selected, instead of if the selection was partial or not (if 10k logs are selected and the user deselects one, I think we should go trough the async download)

### Backend code improvements
- Refactor the `buildDatasetFromLogs` service logic to support customizable column filtering (as well as a little refactor)
- Add missing tests

### Next steps
While I was developing this, a new feature for exporting asynchronously document logs was introduced, so the column filtering functionality is not available for that case. This is mainly due to the fact that that development doesn't reuse any of the logic present in the core services. Maybe this will need a refactor. What do yo think about?


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DownloadLogsModal/useDownloadLogsModal.ts
```diff
@@ -0,0 +1,211 @@
...
+    showModal,
+    handleDownload,
+    isLoadingPreview: isLoading,
+    isDownloading,
+    fetchPreview,
+    previewStaticColumns: staticColumns,
+    previewParameterColumns: parameterColumns,
+    handleSelectStaticColumn,
+    handleSelectParameterColumn,
+  }
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DownloadLogsModal/useDownloadLogsModal.ts:208]: memoize this return

apps/web/src/actions/documentLogs/downloadLogs.ts
```diff
@@ -10,7 +10,7 @@ export const downloadLogsAsyncAction = withDocument
   .createServerAction()
   .input(
     z.object({
-      selectionMode: z.enum(['ALL', 'ALL_EXCEPT']),
+      selectionMode: z.enum(['ALL', 'ALL_EXCEPT', 'PARTIAL']),
```
geclos, line 13: No this is/shouldn't be possible the action is only called when requesting an async download which implies selectionmode is either all or all_except – partial is download synchronously (can probably rename this action to make this clearer)

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DownloadLogsModal/useDownloadLogsModal.ts
```diff
@@ -0,0 +1,211 @@
...
+    selectableState.getSelectedRowIds,
+  ])
+
+  const handleImmediateDownload = useCallback(async () => {
+    const ids = selectableState.getSelectedRowIds()
+    const selectedStaticColumns = getSelectedColumns(staticColumns)
+    const selectedParameterColumns = getSelectedColumns(parameterColumns)
+    const formData = new FormData()
+    formData.append('ids', JSON.stringify(ids))
+    formData.append('staticColumnNames', JSON.stringify(selectedStaticColumns))
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/DownloadLogsModal/useDownloadLogsModal.ts:89]: you've only added the selected columns feature to synchronous downloads but not async, which will require editing the downloadLogsAsyncAction – take into account query performance when doing this.

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/SaveLogsAsDatasetModal/useSaveLogsAsDatasetModal.ts
```diff
@@ -186,11 +114,11 @@ export function useSelectedLogs({
   )
   return useMemo(
     () => ({
-      previewData,
+      data: previewData,
+      state: previewModalState,
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/SaveLogsAsDatasetModal/useSaveLogsAsDatasetModal.ts:118]: add these properties to the dependency array of the memoizer

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/index.tsx
```diff
@@ -215,14 +220,18 @@ export function DocumentLogs({
                 <Button
                   fancy
                   disabled={selectableState.selectedCount === 0}
-                  onClick={previewLogsState.onClickShowPreview}
+                  onClick={saveLogsAsDatasetModalState.onClickShowPreview}
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/index.tsx:223]: much clearer with the renaming 👌🏼

packages/core/src/jobs/job-definitions/documentLogs/downloadLogsJob.ts
```diff
@@ -68,7 +68,7 @@ export const downloadLogsJob = async (
     user: User
     token: string
     workspace: Workspace
-    selectionMode: 'ALL' | 'ALL_EXCEPT'
+    selectionMode: 'ALL' | 'ALL_EXCEPT' | 'PARTIAL'
```
geclos, line 71: same, partial is not a valid value here

packages/core/src/services/datasets/previewFromLogs.ts
```diff
@@ -19,34 +24,68 @@ async function getFirstRowsFromDataset({
...
+  const repo = new DatasetsRepository(workspace.id)
+  const datasets = await repo.findByName(name)
+  if (datasets.length === 0) return
+  return datasets[0]
+}
+
+/**
+ * This service is responsible of obtaining a preview of the dataset that would be generated
+ * including the logs provided. The dataset rows are limited to the first 5 and the logs are
+ * choosen keeping only the ones with different parameters.
```
geclos, line 58: chosen*

geclos (commented):
Thanks for your work!

Main issue is that you did not take into account async downloads – happen whenever you select all logs  and there are more than 25.

Rest are only minor comments.

cesr:
👋 coming with product feedback:

- I love the idea of disabling the columns but sadly it's not a very common pattern and it could be hard to discover
- I would opt for either adding toggles to each column header to make it more explicit or maybe just avoid that pattern and display a muti-select to pick the columns
- It could be interesting to add evaluation results to this exports as well, some users have asked for that

Thank you so much for this PR David 🙏

daavidrgz:
> 👋 coming with product feedback:
> 
>     * I love the idea of disabling the columns but sadly it's not a very common pattern and it could be hard to discover
> 
>     * I would opt for either adding toggles to each column header to make it more explicit or maybe just avoid that pattern and display a muti-select to pick the columns
> 
>     * It could be interesting to add evaluation results to this exports as well, some users have asked for that
> 
> 
> Thank you so much for this PR David 🙏



> 👋 coming with product feedback:
> 
>     * I love the idea of disabling the columns but sadly it's not a very common pattern and it could be hard to discover
> 
>     * I would opt for either adding toggles to each column header to make it more explicit or maybe just avoid that pattern and display a muti-select to pick the columns
> 
>     * It could be interesting to add evaluation results to this exports as well, some users have asked for that
> 
> 
> Thank you so much for this PR David 🙏

Yeah! You're completely right. The UX was the reason why I opened the PR in draft at first.
I really like the "log preview" feature so I'll try to find a better (and more discoverable) way to disable columns. I'll update you for sure!

daavidrgz:
> Thanks for your work!
> 
> Main issue is that you did not take into account async downloads – happen whenever you select all logs and there are more than 25.
> 
> Rest are only minor comments.

Yep, as I commented in the description, the [downloadLogsJob function](https://github.com/latitude-dev/latitude-llm/blob/f3837af6ee0786e210505dc7b3df76af229a697f/packages/core/src/jobs/job-definitions/documentLogs/downloadLogsJob.ts#L66) doesn't reuse any logic present in the core services... I'll try to refactor that function in order to use the [generateCsvFromLogs core function](https://github.com/latitude-dev/latitude-llm/blob/f3837af6ee0786e210505dc7b3df76af229a697f/packages/core/src/services/datasets/generateCsvFromLogs.ts#L37)

packages/core/src/repositories/documentLogsWithMetadataAndErrorsRepository/index.ts
```diff
@@ -86,4 +97,20 @@ export class DocumentLogsWithMetadataAndErrorsRepository extends Repository<Docu
 
     return Result.ok(result[0]!)
   }
+
+  async findManyWithoutErrors(
```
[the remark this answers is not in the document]
daavidrgz: I'm not sure if I get you. This function is in the `DocumentLogsWithMetadataAndErrorsRepository` because I want to retrieve that entity. If I move the function to the `DocumentLogRepository` I'm not retrieving the same entity (losing for example the metadata)

apps/web/src/stores/previewLogs.ts
```diff
@@ -0,0 +1,86 @@
...
+  const {
+    data = EMPTY_DATA,
+    mutate: fetchPreview,
+    isLoading,
+  } = useSWR<OutputItem>(cacheKey, fetcher, {
+    ...opts,
+    revalidateOnMount: false,
+    revalidateOnFocus: false,
+    revalidateIfStale: false,
+    dedupingInterval: 60000,
```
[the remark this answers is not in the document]
daavidrgz: I didn't want to change the previous functionality
image
But I think you're right, I'll test it btw

apps/web/src/app/api/datasets/preview-logs/route.ts
```diff
@@ -17,11 +17,17 @@ export const GET = errorHandler(
     ) => {
       const query = request.nextUrl.searchParams
       const ids = query.get('documentLogIds')?.split(',').map(Number) ?? []
+      const staticColumnNames = query.get('staticColumnNames')?.split(',')
+      const parameterColumnNames = query.get('parameterColumnNames')?.split(',')
       const result = await previewDatasetFromLogs({
         workspace,
         data: {
-          name: query.get('name') ?? '',
+          name: query.get('name') ?? undefined,
```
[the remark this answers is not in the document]
daavidrgz: It turns out that the `query.get()` method returns `string | null` and the newly changed interface accepts `string?`. I changed it for being more semantic. When no dataset is selected, you don't want to go to the database to search for "an empty string dataset name"

apps/web/src/actions/documentLogs/downloadLogs.ts
daavidrgz: I've slightly changed the condition to go through the instantDownload or the asyncDownload. Now it's only based on the selectedElements. I think that if someone selects 10k logs and deselects one of them (converting the selection mode into `PARTIAL`, it should also go through the asyncDownload no?
If the answer is yes, maybe we should increase that limit to 100 logs maybe?

daavidrgz:
Hi again @cesr @geclos!
What do you think about this one?

https://github.com/user-attachments/assets/28de62bb-8358-4d12-b415-83084a547134

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/index.tsx
```diff
@@ -103,9 +110,25 @@ export function DocumentLogs({
     rowIds: documentLogIds,
     totalRowCount: pagination?.count ?? 0,
   })
-  const previewLogsState = useSelectedLogs({
+  const extendedFilterOptions: ExtendedDocumentLogFilterOptions =
```
Comment (daavidrgz) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/logs/_components/DocumentLogs/index.tsx:111]: This is the key point of the "tactical" proposed solution. The `ExtendedDocumentLogFilterOptions` class will track the effective filters used when retrieving the logs + the selection data of the table

apps/web/src/app/api/datasets/preview-logs/route.ts
```diff
@@ -1,11 +1,23 @@
...
-export const GET = errorHandler(
+export const previewLogsRequestSchema = z.object({
+  documentUuid: z.string(),
+  name: z.string().optional(),
+  extendedFilterOptions: extendedDocumentLogFilterOptionsSchema,
+  staticColumnNames: z.array(z.string()).optional(),
+  parameterColumnNames: z.array(z.string()).optional(),
+})
+
+export const POST = errorHandler(
```
daavidrgz, line 20: Another important change. I moved to POST in order to send the `extendedFilterOptions` without a headache. We lose caching (and semantics) but I think that this can be improved in another PR as I think it won't lead to 
 performance issues to end users

packages/core/src/jobs/job-definitions/documentLogs/downloadLogsJob.ts
```diff
@@ -1,243 +1,64 @@
...
-import { buildProviderLogResponse } from '../../../services/providerLogs'
 import { markExportReady } from '../../../services/exports/markExportReady'
 import { findOrCreateExport } from '../../../services/exports/findOrCreate'
-import { buildLogsFilterSQLConditions } from '../../../services/documentLogs/logsFilterUtils'
 import { Readable } from 'stream'
-
-const BATCH_SIZE = 1000
-
-export class CursorState {
-  private redis: Redis | null = null
```
daavidrgz, line 26: I dont really get why we needed redis here

packages/core/src/repositories/documentLogsWithMetadataAndErrorsRepository/index.ts
```diff
@@ -86,4 +106,81 @@ export class DocumentLogsWithMetadataAndErrorsRepository extends Repository<Docu
...
+    const ordering = this.getOrdering(extendedFilterOptions)
+    const result = await this.scope
+      .where(and(this.scopeFilter, ...conditions))
+      .orderBy(...ordering)
+      .limit(size)
+      .offset(offset)
+    return Result.ok(result)
+  }
+
+  async findInDocumentWithCursor(
```
daavidrgz: Aggregate all the database access logic in one place

packages/core/src/services/documentLogs/logsFilterUtils/buildLogsFilterSQLConditions.ts
```diff
@@ -19,21 +22,27 @@ export function buildLogsFilterSQLConditions({
...
-    commitIds.length
-      ? or(inArray(documentLogs.commitId, commitIds))
-      : sql`1 = 0`, // Return none
-    logSources.length
-      ? or(inArray(documentLogs.source, logSources))
-      : sql`1 = 0`, // Return none
+    commitIds.length ? inArray(documentLogs.commitId, commitIds) : sql`1 = 0`, // Return none
+    logSources.length ? inArray(documentLogs.source, logSources) : sql`1 = 0`, // Return none
     createdAt
-      ? and(between(documentLogs.createdAt, createdAt.from, createdAt.to))
```
daavidrgz, line 32: I don't know why the `and`s and `or`s were used here

packages/core/src/repositories/documentLogsWithMetadataAndErrorsRepository/index.ts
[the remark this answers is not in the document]
geclos: oh sorry didn't notice it was in this repo already!

daavidrgz:
image

This is the final UI proposal :smile:

packages/core/src/jobs/job-definitions/documentLogs/downloadLogsJob.ts
geclos: not needed, this is remnants from a previous version of this job

apps/web/src/app/api/datasets/preview-logs/route.ts
geclos: fine tradeoff

apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/logs/route.ts
```diff
@@ -22,26 +20,21 @@ export const GET = errorHandler(
...
-      const repo = new DocumentVersionsRepository(workspace.id)
-      const document = await repo
-        .getSomeDocumentByUuid({ projectId: Number(projectId), documentUuid })
-        .then((r) => r.unwrap())
-
-      const buildQueryFn = queryParams.excludeErrors
-        ? computeDocumentLogs
-        : computeDocumentLogsWithMetadata
 
-      const rows = await buildQueryFn({
```
Comment (daavidrgz) [inline apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/logs/route.ts:40]: Now we always call the `computeDocumentLogsWithMetadata` to avoid more refactors and making the PR more complicated
I'll add a TODO just in case

packages/core/src/services/documentLogs/logsFilterUtils/buildLogsFilterSQLConditions.ts
geclos: probably someone removed some condition and forgot to remove the and/or

cesr:
Something I'm not sure we're taking into account here: a lot of users asked about exporting evaluation results together with the logs, would that be possible?

apps/web/src/app/api/documentLogs/download-logs/route.ts
```diff
@@ -15,34 +26,36 @@ export const POST = errorHandler(
...
-
-      const ids: string[] = JSON.parse(idsRaw as string)
+      try {
+        const body = await request.json()
+        const {
+          documentUuid,
+          extendedFilterOptions,
+          staticColumnNames,
+          parameterColumnNames,
+        } = downloadLogsRequestSchema.parse(body)
```
andresgutgon: You have safeParse which gives `result.success`. So no need to try/catch for the zod errors

apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/logs/route.ts
Comment (andresgutgon) [reply apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/logs/route.ts:40]: When was used `computeDocumentLogs` without metadata?

apps/web/src/stores/previewLogs.ts
```diff
@@ -0,0 +1,102 @@
...
+import { useNavigate } from '$/hooks/useNavigate'
+import { ROUTES } from '$/services/routes'
+import {
+  Dataset,
+  ExtendedDocumentLogFilterOptions,
+  parseRowCell,
+} from '@latitude-data/core/browser'
+import { DatasetRowData } from '@latitude-data/core/schema'
+import { useToast } from 'node_modules/@latitude-data/web-ui/src/ds/atoms/Toast/useToast'
+import { useCallback, useState } from 'react'
```
andresgutgon: stores are things connected to a table in the DB and using local store with [swr](https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/stores/users.ts#L8)

Not need to put this hook here. Move closer to where is used

apps/web/src/stores/previewLogs.ts
```diff
@@ -0,0 +1,102 @@
...
+          method: 'POST',
+          body: JSON.stringify(body),
+        })
+        const response = await handleResponse({
+          response: rawResponse,
+          toast,
+          navigate,
+          serializer: serializeRows,
+        })
+        response && setData(response)
```
andresgutgon: Also this is a POST so no a cacheable thing I guess not even in frontend

geclos:
> Something I'm not sure we're taking into account here: a lot of users asked about exporting evaluation results together with the logs, would that be possible?

for another PR

apps/web/src/stores/previewLogs.ts
daavidrgz: You are right, this is a POST to allow sending the filters in the request. Someday we'll have the [QUERY HTTP method](https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html), idempotent and thus safely cacheable

apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/logs/route.ts
Comment (daavidrgz) [reply apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/logs/route.ts:40]: From [useLogHistoryParams](https://github.com/latitude-dev/latitude-llm/blob/885380997c39499c0877198d90611172014b0e7b/apps/web/src/app/(private)/projects/%5BprojectId%5D/versions/%5BcommitUuid%5D/documents/%5BdocumentUuid%5D/_components/DocumentEditor/Editor/Playground/DocumentParams/HistoryLogParams/useLogHistoryParams.ts#L32) and [useSerializedLogs](https://github.com/latitude-dev/latitude-llm/blob/885380997c39499c0877198d90611172014b0e7b/apps/web/src/app/(private)/projects/%5BprojectId%5D/versions/%5BcommitUuid%5D/documents/%5BdocumentUuid%5D/evaluations/%5BevaluationUuid%5D/editor/_components/EvaluationEditor/Playground/EvaluationParams/HistoryLogParams/useSerializedLogs.ts#L33). I just updated this handler to ensure we are only retrieving logs without errors:
image

PR #1332, opened 2025-05-29
feat: add tracer to gateway

geclos:


geclos:
merging to head branch

PR #1333, opened 2025-05-29
Improve API and SDK error delivering when running an agent and has a compilation error

andresgutgon:
# What?
While working on SDK examples, I was using our SDK.run method in **non-streaming** method of a an [**agentic** prompt](https://github.com/latitude-dev/latitude-llm/blob/main/examples/src/cases/content-moderation-system/example.ts). I forgot to add one of the parameters used in the prompt to the parameters in the SDK, and I noticed that we are returning a generic 500 API error. We should return to the user of the SDK the error that occurred during the run of the prompt.

## Before
```bash
LatitudeApiError: Unexpected API Error: 500 Internal Server Error
    at syncRun (/Users/andres/code/latitude/llm/examples/src/utils/syncRun.ts:72:7)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async run (/Users/andres/code/latitude/llm/examples/src/cases/content-moderation-system/example.ts:69:20) {
  status: 500,
  serverResponse: 'Internal Server Error',
  errorCode: 'internal_server_error',
  dbErrorRef: undefined
```

## After
```bash
LatitudeApiError: Unexpected API Error: 422 Error validating chain:
 Variable 'content_type' is not declared
    at syncRun (/Users/andres/code/latitude/llm/examples/src/utils/syncRun.ts:72:7)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async run (/Users/andres/code/latitude/llm/examples/src/cases/content-moderation-system/example.ts:69:20) {
  status: 422,
  serverResponse: `{"name":"DocumentRunError","message":"Error validating chain:\\n Variable 'content_type' is not declared","details":{"compileCode":"variable-not-declared","message":"Variable 'content_type' is not declared","frame":" 97: <user>  
\\n 98: Content to moderate: {{ content }}  
\\n 99: Content type: {{ content_type }}  
\\n\\n                       ^~~~~~~~~~~~\\n100: Platform context: {{ platform_context }}  \\n101: </user>"}}`,
  errorCode: 'internal_server_error',
  dbErrorRef: undefined
```


packages/core/src/services/agents/run.ts
```diff
@@ -65,7 +66,16 @@ export function runAgent<T extends boolean, C extends SomeChain>({
...
+        // Serialized error is not a ChainError but at this point
+        // we throw the error converted again into a ChainError so the gateway can
+        // catch it with all the details
+        const streamError = value.data.error as ChainError<RunErrorCodes>
+        throw new ChainError({
+          message: streamError.message,
+          statusCode: streamError.statusCode,
+          // @ts-ignore
+          details: streamError.details,
+        })
```
andresgutgon, line 78: This is the fix. In the gateway the Hono `onError` catch all middleware were ignoring these errors because what we get from the stream in `value.data.error` is not a JS Error. It's an object

So the error never arrived here:
https://github.com/latitude-dev/latitude-llm/blob/main/apps/gateway/src/middlewares/errorHandler.ts#L29

Even it looks it should because we do a throw here:
https://github.com/latitude-dev/latitude-llm/blob/main/apps/gateway/src/routes/api/v3/documents/run/run.handler.ts#L88

andresgutgon:
All `runAgent` service is **not covered** by tests. We should make some time to improve this. What do you think?

packages/core/src/lib/chainStreamManager/ChainStreamConsumer/consumeStream.ts
```diff
@@ -41,6 +40,7 @@ export async function consumeStream({
   )) {
     if (chunk.type === 'error') {
       finishReason = 'error'
+      // @ts-ignore
```
geclos, line 43: why this?

PR #1338, opened 2025-05-30
feat: reactivate touch event handlers with some buffering

geclos:
Also added tests

### DIFF
--- a/packages/core/src/utils/bufferOperation.ts
+++ b/packages/core/src/utils/bufferOperation.ts
@@ -1,0 +1,29 @@
+import { cache } from '../cache'
+import { TypedResult } from '../lib/Result'
+
+/**
+ * Buffers an operation to prevent it from being executed too frequently
+ * @param key The base key to use for caching
+ * @param operation The async operation to execute if not buffered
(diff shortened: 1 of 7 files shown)


packages/core/src/utils/bufferOperation.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @returns The result of the operation, or undefined if buffered
+ */
+export async function bufferOperation<T>(
+  key: string,
+  operation: () => Promise<TypedResult<T, Error>>,
+  bufferTimeSeconds: number,
+): Promise<T | undefined> {
+  const redis = await cache()
+
+  const lastExecuted = await redis.get(key)
```
neoxelox, line 18: As this is not a select for update, there could be a universe that this suffers from a data race. But I guess its okay, only a couple of tasks will reach that data race and then lock the the operation.

neoxelox (dismissed):

andresgutgon (dismissed):

packages/core/src/utils/bufferOperation.ts
geclos: Yeah it's not dramatic

PR #1353, opened 2025-06-02
Display dataset file size limit error

andresgutgon:
image


apps/web/src/app/api/datasets/create/route.ts
```diff
@@ -0,0 +1,154 @@
...
+          name,
+          file: dataset_file,
+          csvDelimiter: delimiter,
+        },
+      }).then((r) => r.unwrap())
+
+      return NextResponse.json({ success: true, dataset: result.dataset })
+    },
+  ),
+)
```
geclos, line 154: un testecito o dos?

andresgutgon: Let me copilot this for you

apps/web/src/actions/datasets/create.ts
daavidrgz, line 1: Why move away from server actions? Just for curiosity :smiley:

andresgutgon: limit configured in next.config was not working

PR #1358, opened 2025-06-04
feat: move prompt render to worker threads

geclos:
Prompt rendering is causing cpu bottlenecking in Gateway. This PR moves the prompt rendering to a worker thread in order to keep Gateway responsive while render a prompt.


packages/core/src/services/commits/RunDocumentChecker/index.ts
```diff
@@ -88,10 +87,9 @@ export class RunDocumentChecker {
         if (processedParameters.error) return processedParameters
 
         return Result.ok({
-          chain: new PromptlChain({
+          chain: await createPromptlChain({
             prompt: metadata.resolvedPrompt,
             parameters: processedParameters.unwrap(),
-            adapter: Adapters.default,
```
neoxelox, line 94: Don't delete this. Although it's called "default", its not actually the default adapter, the default one is OpenAI and this is PromptL's.

packages/core/src/services/evaluationsV2/llm/shared.ts
```diff
@@ -98,10 +99,9 @@ export async function buildLlmEvaluationRunFunction<
       ...result.config,
       ...(schema && { schema: zodToJsonSchema(schema, { target: 'openAi' }) }),
     } as LatitudePromptConfig
-    promptChain = new PromptlChain({
+    promptChain = await createPromptlChain({
       prompt: prompt,
       parameters: parameters,
-      adapter: Adapters.default,
```
neoxelox, line 104: Don't delete this. Although it's called "default", its not actually the default adapter, the default one is OpenAI and this is PromptL's.

packages/core/src/services/prompts/run.ts
```diff
@@ -55,7 +55,8 @@ export async function runPrompt({
     metadata = await scan({
       prompt,
     })
-    chain = new PromptlChain({
+
+    chain = await createPromptlChain({
```
neoxelox: Rarete that this does not have the default adapter

PR #1363, opened 2025-06-05
feat: move prompt parsing to worker threads

geclos:


apps/gateway/package.json
```diff
@@ -30,7 +30,7 @@
     "hono": "^4.6.6",
     "ioredis": "^5.4.1",
     "lodash-es": "^4.17.21",
-    "promptl-ai": "^0.6.1",
+    "promptl-ai": "catalog:",
```
andresgutgon, line 33: What's a catalog? xD

geclos: it allows you to define a single dependency versions for all packages using that dependency

andresgutgon: Oh nice

PR #1365, opened 2025-06-05
Improved JSONSchema validation

csansoon:
The `JSONSchema` validation we're performing on tool definition is too basic, which fails on more complex definitions.

A more precise definition of the JSONSchema using Zod has been defined.

### DIFF
--- a/packages/constants/src/latitudePromptSchema/zodJsonSchema.ts
+++ b/packages/constants/src/latitudePromptSchema/zodJsonSchema.ts
@@ -1,0 +1,187 @@
+import { z } from 'zod'
+
+const JSON_SCHEMA_TYPES = [
+  'string',
+  'number',
+  'integer',
+  'boolean',
+  'object',
+  'array',
+  'null',
+] as const
+
+const JSON_ENUM_VALUES = z.union([
+  z.string(),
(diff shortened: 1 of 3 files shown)


packages/constants/src/latitudePromptSchema/zodJsonSchema.ts
```diff
@@ -0,0 +1,187 @@
...
+      // Combining / Logic keywords
+      allOf: z.array(zodJsonSchema).optional(),
+      anyOf: z.array(zodJsonSchema).optional(),
+      oneOf: z.array(zodJsonSchema).optional(),
+      not: zodJsonSchema.optional(),
+
+      //? Conditional (draft‐07+):
+      // if: zodJsonSchema.optional(),
+      // then: zodJsonSchema.optional(),
+      // else: zodJsonSchema.optional(),
```
andresgutgon, line 89: llm comments or you?

csansoon: me xd

these attributes were added on jsonschema draft-07, and I don't think we should support them for now. So I added them as comments to acknowledge them, but didn't enable them for now

PR #1383, opened 2025-06-12
fix: tc

csansoon:


packages/core/src/services/copilot/latte/helpers.ts
```diff
@@ -198,6 +198,7 @@ export async function scanDocuments(
         prompt: document.content,
         fullPath: document.path,
         referenceFn,
+        // @ts-ignore – TODO: fix "Type instantiation is excessively deep and possibly infinite"
```
andresgutgon: Maybe use `@ts-expect-error - Problem with type instantiation is excessively deep`

The difference, I think, is that this will fail by complaining that this comment is not necessary when we fix the underlying problem.

andresgutgon (approved):

PR #1385, opened 2025-06-12
Upgrade eslint to version 9

geclos:
ESLint was upgraded to version 9 across the monorepo.

*   The root `package.json` was updated to ESLint 9.17.0, and `pnpm-lock.yaml` was refreshed.
*   The ESLint configuration was migrated from the legacy `.eslintrc.json` format to the new flat config `eslint.config.cjs`.
*   All package-specific `.eslintrc.json` files were replaced with `eslint.config.cjs` files, importing a new shared configuration from `tools/eslint/library.cjs`.
*   Configuration files were renamed to `.cjs` to ensure CommonJS compatibility in ES module environments.
*   The shared configuration in `tools/eslint/library.cjs` was rewritten to use ESLint 9's flat config API, including `@eslint/js` and `typescript-eslint` v8.
*   The `apps/web/eslint.config.cjs` was updated to integrate `@next/eslint-plugin-next` for Next.js-specific flat configuration.
*   The `eslint-config-turbo` import was removed from the shared config, as it does not yet support the flat config format.
*   Obsolete dependencies were removed from `tools/eslint/package.json`.
*   All original `.eslintrc.json` files were deleted.
*   The upgrade successfully enabled ESLint 9, identifying new linting issues.


tools/eslint/package.json
```diff
@@ -3,22 +3,23 @@
   "version": "0.0.0",
   "private": true,
   "files": [
-    "library.js"
+    "library.cjs"
   ],
   "exports": {
     ".": {
-      "import": "./library.js",
-      "require": "./library.js"
```
andresgutgon, line 11: Why `.cjs`? Why no `.mjs`? use modern ESM

geclos:
giving the job to Codex

PR #1388, opened 2025-06-13
Fix lag* when clicking on tools (Integrations) toggle

andresgutgon:
# What?
There is something wrong in the way we update that state

# TODO
- [x] Take a look and fix it
- [x] When is a custom provider supposed to show a clear input to enter a model name?


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptIntegrations/utils.ts
```diff
@@ -1,202 +0,0 @@
...
-    setLocalConfig(promptConfig as LatitudePromptConfig)
-  }, [promptConfig])
-
-  const setConfig = useCallback(
-    (config: Record<string, unknown>) => {
-      setLocalConfig(config as LatitudePromptConfig)
-      debouncedSetPromptConfig(config)
-    },
-    [debouncedSetPromptConfig, setLocalConfig],
-  )
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptIntegrations/utils.ts:165]: The problem was that we were using a `useEffect` and also this debounce. I refactored to use `useEvents`, which allows me to receive input from outside. However, the internal state that this hook (new hook `useActiveIntegrations`) outputs is only updated when an action occurs.

apps/web/src/components/EditorHeader/ProviderModelSelector/index.tsx
```diff
@@ -156,10 +157,14 @@ export function ProviderModelSelector({
...
-        provider: provider?.name || '',
-        model: selectedModel,
-      })
+      const updatedPrompt = updatePromptMetadata(
+        prompt,
+        {
+          provider: provider?.name || '',
+          model: selectedModel,
+        },
+        { keysToBeRemovedWhenNull: ['model'] },
```
andresgutgon, line 166: I touched this hook to allow removing null values so we don't have `model: ""`

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptIntegrations/useActiveIntegrations.test.ts
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/PromptIntegrations/useActiveIntegrations.test.ts:1]: Vibe tests for the new hook

PR #1400, opened 2025-06-16
Playground simple block editor. AST to Simple blocks

andresgutgon:
# What?
We want to offer a simpler playground experience for building agents where advanced features like for loops or conditionals are not so critical. That's our hypothesis. So, for that, we're going to offer a simple mode experience where the user introduces the prompts as plain text (kind of)

### TODO
- [x] Feature flag to show toggle between current (advance) and simple editors
- [x] Promptl AST -> Simple Editor blocks 
- [x] Simple Editor Blocks -> Promptl Text
- [x] AST to Blocks for `<step />` tag
- [x] AST to Blocks for `<system />` tag
- [x] AST to Blocks for `<user />` tag
- [x] AST to Blocks for `<assistant />` tag
- [x] Simple block for `<prompt />` tag
- [x] Simple block for `<content-image />` tag
- [x] Add tests for `astToSimpleBlocks` function 
- [x] Handle nested message tags inside `<step />` tags
- [x] Handde nested content (prompt, image, file, and tool-call) tags inside `<step />`
- [x] Handle nested content (prompt, image, file, and tool-call) tags inside Message tags
- [x] Simple block for `<content-file />` tag
- [x] Simple block for `<tool-call />` tag
- [x] Map Errors with blocks
- [x] ~Think about performance and unique block IDs~ didn't think much but I think is ok, let's see

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/BlocksEditor/index.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/BlocksEditor/index.tsx
@@ -1,0 +1,29 @@
+import { memo, Suspense } from 'react'
+import { AstError } from '@latitude-data/constants/simpleBlocks'
+import { TextEditorPlaceholder } from '@latitude-data/web-ui/molecules/TextEditorPlaceholder'
+import { AnyBlock } from '@latitude-data/constants/simpleBlocks'
+import { CodeBlock } from '@latitude-data/web-ui/atoms/CodeBlock'
+
+export const PlaygroundBlocksEditor = memo(
+  ({
+    value: _prompt,
+    blocks = [],
+  }: {
+    compileErrors: AstError[] | undefined
+    blocks: AnyBlock[] | undefined
+    value: string
+    defaultValue?: string
+    isSaved: boolean
+    readOnlyMessage?: string
+    onChange: (value: string) => void
+  }) => {
+    if (!blocks.length) return null
+
+    return (
+      <Suspense fallback={<TextEditorPlaceholder />}>
+        Blocks Editor
+        <CodeBlock language='json'>{JSON.stringify(blocks, null, 2)}</CodeBlock>
+      </Suspense>
+    )
+  },
+)
--- a/apps/web/src/workers/readMetadata.ts
+++ b/apps/web/src/workers/readMetadata.ts
@@ -1,17 +1,26 @@
-import {
-  readMetadata,
-  CompileError as CompilerCompileError,
-} from '@latitude-data/compiler'
 import { AgentToolsMap, resolveRelativePath } from '@latitude-data/constants'
 import { latitudePromptConfigSchema } from '@latitude-data/constants/latitudePromptSchema'
+import {
+  astToSimpleBlocks,
+  type AnyBlock,
+} from '@latitude-data/constants/simpleBlocks'
 import type { DocumentVersion } from '@latitude-data/core/browser'
+import { AstError } from '@latitude-data/constants/simpleBlocks'
 
-import { CompileError as PromptlCompileError, scan } from 'promptl-ai'
+import {
+  CompileError as PromptlCompileError,
+  ConversationMetadata as PromptlConversationMetadata,
+  Fragment,
+  scan,
+  parse,
+} from 'promptl-ai'
 
-type CompileError = PromptlCompileError | CompilerCompileError
+type CompileError = PromptlCompileError
 
-export type ReadMetadataWorkerProps = Parameters<typeof readMetadata>[0] & {
+type EditorType = 'code' | 'visual'
+export type ReadMetadataWorkerProps = Parameters<typeof scan>[0] & {
   promptlVersion: number
+  editorType: EditorType
   document?: DocumentVersion
   documents?: DocumentVersion[]
   providerNames?: string[]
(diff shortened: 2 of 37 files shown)


apps/web/src/workers/readMetadata.ts
```diff
@@ -20,6 +31,46 @@ export type ReadMetadataWorkerProps = Parameters<typeof readMetadata>[0] & {
...
+  metadata: CompilerConversationMetadata | PromptlConversationMetadata
+  // TODO: This nullable is only needed until we kill the old compiler
+  ast?: Fragment
+}) {
+  const { setConfig: _, errors: errors, ...returnedMetadata } = metadata
+
+  let simpleBlocks: SimpleBlock[] = []
+
+  if (ast) {
+    simpleBlocks = astToSimpleBlocks({ ast })
```
andresgutgon: My idea is to process the AST here and convert it into simple blocks. Maybe only do this work when simple editor is used. For now I'll just comment this line before merging

☝️ BEFORE MERGE ⚠️

apps/web/src/workers/readMetadata.ts
```diff
@@ -54,32 +105,29 @@ self.onmessage = async function (event: { data: ReadMetadataWorkerProps }) {
...
+  if (isPromptl) {
+    const ast = parse(prompt)
+    const scanParams = {
+      prompt,
+      serialized: ast,
+      fullPath: document?.path,
+      referenceFn: props.referenceFn,
+      withParamters: props.withParameters,
+      requireConfig: true,
+      configSchema: props.configSchema as any,
```
andresgutgon: @csansoon Am I leaving something here? I think I don't but let's check 👀

packages/constants/src/simpleBlocks/astToSimpleBlocks.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    case 'IfBlock':
+      return convertIfBlockToText(node as IParser.IfBlock, insideStep)
+
+    case 'ForBlock':
+      return convertForBlockToText(node as IParser.ForBlock, insideStep)
+
+    case 'ElementTag': {
+      const tag = node as IParser.ElementTag
+      // Always convert element tags to text when they're nested in content
```
andresgutgon: Review this

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/BlocksEditor/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+    isSaved: boolean
+    readOnlyMessage?: string
+    onChange: (value: string) => void
+  }) => {
+    if (!blocks.length) return null
+
+    return (
+      <Suspense fallback={<TextEditorPlaceholder />}>
+        Blocks Editor
+        <CodeBlock language='json'>{JSON.stringify(blocks, null, 2)}</CodeBlock>
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/BlocksEditor/index.tsx:25]: Temporal until implementing blocks editor

packages/constants/src/simpleBlocks/index.test.ts
andresgutgon, line 1: `T` teeeests

packages/constants/src/simpleBlocks/astParsingUtils.ts
```diff
@@ -0,0 +1,335 @@
...
+
+    case 'ElementTag': {
+      const tag = node as ElementTag
+      // Always convert element tags to text when they're nested in content
+      return convertElementToText(tag, insideStep)
+    }
+
+    case 'Config':
+      // Config nodes should not appear in content
+      return ''
```
andresgutgon, line 129: This might change in the future. For now ignore

apps/web/src/workers/readMetadata.ts
```diff
@@ -54,57 +143,27 @@ self.onmessage = async function (event: { data: ReadMetadataWorkerProps }) {
...
-      }
-    }
-
-    const content = documents.find((d) => d.path === fullPath)?.content
-    if (content === undefined) return undefined
-
-    return {
-      path: fullPath,
-      content,
+  if (isPromptl) {
```
geclos: you can remove support from old syntax, there's already a warning in UI that says support for it was deprecated 1st of june.

apps/web/src/workers/readMetadata.ts
```diff
@@ -20,11 +34,86 @@ export type ReadMetadataWorkerProps = Parameters<typeof readMetadata>[0] & {
...
+  if (ast && editorType === 'visual') {
+    blocks = astToSimpleBlocks({ ast, prompt, errors })
+  }
+
+  return {
+    ...(returnedMetadata as PromptlConversationMetadata),
+    errors,
+    blocks,
+  }
+}
```
geclos, line 99: nitpick but i find it makes readability worse having exports at the bottom and all other declarations at the top, it startsl the file with implementation details rather than the full picture

apps/web/src/workers/readMetadata.ts
```diff
@@ -20,11 +34,86 @@ export type ReadMetadataWorkerProps = Parameters<typeof readMetadata>[0] & {
...
+}: {
+  prompt: string
+  editorType: EditorType
+  metadata: CompilerConversationMetadata | PromptlConversationMetadata
+  // TODO: This nullable is only needed until we kill the old compiler
+  ast?: Fragment
+}) {
+  const { setConfig: _, errors: rawErrors, ...returnedMetadata } = metadata
+  const errors = rawErrors.map((error: CompileError) => {
+    // TODO: This is not necessary when we remove old compiler
```
geclos: remove it, support for it has already ended

packages/constants/src/simpleBlocks/astToSimpleBlocks.ts
```diff
@@ -0,0 +1,318 @@
...
+  ast,
+  prompt,
+  errors = [],
+}: {
+  ast: Fragment
+  prompt: string
+  errors?: AstError[]
+}): AnyBlock[] {
+  const generateIdFn = createIdGenerator()
+  return processNodes({ nodes: ast.children, generateIdFn, prompt, errors })
```
geclos, line 317: no point of processNodes function since it's a local function and you are immediately calling it

packages/constants/src/simpleBlocks/astToSimpleBlocks.ts
```diff
@@ -0,0 +1,318 @@
...
+}
```
geclos, line 318: tests

packages/constants/src/simpleBlocks/index.ts
```diff
@@ -0,0 +1,3 @@
+export type { AnyBlock, AstError } from './types'
+export * from './astToSimpleBlocks'
+export * from './simpleBlocksToText'
```
geclos, line 3: don't create index files

packages/web-ui/src/ds/atoms/Switch/index.tsx
```diff
@@ -86,7 +88,9 @@ function SwitchInput({
 
   return (
     <div
-      className={cn('flex flex-col gap-y-2 w-full', className)}
+      className={cn('flex flex-col gap-y-2', className, {
+        'w-full': fullWidth,
```
geclos, line 92: you are changing the devault behaviour of all switch inputs

apps/web/src/workers/readMetadata.ts
andresgutgon: I have the exact nitpick. Just in the opposite direction 😂. We're here in another sort of import dilemma

packages/constants/src/simpleBlocks/astToSimpleBlocks.ts
andresgutgon: image

packages/web-ui/src/ds/atoms/Switch/index.tsx
andresgutgon: No, `fullWidth = true`

packages/constants/src/simpleBlocks/index.ts
andresgutgon: In this case, it is super small, why not?

packages/constants/src/simpleBlocks/astToSimpleBlocks.ts
andresgutgon: It's recursive

geclos: doesn't matter no? just export processNodes

packages/constants/src/simpleBlocks/index.ts
geclos: because it starts small it becomes big and then we end up with the issues of importing more stuff than needed, it's better to just not create them

packages/constants/src/simpleBlocks/astToSimpleBlocks.ts
andresgutgon:
```javascript
const generateIdFn = createIdGenerator()
```
I'm doing a closure here to have a custom ID generator. I can change but I'll leave this way until I see it works with the visual editor. Maybe I need to change it anyway

PR #1403, opened 2025-06-16
Feature/local sync

geclos:


.gitignore
```diff
@@ -76,3 +76,6 @@ bin/connect.sh
 # This is to put there things like promptl source code so we can tell AI how it works.
 # Usually AI editors only pick files from inside the git project.
 ONLY_FOR_COPILOT/
+
+AGENTS.md
```
andresgutgon, line 80: Shouldn't we have this for everyone in the repo with the same rules?

geclos: This is for opencode which I'm testing

geclos:
no longer relevant

PR #1405, opened 2025-06-18
Latte: List and Undo changes

csansoon:
https://github.com/user-attachments/assets/7e8c3fdd-250e-4dda-a830-21b5e87dd745

### How it works
When Latte performs any actions to a project, a new Checkpoint is generated for that document. Each checkpoint contain the state the document had before Latte changed it, which will then be restored when clicking on the "Undo" button.
Latte can do further changes on prompts that it already edited before, and won't create new checkpoints if they already exist.

Accepting the Latte changes will remove all checkpoints from that thread.


packages/constants/src/index.ts
```diff
@@ -74,3 +74,17 @@ export const OPENAI_PROVIDER_ENDPOINTS = [
...
+  UpdatedPath = 'updated_path',
+  Deleted = 'deleted',
+}
+
+export type ChangedDocument = {
+  documentUuid: string
+  path: string
+  errors: number
+  changeType: ModifiedDocumentType
+}
```
geclos: put these somewhere else rather than index

packages/constants/src/models.ts
```diff
@@ -69,15 +69,22 @@ export type ProviderLog = {
...
   documentUuid: string
+  commitId: number
   path: string
   content: string
   resolvedContent: string | null
-  contentHash: string
-  promptlVersion: number
-  commitId: number
+  contentHash: string | null
   datasetId: number | null
```
geclos, line 81: @andresgutgon can we remove this column?

packages/core/src/services/copilot/latte/threads/index.ts
```diff
@@ -1,4 +1,4 @@
 export * from './createThread'
-export * from './createCheckpoint'
-export * from './undoChanges'
-export * from './clearCheckpoints'
+export * from './checkpoints/createCheckpoint'
+export * from './checkpoints/undoChanges'
+export * from './checkpoints/clearCheckpoints'
```
geclos: remove the index file if it's not a massive pita (and don't use more index files please)

packages/core/src/services/copilot/latte/tools/projects/edit.ts
```diff
@@ -210,17 +242,16 @@ const editProject = defineLatteTool(
...
+          .filter(([, errors]) => !!errors?.length),
       )
+
+      return Result.ok({
+        changes: latteChanges,
+        errors,
+      })
     })
   },
   z.object({
```
geclos: huge service, can be split in smaller composable pieces

packages/core/src/services/documents/create.ts
```diff
@@ -61,6 +61,7 @@ export async function createNewDocument(
     content,
     promptlVersion = 1,
     createDemoEvaluation: demoEvaluation = false,
+    includeDefaultContent = true,
```
geclos, line 64: instead of a boolean prop, that leaks implementation details, you could declare a content prop with a default value

geclos (approved):

packages/constants/src/models.ts
andresgutgon: No, why?

geclos: because there's datasetv2id column

packages/core/src/services/documents/create.ts
csansoon: But the default value is generated dynamically inside the method, depending on the workspace provider settings

packages/constants/src/models.ts
andresgutgon: Ah true, yes, this is the old one then. Also these models are fake

PR #1435, opened 2025-06-27
Pipedream Integrations

csansoon:
https://github.com/user-attachments/assets/4a8867c5-88b7-45bc-92dc-6ddde1c5c768


apps/web/src/app/(private)/settings/_components/Integrations/New/_components/Configuration/External.tsx
```diff
@@ -1,69 +1,48 @@
...
+export const ExternalIntegrationConfiguration = forwardRef<
+  {
+    validate: () => Promise<IntegrationConfiguration>
+  },
+  {}
+>((_, ref) => {
+  const { execute, isPending } = useLatitudeAction(pingCustomMcpAction)
+  const [url, setUrl] = useState<string>('')
+
+  useImperativeHandle(ref, () => ({
```
andresgutgon: TIL

apps/web/src/hooks/usePipedreamApp.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+  )
+
+  const { data = undefined, isLoading } = useSWR<AppDto>(
+    ['pipedreamApp', slugName],
+    fetcher,
+  )
+
+  return { data, isLoading }
+}
```
andresgutgon: This should be in stores, no?

apps/web/src/app/(private)/settings/_components/Integrations/New/index.tsx
```diff
@@ -105,6 +44,48 @@ export default function NewIntegration() {
...
+      setIsCreating(false)
+      setError(err as Error)
+      return
+    }
+
+    const [_, createError] = await create({
+      ...integration,
+      name: integrationName,
+    })
+    setIsCreating(false)
```
geclos, line 75: this state should come from the create action rather than being local state

andresgutgon (approved):

apps/web/src/hooks/usePipedreamApp.ts
geclos: +1

apps/web/src/hooks/usePipedreamApp.ts
```diff
@@ -0,0 +1,29 @@
...
+      },
+    },
+  )
+
+  const { data = undefined, isLoading } = useSWR<AppDto>(
+    ['pipedreamApp', slugName],
+    fetcher,
+  )
+
+  return { data, isLoading }
```
geclos: usememo

apps/web/src/hooks/useConnectToPipedreamApp.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    return promise
+  }, [app, token, tokenExpiresAt])
+
+  return {
+    connect,
+    isLoading: !token || isConnecting,
+    connectionId,
+    externalUserId,
+  }
```
geclos: usememo

apps/web/src/stores/integrationTools.ts
```diff
@@ -27,13 +36,51 @@ export default function useIntegrationTools(
...
+  const displayNameFetcher = useFetcher<AppDto | undefined, AppResponse>(
+    integration?.type === IntegrationType.Pipedream
+      ? ROUTES.api.integrations.pipedream.detail(
+          integration.configuration.appName,
+        ).root
+      : undefined,
+    {
+      serializer: (response) => {
+        if (!response.ok) return undefined
+        return response.data
```
geclos, line 56: this serializer is doing nothing no?

csansoon: The request does not fail when there's an error, intead it returns an `ok` and `data` or `error` attributes. This serializaer is returning the `data` value when it succeeds, and ignoring any errores if it fails (in this case, an error happening is not critical at all, this request is only being done to improve the readability in tool names)

geclos (dismissed):

apps/web/src/app/(private)/settings/_components/Integrations/New/index.tsx
csansoon: In this case, the create action is only executed AFTER an async validation is performed, which lives outside of the create action and has no reactive loading state. While this is the case, the same loading state must be displayed to the user, where the button and input fields are disabled.

PR #1444, opened 2025-07-01
Implement blocks editor update cycle

andresgutgon:
# What?
Connect the dots.
Go: Text -> AST -> Blocks -> Lexical Editor
Back: Lexical Editor -> Blocks -> Text -> Repeat

## TODO
- [x] How to handle `<prompt path="./relative/prompt" />`?
- [x] Fix developer message parsing
- [x] Fix generating empty lines between elements `\n` from AST 🐛 
- [x] Convert from AST to Lexical
- [x] Convert from Lexical blocks to TEXT
- [x] Implement parsing of Promptl variables
- [x] 🐛 It's broken the update 
- [x] THINK how to do the changes in a non-intrusive way. Now, even when typing `@`, the typeahead list is re-rendered because we triggered a change. 🔥 
- [x] What to do with `Config`? Write it to Lexical or merge with prev config?. Just do a hidden block and when saving is serialized back

## References
[Nice article explaining Lexical](https://dio.la/article/lexical-state-updates)

## Next
- [ ] Read step `as` prop and convert into lexical variables. Or read from all steps when auto completing. 
- [ ] Style step block + Edit name and `isolate` prop


packages/web-ui/src/ds/molecules/BlocksEditor/Editor/index.tsx
```diff
@@ -217,9 +90,26 @@ export function BlocksEditor({
           className,
         )}
       >
-        <LexicalComposer initialConfig={initialConfig}>
+        <LexicalComposer
+          initialConfig={{
```
andresgutgon: Wrap in a useMemo, maybe

PR #1465, opened 2025-07-13
feat: Add feature toggles for workspaces

geclos:
Implements a db-backed features-toggle functionality.


AGENTS.md
```diff
@@ -0,0 +1,203 @@
...
+
+1. **Database Schema**: Create tables, generate migration
+2. **Services**: Implement business logic with Result pattern
+3. **Repositories**: Create data access layer
+4. **Actions**: Create server actions with validation
+5. **API Routes**: Expose endpoints with proper auth
+6. **Stores**: Create SWR hooks with optimistic updates
+7. **UI Components**: Build interface following design patterns
+8. **Routes**: Add navigation and routing
+9. **Testing**: Validate with linting and type checking
```
geclos, line 207: most AI tools are adoptiong this standard to learn patterns about codebases.

apps/web/src/actions/features/create.ts
```diff
@@ -0,0 +1,22 @@
...
+    }),
+  )
+  .handler(async ({ input }) => {
+    const result = await createFeature({
+      name: input.name,
+      description: input.description,
+    })
+
+    return result.unwrap()
+  })
```
andresgutgon: Move these under actions/admin because are for back-office no?

apps/web/src/app/(admin)/backoffice/features/_components/FeaturesManager.tsx
```diff
@@ -0,0 +1,186 @@
...
+                    Delete
+                  </Button>
+                </TableCell>
+              </TableRow>
+            ))}
+          </TableBody>
+        </Table>
+      </div>
+
+      {/* Workspace Features Section */}
```
andresgutgon: 🔥

packages/core/src/services/workspaceFeatures/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './toggle'
+export * from './findByWorkspace'
+export * from './isFeatureEnabled'
+export * from './findAllWithWorkspaceStatus'
+export * from './toggleForMultipleWorkspaces'
+export * from './getEnabledWorkspaceIds'
```
andresgutgon: Didn't we want to avoid barrel files?

geclos: yeah, AI doing stuff

apps/web/src/stores/useFeature.ts
```diff
@@ -0,0 +1,34 @@
...
+  workspaceId: number | undefined,
+  featureName: string,
+  opts?: SWRConfiguration,
+) {
+  const key =
+    workspaceId && featureName ? `api/workspaceFeatures/${featureName}` : null
+
+  const fetcher = useFetcher<FeatureStatus>(
+    workspaceId && featureName
+      ? API_ROUTES.workspaceFeatures.byName(featureName)
```
andresgutgon, line 19: Why N requests? 

This could be a single fetch with all features. N is going to be super small. Also this could be fetched in the server and pass to the store. 

Don't think even makes sense to have a store for this just put in a zustand state and ask if X feature is there. If features change just refresh. Probably more things have changed

geclos: It... doesnt' really matter? Like yes I agree it would be more efficient to fetch the features with their enabled state but we are splitting hairs here and I want to move on from this – which ain't critical. AI slop is good enough this time.

andresgutgon: Is worst code and more code but ok

PR #1471, opened 2025-07-14
Improve Pipedream tool runs

csansoon:
### Ignored properties
Some arguments for Pipedream tools are hidden because they require other properties to be configured first. 
Even if you hardcode these hidden properties into the tool request, running the tools via Pipedream's MCP implementation will not use those values. 
Now, Pipedream Integrations run tools directly via the SDK, bypassing their MCP.

### Default values
I have added default values for common properties from their integrations. Sending messages using Pipedream's integration for Slack and Discord will no longer include Pipedream's watermark by default. In addition to this, messages sent via Pipedream's Slack integration will use Latitude as the name and Latitude's logo as the image by default.


packages/core/src/services/integrations/pipedream/components.ts
```diff
@@ -156,3 +204,40 @@ export async function configureComponent({
...
+    return Result.error(configuredPropsResult.error)
+  }
+
+  const result = await pipedream.runAction({
+    externalUserId: integration.configuration.externalUserId,
+    actionId: {
+      key: toolName,
+    },
+    configuredProps: configuredPropsResult.unwrap(),
+  })
```
geclos, line 240: can this await throw? if so handle it and return a result error

packages/core/src/services/integrations/pipedream/components.ts
```diff
@@ -156,3 +204,40 @@ export async function configureComponent({
...
+    return Result.error(pipedreamEnv.error!)
+  }
+
+  const pipedream = createBackendClient(pipedreamEnv.unwrap())
+  const configuredPropsResult = await fillConfiguredProps({
+    pipedream,
+    integration,
+    componentId: toolName,
+    configuredProps: args,
+  })
```
geclos, line 228: ditto if the await can fail handle it

PR #1474, opened 2025-07-15
Prevent removing integrations with configured triggers

csansoon:
Removing an Integration that has Triggers already configured in some documents will break stuff. We must make sure all triggers are deleted before removing the integration.


packages/core/src/services/integrations/references.ts
```diff
@@ -0,0 +1,33 @@
...
+    'REFERENCES FOR INTEGRATION',
+    integration.name,
+    references.map((r) => r.documentUuid),
+  )
+
+  return Result.ok(
+    references.map((trigger) => ({
+      projectId: trigger.projectId,
+      documentUuid: trigger.documentUuid,
+      path: 'daaaamnnn',
```
learningbizz: Whats this? xd

Can't we return the real path available in Document row?

apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx
```diff
@@ -1,11 +1,61 @@
...
+      <div className='flex flex-row items-center gap-2 overflow-hidden truncate'>
+        <Icon name='file' />
+        {project ? (
+          <Text.H6 noWrap ellipsis>
+            {project.name}
+          </Text.H6>
+        ) : (
+          <Skeleton className='w-32' height='h6' />
+        )}
+        <Text.H6 color='foregroundMuted'>|</Text.H6>
```
Comment (learningbizz) [inline apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx:46]: Any tests for this new logic?

packages/core/src/services/integrations/references.ts
```diff
@@ -0,0 +1,33 @@
...
+  )
+  const references = await triggersScope.findByIntegrationId(integration.id)
+
+  console.log(
+    'REFERENCES FOR INTEGRATION',
+    integration.name,
+    references.map((r) => r.documentUuid),
+  )
+
+  return Result.ok(
```
learningbizz, line 19: Tests to make sure the new endpoint works as expected?

apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx
```diff
@@ -29,6 +81,24 @@ export default function DestroyIntegration({
...
       model={integration}
       onSuccess={() => navigate.push(ROUTES.settings.root)}
-    />
+      disabled={isLoading || references.length > 0}
+    >
+      {references.length > 0 && (
+        <div className='flex flex-col gap-4'>
+          <Text.H5>
+            This integration is currently being used in {references.length}{' '}
+            prompts:
```
Comment (learningbizz) [inline apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx:90]: We're implementing this logic in the frontend, but the backend has no safety guard to check this.

If by any reason we delete an integration somewhere else throughout the code, this might become a bug. For this reason, I would also add the safeguard for the destroy class in core

apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx
Comment (learningbizz) [reply apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx:46]: We have no tests for frontend components, so we can disregard this comment

packages/constants/src/integrations.ts
```diff
@@ -108,3 +108,10 @@ export enum HostedIntegrationType {
   // Hubspot = 'hubspot', // Docker based
   // Loops = 'loops', // Does not exist
 }
+
+export type IntegrationReference = {
+  projectId: number
+  documentUuid: string
+  asTrigger: boolean
+  // asTool: boolean // TODO: This would be really useful
```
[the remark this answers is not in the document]
csansoon: nono that's not implemented

learningbizz (dismissed):
Looks all good now

apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx
```diff
@@ -29,6 +81,24 @@ export default function DestroyIntegration({
...
+          <div className='flex flex-col gap-2'>
+            {references.map((reference, idx) => (
+              <ReferenceItem key={idx} reference={reference} />
+            ))}
+          </div>
+          <Text.H5>
+            You must remove them all before deleting the integration.
+          </Text.H5>
+        </div>
+      )}
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx:101]: A better approach is what Manu said. Let it fail by going to the service where the validation is. Never trust the client.

Also this approach to improve DX and let users know the can not delete this integration. 
https://github.com/latitude-dev/latitude-llm/pull/1474/files#r2209791130

apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx
```diff
@@ -17,6 +67,8 @@ export default function DestroyIntegration({
   const { data, destroy, isDestroying } = useIntegrations()
   const integration = data.find((p) => p.id === Number(integrationId))
 
+  const { data: references, isLoading } = useIntegrationReferences(integration)
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx:70]: Wait wait. Also this. You're forced to implement a full store just because this nextjs page is client side. Remove the `use client`

Now this will be a server page where you can directly use our core collections and do these 2 queries

1. Get integrations
2. Get integrations references 


Then you show a [ConfirmModal](https://github.com/latitude-dev/latitude-llm/blob/main/packages/web-ui/src/ds/atoms/Modal/index.tsx#L146) if references exists with variant `destructive`. The action is just close the modal

Otherwise, show the destroy modal. No need to change the behaviour and implement `children` prop. Destroy modal is simple for a reason. 

To handle the client on click you just need to move the Destroy modal in another component where `use client` is set.

Setting `use client` in nextjs `page.tsx` is anti-pattern

apps/web/src/app/api/integrations/[integrationName]/references/route.ts
```diff
@@ -0,0 +1,70 @@
...
+      return NextResponse.json(
+        {
+          ok: true,
+          data: result.unwrap(),
+        },
+        { status: 200 },
+      )
+    },
+  ),
+)
```
Comment (andresgutgon) [inline apps/web/src/app/api/integrations/[integrationName]/references/route.ts:70]: 🔥 No need to implement client side store or api

apps/web/src/stores/integrationReferences.ts
```diff
@@ -0,0 +1,46 @@
...
+    fetcher,
+    opts,
+  )
+
+  return {
+    data,
+    isLoading,
+    error,
+  }
+}
```
andresgutgon, line 46: 🔥 ditto

apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx:101]: I still want to show the button as disabled, I don't want to display this information AFTER clicking delete, that's a bad experience for the user.

Still, I also added the check in the backend

apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/settings/integrations/[integrationId]/destroy/page.tsx:70]: Nah but I like the hook, I'll be using it more overs other places in the ui :)

PR #1475, opened 2025-07-16
Tracing frontend

neoxelox:
Related to: https://github.com/latitude-dev/latitude-llm/issues/1273
Screenshot 2025-07-15 at 23 38 23


apps/web/src/app/_lib/formatUtils.ts
```diff
@@ -3,7 +3,7 @@ const MINUTES = 60 * SECONDS
 const HOURS = MINUTES * 60
 
 export function formatDuration(duration?: number | null) {
-  if (!duration) return '-'
+  if (!duration) return '0.000s'
```
geclos, line 6: this is misleading, the duration wasn't 0 we just don't have it

neoxelox: Ah, yeah, this was conflicting with 0 second durations on the spans, I guess we could check for undefined/null and then for 0 value

PR #1479, opened 2025-07-16
fix: latte (again)

csansoon:
Fixes Latte logic that has been broken from the Big Beautiful PR.

Things that broke latte:
 - Tool handling had been removed
 - Streamed messages had a different shape, which was not handled correctly in frontend


apps/web/src/hooks/latte/helpers.ts
```diff
@@ -1,12 +1,12 @@
-import type { ToolCall } from '@latitude-data/constants/legacyCompiler'
 import { LatteTool } from '@latitude-data/constants/latte'
 import { LatteToolStep } from './types'
+import { ToolRequestContent } from '@latitude-data/compiler'
```
geclos: Do not import from compiler please, this should be in legacy compiler and if not move it there

apps/web/src/hooks/latte/index.ts
```diff
@@ -20,6 +22,13 @@ import {
 import { acceptLatteChangesAction } from '$/actions/latte/acceptChanges'
 import { discardLatteChangesActions } from '$/actions/latte/discardChanges'
 import { addFeedbackToLatteChangeAction } from '$/actions/latte/addFeedbackToLatteChange'
+import {
+  ContentType,
+  Message,
+  MessageRole,
+  TextContent,
+} from '@latitude-data/compiler'
```
geclos: No, never import from compiler

apps/web/src/hooks/latte/index.ts
```diff
@@ -20,6 +22,13 @@ import {
...
+import { ToolRequestContent } from '../../../../../packages/compiler/src/types/message'
```
geclos: Nope

packages/core/src/websockets/constants.ts
```diff
@@ -3,7 +3,7 @@
 // All this can be seen in the browser. If you want something private
 // put in other place.
 
-import type { Message } from '@latitude-data/constants/legacyCompiler'
+import type { Message } from '@latitude-data/compiler'
```
geclos: Nope, escape the types if it gives you incompatibilities but don't import from compiler

geclos (dismissed):
Remove any import from compiler package

PR #1482, opened 2025-07-17
feat: fixes from the big beautiful commit

geclos:
Fixes several issues from the big beautiful commit.

### DIFF
--- a/packages/core/src/lib/streamManager/ChainErrors/index.ts
+++ b/packages/core/src/lib/streamManager/ChainErrors/index.ts
@@ -8,14 +8,12 @@ export async function createChainRunError({
   error,
   errorableUuid,
   errorableType,
-  persistErrors,
 }: {
   errorableUuid: string
   error: ChainError<RunErrorCodes> | Error
-  persistErrors: boolean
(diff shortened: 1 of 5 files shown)


csansoon (dismissed):

packages/core/src/lib/streamManager/ChainErrors/index.ts
```diff
@@ -8,14 +8,12 @@ export async function createChainRunError({
...
   errorableType,
-  persistErrors,
 }: {
   errorableUuid: string
   error: ChainError<RunErrorCodes> | Error
-  persistErrors: boolean
   errorableType?: ErrorableEntity
 }) {
-  if (!persistErrors || !errorableType) return error
+  if (!errorableType) return error
```
neoxelox, line 16: Why this change, i think llm evaluation put persistErrors to false and handle themselves the errors!

geclos: nope, llm evals call directly runChain which is not responsible of storing errors anymore, so you don't really need this leakly argument anymore. This argument wasn't even used no more since the big beautiful commit.

PR #1484, opened 2025-07-17
quickfix(chat.handler): bring back toolCalls and trace to messages API

geclos:


apps/gateway/src/routes/api/v3/conversations/chat/chat.handler.ts
```diff
@@ -16,21 +16,28 @@ import { compareVersion } from '$/utils/versionComparison'
...
+  } = c.req.valid('json')
   const workspace = c.get('workspace')
 
   const sdkVersion = c.req.header('X-Latitude-SDK-Version')
+  const isLegacy = !compareVersion(sdkVersion, '5.0.0')
 
   const result = (
     await _addMessages({
-      context: BACKGROUND({ workspaceId: workspace.id }),
+      context: trace ?? BACKGROUND({ workspaceId: workspace.id }),
```
neoxelox: you have to telemetry.resume the incoming trace, ex:

```
 let context = BACKGROUND({ workspaceId: workspace.id })
 if (trace) context = telemetry.resume(trace)
 ```

geclos: uhm, weird that the types don't complain about this?/

PR #1488, opened 2025-07-17
frefactor: Add database parameter to revertCommit function

geclos:


packages/core/src/services/history/revertCommit.ts
```diff
@@ -185,50 +189,61 @@ export async function revertCommit({
...
+      }),
+    )
+
+    for (const result of computedChanges) {
+      if (result.error) return Result.error(result.error)
+    }
 
-  return Result.ok(finalDraft.value)
+    return Result.ok(finalDraft.value)
+  }, db)
```
andresgutgon, line 248: A test no?

geclos: the service is already tested

PR #1490, opened 2025-07-17
feat(core): Do not throw on builtin tool execution errors

geclos:
Builtin tools such as integration, latitude, and subagent tools can error but won't throw so that llm execution can continue. In case of client errors, the client can choose to either throw – which will halt execution – or return an error response and let the llm continue to generate text.


packages/core/src/lib/streamManager/resolveTools/integrationTools.ts
```diff
@@ -138,10 +138,17 @@ async function addIntegrationTools({
...
+              isError: false,
+            }
           } catch (err) {
             $tool?.fail(err as Error)
-            throw err
+
+            return {
+              value: (err as Error).message,
+              isError: true,
+            }
```
csansoon: I'd include the error name too, like a generic Error object: `{ name: string, message: string }`

csansoon (approved):

packages/core/src/lib/streamManager/resolveTools/integrationTools.ts
```diff
@@ -138,10 +138,17 @@ async function addIntegrationTools({
             }).then((r) => r.unwrap())
 
             $tool?.end({ result: { value, isError: false } })
-            return value
+            return {
+              value,
+              isError: false,
+            }
           } catch (err) {
             $tool?.fail(err as Error)
```
neoxelox: change this too

packages/core/src/services/agents/agentsAsTools.ts
```diff
@@ -94,10 +94,17 @@ export async function getToolDefinitionFromDocument({
 
         $span?.end({ result: { value, isError: false } })
 
-        return value
+        return {
+          value,
+          isError: false,
+        }
       } catch (e) {
         $span?.fail(e as Error)
```
neoxelox: change this too

packages/core/src/services/latitudeTools/telemetryWrapper.ts
```diff
@@ -39,9 +54,17 @@ export async function withTelemetryWrapper<
...
     const value = await executeFn(args, toolCall).then((r) => r.unwrap())
     $tool?.end({ result: { value, isError: false } })
-    return value
+
+    return {
+      isError: false,
+      value,
+    }
   } catch (e) {
     $tool?.fail(e as Error)
```
neoxelox: change this too

geclos: protip: you can use `ditto` to not repeat long comments everywhere

packages/core/src/lib/streamManager/resolveTools/integrationTools.ts
geclos: including the full event akshually

Issue #1421, opened 2025-06-23
Make evaluation prompts google compatible


geclos:
They fail because google llms require at least one user message in the prompt

PR #1492, opened 2025-07-18
Clean things after big PR

neoxelox:
As a side quest closes: https://github.com/latitude-dev/latitude-llm/issues/1421

### DIFF
--- a/packages/core/src/services/tracing/spans/completion.ts
+++ b/packages/core/src/services/tracing/spans/completion.ts
@@ -62,6 +62,7 @@ import {
   GEN_AI_RESPONSE_FINISH_REASON_VALUE_OTHER,
   GEN_AI_RESPONSE_FINISH_REASON_VALUE_STOP,
   GEN_AI_RESPONSE_FINISH_REASON_VALUE_TOOL_CALLS,
+  GEN_AI_RESPONSE_FINISH_REASON_VALUE_UNKNOWN,
   Providers,
   SPAN_SPECIFICATIONS,
   SpanAttribute,
@@ -704,6 +705,7 @@ const FINISH_REASON_CONTENT_FILTER = toCamelCase(GEN_AI_RESPONSE_FINISH_REASON_V
 const FINISH_REASON_TOOL_CALLS = toCamelCase(GEN_AI_RESPONSE_FINISH_REASON_VALUE_TOOL_CALLS) // prettier-ignore
 const FINISH_REASON_ERROR = toCamelCase(GEN_AI_RESPONSE_FINISH_REASON_VALUE_ERROR) // prettier-ignore
 const FINISH_REASON_OTHER = toCamelCase(GEN_AI_RESPONSE_FINISH_REASON_VALUE_OTHER) // prettier-ignore
+const FINISH_REASON_UNKNOWN = toCamelCase(GEN_AI_RESPONSE_FINISH_REASON_VALUE_UNKNOWN) // prettier-ignore
 
 function extractFinishReason(
   attributes: Record<string, SpanAttribute>,
@@ -732,7 +734,9 @@ function extractFinishReason(
       return Result.ok('error')
     case FINISH_REASON_OTHER:
       return Result.ok('other')
-    default:
+    case FINISH_REASON_UNKNOWN:
       return Result.ok('unknown')
   }
+
+  return Result.ok('unknown')
 }
--- a/packages/sdks/python/tests/sdk/client_test.py
+++ b/packages/sdks/python/tests/sdk/client_test.py
@@ -1,0 +1,110 @@
+from typing import List, cast
+
+import httpx
+
+from latitude_sdk import GetOrCreatePromptResult, GetPromptResult, __version_semver__
+from tests.utils import TestCase, fixtures
+
+
+class TestClient(TestCase):
+    async def test_success_get(self):
+        path = "prompt-path"
+        endpoint = f"/projects/{self.project_id}/versions/{self.version_uuid}/documents/{path}"
+        endpoint_mock = self.gateway_mock.get(endpoint).mock(
+            return_value=httpx.Response(200, json=fixtures.PROMPT_RESPONSE)
+        )
+
+        result = await self.sdk.prompts.get(path)
+        request, _ = endpoint_mock.calls.last
+
+        self.assert_requested(
+            request,
+            method="GET",
+            endpoint=endpoint,
+            headers={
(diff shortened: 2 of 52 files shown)


packages/core/src/services/tracing/spans/completion.ts
```diff
@@ -732,7 +734,9 @@ function extractFinishReason(
       return Result.ok('error')
     case FINISH_REASON_OTHER:
       return Result.ok('other')
-    default:
+    case FINISH_REASON_UNKNOWN:
       return Result.ok('unknown')
   }
+
+  return Result.ok('unknown')
```
geclos, line 741: toc? your are changing nothing here

packages/telemetry/typescript/tests/instrumentations/latitude.test.ts
```diff
@@ -37,8 +37,7 @@ describe('latitude', () => {
     gatewayMock.close()
   })
 
-  // TODO(compiler): unskip and fix this test — Alex should know how to do this
```
geclos, line 40: thanks

geclos (dismissed):

packages/core/src/services/tracing/spans/completion.ts
neoxelox: yes sorry, it makes no sense, will remove this change

packages/sdks/typescript/src/index.ts
```diff
@@ -549,18 +550,27 @@ class Latitude {
...
+
+    return { messages }
   }
 
   protected async renderChain<M extends AdapterMessageType = PromptlMessage>({
     prompt,
     parameters,
     adapter: _adapter,
     onStep,
-    tools = {},
```
geclos, line 563: you are making the prop mandatory? if not you can just leave the default value, it's better than [handling nulls](https://www.notion.so/latitudedata/Aphorisms-0ed50a6dccdb4aa2a7dec86812b30cc8?source=copy_link#e2e97306e73e495b89ca2447074d022e)

packages/sdks/typescript/src/index.ts
```diff
@@ -621,34 +625,58 @@ class Latitude {
...
+                ],
+                toolId: t.toolCallId,
+                toolName: t.toolName,
+                isError: isError,
+              },
+            ],
+            config: {},
+          }).messages[0]!
+        }),
+    )
```
geclos, line 677: could you explain the chain in this file? it looks as if you've changed the tool result format

packages/sdks/python/tests/sdk/client_test.py
```diff
@@ -0,0 +1,110 @@
...
+                    "authorization": f"Bearer {self.api_key}",
+                    "x-latitude-sdk-version": __version_semver__,
+                    "content-type": "application/json",
+                    "accept": "application/json",
+                },
+                body={"path": path},
+            )
+            for request in requests
+        ]
+        self.assertEqual(endpoint_mock.call_count, self.internal_options.retries)
```
geclos, line 110: fantastic

packages/sdks/python/tests/sdk/memory_test.py
```diff
@@ -0,0 +1,376 @@
...
+        finally:
+            await metrics.stop_tracking()
+            metrics.print_summary()
+
+        # Assert no significant leaks with concurrent access
+        memory_leak_mb = metrics.get_memory_leak_mb()
+        connection_leak = metrics.get_connection_leak()
+
+        assert memory_leak_mb <= 0, f"Memory leak with concurrent requests: {memory_leak_mb:.2f} MB"
+        assert connection_leak <= 0, f"Connection leak with concurrent requests: {connection_leak} connections"
```
geclos, line 376: fantastic

packages/web-ui/src/ds/atoms/Alert/index.tsx
```diff
@@ -21,10 +21,10 @@ type Props = {
 }
 
 const IconColor: Record<string, TextColor> = {
-  destructive: 'destructive',
-  success: 'success',
+  destructive: 'destructiveMutedForeground',
+  success: 'successMutedForeground',
   warning: 'warningMutedForeground',
-  default: 'foreground',
+  default: 'foregroundMuted',
```
geclos, line 27: why these changes? did you talk it with someone?

packages/sdks/typescript/src/index.ts
neoxelox: tools is not mandatory no? it is an optional parameter. The only thing is that i handle the parameter in another function now.

```
export type RenderChainOptions<
  M extends AdapterMessageType = PromptlMessage,
  Tool extends ToolSpec = ToolSpec,
> = {
  prompt: Prompt
  parameters: Record<string, unknown>
  adapter?: ProviderAdapter<M>
  onStep: (args: {
    config: Config
    messages: M[]
  }) => Promise<string | Omit<M, 'role'>>
  tools?: RenderToolCalledFn<Tool> // <--
  logResponses?: boolean
}
```

packages/sdks/typescript/src/index.ts
neoxelox: I used the corresponding adapter `adapter.fromPromptl` to include the tool result message correctly when using other adapters (openai/anthropic/default), instead of hardcoding a message which could be incompatible with the other providers. This is how it was before, at least.

packages/web-ui/src/ds/atoms/Alert/index.tsx
neoxelox: The color of the icon was actually off. The foreground color of the alerts was different (muted) than the color of the icon, and was actually noticeable. Now, both are the same color.

packages/sdks/typescript/src/index.ts
geclos: yes what i mean is that it's always better to leave the default value to avoid nulls.

packages/sdks/typescript/src/index.ts
geclos: hmm looks like we need some tests here, leave a todo or something

packages/web-ui/src/ds/atoms/Alert/index.tsx
geclos: 👌🏼

packages/sdks/typescript/src/index.ts
neoxelox: There are tests: packages/sdks/typescript/src/tests/render.test.ts "automatically runs tool handlers when the response contains a tool call request"

PR #1493, opened 2025-07-18
Fix span ingestion memory issues

neoxelox:


packages/web-ui/src/ds/atoms/Alert/index.tsx
```diff
@@ -52,7 +52,10 @@ export function Alert({
           'justify-center': centered,
         })}
       >
-        <div className='flex flex-col gap-2 whitespace-pre-wrap'>
+        <div
+          className='flex flex-col gap-2 whitespace-pre-wrap'
+          style={{ wordBreak: 'break-word' }}
```
neoxelox, line 57: This class is only in tailwind v4 and we have v3 💀

geclos (approved):
looks good 👌🏼

PR #1501, opened 2025-07-21
refactor: Fix side effects in Transactions

geclos:
Fixes Transaction implementation by ensuring a single transaction is open per instance. Also it implements a callback queue that is consumed once the single transaction is closed, ensuring side effects are executed only when the full transaction has been committed.


packages/core/src/lib/Transaction.ts
```diff
@@ -24,30 +25,52 @@ const DB_ERROR_CODES = {
...
 
 export default class Transaction {
-  public static async call<ResultType>(
-    callback: (trx: Database) => PromisedResult<ResultType>,
-    db = database,
-  ): PromisedResult<ResultType> {
-    return new Transaction().call(callback, db)
-  }
+  private db?: PgTransaction<any, any, any> | Database
+  private callbacks: Array<(result: unknown, db: Database) => void> = []
```
csansoon, line 29: that is actually pretty cool

packages/core/src/services/tracing/spans/process.ts
```diff
@@ -76,154 +76,154 @@ export async function processSpan(
...
-
-  const convertingst = extractSpanType(attributes)
-  if (convertingst.error) return Result.error(convertingst.error)
-  const type = convertingst.value
-
-  const specification = SPAN_SPECIFICATIONS[type]
-  if (!specification) {
-    return Result.error(new UnprocessableEntityError('Invalid span type'))
-  }
+  return await transaction.call(
```
neoxelox, line 82: Why change this to wrap everything in a transaction? The idea is to compute outside of the transaction and then save it under transaction to avoid long running blocking transactions

packages/core/src/services/tracing/segments/process.ts
```diff
@@ -54,85 +54,85 @@ export async function processSegment(
...
+  transaction = new Transaction(),
   disk: DiskWrapper = diskFactory('private'),
 ) {
-  const validating = validateSegmentChain(args.segment, args.chain)
-  if (validating.error) return Result.error(validating.error)
-
-  const getting = await getState(args, db, disk)
-  if (getting.error) return Result.error(getting.error)
-  const state = getting.value
+  return await transaction.call(async (tx) => {
```
neoxelox, line 60: same as with spans, why change this to wrap the whole compute with a transaction?

packages/core/src/services/evaluationsV2/clone.ts
```diff
@@ -28,72 +28,74 @@ export async function cloneEvaluationV2<
...
+  transaction = new Transaction(),
 ) {
-  const documentsRepository = new DocumentVersionsRepository(workspace.id, db)
-  const document = await documentsRepository
-    .getDocumentAtCommit({
-      commitUuid: commit.uuid,
-      documentUuid: evaluation.documentUuid,
-    })
-    .then((r) => r.unwrap())
+  return transaction.call(async (db) => {
```
neoxelox, line 33: why wrap everything under a transaction? before we were just calling createEvaluationV2, shouldnt createEvaluationV2 create a tx itself if needed?

packages/core/src/services/evaluationsV2/createDemoEvaluation.ts
```diff
@@ -22,56 +22,59 @@ export async function createDemoEvaluation(
...
     workspace: Workspace
   },
-  db = database,
+  transaction = new Transaction(),
 ) {
-  // Note: failing silently to avoid not letting the user create the document
-  const result = await findDefaultEvaluationProvider(workspace!, db)
-  if (result.error || !result.value) return Result.nil()
-  const provider = result.value
+  return transaction.call(async (trx) => {
```
neoxelox, line 27: why wrap everything under a transaction??

packages/core/src/services/evaluationsV2/llm/index.ts
```diff
@@ -149,7 +154,8 @@ async function run<M extends LlmEvaluationMetric>(
             details: error.details,
           },
         },
-        db,
+        // TODO: We are stepping out of the db instance. This service should accept an instance of Transaction instead.
```
neoxelox, line 157: I dont understand this, what about the other methods? what if the calling method starts a transaction, shouldnt it pass the tx down?

packages/core/src/services/tracing/spans/process.ts
geclos: Previously you really didn't know if you were outside of a transaction, the db instance could have already been in an open transaction! Assuming because I'm outside of a Transaction.call I can do long running operations was a mistake. The change I'm doing that just acknowledges this (in this instance we acknowledge the prior operations to the transaction are not that long and can be safely wrapped in the transaction). If you have a service that performs a true long running operation it should not receive any instance of db nor transaction.

packages/core/src/services/evaluationsV2/run.ts
```diff
@@ -55,145 +54,147 @@ export async function runEvaluationV2<
...
-    evaluatedLogId: providerLog.id,
-    evaluationUuid: evaluation.uuid,
-  })
-
-  if (result.ok) {
-    return Result.error(
-      new UnprocessableEntityError(
-        'Cannot evaluate a log that is already evaluated for this evaluation',
-      ),
+  return await transaction.call(async (tx) => {
```
neoxelox, line 59: just like the other comments, why? i wanna understand

packages/core/src/services/tracing/segments/process.ts
geclos: ditto

packages/core/src/services/evaluationsV2/clone.ts
geclos: ditto

packages/core/src/services/evaluationsV2/createDemoEvaluation.ts
geclos: ditto

packages/core/src/services/evaluationsV2/llm/index.ts
geclos: > I dont understand this, what about the other methods? what if the calling method starts a transaction, shouldnt it pass the tx down?

yes, by passing an instance of transaction

learningbizz (approved):

PR #1505, opened 2025-07-21
latte search integrations

csansoon:
Some tools for latte to find what apps are available to connect as integrations, and which tools/triggers each of them has

### DIFF
--- a/packages/core/src/services/copilot/latte/tools/settings/searchIntegrationApps.ts
+++ b/packages/core/src/services/copilot/latte/tools/settings/searchIntegrationApps.ts
@@ -1,0 +1,26 @@
+import { z } from 'zod'
+import { defineLatteTool } from '../types'
+import { listApps } from '../../../../integrations/pipedream/apps'
+import { Result } from '../../../../../lib/Result'
+
+const searchIntegrationResources = defineLatteTool(
+  async ({ query }) => {
+    const result = await listApps({
+      query,
+    })
+    if (!Result.isOk(result)) return result
+
+    const apps = result.unwrap().apps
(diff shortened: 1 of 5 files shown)


packages/core/src/services/copilot/latte/tools/settings/searchIntegrationApps.ts
```diff
@@ -0,0 +1,26 @@
+import { z } from 'zod'
+import { defineLatteTool } from '../types'
+import { listApps } from '../../../../integrations/pipedream/apps'
+import { Result } from '../../../../../lib/Result'
+
+const searchIntegrationResources = defineLatteTool(
```
learningbizz, line 6: Could we have a few tests for these services?

geclos: carlos thought he was free because @andresgutgon was on vacation, but we now have manu 👿

geclos (approved):

packages/core/src/services/copilot/latte/tools/settings/searchIntegrationApps.ts
csansoon: this is just a latte tool, which is just a wrapper of a service. Most of these are not services of their own and do not contain any logic other than calling an actual service. It's similar to our current NextJS actions, that do not perform any heavy logic on their own

learningbizz: It starts like this, then over time it gets more complex and we get bugs 🤷

PR #1508, opened 2025-07-22
Unconfigured Pipedream Integrations

csansoon:
This is a strange one.

This PR allows creating and handling Integrations of type Pipedream without having configured an actual connection.

### Why?
Quick answer: Latte
Latte should be able to build full Latitude projects, including configuring integrations, without requiring any additional user input. 
To add tools to a prompt, an integration must exist, as otherwise, the prompt will display an error. 
Latte should be able to create this integration with any app it wants, without actually configuring an account. 
Once everything is created, the user can then configure it afterwards.

### Changes
Some changes requried to make it work:
We were already not using Pipedream's MCP implementation to run tools due to missing feature parity with the SDK. Instead, we used the SDK directly.
However, we were using Pipedream's MCP to list tools. This was necessary because the SDK returns tool properties in a custom type, while LLM tools require them in JSON Schema.
The issue with Pipedream's MCP is that it only works with active connections. In contrast, the SDK provides methods to list tools from any app, regardless of connection status. To address this, we now use the SDK methods to list tools and translate the tool properties to JSON Schema.

Additionally, checks and errors have been added everytime a connection is required, such as running tools or configuring triggers

image


geclos (approved):
I know you are in ship mode but please add tests – it compounds over time

PR #1511, opened 2025-07-23
Feature/Latte-CRUD-inhouse-triggers

learningbizz:
PR for Latte to CRUD schedule and email triggers, more specification in [this page](https://www.notion.so/latitudedata/Latte-creates-inhouse-triggers-231a92110b9780748e09e23f177ac1b7?source=copy_link)


.gitignore
```diff
@@ -79,3 +79,9 @@ ONLY_FOR_COPILOT/
 
 .claude
 __testing__
+
+
+# New spans
+apps/web/ingest
+apps/workers/workspaces/*/traces
```
geclos, line 90: @neoxelox ?

learningbizz: Lo hemos estado hablando but we don't really understand why this is ocurring. Its also happening to Sanson

packages/core/src/services/integrations/list.ts
```diff
@@ -0,0 +1,29 @@
...
+        updatedAt: workspace.updatedAt,
+        authorId: workspace.creatorId!,
+        lastUsedAt: workspace.updatedAt,
+        configuration: null,
+        deletedAt: null,
+        mcpServerId: null,
+    }
+
+    return Result.ok([latitudeIntegration, ...integrations])
+}
```
csansoon: If we're treating "email" and "schedule" as triggers from the Latitude integration, we should return this fake integration with `hasTriggers: true`

packages/constants/src/trigger.ts
```diff
@@ -0,0 +1,5 @@
+
+export type Trigger = {
+    name: string
+    integrationName: string
+}
```
csansoon: What is this type, and what is it used for? `Trigger` is way too generic if this is being used for a specific component only. We're usually using `DocumentTrigger`

packages/core/src/services/copilot/latte/tools/triggers/executeTriggerActions.test.ts
```diff
@@ -0,0 +1,363 @@
...
+        workspace,
+        commit,
+        documents,
+        action,
+      })
+
+      // Assert
+      expect(result.ok).toBe(false)
+      expect(result.error).toStrictEqual(expectedError)
+    })
```
csansoon, line 117: Remove copilot's comments xd

packages/core/src/services/copilot/latte/tools/triggers/executeTriggerActions.ts
```diff
@@ -0,0 +1,165 @@
...
+    documents,
+    action,
+  }: {
+    workspace: Workspace
+    commit: Commit
+    documents: DocumentVersion[]
+    action: LatteTriggerAction
+  },
+  transaction = new Transaction(),
+): PromisedResult<LatteTriggerChanges> {
```
csansoon, line 32: Why does this tool require a `commit`, if trigger actions can only happen in the Live version?

packages/core/src/services/copilot/latte/tools/triggers/triggerActions.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (!commitResult.ok) return commitResult
+    const commit = commitResult.unwrap()
+
+    if (!commit.mergedAt) {
+      return Result.error(
+        new BadRequestError(
+          `Cannot modify/add/delete triggers on a draft commit. Select a previous live commit or publish the draft.`,
+        ),
+      )
+    }
```
csansoon: Well, about the previous comment: I guess having the tool fail in a draft can teach Latte how to modify triggers correctly

csansoon: However, it should not just check if the commit is merged or not. It should only check if the commit is the Live commit, because that's the only version these changes will have an effect to. Previously merged commit will not be affected.

packages/core/src/services/integrations/pipedream/triggers.ts
```diff
@@ -76,21 +76,26 @@ export async function deployPipedreamTrigger({
...
     updatedConfig,
   }: {
     workspace: Workspace
-    originalConfig: IntegrationTriggerConfiguration
-    updatedConfig: IntegrationTriggerConfiguration
+    trigger: Extract<DocumentTrigger, { triggerType: DocumentTriggerType.Integration }>
+    updatedConfig: InsertIntegrationTriggerConfiguration
   },
   db = database,
 ): PromisedResult<IntegrationTriggerConfiguration, Error> {
```
csansoon: 👍

packages/core/src/services/copilot/latte/tools/triggers/executeTriggerActions.test.ts
learningbizz: They were mine 😂

packages/constants/src/trigger.ts
learningbizz: You're right, this will be deleted when doing the integrations triggers for Latte, as it won't be useful anymore

packages/core/src/services/copilot/latte/tools/triggers/triggerActions.ts
learningbizz: Will change that

packages/core/src/services/copilot/latte/tools/triggers/executeTriggerActions.test.ts
andresgutgon: loool

packages/constants/src/latte/index.ts
```diff
@@ -16,4 +17,8 @@ export enum LatteTool {
 
   searchIntegrationApps = 'search_integration_apps',
   searchIntegrationResources = 'search_integration_resources',
+  listIntegrationTriggers = 'list_integration_triggers',
+  listExistingTriggers = 'list_existing_triggers',
+
+  triggerActions = 'trigger_actions',
```
geclos, line 24: what's this? the other tools have self explanatory names but I'm not sure what this is doing

packages/core/src/services/copilot/latte/tools/documents/list.ts
```diff
@@ -20,17 +21,26 @@ const listPrompts = defineLatteTool(
...
-      documents.map((document) =>
-        promptPresenter({
-          document,
+
+    const documentTriggerScope = new DocumentTriggersRepository(workspace.id)
+
+    const promptObjects = await Promise.all(
+      documents.map(async (doc) => {
+        const existingTriggers = await documentTriggerScope.findByDocumentUuid(
+          doc.documentUuid,
```
geclos: n+1 query – it's better to fetch in memory all triggers from all documents (findTriggersAtCommit?) and then find them from memory in the iteration

packages/core/src/services/copilot/latte/tools/triggers/executeTriggerActions.test.ts
```diff
@@ -0,0 +1,363 @@
...
+      workspace,
+      commit,
+      documents,
+      action,
+    })
+    // Assert
+    expect(result.ok).toBe(false)
+    expect(result.error).toStrictEqual(expectedError)
+  })
+})
```
geclos, line 363: lovely

packages/core/src/services/copilot/latte/tools/triggers/executeTriggerActions.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    return Result.ok({
+      projectId: commit.projectId,
+      draftUuid: commit.uuid,
+      promptUuid: action.promptUuid,
+      triggerType: action.triggerType,
+    })
+  }
+  return Result.error(new Error(`Unsupported action operation type`))
+}
```
geclos, line 163: an obvious split here would be to have a service for each write op (create/update/delete)

packages/core/src/services/integrations/list.ts
```diff
@@ -0,0 +1,27 @@
+import { IntegrationDto, Workspace } from '@latitude-data/core/browser'
+import { IntegrationType } from '@latitude-data/constants'
+import { IntegrationsRepository } from '@latitude-data/core/repositories'
+import { Result } from '@latitude-data/core/lib/Result'
+
+export async function listIntegrations(workspace: Workspace) {
+  const integrationsScope = new IntegrationsRepository(workspace.id)
+  const integrations = await integrationsScope.findAll().then((r) => r.unwrap())
+  // Adding a fake integration for Latitude, to later be used to its their tools/triggers
```
geclos, line 9: I don't understand why we need this hack

packages/constants/src/latte/index.ts
learningbizz: This tool is to do CRUD actions for triggers (creating, updating, and deleting them)

packages/core/src/services/integrations/list.ts
learningbizz: To link the triggers of schedule and email to the Latitude integration

packages/core/src/services/copilot/latte/tools/triggers/executeTriggerActions.ts
learningbizz: Thought about it, but as its a small file decided to leave it in one.

packages/constants/src/latte/index.ts
geclos: why did we choose this barrel approach rather than a tool per action?

packages/core/src/services/copilot/latte/tools/triggers/executeTriggerActions.ts
geclos: there's actually quite some logic in this service though, consider starting with separation of concerns from the start in next ocasions, might seem not worth now but it leaves the field ready for the next developer that needs to expand this service in the future.

geclos (dismissed):

packages/core/src/services/copilot/latte/tools/documents/list.ts
learningbizz: Changed

csansoon (dismissed):

packages/core/src/services/copilot/latte/tools/triggers/executeTriggerActions.ts
learningbizz: Responded below

PR #1516, opened 2025-07-24
feat: latte redesign #1

geclos:
First iteration of the redesign, still missing many parts


packages/core/drizzle/meta/0194_snapshot.json
csansoon, line 1: Why is there a new snapshot without a migration script?

csansoon (dismissed):

packages/core/drizzle/meta/0194_snapshot.json
geclos: woopsie good catchio, just my linter formatting stuff

PR #1528, opened 2025-07-31
Integration triggers break when includes reloadProps

### LINKED ISSUE #1527: Integration triggers/tools break when includes reloadProps
When you try to create a trigger that has a reloadProps = true, the app breaks.

This will also be affecting tools with reloadProps, so its a 2x1.

https://github.com/user-attachments/assets/978483b9-9550-4713-8ac3-b3e96b029b35

learningbizz:
Fixes #1527
Closes #1528


packages/core/src/services/integrations/pipedream/components/reloadComponentProps.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const configuredPropsResult = await fillConfiguredProps({
+    pipedream,
+    integration,
+    componentId,
+    configuredProps: configuredClientProps ?? {},
+  })
+  if (!Result.isOk(configuredPropsResult)) {
+    return Result.error(configuredPropsResult.error)
+  }
```
learningbizz, line 53: This is what was missing, we weren't adding the app configurableProp with the authId

geclos (approved):
tests and 🚀

PR #1530, opened 2025-08-04
Fix UI bugs adding and deleting integration triggers

learningbizz:
Fix two bugs of the integration triggers:

Number 1 - When adding a new integration trigger when not having any, it would open the modal in "update trigger" mode and not in "add trigger" mode. This was the reason why we couldnt create a new trigger in last fridays demo

https://github.com/user-attachments/assets/325ebd17-478b-47bd-afe9-dd9921ecf97b

Number 2 - When clicking the trash icon, it wouldnt open the delete trigger modal

https://github.com/user-attachments/assets/c8e81cb2-ba07-40f3-b974-32407127533a

Now works like this:

https://github.com/user-attachments/assets/af5aa94a-4266-40ab-8bfe-9f35edb9cd04

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/BlankSlate.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/BlankSlate.tsx
@@ -25,7 +25,7 @@ export function TriggersBlankSlate({
           fullWidth
           fancy
           variant='default'
-          onClick={openTriggerModal}
+          onClick={() => openTriggerModal()}
           disabled={!isHead}
         >
           Add Trigger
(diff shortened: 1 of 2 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/BlankSlate.tsx
```diff
@@ -25,7 +25,7 @@ export function TriggersBlankSlate({
           fullWidth
           fancy
           variant='default'
-          onClick={openTriggerModal}
+          onClick={() => openTriggerModal()}
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/BlankSlate.tsx:28]: not really needed no?

geclos (approved):

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/BlankSlate.tsx
Comment (learningbizz) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/BlankSlate.tsx:28]: This is the origin of the first bug, as by not adding it, React sends the click event as an argument to the openTriggerModal event handler, making it always have a trigger, and therefore, open "update trigger" mode always instead of "create trigger" mode

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/BlankSlate.tsx:28]: remove this please

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/BlankSlate.tsx:28]: oh I see, then wrap it in a usecallback please

PR #1532, opened 2025-08-04
feat(triggers): new triggers UI

geclos:
Implements UI to create integration triggers (no chat/schedule/email for now)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/Preview.tsx
```diff
@@ -65,6 +65,11 @@ export function SlidingCarousel<T>({
     return () => clearInterval(id)
   }, [items.length, interval])
 
+  const item = items[currentIndex]
+  const nextItem = items[nextIndex]
+
+  if (!item || !nextItem) return null
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/IntegrationTriggers/Preview.tsx:71]: bug

apps/web/src/app/(private)/settings/_components/Integrations/New/_components/Configuration/Pipedream/AppSelector.tsx
```diff
@@ -14,10 +13,6 @@ export function AppSelector({
   onChange: (value: App | undefined) => void
   isLoading?: boolean
 }) {
-  const { execute: executeListPipedreamApps } = useServerAction(
-    listPipedreamIntegrationsAction,
```
geclos, line 18: do not use post actions to fetch data, leverage the browser's cache

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/triggers/new/_components/contexts/column-1-context.tsx
```diff
@@ -0,0 +1,53 @@
...
+  )
+}
+
+export function useColumn1Context() {
+  const context = useContext(Column1Context)
+  if (context === undefined) {
+    throw new Error('useColumn1Context must be used within a Column1Provider')
+  }
+  return context
+}
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/triggers/new/_components/contexts/column-1-context.tsx:53]: i feel we don't leverage contexts and/or zustand stores close to enough and we end up with components with a gazillion props

packages/core/src/schema/models/integrations.ts
```diff
@@ -37,7 +33,8 @@ export const integrations = latitudeSchema.table(
       .notNull(),
     hasTools: boolean('has_tools').notNull().default(true),
     hasTriggers: boolean('has_triggers').notNull().default(false),
-    configuration: jsonb('configuration').$type<IntegrationProviderConfig>(),
+    configuration:
+      jsonb('configuration').$type<IntegrationConfiguration['configuration']>(),
```
geclos, line 37: type was wrong

apps/web/src/actions/integrations/pipedream/list.ts
```diff
@@ -1,19 +0,0 @@
...
-    z.object({
-      query: z.string().optional(),
-      cursor: z.string().optional(),
-    }),
-  )
-  .handler(async ({ input }) => {
-    return listApps({ query: input.query, cursor: input.cursor }).then((r) =>
-      r.unwrap(),
-    )
-  })
```
geclos, line 19: nein

geclos:
> For pure non-coder Latte users, they will ask Latte to create the project and won't know what prompts it created. Therefore, when setting a trigger, they won't know which one to add. We can avoid this by adding a default prompt which is the main agent/bot (we could talk with Carlos to see if Latte could always create the main bot/agent with the same name like index, so we could add it always as a default, or another random one if it doesnt exist). Thoughts?

I think it's very hacky to filter by prompt name, you never know what the user might create with what names. We can give instructions to latte to create entrypoint agent prompts with clear names, that will help the user.

PR #1541, opened 2025-08-05
Improve trigger modal UX by merging selecting existing integrations and connecting new integrations.

andresgutgon:
# What?

Now, we display three columns on the triggers modal, with the first column showing Latitude Integrations (not yet), connected integrations, and all the rest of the Pipedream integrations.
In this commit, we merge connected and all the other integrations, also removing the triggers that don't have events

## TODO
- [x] Filter triggers without events
- [x] Remove connected integrations and do a selector of connected integrations when configuring it in a new column.  
- [x] 🐛 Parameters selector does not show selected parameter
- [x] 🐛 When changing triggers the trigger form is not reset
- [x] Fix multiple select selector styles. Text to small. Also inner padding in popover
- [x] QA trigger is created
- [x] QA trigger modal is closed
- [x] 🐛 In the preview page when removing a trigger the list is not updated
- [x] QA change propmt and see how the other blocks are rest or updated. I think there are bugs 🐛 
- [x] 🐛 Fix popover in full mode entrance animation. Should be the same. Also pick better names for `height` prop
- [x] Do NOT show `Cancel` button on connect account when no accounts

## Next
- [ ] Create email trigger 
- [ ] Create schedule trigger
- [ ] Create chat trigger
- [ ] Parse markdown links in trigger description (maybe)

## Bug: Select options same key
image

☝️ This is not an error. The mentions selector has duplicated options


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/triggers/new/_components/IntegrationsList/IntegrationsListProvider.tsx
```diff
@@ -0,0 +1,93 @@
...
+    )
+    return {
+      isLoading,
+      isLoadingPipedreamApps,
+      isLoadingConnectedIntegrations,
+      pipedreamApps,
+      connectedIntegrations,
+      searchQuery,
+      setSearchQuery,
+      allTriggersCount: pipedreamApps.length + connectedIntegrations.length,
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/triggers/new/_components/IntegrationsList/IntegrationsListProvider.tsx]: don't add it to the context if it's only gonna be used in one component please

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/triggers/new/_components/IntegrationsList/integrations.tsx
```diff
@@ -42,9 +47,7 @@ export function Integrations() {
                       name: app.name,
                       type: IntegrationType.Pipedream,
                       pipedream: {
-                        app: {
-                          name: app.name,
-                        },
+                        name_slug: app.name_slug as unknown as PipedreamSlug,
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/triggers/new/_components/IntegrationsList/integrations.tsx]: how is it you need to do this type casting now, shouldn't be needed

Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/triggers/new/_components/IntegrationsList/integrations.tsx]: Is a bit stupid. PipedreamSlug is a string, but TypeScript does not understand it

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/triggers/new/_components/IntegrationsList/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+export function IntegrationsList() {
+  const { data: connectedApps, isLoading: isLoadingConnectedIntegrations } =
+    useConnectedIntegrationsByPipedreamApp({
+      withTriggers: true,
+    })
+  const { data: pipedreamApps, isLoading: isLoadingPipedreamApps } =
+    usePipedreamApps({ query: '', withTriggers: true })
+
+  const isLoading = isLoadingPipedreamApps || isLoadingConnectedIntegrations
+  const optionGroups = useMemo<SearchableOption[]>(() => {
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/triggers/new/_components/IntegrationsList/index.tsx]: put this in its own local hook

PR #1542, opened 2025-08-06
feat(latte): do not share latte sessions

geclos:
I thought it was cool, someone else thought it was cool, and now we are all thinking it's stupid.


packages/core/src/repositories/latteThreadsRepository.ts
```diff
@@ -26,16 +25,13 @@ export class LatteThreadsRepository extends Repository<LatteThread> {
     threadUuid,
   }: {
     threadUuid: string
-  }): PromisedResult<LatteThread, LatitudeError> {
+  }): Promise<LatteThread | undefined> {
```
geclos, line 28: in my personal xp it's better DX to have repository find methods return empty arrays/null values rather than result objects since their api is already very established

geclos (commented):
🐮 👦🏼 this one since it's minor changes in a feature-flagged UI

packages/core/src/repositories/latteThreadsRepository.ts
andresgutgon: Agree. Although is nice sometimes using the collection in api endpoint to do `.unwrap` and get the error for free

PR #1547, opened 2025-08-07
Feat/obsevability redux

geclos:


packages/constants/src/tracing/span.ts
```diff
@@ -19,6 +33,8 @@ export enum SpanType {
   Http = 'http', // Note: raw HTTP requests and responses
   Segment = 'segment', // (Partial) Wrappers so spans belong to the same trace
```
neoxelox: remove this segment type

packages/constants/src/tracing/span.ts
```diff
@@ -1,6 +1,20 @@
 import { FinishReason } from 'ai'
 import { Message } from 'promptl-ai'
 
+export enum SpanSource {
+  API = 'api',
+  AgentAsTool = 'agent_as_tool', // TODO(tracing): deprecated, use SpanType.Segment with DocumentType.Agent instead
```
neoxelox: remove this comment

packages/core/src/schema/models/spans.ts
```diff
@@ -18,7 +18,7 @@ export const spans = latitudeSchema.table(
   {
     id: varchar('id', { length: 16 }).notNull(),
     traceId: varchar('trace_id', { length: 32 }).notNull(),
-    segmentId: uuid('segment_id'),
+    documentLogUuid: uuid('document_log_uuid'),
```
neoxelox, line 21: i think we should call this conversation_id, more fitting

packages/core/src/services/__deprecated/commits/runDocumentAtCommit.ts
```diff
@@ -112,7 +112,6 @@ export async function runDocumentAtCommitLegacy({
     externalId: customIdentifier,
     template: result.value,
     parameters: parameters,
-    _internal: { source },
```
neoxelox, line 115: if you are deleting the source why keep "SpanSource" and not just move it back to "LogSource"

apps/gateway/src/presenters/runPresenter.ts
```diff
@@ -13,7 +12,6 @@ type DocumentResponse = ChainStepObjectResponse | ChainStepTextResponse
 
 export function v2RunPresenter(
   response: DocumentResponse,
-  trace: TraceContext,
```
geclos, line 16: Since we've removed segments tracing in old prompt engine does not work so I'm entirely removing it.

PR #1552, opened 2025-08-08
feat(providerLog): store llm output in provider logs

geclos:
Previously, we denormalized the response into separate fields of the provider log. We now simply store the raw output of messages from the llm. This fixes an issue where conversations that ended in a tool result coult not continue in chat.


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/hooks/useRunPlaygroundPrompt.ts
```diff
@@ -62,6 +62,8 @@ export function useRunPlaygroundPrompt({
       toolCalls?: ToolCall[]
       trace?: TraceContext
     }) => {
+      console.log(messages)
```
Comment (neoxelox) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Playground/hooks/useRunPlaygroundPrompt.ts]: console loggg

learningbizz:
So what will be the meta now? Always parsing those raw messages?

PR #1553, opened 2025-08-08
Fix bug json parsing

learningbizz:
Creating this as a PR to confirm my understanding of this situation:

The getResult() method returns a parsed JSON object if its a JSON string, or already an object, or double encoded JSON string, passing the isString to false, and a string if its just a simple string.

If this is the case, then the result from this getResult should be the one used to calculate the length of the JSON, not the previous object passed from the backend. Also, there is no need to stringify again if its done above

Do u also agree?

### DIFF
--- a/apps/web/src/components/ChatWrapper/Message/ToolResult.tsx
+++ b/apps/web/src/components/ChatWrapper/Message/ToolResult.tsx
@@ -67,13 +67,13 @@ export function ToolResultContent({
       )
     }
 
-    const strResult = JSON.stringify(toolResponse.result, null, 2)
+    const strResult = JSON.stringify(result, null, 2)
     return (
       <CodeBlock
         language={strResult?.length > MAX_LENGTH_JSON_PREVIEW ? '' : 'json'}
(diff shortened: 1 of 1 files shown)


apps/web/src/components/ChatWrapper/Message/ToolResult.tsx
```diff
@@ -67,13 +67,13 @@ export function ToolResultContent({
...
 
-    const strResult = JSON.stringify(toolResponse.result, null, 2)
+    const strResult = JSON.stringify(result, null, 2)
     return (
       <CodeBlock
         language={strResult?.length > MAX_LENGTH_JSON_PREVIEW ? '' : 'json'}
         bgColor={toolResponse.isError ? 'bg-destructive-muted' : undefined}
       >
-        {JSON.stringify(result, null, 2)}
+        {strResult}
```
learningbizz, line 76: Still parses well the tool response

Image

andresgutgon:
Ping

geclos:
@learningbizz close this?

PR #1558, opened 2025-08-11
Feature/latte create integration triggers

learningbizz:
Tools for Latte to create integration triggers

## TODOS
- [x] Modify list_integration_triggers to list triggers for a specific integration
- [x] Create tools for Latte to create valid trigger configs and create document trigger from them
- [x] Fixed a bug createDocumentTrigger when reloadProps appears
- [x]  refactored some services to be reused and separated for single responsibility / better org


andresgutgon (approved):

packages/core/src/services/copilot/latte/tools/triggers/triggerActions.ts
```diff
@@ -22,7 +22,10 @@ const triggerActions = defineLatteTool(
       .getHeadCommit(projectId)
       .then((r) => r.unwrap())
 
-    if (headCommit == undefined || versionUuid === headCommit.uuid) {
+    if (
+      headCommit == undefined ||
+      (versionUuid !== headCommit.uuid && versionUuid !== 'live')
```
geclos: use HEAD_COMMIT constant rather than 'live'

geclos (dismissed):

PR #1567, opened 2025-08-14
Feature/1559 create agent from landing

neoxelox:
Closes: #1559


packages/constants/src/helpers.ts
```diff
@@ -66,13 +66,26 @@ export function simplifyDocument(
...
   return false
 }
+
+export function isLatitudeUrl(url: unknown): url is string | URL {
+  if (!isSafeUrl(url)) return false
+
+  if (url.toString().startsWith('/')) return true
+  if (url.toString().startsWith('http://localhost')) return true
+  if (url.toString().startsWith('https://latitude.so')) return true
+  if (url.toString().startsWith('https://app.latitude.so')) return true
```
andresgutgon, line 88: Ok for now but domain is configurable. Remember people can deploy this code

apps/web/src/app/global-error.tsx
```diff
@@ -19,19 +19,27 @@ export default function GlobalError({
...
         <ErrorComponent
           message={error.message}
           type='red'
           submit={
             <Link href={ROUTES.root}>
-              <Button>Go back to homepage</Button>
+              <Button
+                variant='link'
+                iconProps={{ name: 'arrowRight', placement: 'right' }}
+                className='p-0'
```
andresgutgon, line 39: size='nope'

packages/core/src/lib/RateLimiter.ts
```diff
@@ -0,0 +1,53 @@
...
+      if (error instanceof RateLimiterRes) {
+        throw new RateLimitError('Too many requests')
+      }
+
+      if (!(error instanceof ReplyError)) {
+        throw error
+      }
+    }
+  }
+}
```
andresgutgon, line 56: Didn't we have a rate limiter implemented somewhere?

packages/core/src/services/actions/execute.ts
```diff
@@ -0,0 +1,84 @@
...
+    user: User
+    workspace: Workspace
+  },
+  tx = new Transaction(),
+  limiter = new RateLimiter({
+    limit: 10,
+    period: 60,
+  }),
+) {
+  await limiter.consume(user.id)
```
andresgutgon: Is this throwing? Should not return a result?

packages/core/src/services/actions/cloneAgent.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+  const document = gettingdo.unwrap()
+  if (!document) {
+    return Result.error(
+      new UnprocessableEntityError('Sample Agent document not found'),
+    )
+  }
+
+  return Result.ok({ workspace, project, commit, document })
+}
```
andresgutgon, line 118: Some tests here no?

packages/core/src/lib/RateLimiter.ts
neoxelox: for actions, this is not an action but a service executed on SSR

packages/core/src/services/actions/execute.ts
neoxelox: true

packages/core/src/services/actions/cloneAgent.ts
neoxelox: ok

PR #1568, opened 2025-08-14
Fix document editor cache issue

andresgutgon:
# What?
The editors get the value from the `useState` inside `useDocumentValueContext`.
This state is set from the props coming from the backend. **Somehow** editing a document, going to another page, and coming back is showing stale text in the text editor. Now we get that info fresh from the `swr` store for the old and new editors

* Somehow
We think is React or Next caching the page and the props


andresgutgon:
DO_NOT_MERGE 🙏 

This is a fix but we feel the issue is related with next.js cache (any of the 1000 caches)

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/DocumentEditor.tsx
```diff
@@ -65,34 +67,14 @@ const HISTORY_ELM_HEIGHT = 44
  * @param props.initialDiff - Initial diff data for the document
  */
 function DocumentEditorContent({
-  document: _document,
-  documents: _documents,
   freeRunsCount,
   initialDiff,
 }: DocumentEditorProps) {
-  const { updateDocumentContent } = useDocumentValue()
+  const { updateDocumentContent, document } = useDocumentValue()
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/DocumentEditor.tsx:73]: better not to export document from this context, the document source of truth is the document store exporting it from other places adds duplicity which leads to confusion

PR #1580, opened 2025-08-21
quickfix(sdk): run prompt response type union was not asserted

geclos:
image

### DIFF
--- a/packages/constants/src/ai.ts
+++ b/packages/constants/src/ai.ts
@@ -61,9 +61,13 @@ export type ProviderData = TextStreamPart<any>
 
 export type ChainEventDto = ProviderData | LatitudeEventData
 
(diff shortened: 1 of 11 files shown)


packages/constants/src/ai.ts
```diff
@@ -84,10 +88,11 @@ export type ChainStepTextResponse = BaseResponse & {
   toolCalls: ToolCall[]
 }
 
-export type ChainStepObjectResponse = BaseResponse & {
-  streamType: 'object'
-  object: any
-}
+export type ChainStepObjectResponse<S extends Record<string, unknown> = any> =
```
geclos, line 91: you can now assert the object the ai is supposed to return

andresgutgon (approved):

PR #1593, opened 2025-08-26
Pending Triggers

csansoon:
Adds a `triggerStatus` to triggers.

There are 3 different statuses:
 - **Pending**: The trigger has been created by Latte from a new Integration that still needs to be configured.
 - **Deployed**: The trigger is correctly deployed and listening to events
 - **Deprecated**: The trigger has been undeployed due to a new version being merged

Merging a commit will fail if any there is any Pending trigger.

UI has also been added to see pending triggers and integrations from the Preview

image

---

## WIP
- [x] Open up Pipedream modal when clicking on "Set Up" integration

### DIFF
--- a/apps/web/src/stores/integrations.ts
+++ b/apps/web/src/stores/integrations.ts
@@ -15,6 +15,33 @@ import { useMemo } from 'react'
 
 const EMPTY_ARRAY: IntegrationDto[] = []
 
+function filterIntegrations({
+  integrations,
+  includeLatitudeTools,
+  withTools,
+  withTriggers,
+}: {
+  integrations: IntegrationDto[]
+  includeLatitudeTools?: boolean
+  withTools?: boolean
+  withTriggers?: boolean
+}) {
+  return integrations.filter((item) => {
+    if (!includeLatitudeTools && item.type === IntegrationType.Latitude) {
+      return false
+    }
+
+    if (withTools !== undefined && item.hasTools !== withTools) {
+      return false
+    }
+    if (withTriggers !== undefined && item.hasTriggers !== withTriggers) {
+      return false
+    }
+
+    return true
+  })
+}
+
 export default function useIntegrations({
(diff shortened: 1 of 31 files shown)


packages/core/src/services/documentTriggers/handleMerge.ts
```diff
@@ -27,39 +28,61 @@ export async function handleTriggerMerge(
...
+  if (
+    updatedTriggers.some(
+      (trigger) => trigger.triggerStatus === DocumentTriggerStatus.Pending,
+    )
+  ) {
+    return Result.error(
+      new BadRequestError(
+        'Cannot merge a commit that contains pending triggers',
+      ),
+    )
```
andresgutgon, line 83: TESTS

csansoon: done!

packages/core/drizzle/0201_pending_triggers.sql
andresgutgon, line 1: Too many SQL files. Something tinky winky

apps/web/src/stores/integrations.ts
```diff
@@ -28,37 +55,23 @@ export default function useIntegrations({
...
-            return false
-          }
-          if (withTriggers !== undefined && item.hasTriggers !== withTriggers) {
-            return false
-          }
-
-          return true
-        })
-        .sort((a, b) => a.id - b.id),
+      rows.map(deserialize).sort((a, b) => a.id - b.id),
```
andresgutgon: Do this sorting in SQL and maybe also the filtering. No saying to do now, but it's weird to do it in the frontend

packages/core/drizzle/0201_pending_triggers.sql
csansoon: whoops, missed it from the rebase! fixed

andresgutgon (approved):

PR #1595, opened 2025-08-26
Feat/latte-stops-chats

learningbizz:
This is a highly requested feature to be able to stop Latte chats, as sometimes it rambles or goes off into long executions we don't want. 

This change includes:
- Revamp of Latte chat (now with the new UX designed by Danil with start/stop button)
- New stop function in Latte Chat when waiting for Latte to respond
  - If the message is stopped, you can still continue talking in the same thread, no need to start a new chat


apps/web/src/actions/latte/stopChat.ts
```diff
@@ -0,0 +1,49 @@
...
+  'waiting',
+  'waiting-children',
+]
+const NON_REMOVABLE_STATES = ['active', 'stalled']
+
+export const stopChatLatteAction = authProcedure
+  .createServerAction()
+  .input(
+    z.object({
+      bullMQjobId: z.string().optional(),
```
andresgutgon: No need to change, but better not to let traces of the underlying tech. If tomorrow we swap the jobs queues system from bullMQ to something else, this won't make sense. 

Just `jobId` is better

apps/web/src/actions/latte/stopChat.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    const job = await documentsQueue.getJob(bullMQjobId)
+    if (!job) {
+      return
+    }
+
+    let state = await job.getState()
+
+    try {
+      if (REMOVABLE_STATES.includes(state)) await job.remove()
```
andresgutgon: This states probably are already defined in bullmq maybe?

andresgutgon: Also use `{}` in the `if` always please

apps/web/src/actions/latte/stopChat.ts
```diff
@@ -0,0 +1,49 @@
...
+      else if (NON_REMOVABLE_STATES.includes(state))
+        await cancelJob(bullMQjobId)
+    } catch (error) {
+      // to catch potential race condition between getting the state and removing
```
andresgutgon: Not sure this is necessary. Remove the `try/catch` until we prove this is the case. 

Simpler code

apps/web/src/components/LatteChat/LatteChatInput.tsx
```diff
@@ -147,42 +149,61 @@ export function LatteChatInput({
...
             }}
-            className='text-latte-input-foreground group-hover:text-latte-input-foreground/75'
             userSelect={false}
+            fancy={true}
+            roundy={true}
+          >
+            Send
+          </Button>
+        )}
+        {isLoading && (
```
andresgutgon: `isLoading` is a bit confusing. Should be `isRunning`

packages/core/src/jobs/job-definitions/copilot/chat.ts
```diff
@@ -44,42 +50,71 @@ export const runLatteJob = async (job: Job<RunLatteJobData>) => {
...
+  }, 500)
 
-    return copilotResult
-  }
-  const {
-    workspace: copilotWorkspace,
-    commit: copilotCommit,
-    document: copilotDocument,
-  } = copilotResult.unwrap()
+  try {
```
andresgutgon, line 58: What's doing this `try`?

packages/core/src/lib/streamManager/ChainStreamConsumer/consumeStream.ts
```diff
@@ -56,7 +58,7 @@ export async function consumeStream({
         )
       }
     }
-
+    if (chunk.type == 'text-delta') _accumulatedText.text += chunk.textDelta
```
andresgutgon, line 61: Why this?

packages/core/src/lib/streamManager/step/streamAIResponse.ts
```diff
@@ -68,11 +72,37 @@ export async function streamAIResponse({
...
+          provider,
+          conversation: {
+            messages,
+            config,
+          },
+          stepStartTime: startTime,
+          errorableUuid: documentLogUuid,
+          response: response,
+        }),
+      }).then((r) => r.unwrap())
```
andresgutgon: I would move to a method called `createProviderLogForAbortedRequest`

packages/core/src/services/bullmq/cancelJob.ts
```diff
@@ -0,0 +1,15 @@
...
+  await redisClient.set(`cancel:${BullMQjobId}`, '1')
+}
+
+export async function isJobCancelled(BullMQjobId: string): Promise<boolean> {
+  return (await redisClient.get(`cancel:${BullMQjobId}`)) === '1'
+}
+
+export async function clearJobCancellation(BullMQjobId: string) {
+  await redisClient.del(`cancel:${BullMQjobId}`)
+}
```
andresgutgon, line 31: Move this to `src/lib`

andresgutgon: Probably we have other things related with redis there

packages/core/src/services/chains/ProviderProcessor/index.ts
```diff
@@ -61,3 +64,44 @@ async function buildOutput(
...
+  }
+}
+
+const nullLanguageModelUse = {
+  promptTokens: 0,
+  completionTokens: 0,
+  totalTokens: 0,
+}
+
+const fake_assistant_message = (accumulatedText: string): AssistantMessage => {
```
andresgutgon: `camelCaseAlways` please

packages/core/src/services/copilot/latte/createLatteJob.ts
```diff
@@ -39,5 +39,11 @@ export async function createLatteJob({
     debugVersionUuid,
   } as RunLatteJobData)
 
-  return Result.nil()
+  if (!job.id) {
```
andresgutgon, line 42: Is this possible?

packages/core/src/services/copilot/latte/run.ts
andresgutgon, line 1: Doble check this is not used 🙏

packages/web-ui/src/lib/hooks/useLocalStorage.ts
```diff
@@ -19,6 +19,7 @@ export enum AppLocalStorage {
   playgroundActions = 'playgroundActions',
   expandParameters = 'expandParameters',
   latteThreadUuid = 'latteThreadUuid',
+  latteBullMqJobId = 'latteBullMqJobId',
```
andresgutgon: See no brands here either 😂 just `latteJobId`

packages/core/src/lib/streamManager/ChainStreamConsumer/consumeStream.ts
learningbizz: To save the text in case it aborts, to later save it as the text for the fake provider log

apps/web/src/actions/latte/stopChat.ts
learningbizz: After talking live, we're leaving this implementation as this can occur

packages/core/src/jobs/job-definitions/copilot/chat.ts
learningbizz: To do a finally after, and ensure the clear interval and the job cancellation flag are removed

packages/core/src/services/copilot/latte/createLatteJob.ts
learningbizz: If the queue system fails to create the job, or if there is a bug or misconfiguration, [job.id] could be undefined or missing. This

apps/web/src/actions/latte/stopChat.ts
```diff
@@ -0,0 +1,47 @@
...
+    try {
+      if (REMOVABLE_STATES.includes(state)) await job.remove()
+      else if (NON_REMOVABLE_STATES.includes(state)) await cancelJob(jobId)
+    } catch (error) {
+      // to catch potential race condition between getting the state and removing
+      state = await job.getState()
+      if (REMOVABLE_STATES.includes(state)) await job.remove()
+      else if (NON_REMOVABLE_STATES.includes(state)) await cancelJob(jobId)
+    }
+  })
```
andresgutgon, line 21: If not done now add a todo here to move to services and add tests for this flow

apps/web/src/components/LatteChat/LatteChatInput.tsx
```diff
@@ -145,42 +147,61 @@ export function LatteChatInput({
...
+            color: 'latteInputForeground',
+            className:
+              'flex-shrink-0 group-hover:text-latte-input-foreground/75',
+          }}
+          className='text-latte-input-foreground group-hover:text-latte-input-foreground/75'
+          userSelect={false}
+        >
+          <Text.H5 noWrap color='latteInputForeground'>
+            New chat
+          </Text.H5>
```
andresgutgon, line 203: This was already this way. But having to use a <Text> inside the button is a smell that the Latte variant is not configured correctly.

PR #1598, opened 2025-08-27
Latte diffs in dev editor

neoxelox:


packages/web-ui/src/ds/molecules/DocumentTextEditor/ReadOnlyMessage/index.tsx
```diff
@@ -2,6 +2,8 @@ import { cn } from '../../../../lib/utils'
 import { Icon } from '../../../atoms/Icons'
 import { Text } from '../../../atoms/Text'
 
+// TODO: FIX MESSAGE CANNOT BE HARDCODED
```
andresgutgon: What does this mean?

PR #1601, opened 2025-08-27
Replace draftUuid for latte tools

learningbizz:
Latte is getting confused when reading draftUuid in Latte tools, as it doesnt know what it is. To fix this, we will reference in Lattes tools the versionUuid as only versionUuid and not draftUuid

### DIFF
--- a/packages/core/src/services/copilot/latte/threads/checkpoints/undoChanges.test.ts
+++ b/packages/core/src/services/copilot/latte/threads/checkpoints/undoChanges.test.ts
@@ -126,7 +126,7 @@ describe('undo latte changes', () => {
     await LATTE_TOOLS[LatteTool.editProject](
       {
         projectId: project.id,
-        draftUuid: draft.uuid,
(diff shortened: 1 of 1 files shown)


apps/web/src/components/LatteChat/LatteChatInput.tsx
```diff
@@ -193,14 +193,12 @@ export function LatteChatInput({
             name: 'rotate',
             color: 'latteInputForeground',
             className:
-              'flex-shrink-0 group-hover:text-latte-input-foreground/75',
+              'flex-shrink-0 group-hover:text-latte-input-foreground/75 -mt-px',
           }}
           className='text-latte-input-foreground group-hover:text-latte-input-foreground/75'
           userSelect={false}
         >
-          <Text.H5 noWrap color='latteInputForeground'>
```
learningbizz, line 201: A smell that removing it doesnt affect its performance, so removing it

PR #1604, opened 2025-08-27
fix(latte-steps)/adapting-to-stop-mechanism

learningbizz:
Now the step interactions are adapted to the stop mechanism.

We now:
- Don't show "Brewing..." default step when its an older interaction of the message chain (this occurs if you stop the response of Latte before it start writing)
- When stopping an agent while using a tool, the loading animation stops.
- Only animates the interaction steps of the current interaction the user is doing with Latte


neoxelox:
Went of a different simpler route on https://github.com/latitude-dev/latitude-llm/pull/1608

PR #1605, opened 2025-08-27
fix(latte-stop)/subagents-dont-stop

learningbizz:
We used to not propagate the abortSignal to the subagents, now we do


packages/core/src/lib/cancelJobs.ts
```diff
@@ -1,9 +1,11 @@
 import { cache as redis } from '@latitude-data/core/cache'
 
+const CANCEL_JOB_FLAG_TTL = 60 * 60 // 1 hour
+
 const redisClient = await redis()
 
 export async function setCancelJobFlag(jobId: string) {
-  await redisClient.set(`cancel:${jobId}`, '1')
+  await redisClient.set(`cancel:${jobId}`, '1', 'EX', CANCEL_JOB_FLAG_TTL)
```
andresgutgon, line 8: What's `EX`?

learningbizz: EX is a type for "expire" within the `cancel_job_flag_ttl` time.

PR #1606, opened 2025-08-27
Fork triggers

csansoon:
Clone triggers and import integrations when forking a document into a new Project.

When importing integrations, it will try to match them to existing valid integrations, or create unconfigured ones if none is found.

WIP, missing tests


packages/core/src/services/documents/forkDocument/cloneTriggers.ts
```diff
@@ -0,0 +1,261 @@
+import { DocumentTriggerType, IntegrationType } from '@latitude-data/constants'
```
andresgutgon, line 1: Tests here

PR #1612, opened 2025-08-29
Fix LATITUDE-WEB-1EM cron input generate infinite updates

andresgutgon:
When clicking on the cron input compoment we had a `useEffect` (PLEASE_NO_PLEASE) syncing 2 inputs that was causing a maximum call stack


packages/web-ui/src/ds/organisms/CronInput/Interval.tsx
```diff
@@ -1,6 +1,6 @@
-import { useEffect, useState } from 'react'
```
andresgutgon, line 1: 🔥

PR #1614, opened 2025-08-29
fix(delete-doc)/propagate-with-triggers

learningbizz:
Now triggers will delete together with documents, as before they werent synchronized.

### DIFF
--- a/packages/core/src/services/documentTriggers/delete.test.ts
+++ b/packages/core/src/services/documentTriggers/delete.test.ts
@@ -12,10 +12,7 @@ import { Result } from '../../lib/Result'
 import * as factories from '../../tests/factories'
 import { mergeCommit } from '../commits'
 import { createDocumentTrigger } from './create'
-
-import { deleteDocumentTrigger } from './delete'
 import { DocumentTriggersRepository } from '../../repositories'
-import { BadRequestError, NotFoundError } from '@latitude-data/constants/errors'
 
 const mocks = vi.hoisted(() => ({
   deployDocumentTrigger: vi.fn(),
@@ -34,15 +31,24 @@ vi.mock('../../events/publisher', () => ({
   publisher: mocks.publisher,
 }))
 
-describe.sequential('deleteDocumentTrigger', () => {
+vi.mock('../../events/publisher', () => ({
+  publisher: mocks.publisher,
+}))
+
+describe.sequential('deleting documents...', () => {
   let workspace: Workspace
   let project: Project
   let draft: Commit
   let document: DocumentVersion
   let user: User
+  let deleteDocumentTrigger: typeof import('./delete').deleteDocumentTrigger
 
   beforeEach(async () => {
     vi.clearAllMocks()
+    vi.resetAllMocks()
+    await vi.resetModules()
+    const deleteModule = await import('./delete')
+    deleteDocumentTrigger = deleteModule.deleteDocumentTrigger
 
     const {
       workspace: w,
@@ -75,7 +81,7 @@ describe.sequential('deleteDocumentTrigger', () => {
     })
 
     expect(result.ok).toBeFalsy()
-    expect(result.error).toBeInstanceOf(BadRequestError)
+    expect(result.error?.constructor.name).toBe('BadRequestError')
     expect(result.error?.message).toBe(
       'Cannot delete a trigger in a live version',
     )
@@ -90,7 +96,7 @@ describe.sequential('deleteDocumentTrigger', () => {
     })
 
     expect(result.ok).toBeFalsy()
-    expect(result.error).toBeInstanceOf(NotFoundError)
+    expect(result.error?.constructor.name).toBe('NotFoundError')
     expect(result.error?.message).toContain(
       "Trigger with uuid 'non-existent-uuid' not found in commit",
     )
@@ -181,6 +187,8 @@ describe.sequential('deleteDocumentTrigger', () => {
 
     // New draft
     const { commit: newDraft } = await factories.createDraft({ project, user })
+
+    mocks.undeployDocumentTrigger.mockResolvedValue(Result.ok(undefined))
 
     const result = await deleteDocumentTrigger<DocumentTriggerType.Email>({
       workspace,
--- a/packages/core/src/services/documentTriggers/deleteAllDocumentTriggersFromCommit.test.ts
+++ b/packages/core/src/services/documentTriggers/deleteAllDocumentTriggersFromCommit.test.ts
@@ -1,0 +1,257 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { DocumentTriggerType, Providers } from '@latitude-data/constants'
+import {
+  Commit,
+  Project,
+  Workspace,
+  DocumentTrigger,
+  DocumentVersion,
+  User,
+} from '../../browser'
+import { Result } from '../../lib/Result'
+import * as factories from '../../tests/factories'
+import { mergeCommit } from '../commits'
+import { createDocumentTrigger } from './create'
+import { DocumentTriggersRepository } from '../../repositories'
+
+const mocks = vi.hoisted(() => ({
+  deployDocumentTrigger: vi.fn(),
+  undeployDocumentTrigger: vi.fn(),
+}))
+
+vi.mock('./deploy', () => ({
+  deployDocumentTrigger: mocks.deployDocumentTrigger,
+  undeployDocumentTrigger: mocks.undeployDocumentTrigger,
+}))
+
+describe('deleting documents...', () => {
+  let workspace: Workspace
+  let project: Project
+  let draft: Commit
+  let document: DocumentVersion
+  let user: User
+  let deleteDocumentTriggersFromDocuments: typeof import('./deleteDocumentTriggersFromDocuments').deleteDocumentTriggersFromDocuments
+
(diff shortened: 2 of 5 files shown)


packages/core/src/services/documentTriggers/delete.ts
```diff
@@ -185,3 +185,30 @@ export async function deleteDocumentTrigger<T extends DocumentTriggerType>(
...
+  commit: Commit
+}): PromisedResult<undefined> {
+  const documentTriggerScope = new DocumentTriggersRepository(workspace.id)
+  const activeDocumentTriggersResult =
+    await documentTriggerScope.getTriggersInProject({
+      projectId: commit.projectId,
+      commit,
+    })
+  const activeDocumentTriggers = activeDocumentTriggersResult.unwrap()
+  await Promise.all(
```
andresgutgon: Don't do this
image


Use results:
``` javascript
    const results = activeDocumentTriggers.map((trigger) =>
      deleteDocumentTrigger({
        workspace,
        commit,
        triggerUuid: trigger.uuid,
      }),
    )
    
    // Here you can see if there are any errors and those don't delete the associated documents . Or even make all fail and let user try again

```

packages/core/src/services/documentTriggers/delete.ts
```diff
@@ -185,3 +185,44 @@ export async function deleteDocumentTrigger<T extends DocumentTriggerType>(
     return Result.ok(currentDocumentTrigger)
   })
 }
+
+export async function deleteAllDocumentTriggersFromCommit(
```
andresgutgon: This is a new file called `deleteAllDocumentTriggersFromCommit.ts`.

packages/core/src/services/documentTriggers/delete.ts
```diff
@@ -185,3 +185,44 @@ export async function deleteDocumentTrigger<T extends DocumentTriggerType>(
...
+  documentTriggerScope = new DocumentTriggersRepository(workspace.id),
+): PromisedResult<undefined> {
+  const activeDocumentTriggersResult =
+    await documentTriggerScope.getTriggersInProject({
+      projectId: commit.projectId,
+      commit,
+    })
+
+  if (!Result.isOk(activeDocumentTriggersResult))
+    return Result.error(activeDocumentTriggersResult.error)
```
andresgutgon: Shorter and is the same
```javascript
if (activeDocumentTriggersResult.error) return activeDocumentTriggersResult.error
```

packages/core/src/services/documents/destroyOrSoftDeleteDocuments.ts
```diff
@@ -140,6 +138,18 @@ export async function destroyOrSoftDeleteDocuments(
...
   transaction = new Transaction(),
 ): Promise<TypedResult<boolean, Error>> {
+  const deleteAllTriggerDocumentsResult =
+    await deleteAllDocumentTriggersFromCommit(
+      {
+        commit,
+        workspace,
+      },
+      transaction,
+    )
```
andresgutgon: Coordinate with @csansoon this should be fine with what he's doing but check with him.

packages/core/src/services/documentTriggers/delete.test.ts
```diff
@@ -316,4 +325,220 @@ describe.sequential('deleteDocumentTrigger', () => {
     // Still present since deletion did not proceed
     expect(triggers.find((t) => t.uuid === created.uuid)).toBeTruthy()
   })
+
+  describe('delete all triggers from commit', () => {
```
andresgutgon: All these is a new file for a new service. Services only have one exported function that does one thing.

packages/core/src/services/documentTriggers/delete.ts
learningbizz: This is not the same, this would return an error and not a result

andresgutgon: Sorry without `.error`. My point is you don't need to wrap `result.error` in `Result.error`. Just return the result

andresgutgon (approved):

packages/core/src/services/documents/destroyOrSoftDeleteDocuments.ts
```diff
@@ -162,6 +160,18 @@ export async function destroyOrSoftDeleteDocuments(
...
+        {
+          commit,
+          workspace,
+          documents,
+        },
+        transaction,
+      )
+
+    if (!Result.isOk(deleteAllTriggerDocumentsResult))
+      return deleteAllTriggerDocumentsResult
```
csansoon: Wrap this in `{}`

Long single-line if statements get automatically transformed to two lines by Prettier, and now it's a bit harder to see what the if contains at first glance

packages/core/src/services/documentTriggers/delete.test.ts
```diff
@@ -90,7 +96,7 @@ describe.sequential('deleteDocumentTrigger', () => {
     })
 
     expect(result.ok).toBeFalsy()
-    expect(result.error).toBeInstanceOf(NotFoundError)
+    expect(result.error?.constructor.name).toBe('NotFoundError')
```
csansoon, line 99: hmm why is this no longer working

packages/core/src/services/documentTriggers/deleteAllDocumentTriggersFromCommit.test.ts
```diff
@@ -0,0 +1,257 @@
...
+  deployDocumentTrigger: vi.fn(),
+  undeployDocumentTrigger: vi.fn(),
+}))
+
+vi.mock('./deploy', () => ({
+  deployDocumentTrigger: mocks.deployDocumentTrigger,
+  undeployDocumentTrigger: mocks.undeployDocumentTrigger,
+}))
+
+describe.sequential('deleting documents...', () => {
```
csansoon: `describe.sequential`?

packages/core/src/services/documentTriggers/deleteDocumentTriggersFromDocuments.ts
```diff
@@ -0,0 +1,66 @@
...
+      documents.map(async (document) => {
+        return documentTriggerScope.getTriggersInDocument({
+          documentUuid: document.documentUuid,
+          commit,
+        })
+      }),
+    )
+
+    if (Result.includesAnError(activeTriggersForDocumentsInCommitResult))
+      return Result.findError(activeTriggersForDocumentsInCommitResult)!
```
csansoon: ditto

packages/core/src/services/documentTriggers/deleteDocumentTriggersFromDocuments.ts
```diff
@@ -0,0 +1,66 @@
...
+    )
+
+    const activeTriggersForDocumentsInCommitResult = await Promise.all(
+      documents.map(async (document) => {
+        return documentTriggerScope.getTriggersInDocument({
+          documentUuid: document.documentUuid,
+          commit,
+        })
+      }),
+    )
```
csansoon, line 32: Maybe it is just easier (and faster) to add a `getTriggersInProject` method into the `DocumentTriggersRepository`

packages/core/src/services/documentTriggers/deleteDocumentTriggersFromDocuments.ts
```diff
@@ -0,0 +1,66 @@
...
+                  commit,
+                  triggerUuid: trigger.uuid,
+                },
+                transaction,
+              )
+            }),
+          )
+        },
+      ),
+    )
```
csansoon: If any ONE of these fails, ALL the others will still try to execute, even though the db changes will be rolled back. 

Instead, I think it is a better option to just iterate over the regular list, and return an error as soon as it appears. Otherwise we're doing more work that we need and we might introduce errors.

packages/core/src/services/documentTriggers/deleteDocumentTriggersFromDocuments.ts
learningbizz: There is already, but we don't want all the triggers in a project, we just want those of the documents

packages/core/src/services/documents/destroyOrSoftDeleteDocuments.ts
learningbizz: Disagree, but I'll add it as ur not the only one who said so

packages/core/src/services/documentTriggers/delete.test.ts
learningbizz: Due to dynamic importing, as I'm resetting the modules in the beforeEach, its importing the NotFoundError from the delete module and not the test module, so the comparison is comparing different instances of the NotFoundError, and it fails

packages/core/src/services/documentTriggers/deleteDocumentTriggersFromDocuments.ts
learningbizz: Sure, as we want to stop when one fails anyways, good catch

packages/core/src/services/documentTriggers/deleteDocumentTriggersFromDocuments.ts
csansoon: Then it may be more efficient to retrieve all of them and then just filter it. It's a single transaction and a single point where it can fail

PR #1618, opened 2025-09-01
Undeploy Pipedream triggers asynchronously with jobs

csansoon:


packages/core/src/events/handlers/undeployDocumentTriggerJob.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const pipedream = createBackendClient(pipedreamEnv.unwrap())
+
+  try {
+    await pipedream.deleteTrigger({
+      id: triggerId,
+      externalUserId,
+    })
+  } catch (error) {
+    console.error(
```
learningbizz: console logs won't be captured by sentry, lets use captureException here

packages/core/src/services/documentTriggers/deploy/integrationTrigger.ts
```diff
@@ -84,9 +83,7 @@ export async function deployIntegrationTrigger(
 }
 
 /**
- * Important Note:
- * This service fetches data from an external service.
- * Do not include this service inside a transaction.
+ * Schedules the undeployment of an integration trigger.
  */
 export async function undeployIntegrationTrigger(
```
learningbizz, line 87: Why don't we create another service for this method? The service has too many responsibilities

packages/core/src/services/documentTriggers/deploy/integrationTrigger.ts
```diff
@@ -110,14 +107,33 @@ export async function undeployIntegrationTrigger(
...
+  }
+
+  // Schedule the undeployment to happen asynchronously via the event system
+  publisher.publishLater({
+    type: 'documentTriggerUndeployRequested',
+    data: {
+      workspaceId: workspace.id,
+      triggerId: documentTrigger.deploymentSettings.triggerId,
+      externalUserId: integration.configuration.externalUserId,
     },
```
andresgutgon, line 126: Tests

packages/core/src/services/documentTriggers/deploy/integrationTrigger.ts
csansoon: It just has the "deploy" and "undeploy" requirements and logic

packages/core/src/services/documentTriggers/deploy/integrationTrigger.ts
```diff
@@ -110,14 +107,33 @@ export async function undeployIntegrationTrigger(
...
+  }
+
+  // Check if the trigger has deployment settings (if not, nothing to undeploy)
+  if (!documentTrigger.deploymentSettings) {
+    return Result.ok(undefined)
+  }
+
+  // Get the first user in the workspace for the event
+  const user = await findFirstUserInWorkspace(workspace)
+  if (!user) {
```
andresgutgon: For what is used `user`? I don't see it

andresgutgon (approved):

packages/core/src/services/documentTriggers/deploy/integrationTrigger.ts
csansoon: whoops nope, that's not needed anymore

packages/core/src/services/documentTriggers/deploy/integrationTrigger.ts
learningbizz: Two different use cases :)

PR #1624, opened 2025-09-01
Improve Latte step descriptions

learningbizz:
Latte now has better descriptions, aligning better with our brand identity


learningbizz:
@Danil reviewed this

PR #1631, opened 2025-09-02
Divide trigger actions in separate tools

learningbizz:
We figured out after testing that Latte performed better with smaller, specific tools with clear namings than an abstract tool with many configurations, so we're adapting the tools for that


packages/core/src/services/copilot/latte/tools/triggers/actions/createTrigger.ts
```diff
@@ -0,0 +1,102 @@
...
+      )
+    }
+
+    const documentsScope = new DocumentVersionsRepository(workspace.id)
+    const documents = await documentsScope
+      .getDocumentsAtCommit(currentVersionCommit)
+      .then((r) => r.unwrap())
+
+    const document = documents.find((doc) => doc.documentUuid === promptUuid)
+
```
geclos, line 44: 👾

packages/core/src/services/copilot/latte/tools/triggers/actions/createTrigger.ts
```diff
@@ -0,0 +1,102 @@
...
+    const projectRepository = new ProjectsRepository(workspace.id)
+    const project = await projectRepository
+      .getProjectById(currentVersionCommit.projectId)
+      .then((r) => r.unwrap())
+
```
geclos, line 49: 👾

geclos (approved):
looks good, nitpicking on your use of blank lines

PR #1632, opened 2025-09-02
feat: refactors diff management of experiments and latte

geclos:
They now both use context and a simple useXDiff hook


apps/web/src/actions/latte/acceptChanges.ts
```diff
@@ -21,14 +22,10 @@ export const acceptLatteChangesAction = authProcedure
...
       threadUuid,
       accepted: true,
     })
 
-    const { result } = evaluationResult.unwrap()
-
-    return {
-      evaluationUuid: result.uuid,
-    }
+    return Result.ok({})
```
geclos: we don't do anything with these in frontend so i've removed the return value (side effect you no longer need an evaluation uuid in dev env)

apps/web/src/actions/latte/discardChanges.ts
```diff
@@ -21,14 +22,10 @@ export const discardLatteChangesActions = authProcedure
       threadUuid,
     }).then((r) => r.unwrap())
 
-    const evaluationResult = await evaluateLatteThreadChanges({
+    await evaluateLatteThreadChanges({
       threadUuid,
       accepted: false,
     })
 
-    const { result } = evaluationResult.unwrap()
```
geclos, line 29: we don't fail anymore if eval fails

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Editors/index.tsx
```diff
@@ -8,32 +9,23 @@ import {
...
-    isSaved,
-    diff: latteDiff,
-  } = useDocumentValue()
-  const { diff: editorDiff, setDiff: setEditorDiff } = useDiffState(
-    initialDiff,
-    updateDocumentContent,
-  )
+  const { value, updateDocumentContent, isSaved } = useDocumentValue()
+  const { diff: latteDiff } = useSyncLatteChanges()
+  const { diff: experimentDiff, setDiff: setEditorDiff } = useExperimentDiff()
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/Editors/index.tsx:28]: simple

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/OldDocumentEditor.tsx
```diff
@@ -32,7 +37,7 @@ export type DocumentEditorProps = {
   freeRunsCount?: number
   copilotEnabled: boolean
   refinementEnabled: boolean
-  initialDiff?: string
+  experimentDiff?: string
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/OldDocumentEditor.tsx:40]: be explicit with the names

apps/web/src/hooks/useDocumentValueContext.tsx
```diff
@@ -119,16 +110,6 @@ export function DocumentValueProvider({
...
-  const { diff } = useSyncLatteChanges({
-    content: value,
-    updateContent: setContentValue,
-    document: document,
-    commit: commit,
-    project: project,
-    devMode: devMode,
-    setDevMode: setDevMode,
-    newLatte: newLatte,
-  })
```
geclos, line 131: use  composition as much as possible instead of prop drilling, this hook had many props, it needed none

apps/web/src/hooks/useDocumentValueContext.tsx
```diff
@@ -146,7 +127,6 @@ export function DocumentValueProvider({
         updateDocumentContent,
         isSaved: !isUpdatingContent,
         document,
-        diff,
```
geclos, line 149: keep hooks public api small, diff can be its own hook, no need to polute this hook with it.

apps/web/src/hooks/useExperimentDiffContext.tsx
```diff
@@ -0,0 +1,66 @@
...
+          onReject: () => {
+            setDiff(undefined)
+            removeApplyExperimentIdFromUrl()
+          },
+        }
+      : undefined,
+  )
+
+  return { diff, setDiff }
+}
```
geclos, line 66: removed useDiffState and merged it in a single useExperimentDiff – let's not pretend they weren't coupled

apps/web/src/hooks/useSyncLatteChanges.ts
```diff
@@ -0,0 +1,98 @@
...
+} from '@latitude-data/web-ui/providers'
+import { useCallback, useMemo, useRef } from 'react'
+import { useDocumentValue } from './useDocumentValueContext'
+import { useDevMode } from './useDevMode'
+
+export function useSyncLatteChanges() {
+  const router = useNavigate()
+  const { commit } = useCurrentCommit()
+  const { project } = useCurrentProject()
+  const { document } = useCurrentDocument()
```
geclos, line 18: composition! this hook assumes it's instantiated in the right context where all this information is present.

andresgutgon (approved):

apps/web/src/actions/latte/acceptChanges.ts
learningbizz: wouldnt it make more sense here to return undefined then? or nil?

geclos: i was wrong it's actually used in frontend, i'm changing the code a bit to just not fail if the eval fails

PR #1633, opened 2025-09-03
Skip deployments on Cloned Triggers

csansoon:
When cloning triggers, it will now skip the deployment.

This has been added because the trigger configuration may not be copied when cloning it (to avoid personal information being leaked), and so it fails when trying to deploy it.

### DIFF
--- a/packages/core/src/services/documentTriggers/deploy.ts
+++ b/packages/core/src/services/documentTriggers/deploy.ts
@@ -34,18 +34,27 @@ export async function deployDocumentTrigger<T extends DocumentTriggerType>(
     triggerUuid,
     triggerType,
     configuration,
+    skipDeployment = false,
   }: {
     workspace: Workspace
     commit: Commit
     triggerUuid: string
     triggerType: T
     configuration: DocumentTriggerConfiguration<T>
+    skipDeployment?: boolean
   },
   transaction = new Transaction(),
 ): PromisedResult<{
   deploymentSettings: DocumentTriggerDeploymentSettings<T>
   triggerStatus: DocumentTriggerStatus
 }> {
+  if (skipDeployment) {
+    return Result.ok({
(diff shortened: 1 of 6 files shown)


packages/core/src/services/documentTriggers/deploy.ts
```diff
@@ -34,18 +34,27 @@ export async function deployDocumentTrigger<T extends DocumentTriggerType>(
...
   deploymentSettings: DocumentTriggerDeploymentSettings<T>
   triggerStatus: DocumentTriggerStatus
 }> {
+  if (skipDeployment) {
+    return Result.ok({
+      deploymentSettings: {} as DocumentTriggerDeploymentSettings<T>,
+      triggerStatus: DocumentTriggerStatus.Pending,
+    })
+  }
+
```
andresgutgon, line 57: Test me baby

csansoon: done! 😃👍

packages/core/src/tests/factories/documentTriggers.ts
```diff
@@ -189,3 +200,81 @@ export async function createIntegrationDocumentTrigger({
...
+      projectId,
+      commitId,
+      documentUuid,
+      triggerType: DocumentTriggerType.Chat,
+      triggerStatus: 'deployed',
+      configuration,
+      deploymentSettings,
+      enabled,
+    })
+    .returning()
```
andresgutgon, line 240: You have the create service

andresgutgon (approved):

packages/core/src/tests/factories/documentTriggers.ts
csansoon: But this factory skips stuff like deployment and such, it is just for easy creation in testing environment

PR #1637, opened 2025-09-04
Fix/filter empty aborted assistant messages

neoxelox:


packages/core/src/services/ai/providers/rules/vercel.ts
```diff
@@ -210,5 +243,7 @@ export function vercelSdkRules(
     return [{ ...extracted, content } as Message]
   }) as Message[]
 
+  messages = filterEmptyMessages(messages)
+
   return { ...rules, messages }
 }
```
csansoon, line 249: No need to reassign the variable, just

```ts
return {
  ...rules,
  messages: filterEmptyMessages(messages)
}
```

csansoon (approved):

packages/core/src/services/ai/providers/rules/vercel.ts
neoxelox: i like to reassign variables, it is like an assembly line

PR #1640, opened 2025-09-05
feat: partial latte changes

geclos:
Allows to accept files one by one rather than all or nothing. Refactors how changes are handled in frontend removing the latte store changes indirection and removing most side effects via events.


apps/web/src/hooks/useSyncLatteChanges.ts
```diff
@@ -1,98 +1,34 @@
...
   const backToPrevEditor = useCallback(() => {
     if (!simpleMode.current) return
     simpleMode.current = false
     setDevMode(false)
   }, [setDevMode])
+  const onLatteChangesRejected = useCallback(() => {
+    backToPrevEditor()
+  }, [backToPrevEditor])
+  const onLatteChangesAccepted = useCallback(() => {
+    backToPrevEditor()
```
geclos: I'm not mega convinced about this tbh, it results in the ui flickering a lot automatically (changing from simple to dev mode for the diff then back to simple when you accept changes). I've disabled it for now but feel free to discuss it

apps/web/src/stores/documentVersions.ts
```diff
@@ -340,36 +339,6 @@ export default function useDocumentVersions(
...
-            }
-          } else {
-            newDocuments.push(change.current as unknown as DocumentVersion)
-          }
-        })
-
-        return newDocuments
-      })
-    },
-  })
```
geclos, line 371: in terms of design this is actually a very good approach but it was resulting in some performance problems at least in my dev mode, so i've switched to imperatively updating local state whenever it's needed — I will revisit this when we have some more bandwidth

AGENTS.md
```diff
@@ -10,7 +10,8 @@
 - `pnpm test:watch` - Run tests in watch mode (specific packages)
 - `pnpm --filter @latitude-data/core db:migrate` - Run database migrations
 - `pnpm --filter @latitude-data/core db:generate` - Generate database migrations
-- `pnpm prettier` - **ALWAYS run this when completing tasks** - Formats all code to standardize patterns
```
andresgutgon, line 13: This is annoying. Cursor really does this all the time. I prefer to run prettier by myself.

apps/web/src/actions/latte/partialRejectChanges.ts
```diff
@@ -0,0 +1,35 @@
...
+    await partialRejectLatteChanges({
+      workspace,
+      threadUuid,
+      documentUuidsToReject,
+    }).then((r) => r.unwrap())
+
+    // Evaluate the remaining changes (if any)
+    const result = await evaluateLatteThreadChanges({
+      threadUuid,
+      accepted: false, // Mark as rejected since we're rejecting specific changes
```
andresgutgon, line 29: This is contradictory to the partial acceptance. If the user accepts some of the changes is accepted for me. Does not matter if they partially reject some of them.

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/DocumentEditor.tsx
```diff
@@ -77,12 +78,13 @@ function DocumentEditorContent({
...
   refinementEnabled,
 }: Omit<DocumentEditorProps, 'experimentDiff'>) {
-  const { updateDocumentContent, document } = useDocumentValue()
+  const { updateDocumentContent } = useDocumentValue()
   const [mode, setMode] = useState<'preview' | 'chat'>('preview')
   const { metadata } = useMetadata()
   const { commit } = useCurrentCommit()
   const { project } = useCurrentProject()
   const { toggleDocumentation } = useDeployPrompt()
+  const { document } = useCurrentDocument()
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/DocumentEditor.tsx:87]: Why this change. What's different from the these 2 `document`?

AGENTS.md
geclos: that's why i removed it

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/DocumentEditor.tsx
Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/DocumentEditor.tsx:87]: the other hook needlessly exported document instance, i've just removed that export and favoured useCurrentDocument, which is what we use everywhere. Keep hooks APIs as small as possible.

apps/web/src/actions/latte/partialRejectChanges.ts
geclos: but... this is rejecting the changes no? We do not know if the user will accept the rest of the changes, we only know it rejected some at this point

apps/web/src/app/providers/DocumentProvider.tsx
```diff
@@ -24,20 +24,15 @@ const DocumentVersionProvider = ({
...
+    projectId: projectId,
+    commitUuid: commitUuid,
   })
-
-  const document = useMemo(() => {
-    return (
-      documents?.find((d) => d.id === fallbackDocument.id) ?? fallbackDocument
-    )
-  }, [documents, fallbackDocument])
+  const document = documents.find((d) => d.id === fallbackDocument.id)
```
andresgutgon: Why the memo removal?

apps/web/src/hooks/latte/index.ts
```diff
@@ -0,0 +1,9 @@
+'use client'
+
+export { useLatteChangeActions } from './useLatteChangeActions'
+export { useLatteChatActions } from './useLatteChatActions'
+export { useLatteDebugMode } from './useLatteDebugMode'
+export { useLatteProjectChanges } from './useLatteProjectChanges'
+export { useLatteThreadUpdates } from './useLatteThreadUpdates'
+export { useLoadThreadFromProviderLogs } from './useLoadThreadFromProviderLogs'
+export { useSyncLatteUrlState } from './useSyncLatteUrlState'
```
andresgutgon, line 9: Why do we need barrel files? I find it unnecessary.

apps/web/src/app/providers/DocumentProvider.tsx
geclos: mistake

apps/web/src/hooks/latte/index.ts
geclos: we don't really, was just easier for me, can remove

apps/web/src/hooks/latte/useLatteDebugMode.ts
```diff
@@ -1,11 +1,13 @@
-import { useEffect, useMemo } from 'react'
-import useSWR from 'swr'
-import useFeature from '$/stores/useFeature'
-import { LatteVersion } from '@latitude-data/core/services/copilot/latte/debugVersions'
+'use client'
```
andresgutgon, line 1: Why add this everywhere? As I understand it's only needed in the file that's the boundary between backend <-> frontend. Hooks are usually already inside a `use client` component

andresgutgon: Let's clear this up to avoid adding it if not necessary

geclos: doesn't hurt 🤷🏼 , just don't bother about it

apps/web/src/stores/documentVersions.ts
```diff
@@ -60,29 +59,29 @@ export default function useDocumentVersions(
   )
 
   const router = useRouter()
-  const { execute: executeCreateDocument } = useServerAction(
+  const { execute: executeCreateDocument } = useLatitudeAction(
```
learningbizz, line 62: When should we use ServerAction or LatitudeAction?

apps/web/src/stores/documentVersions.ts
```diff
@@ -60,29 +59,29 @@ export default function useDocumentVersions(
...
     createDocumentVersionAction,
     {
       onSuccess: ({ data: document }) => {
         onSuccessCreate?.(document)
       },
     },
   )
-  const { execute: executeUploadDocument } = useServerAction(
+  const { execute: executeUploadDocument } = useLatitudeAction(
```
learningbizz, line 70: Also isnt useLatitudeAction not imported?

apps/web/src/stores/latteThreadCheckpoints.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export default function useLatteThreadCheckpoints({
+  threadUuid,
+  commitId,
+}: {
+  threadUuid?: string
+  commitId?: number
+} = {}) {
+  const enabled = !!threadUuid && !!commitId
+  const fetcher = useFetcher<LatteThreadCheckpoint[]>(
```
andresgutgon, line 17:
```javascript 
const route = ROUTES.api.latte.threads.detail(threadUuid).checkpoints.root
useFetcher<LatteThreadCheckpoint[]>(
  enabled ? route : undefined,
 { searchParams: { commitId} }
)
```

apps/web/src/services/routes/api.ts
```diff
@@ -398,5 +398,13 @@ export const API_ROUTES = {
         root: `/api/latte/debug/versions`,
       },
     },
+    threads: {
+      detail: (threadUuid: string) => ({
+        checkpoints: {
+          detail: (commitId: number, documentUuid?: string) =>
+            `/api/latte/threads/${threadUuid}/checkpoints?commitId=${commitId}&documentUuid=${documentUuid}`,
```
andresgutgon: This is not necessary. You can pass the query params in the `useFetecher` as `searchParams`. Doing that way keep the URL clean. No need to pass pass in the url
https://github.com/latitude-dev/latitude-llm/pull/1640/files#r2324361765

apps/web/src/app/providers/DocumentProvider.tsx
learningbizz: Won't removing the null-safe access of documents potentially break when documents is undefined in the first render of DocumentProvider?

packages/core/src/services/copilot/latte/threads/checkpoints/partialAcceptChanges.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const threadsScope = new LatteThreadsRepository(workspace.id)
+  const checkpointsToUpdate = await threadsScope
+    .findAllCheckpoints(threadUuid)
+    .then((r) => r.unwrap())
+    .then((checkpoints) =>
+      checkpoints.filter((cp) =>
+        documentUuidsToAccept.includes(cp.documentUuid),
+      ),
+    )
```
andresgutgon, line 36: This can be done easily in SQL no?
```javascript
import { inArray, and, eq } from "drizzle-orm"

async findAllCheckpoints(
  threadUuid: string,
  documentUuids?: string[], // optional param
): Promise<LatteThreadCheckpoint[]> {
  let whereClause = and(
    this.scopeFilter,
    eq(latteThreads.uuid, threadUuid),
  )

  if (documentUuids && documentUuids.length > 0) {
    whereClause = and(
      whereClause,
      inArray(latteThreadCheckpoints.documentUuid, documentUuids),
    )
  }

  return await this.db
    .select(getTableColumns(latteThreadCheckpoints))
    .from(latteThreadCheckpoints)
    .leftJoin(
      latteThreads,
      eq(latteThreadCheckpoints.threadUuid, latteThreads.uuid),
    )
    .where(whereClause)
}
```

Also if you do this remove the Result does not makes sense. This will never fail

learningbizz: Why do you prefer doing it in SQL?

apps/web/src/app/providers/DocumentProvider.tsx
geclos: there's a fallback in the context setting

packages/core/src/services/copilot/latte/threads/checkpoints/partialRejectChanges.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    const threadsScope = new LatteThreadsRepository(workspace.id, tx)
+    const checkpoints = await threadsScope
+      .findAllCheckpoints(threadUuid)
+      .then((r) => r.unwrap())
+      .then((checkpoints) =>
+        checkpoints.filter((cp) =>
+          documentUuidsToReject.includes(cp.documentUuid),
+        ),
+      )
```
andresgutgon: Same thing

apps/web/src/stores/documentVersions.ts
geclos: uselatitudeaction always if possible

apps/web/src/stores/documentVersions.ts
geclos: it is

packages/core/src/services/copilot/latte/threads/checkpoints/partialRejectChanges.ts
```diff
@@ -0,0 +1,51 @@
...
+    await tx.delete(latteThreadCheckpoints).where(
+      inArray(
+        latteThreadCheckpoints.id,
+        checkpoints.map((c) => c.id),
+      ),
+    )
+
+    return Result.nil()
+  })
+}
```
andresgutgon, line 48: Tests

apps/web/src/services/routes/api.ts
geclos: lovely

apps/web/src/actions/latte/partialRejectChanges.ts
andresgutgon: Exactly. We don't know if they will reject all. So marking as rejected the eval is not ok imo. Only when they use the `undo` button or when is the last partial reject present

packages/core/src/services/copilot/latte/threads/checkpoints/partialAcceptChanges.ts
andresgutgon: Faster, always in SQL if possible, and not too convoluted. The result consumes less memory. Also is less code from the user of this method perspective

geclos: always better to filter data out in db  to not overload runtime memory with unnecessary data.

packages/core/src/repositories/latteThreadsRepository.ts
```diff
@@ -69,4 +69,50 @@ export class LatteThreadsRepository extends Repository<LatteThread> {
...
+      .from(latteThreadCheckpoints)
+      .leftJoin(
+        latteThreads,
+        eq(latteThreadCheckpoints.threadUuid, latteThreads.uuid),
+      )
+      .where(
+        and(
+          this.scopeFilter,
+          eq(latteThreads.uuid, threadUuid),
+          inArray(latteThreadCheckpoints.documentUuid, documentUuids),
```
andresgutgon, line 114: What if the Array is empty?

geclos: that's fine it won't match

apps/web/src/hooks/latte/index.ts
andresgutgon: 🔥

PR #1641, opened 2025-09-05
feat: google gcp file adapter

geclos:
Implements flydrive GCS driver


andresgutgon (approved):

packages/core/src/lib/disk/config.ts
```diff
@@ -0,0 +1,81 @@
+import { env } from '@latitude-data/env'
+import { FSDriver } from 'flydrive/drivers/fs'
+import { S3Driver } from 'flydrive/drivers/s3'
+import { GCSDriver } from 'flydrive/drivers/gcs'
```
andresgutgon, line 4: Super proud of this tech choice

PR #1645, opened 2025-09-05
fix-websockets-order

csansoon:
### Problem

Currently, when a worker needs to send a message to a browser client, it must go through our WebSocket Server first:

```
WORKER ---HTTP---> WEBSOCKET SERVER ---WebSocket---> BROWSER
```

However, this approach has a major drawback:
Each message is sent as a separate HTTP request. Because the WebSocket server handles these requests independently, message order is **not guaranteed**. Network latency, payload size, and request handling can all cause messages to arrive out of sequence.

---

### Solution

Instead of sending individual HTTP requests, workers now publish messages to **Redis PubSub**.
The WebSocket Server subscribes to the Redis channel and forwards messages to the browser:

```
WORKER ---PubSub---> REDIS ---Subscription---> WEBSOCKET SERVER ---WebSocket---> BROWSER
```

Redis PubSub preserves **FIFO (first-in, first-out)** order, ensuring that messages always reach the client in the correct sequence.

---

### Benefits

* ✅ Guaranteed message ordering
* ✅ Reduced overhead on the WebSocket Server
* ✅ Cleaner separation of responsibilities between workers and message delivery

### DIFF
--- a/apps/websockets/src/server.ts
+++ b/apps/websockets/src/server.ts
@@ -9,12 +9,12 @@ import {
 import {
   buildWorkspaceRoom,
   verifyWebsocketToken,
-  verifyWorkerWebsocketToken,
 } from '@latitude-data/core/websockets/utils'
 import { env } from '@latitude-data/env'
 import cookieParser from 'cookie-parser'
 import express from 'express'
 import { Namespace, Server, Socket } from 'socket.io'
+import Redis from 'ioredis'
 
 function parseCookie(cookieString: string): Record<string, string> {
   return cookieString.split(';').reduce(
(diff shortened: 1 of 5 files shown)


apps/websockets/src/server.ts
```diff
@@ -9,12 +9,12 @@ import {
 import {
   buildWorkspaceRoom,
   verifyWebsocketToken,
-  verifyWorkerWebsocketToken,
 } from '@latitude-data/core/websockets/utils'
 import { env } from '@latitude-data/env'
 import cookieParser from 'cookie-parser'
 import express from 'express'
 import { Namespace, Server, Socket } from 'socket.io'
+import Redis from 'ioredis'
```
geclos, line 17: import queues from core instead no?

csansoon: why? this is not using any workers, it's just subscribing to redis in the websockets server itself

apps/websockets/src/server.ts
```diff
@@ -134,6 +104,54 @@ web.on('connection', (socket) => {
...
+      try {
+        // Extract workspace ID from channel name
+        const workspaceId = parseInt(channel.split(':')[2])
+
+        // Parse message
+        const { event, data } = JSON.parse(message)
+
+        // Emit to workspace room
+        const room = buildWorkspaceRoom({ workspaceId })
+        web.to(room).emit(event as any, data)
```
andresgutgon, line 130: Can this be typed?

apps/websockets/src/server.ts
geclos: nevermind i thought core exported a queues redis instance

PR #1646, opened 2025-09-05
Scope Latte threads by project

andresgutgon:
# What?
We want to allow users to have different open conversations with Latte in each project

## TODO
- [x] Finish rendering from DB the tools
- [x] QA with multiple projects
- [x] Re-Review the migration


packages/core/src/services/copilot/latte/tools/triggers/actions/deleteTrigger.ts
```diff
@@ -45,22 +42,6 @@ const deleteTrigger = defineLatteTool(
...
-        .then((r) => r.unwrap())
-
-      if (!project) {
-        return Result.error(
-          new NotFoundError(
-            `Project with ID ${currentVersionCommit.projectId} not found.`,
-          ),
-        )
-      }
-    }
```
andresgutgon, line 62: I think this check is not necessary now

apps/web/src/components/LatteChat/index.tsx
```diff
@@ -33,15 +34,9 @@ import { LatteMessageList } from './_components/MessageList'
 import { LatteChatInput } from './LatteChatInput'
 import { useLatteEventHandlers } from '$/hooks/latte/useLatteEventHandlers'
 
-export function LatteChat() {
+export function LatteChatUI({ project }: { project: Project }) {
```
geclos: useCurrentProject instead

apps/web/src/hooks/latte/useLoadThreadFromProviderLogs.ts
```diff
@@ -1,21 +1,105 @@
...
+  AppLocalStorage,
+  useLocalStorage,
+} from '@latitude-data/web-ui/hooks/useLocalStorage'
 
 import { sortBy } from 'lodash-es'
 import { LatteInteraction } from './types'
+import { useOnce } from '$/hooks/useMount'
+import { Project, ProviderLogDto } from '@latitude-data/core/browser'
+
+async function fetchProviderLogHydratated({
```
geclos: typo

apps/web/src/hooks/latte/useLoadThreadFromProviderLogs.ts
```diff
@@ -1,21 +1,105 @@
...
+    : undefined
+
+  if (!latestProviderLogId) return undefined
+
+  const providerLog = await executeFetch<ProviderLogDto | undefined>({
+    route: ROUTES.api.providerLogs.detail(latestProviderLogId).root,
+  })
+
+  return providerLog
+}
```
geclos: weird, use stores instead

andresgutgon: This is fundamental. Is inside a `useOnce`. The data is set in interactions. This is just the fetching of the hydrated provider log. All is inside this useOnce -> useTransaction process. 

I need to finish testing it but I find very clean

packages/core/drizzle/0205_add_project_id_scope.sql
```diff
@@ -0,0 +1,8 @@
+-- Empty these tables to avoid foreign key constraint issues
```
andresgutgon, line 1: ⚠️ ⚠️ PLEASE TAKE A GOOD LOOK ⚠️ ⚠️ ⚠️

apps/web/src/hooks/latte/useLoadThread/fetchProviderLogHydrated.ts
```diff
@@ -0,0 +1,30 @@
...
+    ? sortBy(providerLogs, 'generatedAt').at(-1)?.id
+    : undefined
+
+  if (!latestProviderLogId) return undefined
+
+  const providerLog = await executeFetch<ProviderLogDto | undefined>({
+    route: ROUTES.api.providerLogs.detail(latestProviderLogId).root,
+  })
+
+  return providerLog
```
andresgutgon, line 29: All this could be collapsed in a backend fetch
```js
api/projects/:project_id/latte/threads/:thead_uuid/interactions/route.ts
```

Would be faster and cleaner. Backend
1. Fetch the correct thread (aka documentLog)
2. Get the latest provider log and hydratates it
3. Transform messages -> interactions

On the frontend interactions should be a swr store scoped by project so we dont need custom logic just fetch this store and add items to this store when websockets happen.

I didn't want to do this in this PR to avoid more enthropy but I think would simplify things

apps/web/src/services/routes/api.ts
```diff
@@ -81,6 +81,9 @@ export const API_ROUTES = {
   },
   providerLogs: {
     root: '/api/providerLogs',
+    detail: (providerLogId: string | number) => ({
+      root: `/api/providerLogs/${providerLogId}`,
+    }),
```
andresgutgon, line 86: Missing route

packages/core/src/schema/models/latteThreads.ts
```diff
@@ -27,14 +20,14 @@ export const latteThreads = latitudeSchema.table(
     workspaceId: bigint('workspace_id', { mode: 'number' })
       .notNull()
       .references(() => workspaces.id, { onDelete: 'cascade' }),
+    projectId: bigint('project_id', { mode: 'number' }).notNull(),
```
andresgutgon, line 23: Not really used to be honest. Scoping is done in frontend as it was before with the local storage saved there but scoped by projectId

apps/web/src/stores/latte/index.ts
```diff
@@ -0,0 +1,139 @@
...
+      saveInLocalStorage('threadUuid', uuid)
+    },
+    [store, addOrRemoveValue, saveInLocalStorage],
+  )
+
+  const setJobId = useCallback(
+    (jobId: string | undefined) => {
+      store.setJobId(jobId)
+      addOrRemoveValue('jobId', jobId)
+      saveInLocalStorage('jobId', jobId)
```
andresgutgon, line 96: Same as before but now job/thread live under the same object:

```js
{ [projectId]: { jobId?: string; threadUuid?: string }}
```
As before we double save it in zustand and localStorage

packages/core/drizzle/0205_add_project_id_scope.sql
neoxelox: why didnt you used TRUNCATE?

andresgutgon: Didn't think much about it. I know truncate is more performant. But does it matter in this case?

apps/web/src/components/LatteChat/index.tsx
```diff
@@ -1,17 +1,17 @@
...
 import { useLatteChatActions } from '$/hooks/latte/useLatteChatActions'
-import { useSyncLatteUrlState } from '$/hooks/latte/useSyncLatteUrlState'
 import { useOnce } from '$/hooks/useMount'
 import {
   PlaygroundAction,
   usePlaygroundAction,
 } from '$/hooks/usePlaygroundAction'
 import useCurrentWorkspace from '$/stores/currentWorkspace'
-import { useLatteStore } from '$/stores/latte'
+import { useLatteStore } from '$/stores/latte/index'
```
geclos: weird index import

apps/web/src/hooks/latte/useLoadThread/buildInteractionsFromProviderLog.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  if (lastInteraction && providerLog.response) {
+    lastInteraction.steps.push({
+      type: 'text',
+      text: providerLog.response,
+    })
+  }
+
+  return interactions
+}
```
geclos, line 145: probably not now but it'd be cool to test this hook

apps/web/src/hooks/latte/useLoadThread/fetchProviderLogHydrated.ts
geclos: Why don't you use `useProviderLog`? it already returnst he hydrated log

apps/web/src/hooks/latte/useLoadThread/index.ts
```diff
@@ -0,0 +1,63 @@
...
+    lastLoadedProjectId.current = effectiveProjectId
+
+    startTransition(async () => {
+      const threadUuid = syncAndGetThreadUuid()
+      if (!threadUuid) return
+
+      const providerLog = await fetchProviderLogHydrated({ threadUuid })
+      if (!providerLog) return
+
+      const _interactions = buildInteractionsFromProviderLog({ providerLog })
```
geclos, line 48: all this could be a reactive useMemo that react to interactions and hydrated provider log being present afaict

apps/web/src/hooks/latte/useLoadThread/fetchProviderLogHydrated.ts
andresgutgon: That add 2 hooks in the middle with an intermediary state. This is sequential all happening in the same transaction. 

I usually prefer swr stores. But this is not a store. This is one time each page refresh setup of interactions

This interactions end in a zustand store. I think the real fix would be to move all this to the backend as commented and have its own store

apps/web/src/hooks/latte/useLatteChangeActions.ts
```diff
@@ -6,21 +6,23 @@ import { discardLatteChangesActions } from '$/actions/latte/discardChanges'
...
  */
 export function useLatteChangeActions() {
+  const { project } = useCurrentProject()
   const {
     threadUuid,
     latteActionsFeedbackUuid,
     setLatteActionsFeedbackUuid,
     setError,
-  } = useLatteStore()
+  } = useLatteStore({ project })
```
geclos: use the context in the store as well instead of passing this parameter, that store can't be used anywhere where there's no project anyways

apps/web/src/hooks/latte/useLatteChatActions.ts
```diff
@@ -16,8 +17,9 @@ import { LatteInteraction } from './types'
  *   - `sendMessage`: Function to send a message to the current or new Latte thread
  *   - `stopChat`: Function to stop the current active Latte response
  */
-export function useLatteChatActions() {
+export function useLatteChatActions({ project }: { project: Project }) {
```
geclos: don't prop drill this prop, there's no need, the context is meant to avoid this

apps/web/src/services/routes/api.ts
geclos: what is useProviderLog store using then?

apps/web/src/hooks/latte/useLoadThread/index.ts
andresgutgon: But this is pure input/output. Don't need to be reactive

This is the sequence:
1. Fist provider logs for document log
2. Find the latest by generated at
3. Fetch the hydrated provider log
4. Build the interactions
5. set the built interaction in zustand latte store `interactions` for the project

This only happens once per project. Does not need to be reactive to interactions coming from websockets no?

apps/web/src/stores/latte/index.ts
```diff
@@ -0,0 +1,139 @@
...
+  const resetAll = useCallback(() => {
+    store.resetAll()
+    setStoredLatteData((prev) => {
+      const { [projectId]: _, ...rest } = prev
+      return rest
+    })
+  }, [store, setStoredLatteData, projectId])
+
+  return useMemo(() => {
+    const currentProjectState = store.getCurrentProjectState()
```
geclos, line 110: that's weird why do you need to do this imperative fetching of state? just return the store it should automatically update when needed, if something is not updating when you want to the problem should be somewhere else, not in the store.

apps/web/src/stores/providerLogs.ts
```diff
@@ -20,9 +20,14 @@ export default function useProviderLogs(
   opts?: SWRConfiguration,
 ) {
   const fetcher = useFetcher<ProviderLogDto[], ProviderLogDto[]>(
-    buildRoute({ documentUuid, documentLogUuid, documentLogId }),
+    ROUTES.api.providerLogs.root,
```
geclos, line 23: much better

packages/core/drizzle/0205_add_project_id_scope.sql
geclos: what happens with checkpoints associated to the threads table

packages/core/src/schema/models/latteThreads.ts
geclos: better to store it in db anyways

csansoon (approved):

apps/web/src/hooks/latte/useLoadThread/fetchProviderLogHydrated.ts
geclos: move this method to the useProviderLog store and call it fetchProviderLog then, at least we centralize all provider log fetching logic in one place

apps/web/src/stores/latte/index.ts
andresgutgon: This has some logic inside so the store always return the same shape no matter if there is a current project or not

```js
const EMPTY_PROJECT_STATE = {
  threadUuid: undefined,
  jobId: undefined,
  usage: undefined,
  latteActionsFeedbackUuid: undefined,
  interactions: [],
  isBrewing: false,
  error: undefined,
} satisfies ProjectLatteState
//...
  getCurrentProjectState: () => {
    const state = get()
    if (!state.currentProjectId) return EMPTY_PROJECT_STATE

    return state.projectStates[state.currentProjectId] ?? EMPTY_PROJECT_STATE
  },
```

packages/core/drizzle/0205_add_project_id_scope.sql
andresgutgon: They have a delete cascade
https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/schema/models/latteThreadCheckpoints.ts#L14

apps/web/src/stores/latte/index.ts
geclos: hmm ok, i think this is breaking the memoization altogether though=

PR #1647, opened 2025-09-05
fix: Latte double message

csansoon:
Fixes this:
image


apps/web/src/hooks/latte/useLatteThreadUpdates.ts
```diff
@@ -142,17 +142,13 @@ export function useLatteThreadUpdates() {
         const lastInteraction = [...prev.slice(-1)][0]!
 
         if (update.type === 'fullResponse') {
-          lastInteraction.steps.push({
-            type: 'text',
-            text: update.response,
-          })
```
andresgutgon, line 148: So we don't override text with a full response just in case?

andresgutgon (approved):

PR #1648, opened 2025-09-05
Refactor Latte Integration Tools

learningbizz:
We have too many tools with similar functionalities for finding integration, tools and triggers. This PR simplifies this situation.

We will now have the following tools:
- SearchAvailableIntegrations - Search apps of pipedream using a query
- ListExistingIntegrations - List existing integrations of the workspace
- ListIntegrationTools - List the tools of an integration (existing in the workspace or not)
- ListIntegrationTriggers - List the triggers of an integration (existing in the workspace or not)


packages/core/src/services/copilot/latte/tools/settings/listIntegrationTools.ts
```diff
@@ -1,19 +1,28 @@
...
+import { Result } from '../../../../../lib/Result'
 
 const listIntegrationTools = defineLatteTool(
-  async ({ name }, { workspace }) => {
-    const integrationsScope = new IntegrationsRepository(workspace.id)
-    const integrationResult = await integrationsScope.findByName(name)
-    if (!integrationResult.ok) return integrationResult
-    const integration = integrationResult.unwrap()
+  async ({ integrationAppName }) => {
+    const toolsResult = await listPipedreamIntegrationTools(integrationAppName)
```
andresgutgon, line 11: Careful. This method does an `unwrap` and it will 💥 . Change that service to return a result instead of `unwrap`.

packages/core/src/services/copilot/latte/tools/settings/listIntegrationTools.ts
```diff
@@ -1,19 +1,28 @@
...
+    if (!Result.isOk(toolsResult)) {
+      return Result.error(new LatitudeError(toolsResult.error.message))
+    }
```
andresgutgon: When you fix that service just do
```javascript
if (toolsResult.error) return toolResult
```

packages/core/src/services/copilot/latte/tools/triggers/listIntegrationTriggers.ts
```diff
@@ -1,34 +1,35 @@
...
-    if (!integrationResult.ok) return integrationResult
-    const integration = integrationResult.unwrap()
-
-    const triggersResult = await listTriggers(integration)
-    if (!triggersResult.ok) return triggersResult
+    const triggersResult = await listTriggers({
+      integrationType: IntegrationType.Pipedream,
+      integrationAppName,
+    })
+    if (!Result.isOk(triggersResult)) return triggersResult
```
andresgutgon, line 26:
```javascript
if (triggersResult.error) return triggerResult
```

packages/core/src/services/copilot/latte/tools/settings/listIntegrationTools.ts
learningbizz: it does an unrwap if the appResult is not an error, so we should be fine

packages/core/src/services/copilot/latte/tools/triggers/listIntegrationTriggers.ts
learningbizz: Whats the difference? If we do the isOk(), the triggerResult automatically becomes an ok which is useful

andresgutgon: The same with checking that way, and it is shorter. That's the main advantage.

packages/core/src/services/copilot/latte/tools/settings/listIntegrationTools.ts
andresgutgon: Where are these `.unwrap` caught?

packages/core/src/services/copilot/latte/tools/triggers/listIntegrationTriggers.ts
learningbizz: Then I dont understand why we added the isOk :/

packages/core/src/services/copilot/latte/tools/settings/listIntegrationTools.ts
learningbizz: why does it matter?

packages/core/src/services/copilot/latte/tools/triggers/listIntegrationTriggers.ts
geclos: result.ok will return falsy boolean if the underlying value is falsy (null, undefined, etc.), Result.isOk prevents this. We should unify the apis and remove one of the methods.

PR #1650, opened 2025-09-07
feat: improve typescript performance

geclos:
Implements type build steps in major internal packages in order to improve performance of typescript compilation during development.

This reduces type checking times by 10x in apps/web during development.


packages/core/src/lib/streamManager/step/streamAIResponse.ts
```diff
@@ -54,7 +54,12 @@ export async function streamAIResponse({
   schema?: JSONSchema7
   output?: Output
   abortSignal?: AbortSignal
-}) {
+}): Promise<{
+  response: any
+  messages: LegacyMessage[]
+  tokenUsage: any
+  finishReason: any
+}> {
```
geclos, line 66: cannot build types otherwise, probably a better fix than this exists.

packages/core/package.json
```diff
@@ -4,34 +4,113 @@
...
+      "import": "./src/workers/*.js"
+    },
+    "./utils/*": {
+      "types": "./dist/types/utils/*.d.ts",
+      "import": "./src/utils/*.ts"
+    },
+    "./telemetry": {
+      "types": "./dist/types/telemetry/index.d.ts",
+      "import": "./src/telemetry/index.ts"
+    }
```
geclos, line 110: tell typescript where to find the type declarations

apps/web/tsconfig.json
```diff
@@ -10,27 +10,33 @@
     "tsBuildInfoFile": ".tsbuildinfo",
     "moduleResolution": "Bundler",
     "jsx": "preserve",
-    "typeRoots": [
-      "./node_modules/@types"
-    ],
+    "isolatedModules": true,
```
andresgutgon, line 13: What does this?

packages/core/package.json
```diff
@@ -4,34 +4,113 @@
...
-    "./schema": "./src/schema/index.ts",
-    "./schema/*": "./src/schema/*.ts",
-    "./services/*": "./src/services/*.ts",
-    "./test": "./src/tests/useTestDatabase.ts",
-    "./websockets/*": "./src/websockets/*.ts",
-    "./workers/*": "./src/workers/*.js",
-    "./utils/*": "./src/utils/*.ts",
-    "./telemetry": "./src/telemetry/index.ts"
+    "./__deprecated/*": {
+      "types": "./dist/types/__deprecated/*.d.ts",
```
andresgutgon, line 8: Buff ⚠️ NO PLEASE

andresgutgon: This is making typescript work too much. Monorepo is prepared to use apps that consume TypeScript packages. This is not the issue here

packages/core/package.json
andresgutgon: Type declarations on a TypeScript code base consumed by an app make in Typescript

packages/core/package.json
```diff
@@ -4,34 +4,113 @@
...
+      "types": "./dist/types/utils/*.d.ts",
+      "import": "./src/utils/*.ts"
+    },
+    "./telemetry": {
+      "types": "./dist/types/telemetry/index.d.ts",
+      "import": "./src/telemetry/index.ts"
+    }
   },
   "scripts": {
+    "build": "tsc -p tsconfig.build.json",
```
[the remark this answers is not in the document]
andresgutgon: Manual build is not the way on a typescript monorepo. Again. I would do smaller PRs for Telemetry and SDK clean barrel imports. Measure perf improvements

Then I would check our own Zod type inference. We're doing some infinite reference that's making typescript work too much

With those changes memory and time processing will improve

andresgutgon (changes requested):
I already commented to hard in this PR 😂. I'll say no more here

andresgutgon (approved):
🙈

apps/web/tsconfig.json
geclos: some extra checks https://www.typescriptlang.org/tsconfig/#isolatedModules

PR #1653, opened 2025-09-08
Trigger Hash

csansoon:
### Problem

Although we are recording every event from triggers, displaying it was a bit tricky.

Previously, each event stored both a `triggerUuid` and a `commitId`. In theory, this allowed us to look up for the specific trigger version from a single event. However, this approach had two key issues:

1. Looking up a trigger version from a trigger event was straightforward with version control, but doing the reverse (finding events for a specific trigger version across multiple commits) was much more complex.
2. When a trigger was modified multiple times within the same draft, all events from all edited versions of that trigger would appear every time, rather than only showing the events corresponding to the latest configuration.

### Solution

We introduced a new field: `triggerHash`. This hash represents the configuration of a trigger at a specific point in time.

Now, each event records both the `triggerUuid` and the `triggerHash` at the moment the trigger fired. This makes it significantly easier to:

* Associate events with a specific trigger configuration, independent of commits or version history.
* Avoid mixing events from outdated trigger edits.


packages/core/src/repositories/documentTriggerEventsRepository.ts
```diff
@@ -17,113 +17,40 @@ export class DocumentTriggerEventsRepository extends Repository<DocumentTriggerE
...
-      )
+    const results = await query
 
-    const event = results[0]
-    if (!event) {
-      return Result.error(
-        new NotFoundError(`Trigger event with id '${id}' not found`),
-      )
-    }
+    console.log('Found', results.length, 'events')
```
geclos: 👀

packages/core/src/repositories/documentTriggerEventsRepository.ts
```diff
@@ -17,113 +17,40 @@ export class DocumentTriggerEventsRepository extends Repository<DocumentTriggerE
...
 
-  getAllTriggerEventsInWorkspace(): PromisedResult<
-    DocumentTriggerEvent[],
-    LatitudeError
-  > {
-    return this.getTriggerEvents()
-  }
+    console.log('workspaceId', this.workspaceId)
+    console.log('triggerUuid', trigger.uuid)
+    console.log('triggerHash', trigger.triggerHash)
```
geclos: 👁️

packages/core/src/tests/factories/documentTriggers.ts
```diff
@@ -128,17 +139,26 @@ export async function createEmailDocumentTrigger({
...
+    triggerUuid,
+    triggerType: DocumentTriggerType.Email,
+    configuration,
+  })
+
   const [trigger] = await database
     .insert(documentTriggers)
     .values({
       workspaceId,
+      uuid: triggerUuid,
```
geclos: why do we need this?

packages/core/src/tests/factories/documentTriggers.ts
```diff
@@ -261,16 +299,26 @@ export async function createDocumentTrigger<T extends DocumentTriggerType>({
   triggerStatus?: DocumentTriggerStatus
   deploymentSettings?: DocumentTriggerDeploymentSettings<T>
 }): Promise<DocumentTrigger<T>> {
+  const triggerUuid = generateUUIDIdentifier()
+
```
geclos: 👾

geclos (approved):

packages/core/src/repositories/documentTriggerEventsRepository.ts
csansoon: whoopss

packages/core/src/tests/factories/documentTriggers.ts
csansoon: yeah actually it's not reaaally needed

PR #1663, opened 2025-09-08
Grants - part 1

neoxelox:


packages/core/src/constants.ts
```diff
@@ -26,27 +28,26 @@ import type {
...
   type LegacyChainEvent,
-  EMAIL_REGEX,
-  isSafeUrl,
-  isLatitudeUrl,
 } from '@latitude-data/constants'
-export * from '@latitude-data/constants/evaluations'
-export * from '@latitude-data/constants/tracing'
 export * from '@latitude-data/constants/actions'
+export * from '@latitude-data/constants/evaluations'
+export * from '@latitude-data/constants/grants'
```
andresgutgon, line 48: Why do we re-export everything from here? This is not necessary, afaik

packages/core/src/data-migrations/backfillSubscriptionGrants.ts
```diff
@@ -0,0 +1,34 @@
...
+export async function backfillSubscriptionGrants(real: boolean = false) {
+  const workspaceIds = await database
+    .select({ id: workspaces.id })
+    .from(workspaces)
+    .then((result) => result.map((r) => r.id))
+
+  console.log(`Found ${workspaceIds.length} workspaces to backfill`)
+  if (!real) return
+
+  for (const workspaceId of workspaceIds) {
```
andresgutgon, line 15: I know this is wip but why look fro all ids and then do a find in each one?

packages/core/src/constants.ts
geclos: yes, don't do it, we wanna remove barrel files

packages/core/src/data-migrations/backfillSubscriptionGrants.ts
neoxelox: getting all workspaces entities would be bad for ram

PR #1674, opened 2025-09-09
ui rebranding

csansoon:
latitude but prettier


packages/web-ui/src/ds/atoms/Button/index.tsx
```diff
@@ -152,6 +147,23 @@ const buttonVariants = cva(
...
+  if (variant === 'destructive') return 'destructiveForeground'
+  if (variant === 'default') return 'background'
+  if (variant === 'secondary') return 'secondaryForeground'
+  if (variant === 'link') return 'accentForeground'
+  if (variant === 'linkOutline') return 'accentForeground'
+  if (variant === 'linkDestructive') return 'destructiveForeground'
+  if (variant === 'shiny') return 'accentForeground'
+  if (variant === 'latte') return 'latteInputForeground'
+  if (variant === 'primaryMuted') return 'primary'
+  return 'foreground'
```
andresgutgon, line 164: I'm sure there is a better way, but not going to say anything now.

packages/web-ui/src/ds/atoms/Button/index.tsx
```diff
@@ -313,7 +325,12 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(function Button(
                 className={cn('flex-shrink-0', iconProps.className)}
               />
             ) : null}
-            {children ? (
+
+            {typeof children === 'string' ? (
+              <Text.H5B noWrap color={textColorVariants({ variant })}>
+                {children}
+              </Text.H5B>
```
andresgutgon, line 332: Why this? Why wrapping in a Text element?

csansoon: homogeneidad

PR #1675, opened 2025-09-09
chore: open the gates

geclos:


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/ProjectSection/index.tsx
```diff
@@ -42,50 +41,29 @@ export default function ProjectSection({
   project: Project
   commit: Commit
 }) {
-  const feature = useFeature('latte')
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/ProjectSection/index.tsx:45]: Optional idea. Instead of changing code we could set an attribute on feature to enableForAll. If it's enable do not check what workspaces have it. This way from backoffice is easier to disable it if we need to

PR #1682, opened 2025-09-10
ui rebranding: a few more fixes

csansoon:


packages/web-ui/src/ds/tokens/font.ts
```diff
@@ -2,7 +2,7 @@ export const font = {
   size: {
     h8: 'text-[8px] leading-[10px]', // 8px/10px
     h7: 'text-[10px] leading-4', // 10px/16px
-    h6: 'text-xs leading-4', // 12px/16px
```
andresgutgon, line 5: Change the comment. Now is more than 12px

packages/web-ui/src/ds/atoms/Button/index.tsx
```diff
@@ -337,12 +340,20 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(function Button(
...
+                  textColor ??
+                  iconProps?.color ??
+                  textColorVariants({ variant })
+                }
                 className={cn('flex-shrink-0', iconProps.className)}
               />
             ) : null}
 
             {typeof children === 'string' ? (
-              <Text.H5B noWrap color={textColorVariants({ variant })}>
```
andresgutgon, line 345: Are we planing on rebuilding button with Danil? 
If so add a big and nice FIXME in this beast

csansoon: Yes we're rebuilding the whole button component soon. It has gotten too big of a beast with too many random attributes added for small fixes over time.

andresgutgon: Also what Danil showed with less variations is super nice 🙌

packages/web-ui/src/ds/atoms/Button/index.tsx
```diff
@@ -30,7 +30,8 @@ const buttonContainerVariants = cva(
         linkDestructive: 'shadow-none underline-offset-4 hover:underline',
         shiny: '',
         shinyLatte: '',
-        latte: 'bg-latte-border hover:bg-latte-border/90 border-latte-border',
+        // TODO: apply a color token for this
+        latte: 'bg-[#E5B217] hover:bg-[#E5B217]/90 border-latte-border',
```
andresgutgon, line 34: OMG this disgusting

PR #1691, opened 2025-09-12
compatible agent names

csansoon:
### Summary

When importing subagents into a prompt, an automatic tool name is generated based on the agent’s path. These names often end up long, repetitive, and unhelpful.

For example:

```yaml
agents:
  - /agents/documentation/ask
```

is turned into:

```
lat_agent_agents_documentation_ask
```

This can cause problems:

* Some providers fail due to excessively long tool names.
* The generated name doesn’t provide meaningful context for the original agent.

---

### What’s Changed

This PR introduces two improvements:

1. **Safer automatic name generation**

   * Tool names will no longer exceed provider limits.
   * If the path is too long, only the last valid segment(s) will be used.

2. **Optional `name` attribute**

   * Agents can now define a custom tool name explicitly, avoiding verbose or confusing defaults.

---

### Next Steps / TODO

* Detect and disallow importing multiple agents that resolve to the same tool name (to prevent provider errors and improve editor feedback).
* Fix `agentAsToolMap` so the correct UI is displayed in chat (currently broken).

### DIFF
--- a/packages/core/src/services/agents/agentsAsTools.ts
+++ b/packages/core/src/services/agents/agentsAsTools.ts
@@ -50,13 +50,14 @@ export async function getToolDefinitionFromDocument({
   ) => Promise<{ path: string; content: string } | undefined>
   streamManager: StreamManager
   context: TelemetryContext
-}): Promise<Tool> {
+}): Promise<{ name: string; toolDefinition: Tool }> {
   const metadata = await scan({
     prompt: document.content,
     fullPath: document.path,
     referenceFn,
   })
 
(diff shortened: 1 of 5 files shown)


packages/core/src/services/agents/agentsAsTools.ts
```diff
@@ -156,6 +157,11 @@ export async function getToolDefinitionFromDocument({
       }
     },
   }
+
+  return {
+    name: name ?? getAgentToolName(document.path),
+    toolDefinition,
+  }
```
andresgutgon, line 164: Do tests for this

Issue #1596, opened 2025-08-26
Make Vertex provider compatible with custom fine tuned models


neoxelox:
Seems Vercel SDK is always calling: `https://${location}-aiplatform.googleapis.com/v1/projects/${project}/locations/${location}/publishers/google` . For custom fine tuned models the url should be `https://${location}-aiplatform.googleapis.com/v1/projects/{project}/locations/${location}/endpoints/${endpoint}`.

Maybe we should create a "Vertex Custom" provider.

PR #1694, opened 2025-09-12
feat: vertex optional base url configuration

geclos:
Fixes https://github.com/latitude-dev/latitude-llm/issues/1596

### DIFF
--- a/packages/core/src/services/ai/helpers.ts
+++ b/packages/core/src/services/ai/helpers.ts
@@ -9,8 +9,8 @@ import {
 import {
   createVertexAnthropic,
   GoogleVertexAnthropicProvider,
-} from '@ai-sdk/google-vertex/anthropic/edge'
-import { createVertex, GoogleVertexProvider } from '@ai-sdk/google-vertex/edge'
+} from '@ai-sdk/google-vertex/anthropic'
+import { createVertex, GoogleVertexProvider } from '@ai-sdk/google-vertex'
 import { createMistral, MistralProvider } from '@ai-sdk/mistral'
 import { createOpenAI, type OpenAIProvider } from '@ai-sdk/openai'
 import { createPerplexity, PerplexityProvider } from '@ai-sdk/perplexity'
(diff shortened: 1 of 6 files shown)


packages/core/src/services/ai/helpers.ts
```diff
@@ -86,15 +86,20 @@ function validateVertexConfig({
   name: string
   maybeConfig: unknown
 }) {
+  console.log('maybeConfig', maybeConfig)
```
andresgutgon: 🔥

packages/core/src/services/ai/helpers.ts
```diff
@@ -86,15 +86,20 @@ function validateVertexConfig({
...
   const result = vertexConfigurationSchema.safeParse(maybeConfig)
+  console.log('result', result)
```
andresgutgon: 🔥

pnpm-workspace.yaml
```diff
@@ -10,7 +10,7 @@ catalog:
   drizzle-kit: 0.31.1
   dd-trace: 5.43.0
   promptl-ai: 0.7.5
-  zod: 3.24.2
+  zod: ^4
```
andresgutgon: This could be a big one. I would do a PR with this upgrade alone

andresgutgon: Or maybe is required for sdk v5? Dont remember

geclos: it's required

pnpm-workspace.yaml
```diff
@@ -10,7 +10,7 @@ catalog:
   drizzle-orm: 0.44.2
   drizzle-kit: 0.31.1
   dd-trace: 5.43.0
-  promptl-ai: 0.8.0
+  promptl-ai: ^0.9.2
```
andresgutgon: Wait
We need to move promptl to Zod V4. 
https://github.com/latitude-dev/promptl/pull/43

Now is working because is using a promptl version that was deployed with Zod but i reverted it in the code

PR #1696, opened 2025-09-14
feat: move last thread read to backend

geclos:
Moved to the backend so that we can server-render the previous chat and improve the UX 10x.


apps/web/src/hooks/latte/useLoadThread/useSyncAndGetThreadUuid.ts
```diff
@@ -1,53 +1,14 @@
...
-  if (!threadUuid && storedThreadUuid) return storedThreadUuid
-
-  return undefined
-}
-
-/**
- * Synchronizes the Latte thread UUID with local storage and returns the threadUuid to fetch.
- * Handles the logic of syncing between store state and localStorage for project-scoped threads.
- */
+/** No-op now that we don't use localStorage; returns current threadUuid. */
```
andresgutgon: Why we keep this file?

packages/core/src/services/providerLogs/getInitialForPreview.ts
```diff
@@ -0,0 +1,16 @@
+import { ProviderLogDto } from '../../browser'
+import { findLastProviderLogFromDocumentLogUuid } from '../../data-access'
+import { buildProviderLogResponse } from './buildResponse'
+
+export async function getInitialProviderLogForPreview({
```
andresgutgon: A bit weird this name. Is used for other places other than preview no?

andresgutgon (approved):

apps/web/src/hooks/latte/useLoadThread/useSyncAndGetThreadUuid.ts
geclos: vibe coding, will remove

packages/core/src/services/providerLogs/getInitialForPreview.ts
geclos: will change, vibe coding

PR #1704, opened 2025-09-15
Scheduled Timezones

csansoon:
**Scheduled Triggers now store and respect the client’s original timezone at the time of creation.**

Previously, triggers could behave inconsistently if they were created in one timezone and executed in another. By persisting the original timezone, scheduling becomes more predictable and easier to reason about.

* **Creating triggers:** No additional UI changes. A new trigger is created in the current user’s timezone.
* **Viewing/editing triggers in the same timezone:** No change in behavior — everything looks as before.
* **Viewing triggers created in a different timezone:** The original timezone is displayed in the trigger description.
  image
* **Editing triggers created in a different timezone:** The timezone is locked (not editable) and a warning is shown.
  image

This ensures triggers always run according to the timezone in which they were originally configured, reducing confusion and preventing accidental misfires.

---

Additionally, there have been some UI changes:
 - We had two different UIs to build schedule configurations. Removed the older one.
 - Improved Trigger List UI to better display what document each trigger runs (linkable and tooltip with the full path), and which integration is selected for integration triggers:
 image
 
- DRY'd a bunch of Trigger and CronInput components


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/ScheduleTrigger/Config.tsx
```diff
@@ -41,95 +18,53 @@ export function ScheduleTriggerConfig({
...
-      {config.type === 'custom' && (
-        <CustomScheduleForm
-          config={config}
-          updateConfig={updateConfig}
-          isLoading={isLoading}
-        />
-      )}
-
-      <div className='p-3 bg-muted rounded-md flex flex-col gap-1'>
+      {/* <div className='p-3 bg-muted rounded-md flex flex-col gap-1'>
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentTabs/DocumentTriggers/Settings/ScheduleTrigger/Config.tsx]: 👀

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/@modal/(.)triggers/new/_components/TriggerTypes/ScheduleTrigger/index.tsx
```diff
@@ -7,17 +7,19 @@ import {
...
-import { ScheduleTriggerForm } from '../../../../_components/TriggerForms/ScheduleTriggerForm'
 import {
   SelectDocument,
   useDocumentSelection,
 } from '../../../../_components/SelectDocument'
 import { TriggerWrapper } from '../TriggerWrapper'
+import { ScheduledTriggerConfiguration } from '@latitude-data/constants/documentTriggers'
+import { CronFormField } from '@latitude-data/web-ui/organisms/CronInput'
+
+const CLIENT_TIMEZONE = Intl.DateTimeFormat().resolvedOptions().timeZone
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/@modal/(.)triggers/new/_components/TriggerTypes/ScheduleTrigger/index.tsx]: Do a `getLocalTimezone` helper

andresgutgon (dismissed):

packages/core/src/services/documentTriggers/handlers/scheduled/registerEvents.ts
```diff
@@ -56,7 +56,7 @@ async function findScheduledTriggersDueToRun(
     const dueTriggers = triggersWithoutNextRunTime.filter((trigger) => {
       return checkCronExpression(
         trigger.configuration.cronExpression,
-        'UTC',
+        trigger.configuration.timezone,
```
geclos, line 59: rather than storing the timezone in the database the proper way to deal with this is to transform to local timezone at the edges (for ui purposes) and keep the backend utc-only

andresgutgon: But in the job, when the trigger is executed, you want to know what was user's local time is. How would you do that without storing the timezone?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/@modal/(.)triggers/new/_components/TriggerTypes/ScheduleTrigger/index.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/@modal/(.)triggers/new/_components/TriggerTypes/ScheduleTrigger/index.tsx]: moved it to a shared `CLIENT_TIMEZONE` constant instead

packages/core/src/services/documentTriggers/handlers/scheduled/registerEvents.ts
csansoon: wdym "at the edges"?

The issue is that transforming cron expressions is actually really hard... Especially with edge cases, where adding or removing hours requires changing the actual dates... This is why I decided to store the timezone along with the expressions, which I think is a much easier approach :)

andresgutgon: "at the edges," I think he means on the browser

geclos: at the edges of the app means at the last possible minute yes, date transformations

> But in the job, when the trigger is executed, you want to know what was user's local time is. How would you do that without storing the timezone?

your can store the timezone in iso format but in utc (classic 13pm +8hours)

geclos: > The issue is that transforming cron expressions is actually really hard...

yeah nevermind i thought we were working with dates but we are really working with cron expressions

PR #1707, opened 2025-09-17
chore: simplified docker compose experience/docs

geclos:


docker-compose.yml
```diff
@@ -1,226 +1,94 @@
-services:
```
andresgutgon, line 1: Version could be the same no?

PR #1713, opened 2025-09-17
feat(maintenance): Add row level sec to document and provider logs

geclos:
Implements workspace_id column to document and provider logs. Updates write services. Implements job to backfill existing data.

### DIFF
--- a/packages/core/src/jobs/job-definitions/maintenance/scheduleWorkspaceLogsMigrationJobs.ts
+++ b/packages/core/src/jobs/job-definitions/maintenance/scheduleWorkspaceLogsMigrationJobs.ts
@@ -1,0 +1,36 @@
+import { Job } from 'bullmq'
+import { database } from '../../../client'
+import { workspaces } from '../../../schema'
+import { queues } from '../../queues'
+
+export type ScheduleWorkspaceLogsMigrationJobsData = Record<string, never>
+
+/**
+ * Job that schedules per-workspace migration jobs to backfill workspaceId
+ * on document_logs and provider_logs tables for rows where it is null.
+ *
+ * This job:
+ * 1. Finds all non-deleted workspaces
+ * 2. Enqueues a migrateWorkspaceLogsJob for each workspace
+ */
+export const scheduleWorkspaceLogsMigrationJobs = async (
+  _: Job<ScheduleWorkspaceLogsMigrationJobsData>,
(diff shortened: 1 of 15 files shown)


packages/core/src/jobs/job-definitions/maintenance/scheduleWorkspaceLogsMigrationJobs.ts
```diff
@@ -0,0 +1,36 @@
...
+ * on document_logs and provider_logs tables for rows where it is null.
+ *
+ * This job:
+ * 1. Finds all non-deleted workspaces
+ * 2. Enqueues a migrateWorkspaceLogsJob for each workspace
+ */
+export const scheduleWorkspaceLogsMigrationJobs = async (
+  _: Job<ScheduleWorkspaceLogsMigrationJobsData>,
+) => {
+  const allWorkspaces = await database
```
andresgutgon, line 19: Would not makes sense to deploy this with a scoped list of workspaces before running on all the db?

packages/core/src/schema/models/documentLogs.ts
```diff
@@ -13,13 +13,15 @@ import { latitudeSchema } from '../db-schema'
...
 import { experiments } from './experiments'
+import { workspaces } from './workspaces'
 import { logSourcesEnum } from './providerLogs'
 
 export const documentLogs = latitudeSchema.table(
   'document_logs',
   {
     id: bigserial('id', { mode: 'number' }).notNull().primaryKey(),
     uuid: uuid('uuid').notNull().unique(),
+    workspaceId: bigint('workspace_id', { mode: 'number' }).references(() => workspaces.id, { onDelete: 'cascade' }),
```
andresgutgon, line 24: Would this be not null in the future no?

andresgutgon (approved):

packages/core/src/schema/models/documentLogs.ts
geclos: yes

packages/core/src/jobs/job-definitions/maintenance/scheduleWorkspaceLogsMigrationJobs.ts
geclos: nah it's fine

PR #1721, opened 2025-09-19
Add commitMerged event to merge service

geclos:
- Add commitMerged event type with userEmail and workspaceId properties
- Publish commitMerged event after successful commit merge
- Fetch user email from database using commit's userId

### DIFF
--- a/packages/core/src/services/commits/merge.ts
+++ b/packages/core/src/services/commits/merge.ts
@@ -17,6 +17,8 @@ import {
   DocumentTriggersRepository,
   EvaluationsV2Repository,
 } from '../../repositories'
+import { publisher } from '../../events/publisher'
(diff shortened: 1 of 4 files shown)


packages/core/src/services/commits/merge.ts
```diff
@@ -112,30 +114,45 @@ export async function mergeCommit(
...
+        }),
+        pingProjectUpdate(
+          {
+            projectId: commit.projectId,
+          },
+          transaction,
+        ).then((r) => r.unwrap()),
+      ])
+    },
+  )
```
andresgutgon, line 157: Test

geclos: para side effects? nah

PR #1727, opened 2025-09-22
Support for Streamable HTTP MCPs

csansoon:
SSE support is still available for server paths that end with `/sse`


packages/core/src/services/integrations/McpClient/utils.ts
```diff
@@ -51,15 +51,19 @@ export const PIPEDREAM_MCP_URL = 'https://remote.mcp.pipedream.net'
...
-      : `${urlWithProtocol}/sse`
-    return Result.ok(new URL(urlWithPath))
+    const urlObject = new URL(urlWithProtocol)
+
+    const isSSE = urlObject.pathname.endsWith('/sse')
+    if (isSSE) {
+      return Result.ok(new SSEClientTransport(urlObject))
+    }
+
+    return Result.ok(new StreamableHTTPClientTransport(urlObject))
```
csansoon, line 66: Here's the deal

packages/core/src/services/integrations/McpClient/external.ts
```diff
@@ -22,18 +23,18 @@ export async function createAndConnectExternalMcpClient(
...
-
+  const transport = transportResult.unwrap()
   const client = new McpClient({
     name: integration.name,
     version: '1.0.0',
   })
 
   const connectResult = await retryWithBackoff(async () => {
-    const transport = new SSEClientTransport(urlResult.value)
+    console.log('connecting to MCP', transport)
```
geclos, line 35: 👀

packages/core/src/services/integrations/McpClient/external.ts
```diff
@@ -14,6 +13,8 @@ import { IntegrationType } from '@latitude-data/constants'
 export async function createAndConnectExternalMcpClient(
   integration: IntegrationDto,
 ): Promise<TypedResult<McpClientConnection, McpConnectionError>> {
+  console.log('createAndConnectExternalMcpClient')
```
geclos: 👀

geclos (approved):
remove console logs and palantir

PR #1728, opened 2025-09-22
Feat/cache providers map

geclos:


packages/core/src/lib/streamManager/chainStreamManager.ts
```diff
@@ -69,6 +69,8 @@ export class ChainStreamManager extends StreamManager implements StreamManager {
         toolsBySource,
       )
 
+      console.log('config', config)
```
neoxelox: console log

packages/core/src/events/handlers/index.ts
```diff
@@ -42,7 +43,9 @@ export const EventHandlers: IEventsHandlers = {
   magicLinkTokenCreated: [sendMagicLinkJob],
   membershipCreated: [sendInvitationToUserJob],
   projectCreated: [],
-  providerApiKeyCreated: [],
+  providerApiKeyCreated: [clearProviderApiKeysCache],
```
learningbizz, line 46: Why do we care about these three events?

packages/core/src/repositories/providerApiKeysRepository.ts
```diff
@@ -19,32 +19,30 @@ import {
...
+      .$dynamic()
   }
 
   async findByName(name: string) {
-    const result = await this.db
-      .select()
-      .from(this.scope)
-      .where(eq(this.scope.name, name))
+    const result = await this.scope.where(
+      and(this.scopeFilter, eq(providerApiKeys.name, name)),
```
learningbizz, line 44: this additional this.scopeFilter is redundant, as you have it in the scope() already, no?

packages/core/src/repositories/providerApiKeysRepository.ts
```diff
@@ -56,10 +54,9 @@ export class ProviderApiKeysRepository extends RepositoryLegacy<
   }
 
   async findAllByNames(names: string[]) {
-    return await this.db
-      .select()
-      .from(this.scope)
-      .where(inArray(this.scope.name, names))
+    return await this.scope.where(
+      and(this.scopeFilter, inArray(providerApiKeys.name, names)),
```
learningbizz, line 58: Same here

packages/core/src/services/providerApiKeys/buildMap.ts
```diff
@@ -10,11 +13,19 @@ export async function buildProvidersMap(
   },
   db = database,
 ) {
-  const scope = new ProviderApiKeysRepository(workspaceId, db)
-  const result = await scope.findAll().then((r) => r.unwrap())
+  const cacheKey = `workspace:${workspaceId}:provider-api-keys-map`
```
learningbizz, line 16: nitpick: this could be a const next to the TTL

packages/core/src/repositories/providerApiKeysRepository.ts
geclos: unfortunately not because drizzle does not support query composition so whenever you have a where you have to reinclude the scope filter.

packages/core/src/lib/streamManager/chainStreamManager.ts
geclos: thx

packages/core/src/events/handlers/index.ts
geclos: we need to clear the cache on any api key write

packages/core/src/services/providerApiKeys/buildMap.ts
geclos: don't understand the comment

learningbizz: it could go outside the function as CACHE_KEY

geclos: no it depends on the workspaceId parameter

PR #1730, opened 2025-09-23
Document Integration References

csansoon:
Okay, so

We need to know what integrations are being included in each prompt as tools. This cannot be calculated on demand, since it requires to compile the prompt over and over again, which not only is expensive but also will return the same results every time while the prompt is not changed.

So, I have added a new `documentIntegrationReferences` table, which includes all integrations references as tools from any document version, and gets updated every time a document gets updated!

This is a bit hacky. Trust me, I know. But it's the cheapest way we can do that for now! The whole tool / trigger definition is going to be revised and changed eventually, so we might work with this for now.

---

Additionally, I have revamped how forking a document works.

Now, it will first go through a process to identify the elements that will need to be imported. This will return:
 - The main document
 - Documents to be imported as agents
 - Document to be imported as snippets (which will be imported without configuration)
 - Integrations (from tools from imported executable docs, or triggers from the imported doc)
 - Triggers
 
 Then, it will first clone the integrations, and build a map between the old and new to match them.

Now, when cloning documents, all prompts using integration tools will now point to the new matching cloned integration.

Finally, triggers are cloned pointing to the new integration too.


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/_components/UnconfiguredIntegrations/index.tsx
```diff
@@ -1,15 +1,69 @@
...
+        if (usedByTool) return true
+
+        return false
+      }),
+    [unconfiguredIntegrations, triggers, toolReferences],
+  )
+
+  return {
+    referencedUnconfiguredIntegrations,
+  }
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/_components/UnconfiguredIntegrations/index.tsx]: memo this

apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/integrationReferences/route.ts
```diff
@@ -0,0 +1,40 @@
...
+      {
+        params: { commitUuid, projectId: _projectId },
+        workspace,
+      }: {
+        params: { commitUuid: string; projectId: string }
+        workspace: Workspace
+      },
+    ) => {
+      const projectId = parseInt(_projectId)
+
```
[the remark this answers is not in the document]
Comment (csansoon) [reply apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/integrationReferences/route.ts:23]: why

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/_components/UnconfiguredIntegrations/index.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/_components/UnconfiguredIntegrations/index.tsx]: The whole object? `referenceUnconfiguredIntegrations` is already memoed

apps/web/src/stores/documentIntegrationReferences.ts
```diff
@@ -0,0 +1,42 @@
...
+    .detail(projectId)
+    .commits.detail(commitUuid).integrationReferences.root
+
+  const searchParams: Record<string, string> = documentUuid
+    ? { documentUuid }
+    : {}
+  const fetcher = useFetcher<DocumentIntegrationReference[]>(apiRoute, {
+    searchParams,
+  })
+
```
geclos, line 29: nitpick but you are not very consistent with your spacings, a usual thumbrule is to group var declarations expresions and return statements, and only jump line between the groups

apps/web/src/stores/documentIntegrationReferences.ts
```diff
@@ -0,0 +1,42 @@
...
+  >(
+    ['documentIntegrationReferences', projectId, commitUuid, documentUuid],
+    fetcher,
+    opts,
+  )
+
+  return {
+    data,
+    isLoading,
+  }
```
geclos: memo

packages/core/src/repositories/documentVersionsRepository/index.ts
```diff
@@ -342,6 +342,10 @@ export class DocumentVersionsRepository extends RepositoryLegacy<
     return Result.ok(totalDocuments)
   }
 
+  get queries() {
+    return {}
+  }
```
geclos: ?

packages/core/src/repositories/documentIntegrationReferencesRepository.ts
```diff
@@ -0,0 +1,238 @@
...
+      .where(
+        and(
+          eq(documentIntegrationReferences.workspaceId, this.workspaceId),
+          isNotNull(this.documentVersionsScope.mergedAt),
+        ),
+      )
+
+    return Result.ok(
+      referencesFromMergedCommits as DocumentIntegrationReference[],
+    )
```
geclos: no need to return result here

packages/core/src/repositories/documentIntegrationReferencesRepository.ts
```diff
@@ -0,0 +1,238 @@
...
+  async getActiveInCommit(
+    commit: Commit,
+  ): PromisedResult<DocumentIntegrationReference[], LatitudeError> {
+    if (!commit) return Result.ok([])
+
+    const referencesFromMergedCommitsResult =
+      await this.getReferencesFromMergedCommits({
+        projectId: commit.projectId,
+      })
+    if (!Result.isOk(referencesFromMergedCommitsResult)) {
```
geclos: fake if this method is never returning a result error

csansoon: If the method never returns a result error, it shuoldn't return a `Result` object in the first place

csansoon: oh this was me

packages/core/src/schema/models/documentIntegrationReferences.ts
```diff
@@ -0,0 +1,37 @@
...
+    projectId: bigint('project_id', { mode: 'number' })
+      .notNull()
+      .references(() => projects.id, { onDelete: 'cascade' }),
+    commitId: bigint('commit_id', { mode: 'number' }).notNull(),
+    documentUuid: uuid('document_uuid').notNull(),
+    integrationId: bigint('integration_id', { mode: 'number' })
+      .notNull()
+      .references(() => integrations.id, { onDelete: 'cascade' }),
+  },
+  (table) => ({
```
geclos: return an array instead of as hash (deprecated api)

csansoon: oh, didn't know!

packages/core/src/services/documents/updateListOfIntegrations.ts
```diff
@@ -0,0 +1,99 @@
...
+      if (!toolName?.length) return undefined // Return undefined if there is no '/' (not an integration tool)
+      return integrationName
+    })
+    .filter((name): name is string => name !== undefined) // Remove instances where there is no '/'
+    .filter((name, index, array) => array.indexOf(name) === index) // Remove duplicates
+}
+
+export async function updateListOfIntegrations(
+  {
+    workspaceId,
```
geclos: should receive a workspace instance rather than workspaceId since you are using it for tenancy check in line 77

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/_components/UnconfiguredIntegrations/index.tsx
Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/_components/UnconfiguredIntegrations/index.tsx]: yup, doesn't matter that object props are memoed, you are instantiating a new object on each return call and react uses identity equality rather than deep equality for it's rerender checks

packages/core/src/repositories/documentIntegrationReferencesRepository.ts
geclos: > If the method never returns a result error, it shuoldn't return a Result object in the first place

100%

geclos (approved):
looks good

PR #1744, opened 2025-09-29
feat: improve types and deprecate old API

geclos:
- Deprecates APIs v1 and v2
- Improves type checking performance through several optimizations
- Removes compiler packages
- Removes static type generation

### DIFF
--- a/apps/web/src/actions/documents/destroyDocumentAction/index.test.ts
+++ b/apps/web/src/actions/documents/destroyDocumentAction/index.test.ts
@@ -113,10 +113,11 @@ describe('destroyDocumentAction', async () => {
         documentUuid: document.documentUuid,
       })
       // TODO: move to core
-      const documents = await database.query.documentVersions.findMany({
-        // @ts-ignore
-        where: and(eq(documentVersions.documentUuid, document.documentUuid)),
-      })
+      const documents = await database
+        .select()
+        .from(documentVersions)
+        // @ts-expect-error - mocking
+        .where(and(eq(documentVersions.documentUuid, document.documentUuid)))
 
       const drafDocument = documents.find(
         (d: DocumentVersion) => d.commitId === draft.id,
--- a/apps/web/src/services/user/setupService.test.ts
+++ b/apps/web/src/services/user/setupService.test.ts
@@ -46,46 +46,56 @@ describe('setupService', () => {
     const { user, workspace } = result.value!
 
     // Check user creation
-    const createdUser = await database.query.users.findFirst({
-      where: utils.eq(users.id, user.id),
-    })
+    const createdUser = await database
+      .select()
+      .from(users)
+      .where(utils.eq(users.id, user.id))
+      .then((rows) => rows[0])
     expect(createdUser).toBeDefined()
     expect(createdUser?.email).toBe('test@example.com')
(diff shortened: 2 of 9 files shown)


andresgutgon (commented):
What about `packages/compiler`?

packages/core/package.json
```diff
@@ -221,16 +197,21 @@
...
     "@types/nodemailer-mailgun-transport": "^1.4.6",
     "@types/pg": "8.15.4",
-    "@types/react": "catalog:",
-    "@types/react-dom": "catalog:",
+    "@types/react": "18.3.0",
+    "@types/react-dom": "18.3.0",
     "@types/safe-regex": "^1.1.6",
     "@types/uuid": "^10.0.0",
+    "@typescript/analyze-trace": "^0.10.1",
+    "ast-types": "^0.15.2",
```
andresgutgon: Can we remove these 2? Were for tracing. Maybe document what you did so we can do it again

packages/core/package.json
```diff
@@ -221,16 +197,21 @@
...
     "drizzle-kit": "catalog:",
     "eslint": "8",
     "eslint-plugin-drizzle": "^0.2.3",
     "pg-transactional-tests": "^1.0.9",
+    "recast": "^0.23.11",
     "supertest": "^7.0.0",
     "svelte": "^4.2.19",
+    "ts-morph": "^27.0.0",
+    "tsx": "^4.20.6",
```
andresgutgon: What about these 3? Are they used?

packages/core/src/constants.ts
```diff
@@ -475,7 +469,7 @@ export type PromptSource = EvaluationV2 | DocumentRunPromptSource
 export type LatitudeToolDefinition = {
   name: LatitudeTool
   internalName: LatitudeToolInternalName
-  definition: (context?: TelemetryContext) => Tool
+  definition: (context?: any) => Tool
```
andresgutgon, line 473: Is this necessary? We still need to import telemetry in core no?

packages/core/src/data-access/documentLogs.ts
```diff
@@ -1,36 +1,9 @@
...
-import {
-  commits,
-  documentLogs,
-  documentVersions,
-  projects,
-  subscriptions,
-  workspaces,
-} from '../schema'
-
-export const findWorkspaceFromDocumentLog = async (
```
andresgutgon, line 15: Check for me: Is this used?

andresgutgon: This was moved +1

packages/core/package.json
geclos: used them for some refactor scripts, can't be removed although they are dev dependencies so no big deal

packages/core/package.json
geclos: yeah same, used them for some refactor stuff, can be removed

packages/core/src/constants.ts
geclos: better to limit telemetry import as much as possible so that next can tree shake

geclos:
> It's ok? I comment here because files section on Github do not load

it's ok yes these 2 were big culprits of performance issues in telemetry package

PR #1748, opened 2025-10-01
feat: add references to prompt spans

geclos:
In order to be able to migrate fully to spans we need to add some missing references to the model


packages/constants/src/tracing/span.ts
```diff
@@ -132,6 +132,17 @@ export type ToolSpanMetadata = BaseSpanMetadata<SpanType.Tool> & {
   }
 }
 
+export type PromptSpanMetadata<T extends SpanType = SpanType> =
```
neoxelox: thats not the correct type, fijate en las otras metadatas

packages/core/src/schema/models/spans.ts
```diff
@@ -34,30 +34,28 @@ export const spans = latitudeSchema.table(
     duration: bigint('duration', { mode: 'number' }).notNull(),
     startedAt: timestamp('started_at').notNull(),
     endedAt: timestamp('ended_at').notNull(),
+    documentUuid: uuid('document_uuid'),
+    commitUuid: uuid('uuid'),
```
neoxelox: mmm why is this named uuid? i think this is a typo!

packages/core/drizzle/0215_flippant_iron_monger.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE "latitude"."spans" ADD COLUMN "document_uuid" uuid;--> statement-breakpoint
+ALTER TABLE "latitude"."spans" ADD COLUMN "uuid" uuid;--> statement-breakpoint
+ALTER TABLE "latitude"."spans" ADD COLUMN "experiment_id" bigint;--> statement-breakpoint
+CREATE INDEX "spans_document_uuid_idx" ON "latitude"."spans" USING btree ("document_uuid");--> statement-breakpoint
```
neoxelox, line 4: please CONCURRENTLY

packages/core/drizzle/0215_flippant_iron_monger.sql
```diff
@@ -0,0 +1,6 @@
...
+CREATE INDEX "spans_commit_uuid_idx" ON "latitude"."spans" USING btree ("uuid");--> statement-breakpoint
```
neoxelox: typo in the commit uuid column name

packages/core/src/schema/models/spans.ts
geclos: typo? why? I want to use the commit uuid rather than the commit id (that duality should be eventually removed if possible)

PR #1749, opened 2025-10-01
chore: hydrate prompt spans with old data

geclos:
Needed for the migration from document logs to spans


packages/core/src/jobs/job-definitions/maintenance/migrateSpansJobs.ts
```diff
@@ -0,0 +1,25 @@
...
+import { workspaces } from '../../../schema'
+import { queues } from '../../queues'
+
+export const migrateSpansJobs = async () => {
+  // Find all workspaces with free plans
+  const freeWorkspaces = await database
+    .select({
+      id: workspaces.id,
+    })
+    .from(workspaces)
```
andresgutgon, line 10: Is this ok?

geclos: yeah the comment is wrong (copy pasted)

PR #1751, opened 2025-10-01
added "isLive" prop to DocumentRunRequestedEvent

csansoon:
added a `isLiveCommit` prop to track whether the requested commit is the current Live one or not for every new `documentRunRequested` request

### DIFF
--- a/apps/web/src/app/api/documents/[documentUuid]/run/route.ts
+++ b/apps/web/src/app/api/documents/[documentUuid]/run/route.ts
@@ -64,12 +64,18 @@ export const POST = errorHandler(
           'runs',
         ).then((r) => r.unwrap())
 
+        const commitsScope = new CommitsRepository(workspace.id)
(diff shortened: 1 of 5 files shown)


apps/web/src/app/api/documents/[documentUuid]/run/route.ts
```diff
@@ -64,12 +64,18 @@ export const POST = errorHandler(
           'runs',
         ).then((r) => r.unwrap())
 
+        const commitsScope = new CommitsRepository(workspace.id)
+        const headCommit = await commitsScope
+          .getHeadCommit(projectId)
+          .then((r) => r.unwrap())
```
Comment (andresgutgon) [inline apps/web/src/app/api/documents/[documentUuid]/run/route.ts:70]: You could move this to the repo

```javascript
await new CommitsRepository(workspace.id).isLive({ projectId, commitUuid })
```
Looks pretty handy and you're already duplicating the same logic

PR #1752, opened 2025-10-02
fix: commit repository method wrong types

geclos:


packages/core/src/repositories/commitsRepository/utils/getHeadCommit.ts
```diff
@@ -27,5 +27,5 @@ export async function getHeadCommitForProject(
     .orderBy(desc(commitsScope.mergedAt))
     .limit(1)
 
-  return Result.ok(result[0])
+  return result[0] as Commit | undefined
```
geclos, line 30: this can clearly be undefined yet types always asserted commit existed, for some reason

packages/core/src/repositories/commitsRepository/utils/getHeadCommit.ts
```diff
@@ -27,5 +27,5 @@ export async function getHeadCommitForProject(
     .orderBy(desc(commitsScope.mergedAt))
     .limit(1)
 
-  return Result.ok(result[0])
```
geclos, line 30: do not return result objects in read methods

packages/core/src/repositories/commitsRepository/utils/getHeadCommit.ts
andresgutgon: Just don't do the `as` no? It will give drizzle schema or `undefined`

geclos: for some reason it doesn't, not sure what is happening here

PR #1755, opened 2025-10-02
Feature/new onboarding step 3 and 4

learningbizz:
This PR includes the last two steps of the new onboarding: Trigger agent and Run Agent.

image

In addition to this, there are a few improvements
- The first step of the onboarding is dynamically calculated in the backend, depending on the pending integrations and triggers you have after cloning the agent. Like this, you don't land on the setupIntegration step if no integrations need configuring
- Move next step is now dynamically calculated based on what pending integrations and triggers you have. Like this, if there is no configurable triggers, you can jump smoothly from step 1 to step 3

### DIFF
--- a/apps/web/src/app/(onboarding)/onboarding/_components/OnboardingClient/triggerAgent/_components/RunTrigger.tsx
+++ b/apps/web/src/app/(onboarding)/onboarding/_components/OnboardingClient/triggerAgent/_components/RunTrigger.tsx
@@ -1,0 +1,177 @@
+import { DocumentTriggerType } from '@latitude-data/constants'
+import {
+  DocumentTrigger,
+  DocumentVersion,
+  IntegrationDto,
+} from '@latitude-data/core/browser'
+import { Button } from '@latitude-data/web-ui/atoms/Button'
+import { Text } from '@latitude-data/web-ui/atoms/Text'
+import { cn } from '@latitude-data/web-ui/utils'
+import { useCallback, useMemo } from 'react'
+import { TriggerEventsList } from '$/app/(private)/projects/[projectId]/versions/[commitUuid]/preview/_components/TriggerEventsList'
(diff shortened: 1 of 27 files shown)


apps/web/src/app/(onboarding)/onboarding/_components/OnboardingClient/triggerAgent/_components/RunTrigger.tsx
```diff
@@ -0,0 +1,177 @@
...
+function isIntegrationTrigger(
+  trigger: DocumentTrigger,
+): trigger is DocumentTrigger<DocumentTriggerType.Integration> {
+  return trigger.triggerType === DocumentTriggerType.Integration
+}
+
+const RUNNABLE_TRIGGERS = [
+  DocumentTriggerType.Scheduled,
+  DocumentTriggerType.Chat,
+]
```
csansoon: We're using a similar constant somewhere else, move it to a shared constant instead

apps/web/src/app/api/documents/[documentUuid]/run/route.ts
```diff
@@ -267,9 +267,18 @@ async function generateAIParameters({
...
+      env.COPILOT_WORKSPACE_ID!,
+    )
+
+    const sdk = await createSdk({
+      workspace: copilotWorkspace,
+      apiKey: env.COPILOT_WORKSPACE_API_KEY!,
       projectId: env.COPILOT_PROJECT_ID,
-    })
+      __internal: { source: LogSources.Copilot },
+    }).then((r) => r.unwrap())
```
Comment (csansoon) [inline apps/web/src/app/api/documents/[documentUuid]/run/route.ts:275]: Why are we now using the the workspace's id instead of its api key?

Comment (learningbizz) [reply apps/web/src/app/api/documents/[documentUuid]/run/route.ts:275]: We have to look at the createSdk, as the workspace is not being used for anything but getting the API key if it doesnt exist. Looking at it for the future

PR #1766, opened 2025-10-05
Vercel sdk v5 maybe this time

andresgutgon:
# What?
This PR introduce Vercel SDK v5 which force us also to upgrade Zod from
v3 to v4 and also to change our server action lib because is not
compatible with Zod 4.

## Fix unhandle Vercel SDK error NoOutputGeneratedError
The way we handle this is by listening to `onError` callback on
`streamText` and throw the error before we start reading from the
streamText promises which will fail anyway because when Vercel gets
provider API error they return the NoOutputGeneratedError content which
is not handle in our side. Throwing on their onError callback we avoid
this problem


packages/core/package.json
```diff
@@ -212,5 +212,8 @@
     "vitest": "^3.1.4",
     "vue": "^3.4.38",
     "yaml": "^2.4.5"
+  },
+  "peerDependencies": {
+    "dd-trace": "^5.69.0"
```
andresgutgon, line 217: Is this ok @geclos? This wasn't here when I did the PR, I think

packages/core/src/services/ai/getLanguageModel.ts
```diff
@@ -45,11 +45,15 @@ export function getLanguageModel({
...
 
   const openAiProvider = llmProvider as OpenAIProvider
+
   const configuration =
     provider.configuration as ProviderConfiguration<Providers.OpenAI>
-  const usingOpenAIResponses = configuration?.endpoint === 'responses'
+  const usingChatCompletions = configuration?.endpoint === 'chat_completions'
 
-  if (usingOpenAIResponses) return openAiProvider.responses(model)
+  if (usingChatCompletions) return openAiProvider.chat(model)
```
andresgutgon, line 55: Check if config is undefined and return chat because this means is an old provider before responses

packages/core/src/services/ai/index.ts
```diff
@@ -146,13 +155,29 @@ export async function ai({
...
+        if (!span) {
+          console.error(error)
+          return
+        }
+
+        span.log({
+          event: '[Latitude]: AI Provider Call Error',
+          message: (error as Error)?.message,
+          error,
+        })
```
andresgutgon: I think we can remove these traces @geclos this is a handle error. What I saw is if we throw this error before reading the failed promise reponses we avoid the error

packages/core/src/services/chains/ProviderProcessor/index.ts
```diff
@@ -25,46 +31,31 @@ export async function processResponse({
...
+    text = await aiResult.text
+    response = await aiResult.response
+    messages = response.messages
+    output = convertResponseMessages({ messages })
+    usage = await aiResult.usage
+    toolCalls = await aiResult.toolCalls
+    reasoning = await aiResult.reasoning
+  } catch (_) {
+    // do nothing
+  }
```
andresgutgon: Remove this try/catch and see if the fix works

packages/core/package.json
geclos: yeah it's ok

packages/core/src/services/ai/index.ts
geclos: let's talk it tomorrow offline

PR #1771, opened 2025-10-06
project main prompt

csansoon:
Added the concept of "Main Prompt"

image

All projects can have a single "Main Prompt".
This PR adds the logic and UI to configure what this main prompt is:
 - Can be identified with an icon next to the file
 - Can be manually set by clicking the icon
 - Is automatically set when creating a file on a project without main prompt
 - Is automatically set when forking a document

This feature is not being used yet, so the UI is currently disabled to avoid confusion


apps/web/src/actions/commits/setCommitMainDocumentAction.ts
```diff
@@ -0,0 +1,49 @@
...
+    const { commitId, documentUuid } = parsedInput
+
+    const commitScope = new CommitsRepository(workspace.id)
+    const commit = await commitScope
+      .getCommitById(commitId)
+      .then((r) => r.unwrap())
+
+    const data: { mainDocumentUuid: string | null } = { mainDocumentUuid: null }
+
+    if (documentUuid) {
```
andresgutgon, line 29: So you set null when you want to remove it?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/providers/ClientCommitProvider.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const { data: commits } = useCommitsFromProject(project.id)
+
+  const currentCommit = useMemo(() => {
+    return commits?.find((c) => c.uuid === commit.uuid) ?? commit
+  }, [commits, commit])
+
+  return (
+    <CommitProvider commit={currentCommit} isHead={isHead}>
+      {children}
+    </CommitProvider>
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/providers/ClientCommitProvider.tsx]: Move to apps/web this provider and only have one

apps/web/src/actions/commits/setCommitMainDocumentAction.ts
csansoon: yes! :D

packages/core/src/schema/models/commits.ts
```diff
@@ -29,29 +29,25 @@ export const commits = latitudeSchema.table(
...
-    deletedAtIdx: index('commits_deleted_at_indx').on(table.deletedAt),
-    projectDeletedAtIdx: index('commits_project_deleted_at_idx').on(
+  (table) => [
+    index('project_commit_order_idx').on(table.mergedAt, table.projectId),
+    uniqueIndex('unique_commit_version').on(table.version, table.projectId),
+    index('user_idx').on(table.userId),
+    index('merged_at_idx').on(table.mergedAt),
+    index('project_id_idx').on(table.projectId),
+    index('commits_deleted_at_indx').on(table.deletedAt),
+    index('commits_project_deleted_at_idx').on(
```
andresgutgon, line 47: What? Why you changed to an array?

andresgutgon (dismissed):
Solve the question and 👍

packages/core/src/schema/models/commits.ts
csansoon: The last way is deprecated
This is how Drizzle shows how to do it now
https://orm.drizzle.team/docs/indexes-constraints#indexes

PR #1778, opened 2025-10-08
Use redis streams with runs

neoxelox:


apps/web/package.json
```diff
@@ -25,7 +25,7 @@
     "@latitude-data/constants": "workspace:*",
     "@latitude-data/core": "workspace:*",
     "@latitude-data/env": "workspace:^",
-    "@latitude-data/sdk": "5.2.0-beta.5",
+    "@latitude-data/sdk": "workspace:^",
```
andresgutgon, line 28: Nice

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunPanel/index.tsx
```diff
@@ -134,6 +141,28 @@ function CompletedRunPanel({
     )
   }, [results, run.log.uuid, manualEvaluations])
 
+  const syncAnnotations: typeof mutateResults = useCallback(
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunPanel/index.tsx:144]: Be careful with abusing `typeof` can have performance penalties for the TS server I think

apps/web/src/app/api/documents/[documentUuid]/run/route.ts
```diff
@@ -230,6 +151,9 @@ export const POST = errorHandler(
           },
         })
       } catch (error) {
+        // When client closes the connection, the SDK will throw an undefined
+        // error or AbortError which we can safely ignore
+        if (!error || (error as Error).name === 'AbortError') return
```
Comment (andresgutgon) [inline apps/web/src/app/api/documents/[documentUuid]/run/route.ts:156]: Better `instanceOf` if possible an Abort error is an specific error class

apps/web/src/hooks/useLatitudeAction.ts
```diff
@@ -93,14 +93,52 @@ export type ExecuteFn<S extends ActionSchema, Data> = (
   | readonly [null, ActionErrorKlass<S> | ServerErrorKlass]
 >
 
+const capitalize = (str: string) => {
+  return str
+    .replace(/([A-Z])/g, ' $1')
+    .toLowerCase()
+    .replace(/^./, (match) => match.toUpperCase())
+    .trim()
+}
```
andresgutgon, line 102: Bro https://lodash.com/docs/#capitalize

apps/web/src/hooks/useLatitudeAction.ts
```diff
@@ -93,14 +93,52 @@ export type ExecuteFn<S extends ActionSchema, Data> = (
...
+
 export function buildActionError<S extends StandardSchemaV1>({
   validationErrors,
   serverError,
 }: {
   validationErrors?: ActionError<S>
   serverError?: unknown
 }) {
-  if (validationErrors) return new ActionErrorKlass<S>(validationErrors)
```
andresgutgon, line 103: Frontend should not be concerned with the shape of the errors. Should receive errors in the right format. 

DON'T DO THIS

packages/core/src/lib/streamManager/ChainStreamConsumer/consumeStream.ts
```diff
@@ -35,6 +35,8 @@ export async function consumeStream({
   const reader = result.fullStream.getReader()
 
   while (true) {
+    if (error) break
```
andresgutgon, line 38: ⚠️ Why are you doing this?

packages/core/src/lib/streamManager/step/streamAIResponse.ts
```diff
@@ -115,9 +115,16 @@ export async function streamAIResponse({
     documentLogUuid,
   })
 
+  let finishReason
+  try {
+    finishReason = await aiResult.finishReason
+  } catch (_) {
+    // do nothing
+  }
```
andresgutgon, line 123: 🔥 Should not be needed

packages/core/src/lib/streamManager/step/streamAIResponse.ts
```diff
@@ -140,6 +147,6 @@ export async function streamAIResponse({
     // FIXME: Make response.output non optional when we remove `__deprecated`
     messages: response.output ?? [],
     tokenUsage: response.usage,
-    finishReason: await aiResult.finishReason,
```
andresgutgon, line 143: Leave like before

packages/core/src/lib/vercelSdkFromV5ToV4/index.ts
```diff
@@ -11,39 +11,37 @@ type LegacyToolContent = Array<
...
 > {
-  const calls = await toolCalls
-  return calls.map((t) => ({
+  return toolCalls?.map((t) => ({
     id: t.toolCallId,
     name: t.toolName,
     // Vercel SDK v4 -> v5 changed the name from `arguments` to `input`
     arguments: t.input as Record<string, unknown>,
-  }))
+  })) || []
```
andresgutgon: I think you didn't rebased with latest changes

packages/core/src/services/chains/ProviderProcessor/index.ts
```diff
@@ -31,21 +31,29 @@ export async function processResponse({
...
+    response = await aiResult.response
+    text = await aiResult.text
+    toolCalls = await aiResult.toolCalls
+    usage = await aiResult.usage
+
+    messages = response.messages
+    output = convertResponseMessages({ messages: response.messages })
+  } catch (_) {
+    // do nothing
+  }
```
andresgutgon, line 44: 🔥 Old

packages/core/src/services/chains/ProviderProcessor/index.ts
```diff
@@ -31,21 +31,29 @@ export async function processResponse({
...
     streamType: aiResult.type,
     documentLogUuid,
     text: text ?? '',
     object: isObject ? parseObject(text ?? '') : undefined,
     output,
-    usage: await vercelSdkFromV5ToV4.convertTokenUsage(aiResult.usage),
+    usage: await vercelSdkFromV5ToV4.convertTokenUsage(usage),
     reasoning,
-    toolCalls: await vercelSdkFromV5ToV4.convertToolCalls(aiResult.toolCalls),
+    toolCalls: await vercelSdkFromV5ToV4.convertToolCalls(toolCalls),
```
andresgutgon, line 54: Old

apps/web/src/hooks/useLatitudeAction.ts
neoxelox: i think this is better than having to rewrite all of the zod schemas in order to say "Field is required" imo

packages/core/src/lib/streamManager/step/streamAIResponse.ts
csansoon: yeah we want to return promises, not await for them before returning

csansoon: right? xd

apps/web/src/hooks/useLatitudeAction.ts
neoxelox: well its not only capitalizing, its humanizing camel case fields. I will change the name

apps/web/src/hooks/useLatitudeAction.ts
andresgutgon: I think is not. We rely on knowledge of zod internals. The other way we use zod as is suppose to be used

packages/core/src/lib/streamManager/step/streamAIResponse.ts
andresgutgon: At this point we were always waiting. This is after the streaming finishes.

apps/web/src/hooks/useLatitudeAction.ts
neoxelox: no time to rewrite every damn schema, maybe if the backend returned a error code it would be better but it does not it seems.... so i think this is good for now

packages/core/src/services/chains/ProviderProcessor/index.ts
geclos: no, it's fine

packages/core/src/services/chains/ProviderProcessor/index.ts
geclos: it's fine

packages/core/src/lib/streamManager/step/streamAIResponse.ts
geclos: nope, we want it like this we do not want finish reason to throw, (nor any other promise for that matter)

packages/core/src/lib/streamManager/ChainStreamConsumer/consumeStream.ts
geclos: because we immediately raise if an error is found so there's no point in consuming the rest of the stream

packages/core/src/lib/streamManager/step/streamAIResponse.ts
geclos: it is, if the stream end with no content it raises, we do not want to raise

PR #1783, opened 2025-10-09
New Agent Page: Part 1

csansoon:
Part 1:

Both the "main prompt" button and the new "agent" page are feature flagged behind "newAgentPage"

Missing:
 - Triggers list
 - Improve chat UI


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/agent/_components/MainSection/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+import { useCurrentProject } from '$/app/providers/ProjectProvider'
+import { Text } from '@latitude-data/web-ui/atoms/Text'
+import { ProjectIcon } from './ProjectIcon'
+import { useMemo } from 'react'
+import { extractLeadingEmoji } from '@latitude-data/web-ui/textUtils'
+import { useCurrentCommit } from '$/app/providers/CommitProvider'
+import { Tooltip } from '@latitude-data/web-ui/atoms/Tooltip'
+import Image from 'next/image'
+import MainPromptImage from './mainPrompt.png'
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/agent/_components/MainSection/index.tsx]: so now it's not failing?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/agent/_components/MainSection/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+            className='rounded-xl'
+            trigger={
+              <div className='border-border border-b-2 border-dotted cursor-text'>
+                <Text.H5 color='foregroundMuted'>main prompt</Text.H5>
+              </div>
+            }
+          >
+            <div className='flex flex-col gap-2 py-1.5'>
+              <Image
+                src={MainPromptImage}
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/agent/_components/MainSection/index.tsx]: I don't understand how this will work in production usually the way to do assets in next is you put them in public folder and just write the correct route to them here

apps/web/src/types/images.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
+declare module '*.webp' {
+  const src: string
+  export default src
+}
+
+declare module '*.svg' {
+  const src: string
+  export default src
+}
```
geclos, line 29: hmm ok i see the hack but I'm not sure this will work in production make sure the turbopack is actually bundling your image as you expect — I still think it's much simpler to do images as recommended in the [next docs](https://nextjs.org/docs/pages/api-reference/components/image)

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/agent/_components/MainSection/index.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/agent/_components/MainSection/index.tsx]: I added an `images.d.ts` file for declaring images as regular modules

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/agent/_components/MainSection/index.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/agent/_components/MainSection/index.tsx]: This is a NextJS static import

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/agent/_components/MainSection/index.tsx]: https://nextjs.org/docs/pages/api-reference/file-conventions/public-folder

PR #1787, opened 2025-10-10
Handle integrations toggle on editor sidebar

andresgutgon:
# What?
Manage state of what integrations, custom tools and sub-agents the document has enabled with the editor sidebar.

### TODO
- [x] Handle active integrations on a zustand state 
- [x] When changing tools from old toolbar or editing the prompt update editor sidebar
- [x] List used integrations on the prompt
- [x] Avoid re-computing the sidebar when a change on the prompt comes from the sidebar toggle
- [x] Toggle all button on active tool-set
- [x] Keep the tool set visible when unchecked all
- [x] Add to active tool-set from connect modal
- [x] 🔥 tool-set option remove
- [x] 🐛 Annoying visual bug while loading tools on modal
- [x] Search typeahead with more than > 10 tools

## NEXT
- [ ] Add sub-agents
- [ ] Add custom-tools
- [ ] Create sub-agents typeahead popover
- [ ] Enable agent behaviour for everyone
- [ ] Enable feature flag

### DIFF
--- a/apps/web/src/hooks/useFetcher.ts
+++ b/apps/web/src/hooks/useFetcher.ts
@@ -93,6 +93,9 @@ export async function executeFetch<
   Raw extends boolean = false,
 >({
   route,
+  method = 'GET',
+  abortSignal,
+  body,
   searchParams,
   toast,
   serializer,
@@ -102,6 +105,9 @@ export async function executeFetch<
   onFail,
 }: {
   route: string
+  method?: 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH'
+  body?: Record<string, unknown> | null
+  abortSignal?: AbortSignal
   searchParams?: ISearchParams
   toast?: ReturnType<typeof useToast>['toast']
   navigate?: ReturnType<typeof useNavigate>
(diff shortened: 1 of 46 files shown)


packages/core/src/services/documents/updateDocumentContent/updateContent.ts
```diff
@@ -0,0 +1,41 @@
...
+import { validateDocumentMetadata } from './validateDocumentMetadata'
+
+type ValidatedMetadata = Awaited<ReturnType<typeof validateDocumentMetadata>>
+
+export type UpdateDocumentContentResponse = {
+  document: DocumentVersion
+  metadata: ValidatedMetadata
+}
+
+export async function updateDocumentContent({
```
andresgutgon, line 13: This deserve tests but I already exhausted my tests credits in this PR

apps/web/src/hooks/useLatitudeAction.ts
```diff
@@ -228,9 +231,13 @@ export default function useLatitudeAction<
         ] as const
       }
 
+      // Manually trigger success callback since executeAsync doesn't trigger it
+      if (successCb && result.data !== undefined) {
+        successCb({ data: result.data })
+      }
```
andresgutgon, line 240: Cursor caught a bug here. Holy shit

apps/web/src/hooks/useFetcher.ts
andresgutgon, line 1: @geclos check these changes when you have time. I added support for other than GET method here and also signal and body suport. Let me know if you see it ok

PR #1792, opened 2025-10-13
Implement sub-agents and custom tools on sidebar editor

andresgutgon:
# What?
Finish implementing editor sidebar by adding sub-agents and custom tools and removing old tool bar if feature flag for sidebar is enabled

## TODO
- [x] Add sub-agents
- [x] Create sub-agents typeahead popover
- [x] Add custom-tools
- [x] Enable feature flag

## Next
- [ ] Enable agent behaviour for everyone
- [ ] Remove all toolbar code
- [ ] Remove feature flag


packages/web-ui/src/ds/molecules/MultiSelectInput/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+            size={size}
+            trigger={trigger}
+          />
+        </div>
+      </FormField>
+    )
+  },
+)
+
+MultiSelectInput.displayName = 'MultiSelectInput'
```
andresgutgon, line 124: Old multi select was tied to forms. So now old multi select is MultiSelectInput and new MultiSelect can be used out of forms

PR #1793, opened 2025-10-13
Agents Page: Part 3 — Return of the Agents

csansoon:
Third and last part.

This PR removes the `preview` page completely from the face of the Earth, replacing it with the brand new `agent` page.
It also removes any existing Chat Trigger, and the concept itself from the codebase.

- Replaced every /preview route to /agents
- Removed every component related to /preview
- Deleted all Chat triggers!
- Removed the "Chat" trigger type and all related logic
- Removed "runAgentPage" feature flag


apps/web/src/app/(onboarding)/onboarding-agents/start/_components/OnboardingClient/TriggerAgent/index.tsx
```diff
@@ -108,28 +86,12 @@ export function TriggerAgentBody({
...
-              project={project}
-              document={activeChatTrigger.active.document}
-              chatTrigger={activeChatTrigger.active.trigger}
-              chatFocused={openChatInput}
-              onRunTrigger={onRunTrigger}
-              options={activeChatTrigger.options}
-              onChange={activeChatTrigger.onChange}
-            />
-          </div>
-        ) : null}
```
andresgutgon, line 132: CHeck this 🔴

packages/core/drizzle/0218_remove_chat_trigger.sql
```diff
@@ -0,0 +1,12 @@
+-- Remove chat trigger type
+ALTER TABLE "latitude"."document_trigger_events" ALTER COLUMN "trigger_type" SET DATA TYPE text;--> statement-breakpoint
+ALTER TABLE "latitude"."document_triggers" ALTER COLUMN "trigger_type" SET DATA TYPE text;--> statement-breakpoint
+DROP TYPE "latitude"."document_trigger_types";--> statement-breakpoint
```
andresgutgon, line 4: Here you're dropping all the type. Is that ok?

csansoon: In order to remove an element from the enum, Drizzle decided to:
- change the column type to just text
- delete the enum
- recreate the enum with fewer elements
- change the column type back to the enum
 🤷

PR #1794, opened 2025-10-13
Remove old agents tool bar and make all prompts agents

andresgutgon:
# What?
This PR 🔥 old code once QA is done in new editor sidebar and change behaviour on our system to act as agents always. 

## TODO
- [x] Remove all toolbar code
- [x] Remove feature flag
- [x] Enable agent behaviour for everyone
- [x] documentType `prompt` and `agent` review what happens.

## New behaviour
If the document doesn't have front matter is a `document` (new type). Old `prompt`s and old `agent`s are all agents.


packages/core/drizzle/0219_change_all_document_to_agents.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE "latitude"."document_versions" ALTER COLUMN "document_type" SET DEFAULT 'agent';
+--> statement-breakpoint
+
+UPDATE "latitude"."document_versions" SET "document_type" = 'agent';
```
andresgutgon, line 3: Changing all the documents to `agent`

packages/core/drizzle/0218_change_all_document_to_agents.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE "latitude"."document_versions" ALTER COLUMN "document_type" SET DEFAULT 'agent';
```
andresgutgon: Setting default as `agent` from now on

packages/core/drizzle/0219_change_all_document_to_agents.sql
geclos: why not remove the document_type column?

andresgutgon: A document without front matter is still possible. In that case is not an agent

PR #1798, opened 2025-10-14
Fix job to insert new workspaceOnboardings

learningbizz:
The last maintenance job we created didn't do what it was supposed to. 

We were updating existing rows when the job was to create new rows for those workspaces that didn't have any. 

This PR fixes that

### DIFF
--- a/packages/core/src/jobs/job-definitions/maintenance/fillOnboardingsAsCompletedJob.ts
+++ b/packages/core/src/jobs/job-definitions/maintenance/fillOnboardingsAsCompletedJob.ts
@@ -24,7 +24,7 @@ export const fillOnboardingsAsCompletedJob = async (
       ),
     )
 
-  const completedOnboardingsCreated = await database
+  await database
     .insert(workspaceOnboarding)
     .values(
       onboardingsToInsert.map(({ workspaceId }) => ({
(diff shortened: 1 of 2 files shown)


packages/core/src/jobs/job-definitions/maintenance/fillOnboardingsAsCompletedJob.ts
```diff
@@ -2,23 +2,42 @@ import { Job } from 'bullmq'
...
     })
-    .where(isNull(workspaceOnboarding.completedAt))
-    .returning()
 
   return Result.ok({
-    success: true,
-    updatedOnboardings: updatedOnboardings.length,
+    success: completedOnboardingsCreated.length === onboardingsToInsert.length,
+    completedOnboardings: completedOnboardingsCreated.length,
   })
```
geclos: do not return please no one listens to job results

learningbizz: We should return nil then?

learningbizz: Will change to nil then

PR #1805, opened 2025-10-14
Allow to publish when only change was changing the main prompt

csansoon:


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/PublishDraftCommitModal/index.tsx
```diff
@@ -93,6 +94,8 @@ export default function PublishDraftCommitModal({
   const somethingSelected = !!selectedChange
   const hasErrors = (!isLoading && changes.hasIssues) || !anyChanges
 
+  console.log('changes', changes)
```
Comment (learningbizz) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/PublishDraftCommitModal/index.tsx]: 👀

packages/core/src/services/commits/merge.ts
```diff
@@ -86,7 +86,9 @@ export async function mergeCommit(
     const totalChanges =
       evaluationChanges.length + triggerChanges.length + documentChanges.length
 
-    if (totalChanges === 0) {
+    const mainDocumentChanged = recomputedResults.unwrap().mainDocumentChanged
```
learningbizz: nitpick: bit ugly to chain the unwrap and then get the mainDocumentChanged

apps/web/src/stores/commitChanges.ts
```diff
@@ -7,6 +7,8 @@ import { Commit } from '@latitude-data/core/schema/types'
 const NO_CHANGES = {
   anyChanges: false,
   hasIssues: false,
+  mainDocumentUuid: null,
```
learningbizz: Maybe I'm not understanding well, but why is the mainDocumentUuid here null when there is no changes?

learningbizz (dismissed):
Remove console log

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/PublishDraftCommitModal/index.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/Sidebar/PublishDraftCommitModal/index.tsx]: whoopsies

packages/core/src/services/commits/merge.ts
csansoon: yeah you're right i hate it

csansoon: fixed xd

apps/web/src/stores/commitChanges.ts
csansoon: yeah i'll make it more obvious

PR #1808, opened 2025-10-15
fix(core): clean active run if no job is found

geclos:
The active runs and job state can get desynced in unexpected conditions. The stop service should clean up the state in that case.

Added tests.

### DIFF
--- a/packages/core/src/services/runs/stop.ts
+++ b/packages/core/src/services/runs/stop.ts
@@ -1,13 +1,16 @@
 import { Run } from '../../constants'
 import { publisher } from '../../events/publisher'
 import { queues } from '../../jobs/queues'
-import { NotFoundError, UnprocessableEntityError } from '../../lib/errors'
+import { UnprocessableEntityError } from '../../lib/errors'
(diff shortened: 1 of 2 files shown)


packages/core/src/services/runs/stop.ts
```diff
@@ -20,13 +23,24 @@ export async function stopRun({
   const { runsQueue } = await queues()
   const job = await runsQueue.getJob(run.uuid)
   if (!job?.id) {
-    return Result.error(
-      new NotFoundError(`Active run job with uuid ${run.uuid} not found`),
-    )
+    const runsRepo = new RunsRepository(workspace.id, project.id)
+    const staleRun = await runsRepo.get({ runUuid: run.uuid }).then((r) => r.value) // prettier-ignore
```
neoxelox: why you are getting the run from the repo again? stopRun service is called with the run object itself, which implies the run repo has already been called

geclos: hmm pues tiene razon el chico

Issue #1092, opened 2025-04-04
Issue migrating for new deployment


webbugt:
Hello, I've tried starting Latitude locally and I get this error:
`pnpm drizzle-kit migrate --config=./drizzle.config.ts
drizzle-kit: v0.22.8
drizzle-orm: v0.33.0

Custom config path was provided, using './drizzle.config.ts'
Reading config file '/home/idv/latitude-llm/packages/core/drizzle.config.ts'
Using 'pg' driver for database querying
[⢿] applying migrations...error: relation "subscriptions" does not exist
    at /home/idv/latitude-llm/node_modules/.pnpm/drizzle-kit@0.22.8/node_modules/drizzle-kit/bin.cjs:77696:15
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async <anonymous> (/home/idv/latitude-llm/node_modules/.pnpm/drizzle-orm@0.33.0_@opentelemetry+api@1.9.0_@types+pg@8.11.11_@types+react@19.0.0_pg@8.14.1_react@19.1.0/node_modules/src/pg-core/dialect.ts:89:7)
    at async NodePgSession.transaction (/home/idv/latitude-llm/node_modules/.pnpm/drizzle-orm@0.33.0_@opentelemetry+api@1.9.0_@types+pg@8.11.11_@types+react@19.0.0_pg@8.14.1_react@19.1.0/node_modules/src/node-postgres/session.ts:155:19)
    at async PgDialect.migrate (/home/idv/latitude-llm/node_modules/.pnpm/drizzle-orm@0.33.0_@opentelemetry+api@1.9.0_@types+pg@8.11.11_@types+react@19.0.0_pg@8.14.1_react@19.1.0/node_modules/src/pg-core/dialect.ts:82:3)
    at async migrate (/home/idv/latitude-llm/node_modules/.pnpm/drizzle-orm@0.33.0_@opentelemetry+api@1.9.0_@types+pg@8.11.11_@types+react@19.0.0_pg@8.14.1_react@19.1.0/node_modules/src/node-postgres/migrator.ts:10:2) {
  length: 113,
  severity: 'ERROR',
  code: '42P01',
  detail: undefined,
  hint: undefined,
  position: '278',
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'parse_relation.c',
[...]

PR #1812, opened 2025-10-15
fix: qualify table names in SQL migration for clarity

felpasl:
same error of #1092


geclos:
there are more than 200 migration files and you will probably encounter this issue in all subsequent migrations. Instead I would suggest changing the default search path of the latitude database to the latitude schema so that migrations work out of the box. Not sure why in some cases drizzle is able to configure that for the migration and, in some other cases, it's not.

[Loose pointer](https://www.perplexity.ai/search/set-default-database-schema-po-K6eAJiiwTmOmniPwcn_weQ#0)

PR #1815, opened 2025-10-15
Fix race condition adding background run after adding the job to Bullmq

andresgutgon:
# What?
A background run job finished, but the UI didn't noticed and tried to attach to it.

## What was the problem?
I conducted a stress test by running multiple background tasks simultaneously. Doing that, I saw that in the enqueue phase, we enqueue first in BullMQ, and later we add it to active runs. 

Also I added backoff retry on the lock mechanism so we have more contention in case of many concurrent writes. But the main fix is the race condition enqueuing first and not adding first to active runs

### DIFF
--- a/packages/core/src/cache/index.ts
+++ b/packages/core/src/cache/index.ts
@@ -74,7 +74,7 @@ export const getOrSet = async <T>(
 export const withCacheLock = async <T>({
   lockKey,
   callbackFn,
-  timeout = 5000,
+  timeout = 1000,
   maxRetries = 5,
 }: {
   lockKey: string


packages/core/src/services/runs/attach.ts
```diff
@@ -40,13 +39,12 @@ export async function attachRun({
...
+  // We still try to read from Redis stream as it has a grace period before expiring
+  // This fixes the race condition where UI tries to attach after job completion
   const { runsQueue } = await queues()
-  const job = await runsQueue.getJob(run.uuid)
-  if (!job?.id) {
-    return Result.error(
-      new NotFoundError(`Active run job with uuid ${run.uuid} not found`),
-    )
-  }
+  const _job = await runsQueue.getJob(run.uuid) // Intentionally unused - just for debugging/logging
```
neoxelox: why are you deleting this check, you cannot attach to a non-active run, there is no stream in redis.

andresgutgon: To be talked offline. WIP. But thanks for the review

packages/core/src/services/runs/enqueue.ts
```diff
@@ -58,16 +65,13 @@ export async function enqueueRun({
...
+    await repository.delete({ runUuid })
     return Result.error(
       new UnprocessableEntityError('Failed to enqueue background run job'),
     )
   }
 
-  const repository = new RunsRepository(workspace.id, project.id)
-  const creating = await repository.create({ runUuid, queuedAt: new Date() })
-  if (creating.error) return Result.error(creating.error)
-  const run = creating.value
```
andresgutgon, line 69: This was the issue. By the time we add this to the active runs the bull mq was already run and start failed

packages/core/src/cache/index.ts
```diff
@@ -60,51 +60,79 @@ export const getOrSet = async <T>(
...
  * @returns Promise that resolves to the return value of the callback function
+ * @throws Error if lock cannot be acquired after all retry attempts
  */
-export const withCacheLock = async <T>(
-  lockKey: string,
-  callback: (cache: Cache) => Promise<T>,
+export const withCacheLock = async <T>({
+  lockKey,
+  callbackFn,
   timeout = 5000,
```
geclos: could we reduce this timeout? 5 seconds is huge for redis, should be one at most 1 second and that's already stretching it

geclos (approved):

packages/core/scripts/background-run/stress-test.ts
```diff
@@ -0,0 +1,191 @@
...
+ *   tsx stress-test-runs.ts 10
+ *   tsx stress-test-runs.ts 40
+ *   tsx stress-test-runs.ts 100
+ */
+
+// Local testing credentials. Use your own.
+const CONFIG = {
+  API_URL:
+    'http://localhost:8787/api/v3/projects/50/versions/ecfda479-b6c4-460e-8340-e1dfba5a0fc9/documents/run',
+  API_KEY: 'd1a87bd5-0732-4e5d-ba95-a8a47eaa3055',
```
neoxelox:
```suggestion
  API_KEY: process.env.TEST_LATITUDE_API_KEY,
```

neoxelox (dismissed):

packages/core/src/cache/index.ts
andresgutgon:
```suggestion
  timeout = 1000,
```

PR #1818, opened 2025-10-16
Feat/remove schema types

geclos:


packages/web-ui/src/lib/getUserInfo.ts
```diff
@@ -1,4 +1,4 @@
-import type { User } from '@latitude-data/core/schema/types'
+import type { User } from '@latitude-data/core/schema/models/User'
```
andresgutgon: Web ui should not know about core. Not saying to change here, but this should not be this way. Maybe was me 😅

geclos: hmm it's imported multiple times (types only)

PR #1819, opened 2025-10-16
feat: file cache

geclos:
Implements a file cache for file downloads in object storage. Disabled by default. This allows self-hosted customers to easily expose private files via publicly signed URLs with short TTLs and do not worry about ttl expiration in long-running conversations.

### DIFF
--- a/packages/core/src/services/ai/index.ts
+++ b/packages/core/src/services/ai/index.ts
@@ -26,6 +26,7 @@ import { handleAICallAPIError } from './handleError'
 import { createProvider } from './helpers'
 import { Providers } from '@latitude-data/constants'
 import { applyAllRules } from './providers/rules'
+import { downloadWithCache } from '../../lib/downloadWithCache'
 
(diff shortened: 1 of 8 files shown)


packages/core/src/services/ai/index.ts
```diff
@@ -171,6 +172,17 @@ export async function ai({
       experimental_output: useSchema
         ? Output.object({ schema: jsonSchema(schema) })
         : undefined,
+      experimental_download: async (
+        options: { url: URL; isUrlSupportedByModel: boolean }[],
+      ) => {
+        try {
+          return Promise.all(options.map(({ url }) => downloadWithCache(url)))
+        } catch (error) {
+          throw new Error(
```
andresgutgon, line 181: So one file failing fails all?

geclos: yeah that's also how it works now i'm not changing behaviour here

andresgutgon (approved):

PR #1821, opened 2025-10-16
Add new sample agent for agent onboarding

learningbizz:
After looking through the metrics, lots of users would skip the onboarding and end up with an empty dashboard without knowing what to do. This PR adds a default agent so they have a generic deep research agent to be able to interact with and understand how Latitude works.


apps/web/src/actions/workspaceOnboarding/createDefaultAgentOnboardingProject.ts
```diff
@@ -0,0 +1,25 @@
...
+    const onboarding = await getWorkspaceOnboarding({
+      workspace,
+    }).then((r) => r.unwrap())
+
+    // We will mark the onboarding as complete so once the user clones the agent, they will be redirected to the home page
+    await markWorkspaceOnboardingComplete({
+      onboarding,
+    }).then((r) => r.unwrap())
+
+    const defaultAgentOnboardingRoute = `/actions/clone-agent?uuid=${envClient.NEXT_PUBLIC_DEFAULT_AGENT_ONBOARDING_UUID}`
```
andresgutgon: Use ROUTES

geclos (approved):

PR #1824, opened 2025-10-16
Fix claiming promocodes stopped working

learningbizz:
The action of claiming stopped working after the vercel ai sdk upgrade to v5 due to not having the authProcedure as part of the action. For this reason, the input parameter of workspace was not found within the action, and the action failed. This PR fixes this

### DIFF
--- a/apps/web/src/actions/promocodes/claim.test.ts
+++ b/apps/web/src/actions/promocodes/claim.test.ts
@@ -1,6 +1,6 @@
 import { createProject } from '@latitude-data/core/factories'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import { Workspace, Promocode } from '@latitude-data/core/schema/types'
+import { Workspace, Promocode, User } from '@latitude-data/core/schema/types'
 import { createCreditsPromocode } from '../../../../../packages/core/src/tests/factories/promocodes'
 import { claimPromocodeAction } from './claim'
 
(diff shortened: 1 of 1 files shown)


apps/web/src/actions/promocodes/claim.ts
```diff
@@ -14,6 +14,7 @@ export const claimPromocodeAction = errorHandlingProcedure
   .action(async ({ ctx, parsedInput }) => {
     const { workspace } = ctx
     const { code } = parsedInput
+    console.log('workspace', workspace)
```
andresgutgon: 🔥

andresgutgon (dismissed):

geclos (dismissed):

apps/web/src/actions/promocodes/claim.ts
```diff
@@ -1,10 +1,10 @@
 'use server'
 
 import { z } from 'zod'
-import { errorHandlingProcedure, withRateLimit } from '$/actions/procedures'
+import { authProcedure, withRateLimit } from '$/actions/procedures'
 import { claimPromocode } from '@latitude-data/core/services/promocodes/claimPromocode'
 
-export const claimPromocodeAction = errorHandlingProcedure
```
geclos, line 7: add a quick test and we'll be protected forever

apps/web/src/actions/promocodes/claim.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    promocode = await createCreditsPromocode(100)
+
+    mocks.getSession.mockResolvedValue({
+      user,
+      session: { userId: user.id, currentWorkspaceId: workspace.id },
+    })
+  })
+
+  it('should successfully claim a promocode', async () => {
```
andresgutgon: I think in this case is more useful just don't mock the user and assert the call fails. It's all we want to test

andresgutgon: example
https://github.com/latitude-dev/latitude-llm/blob/b5b7eec525661e00c45e569c4c0a6a9d5b450272/apps/web/src/actions/commits/createDraftWithContentAction.test.ts#L48-L60

learningbizz: I wanted to kill two birds with one stone with my test, as it would fail if you used errorHandling. However, I do like adding smaller tests to test specific things. Added

PR #1829, opened 2025-10-18
Try Next.js 16 beta + ESlint 9 flat config

andresgutgon:
# What?
Next.js migrated to ESlint 9 so this is good opportunity to move to that version

## React Compiler
React compiler is stable in `react@19.2.2` and `next@16.0.0.beta` so next step will be to enable it and see what we need to change or remove
https://nextjs.org/blog/next-16-beta#react-compiler-support-stable


apps/web/src/app/_lib/buildMetatags.ts
```diff
@@ -6,7 +6,7 @@ const DEFAULT_DESCRIPTION =
 
 // This function is necessary to define default metadata correctly, because
 // Nextjs metadata merging would overwrite the nested objects totally.
-export default function buildMetatags({
+export default async function buildMetatags({
```
geclos, line 9: why?

andresgutgon: The signature is a Promise

PR #1835, opened 2025-10-20
feat: add parameters to prompt span

geclos:
Added proper namespacing to latitude otel parameters and also changed reference between spans and experiments to be consistent and use uuids everywhere


apps/web/src/components/tracing/spans/Prompt.tsx
```diff
@@ -14,43 +14,33 @@ export default {
 }
 
 function DetailsPanel({ span }: DetailsPanelProps<SpanType.Prompt>) {
+  console.log(span?.metadata?.parameters)
```
neoxelox, line 17: console log

PR #1837, opened 2025-10-20
feat: remove evals v1

geclos:


packages/core/src/constants.ts
```diff
@@ -35,7 +34,6 @@ import { type Workspace } from './schema/models/types/Workspace'
 export {
   DocumentType,
   EMAIL_REGEX,
```
neoxelox, line 36: I cannot comment on that line but delete "EvaluationMetadataType"

PR #1838, opened 2025-10-20
Include Source Data in Tool Calls

csansoon:
When running an Agent, we compute all tools and build a "source" object, where we know exactly where tools are coming from, independently of their name, ids, and any other attributes. This source is used to know what callback to execute when a Tool Call is received.

We can include this source to the Tool Call data we send to the stream and store as a Provider Log! This way, we can apply the same logic and print different components based on the tool origin, independently of the tool's actual name.

Before this change, we could only detect if a tool came from an integration if it started with a custom prefix we always added. And even then, we could not know what integration it was even coming from.

Now, since all source data is included in the Tool Call itself, we don't need prefixes. All we need to do is check the source, which contains a `type` (client, integration, subagent, latitude tool...), and relevant information depending on its type, such as its integration id for integration tools, or the document uuid for subagent requests!

### DIFF
--- a/packages/core/src/lib/streamManager/ChainStreamConsumer/consumeStream.ts
+++ b/packages/core/src/lib/streamManager/ChainStreamConsumer/consumeStream.ts
@@ -11,11 +11,13 @@ import {
 } from '../../../constants'
 import { Providers } from '@latitude-data/constants'
 import { AIReturn } from '../../../services/ai'
+import { ResolvedTools } from '../resolveTools/types'
 
 type ConsumeStreamParams = {
   result: AIReturn<StreamType>
   controller: ReadableStreamDefaultController
   accumulatedText: { text: string }
+  resolvedTools?: ResolvedTools
 }
 
 type NoRunError = object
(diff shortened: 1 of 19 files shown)


packages/core/src/lib/streamManager/ChainStreamConsumer/consumeStream.ts
```diff
@@ -53,11 +56,13 @@ export async function consumeStream({
 
       _accumulatedText.text += vercelChunk.text
     } else if (vercelChunk.type === 'tool-call') {
+      const resolvedTool = resolvedTools?.[vercelChunk.toolName]
       chunk = {
         type: 'tool-call',
         toolCallId: vercelChunk.toolCallId,
         toolName: vercelChunk.toolName,
         args: vercelChunk.input,
+        ...(resolvedTool ? { _sourceData: resolvedTool.sourceData } : {}),
```
andresgutgon, line 65: Do tests here. This is complicated enough already. (te toco)

csansoon: done :D

PR #1840, opened 2025-10-20
Remove duplicate tool calls when assistant message included tool call

learningbizz:
When we handle tool calls, we would merge our assistant `content` and `toolCalls` within a single content argument to abide Vercel's AI sdk style, regardless if the content of the assistant message had the tool call type already there. This resulted in an assistant message with two tool calls within the content with the same id. Example:

```
     "role": "assistant",
     "content": [
       {
         "type": "tool-call",
         "toolCallId": "call_jashdkjhsadasd,
         "toolName": "lat_agent_latte_managers_investigator",
         "input": {
           "project_id": "2",
           "version_id": "live",
           "query": "What triggers does this agent have?"
         }
       },
       {
         "type": "tool-call",
         "toolCallId": "call_jashdkjhsadasd",
         "toolName": "lat_agent_latte_managers_investigator",
         "input": {
           "project_id": "2",
           "version_id": "live",
           "query": "What triggers does this agent have?"
         }
       }
     ]
   }
```

Some providers don't mind this, such as OpenAIs GPTs, but Anthropic does mind and throws an error, which is the one Will was facing with Latte. 

This fix fixes that by checking if the content of the assistant already includes that tool call, by comparing the ids, and adding it if its not originally there.

### DIFF
--- a/packages/core/src/services/ai/convertLatitudeMessagesToVercelFormat.test.ts
+++ b/packages/core/src/services/ai/convertLatitudeMessagesToVercelFormat.test.ts
@@ -428,4 +428,126 @@ describe('convertLatitudeMessagesToVercelFormat', () => {
       },
     ])
   })
+
+  it('only adds tool calls from toolCalls that are not already in content', () => {
+    const messages: Message[] = [
+      {
+        role: MessageRole.assistant,
+        content: [
+          {
+            type: 'tool-call',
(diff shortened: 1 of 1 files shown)


geclos (approved):

andresgutgon (changes requested):
Tests

PR #1841, opened 2025-10-21
Fix latte text area cutoff and header commit title max width issue

learningbizz:
The text area of latte's chat input would cutoff when debug mode was on, and the commit title would break the header when it was too long as well. This small PR fixes these issues

### DIFF
--- a/apps/web/src/components/LatteSidebar/LatteChat/LatteChatInput.tsx
+++ b/apps/web/src/components/LatteSidebar/LatteChat/LatteChatInput.tsx
@@ -1,5 +1,6 @@
 import { useCurrentDocumentMaybe } from '$/app/providers/DocumentProvider'
 import { useLatteChangeActions } from '$/hooks/latte/useLatteChangeActions'
+import { useLatteDebugMode } from '$/hooks/latte/useLatteDebugMode'
 import { useDocumentValueMaybe } from '$/hooks/useDocumentValueContext'
 import { useNavigate } from '$/hooks/useNavigate'
 import { ROUTES } from '$/services/routes'
@@ -71,6 +72,7 @@ export function LatteChatInput({
     threadUuid,
     commitId: commit.id,
   })
+  const { enabled: debugModeEnabled, data: debugData } = useLatteDebugMode()
   const checkpoint = useMemo(() => {
     return checkpoints?.find((cp) => cp.documentUuid === document.documentUuid)
   }, [checkpoints, document])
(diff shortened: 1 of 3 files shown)


packages/web-ui/tailwind.config.js
```diff
@@ -19,6 +19,9 @@ export default {
       mono: ['var(--font-mono)'],
     },
     extend: {
+      spacing: {
+        18: '72px',
```
andresgutgon, line 23: Did you talk with product about this?

apps/web/src/components/layouts/AppLayout/Header/Breadcrumb/Projects/Versions/index.tsx
```diff
@@ -49,9 +49,11 @@ export function CommitBreadcrumbItems({
           <BreadcrumbItemSkeleton className='pl-3' />
         ) : (
           <div className='flex flex-row w-full justify-between overflow-hidden pl-3'>
-            <Text.H5 color='foregroundMuted' noWrap ellipsis>
-              {currentCommit?.title ?? commitUuid}
-            </Text.H5>
+            <div className='flex flex-row overflow-hidden max-w-[200px] sm:max-w-[300px] md:max-w-[400px] lg:max-w-[500px] xl:max-w-[600px]'>
```
andresgutgon, line 52: This is not a very elegant way of fixing things. I'm sure using flex in a certain way can help here

apps/web/src/components/LatteSidebar/LatteChat/LatteChatInput.tsx
```diff
@@ -160,10 +165,11 @@ export function LatteChatInput({
       <TextArea
         ref={inputRef}
         className={cn(
-          'bg-background w-full px-3 pt-3 pb-14 resize-none text-sm border-none',
+          'bg-background w-full px-3 pt-3 resize-none text-sm border-none',
           'shadow-sm text-muted-foreground',
           'ring-0 focus-visible:ring-0 outline-none focus-visible:outline-none',
           'focus-visible:animate-glow focus-visible:glow-latte custom-scrollbar scrollable-indicator',
+          isDebugSelectorVisible ? 'pb-18' : 'pb-14',
```
andresgutgon, line 172: This is a smell that the space at the bottom could be handled in another way. Not saying to change but this is a patch

apps/web/src/components/layouts/AppLayout/Header/Breadcrumb/Projects/Versions/index.tsx
learningbizz: Tried with flex, but didnt find another way. any ideas?

packages/web-ui/tailwind.config.js
learningbizz: No, why? I just calculated the middle between pb-16 and pb-20 from tailwind and added it here

PR #1844, opened 2025-10-21
Substitute sentry use for datadog in core

learningbizz:
- Replace Datadog captureMessage/error with Sentry's
- Add Capture message in `convertLatitudeMessagesToVercelFormat`
- Add Error type cast in Transaction


packages/core/src/utils/datadogCapture.ts
```diff
@@ -0,0 +1,80 @@
+import tracer from 'dd-trace'
```
geclos, line 1: make sure we have datadog tracer installed in web/gateway/workers

geclos (approved):

packages/core/src/utils/datadogCapture.ts
andresgutgon: Why would this be necessary on the gateway and workers? Is not installed by the core package?

Not talking about next.js because it has "special needs"

learningbizz: image

We have them in all apps

PR #1845, opened 2025-10-21
Standarize datadog version to latest

learningbizz:
We have different versions across apps, so we should standarize them to the latest one used: 5.69.0.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -45,6 +45,9 @@ catalogs:
     ai:
       specifier: 5.0.44
       version: 5.0.44
+    dd-trace:
+      specifier: 5.69.0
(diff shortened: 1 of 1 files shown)


andresgutgon (dismissed):

apps/gateway/package.json
```diff
@@ -31,7 +31,7 @@
     "promptl-ai": "catalog:",
     "rate-limiter-flexible": "5.0.3",
     "zod": "catalog:",
-    "dd-trace": "catalog:"
+    "dd-trace": "^5.69.0"
```
geclos: why? use catalog

learningbizz: True, changed them all to use catalog now 👍

PR #1846, opened 2025-10-21
Fix syntax highlighter flicker by lazy loading theme and component

learningbizz:
Instead of just dynamically importing the theme and separately loading lazily the prism highlighter, we do it together and we avoid the flickering issue

### DIFF
--- a/packages/web-ui/src/ds/atoms/CodeBlock/SyntaxHightlighter/index.tsx
+++ b/packages/web-ui/src/ds/atoms/CodeBlock/SyntaxHightlighter/index.tsx
@@ -2,7 +2,7 @@
 
 import { ComponentProps, Suspense, lazy } from 'react'
 import { Prism as PrismHighlighter } from 'react-syntax-highlighter'
-import { Text } from '../../Text'
+import { Skeleton } from '../../Skeleton'
 
 type Props = Omit<ComponentProps<typeof PrismHighlighter>, 'style'> & {
   currentTheme?: string
@@ -17,7 +17,7 @@ export function SyntaxHighlighter({
   currentTheme,
 }: Props) {
   return (
-    <Suspense fallback={<Text.H5 animate>Loading code...</Text.H5>}>
+    <Suspense fallback={<Skeleton height='h5' />}>
       <LazyPrism
         className={className}
         language={language}
--- a/packages/web-ui/src/ds/atoms/CodeBlock/SyntaxHightlighter/LazyPrism/index.tsx
+++ b/packages/web-ui/src/ds/atoms/CodeBlock/SyntaxHightlighter/LazyPrism/index.tsx
@@ -2,6 +2,7 @@
 import { Prism as PrismHighlighter } from 'react-syntax-highlighter'
 import { CurrentTheme } from '../../../../../constants'
 import { ComponentProps, use } from 'react'
+import { cn } from '../../../../../lib/utils'
 
 type Props = Omit<ComponentProps<typeof PrismHighlighter>, 'style'> & {
   currentTheme?: string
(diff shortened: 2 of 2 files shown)


packages/web-ui/src/ds/atoms/CodeBlock/SyntaxHightlighter/index.tsx
```diff
@@ -1,73 +1,29 @@
...
-    loadTheme()
-
-    return () => {
-      active = false
-    }
-  }, [currentTheme])
-
   return (
-    <Suspense fallback={<div className='p-4 text-sm'>Loading code...</div>}>
+    <Suspense fallback={<Text.H5 animate>Loading code...</Text.H5>}>
```
andresgutgon: Nicer loader with `<Skeleton />`

packages/web-ui/src/ds/atoms/CodeBlock/SyntaxHightlighter/index.tsx
```diff
@@ -1,73 +1,29 @@
...
-        className={cn('text-xs', className)}
+        className={className}
         language={language}
-        style={style}
-        customStyle={{
-          borderRadius: '0.375rem',
-          padding: '1rem',
-          lineHeight: '1.25rem',
-          margin: '0',
-        }}
```
andresgutgon, line 68: This was used before. Move to Lazy prism

packages/web-ui/src/ds/atoms/CodeBlock/SyntaxHightlighter/index.tsx
```diff
@@ -1,73 +1,29 @@
...
-    return () => {
-      active = false
-    }
-  }, [currentTheme])
-
   return (
-    <Suspense fallback={<div className='p-4 text-sm'>Loading code...</div>}>
+    <Suspense fallback={<Text.H5 animate>Loading code...</Text.H5>}>
       <LazyPrism
-        className={cn('text-xs', className)}
```
andresgutgon, line 60: This also

packages/web-ui/src/ds/atoms/CodeBlock/index.tsx
```diff
@@ -19,8 +19,10 @@ interface CodeBlockProps {
   bgColor?: string
 }
 
-export function useCodeBlockBackgroundColor(override?: string) {
-  const { resolvedTheme } = useTheme()
+function getCodeBlockBackgroundColor(
```
andresgutgon, line 22: Why this change?

learningbizz: To avoid re-rendering twice by subscribing to the context `useTheme` twice.

If you pass the prop as a parameter, it re-renders in one unit of work.

packages/web-ui/src/ds/atoms/CodeBlock/SyntaxHightlighter/index.tsx
```diff
@@ -1,73 +1,29 @@
...
-    loadTheme()
-
-    return () => {
-      active = false
-    }
-  }, [currentTheme])
-
   return (
-    <Suspense fallback={<div className='p-4 text-sm'>Loading code...</div>}>
+    <Suspense fallback={<Skeleton height='h5' />}>
```
andresgutgon, line 20: I mean, this is using the skeleton. But it is not nice 😂 It does not represent the shape of a code block. 

Take, as example main editor loading state

learningbizz: Never satisfied :( hahahah

PR #1856, opened 2025-10-22
Annotation controls settings for HITL evals

neoxelox:


packages/constants/src/evaluations/human.ts
```diff
@@ -6,6 +6,7 @@ import {
 } from './shared'
 
 const humanEvaluationConfiguration = baseEvaluationConfiguration.extend({
+  enableControls: z.boolean().optional(),
```
andresgutgon, line 9: Is this the final? wasn't `enableUiAnnotation`?

PR #1857, opened 2025-10-22
Chat input on playground with no parameters

csansoon:
The Playground's parameters UI is really confusing if you have not added any parameter to your prompt.
To solve this, if you have not added any parameter, we'll display a Chat Input instead

https://github.com/user-attachments/assets/a698b0aa-edee-4e63-b539-33dd79095adb

### DIFF
--- a/apps/web/src/components/ChatWrapper/Message/Content/Text.tsx
+++ b/apps/web/src/components/ChatWrapper/Message/Content/Text.tsx
@@ -116,7 +116,7 @@ export function TextMessageContent({
       size={size}
       text={text}
       parameters={parameters}
-      sourceMap={sourceMap}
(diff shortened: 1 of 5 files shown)


apps/web/src/components/ChatWrapper/Message/Content/Text.tsx
```diff
@@ -116,7 +116,7 @@ export function TextMessageContent({
       size={size}
       text={text}
       parameters={parameters}
-      sourceMap={sourceMap}
+      sourceMap={debugMode ? sourceMap : undefined}
```
csansoon, line 119: (bugfix)

andresgutgon (dismissed):

PR #1859, opened 2025-10-23
Added dropzone in home page

learningbizz:
Added ability to add files and images from the home page if the parameter is an image/file.

image


apps/web/src/components/Agent/AgentInput/index.tsx
```diff
@@ -137,13 +163,15 @@ export function AgentInput({
                 'border-t border-border': i > 0,
               })}
             >
-              <div className='pl-3 pt-2'>
+              <div className={cn('pl-3 pt-2')}>
```
andresgutgon: No need

PR #1865, opened 2025-10-23
Substitute latte credit rewards for workspace runs

learningbizz:
We want to substitute the rewards to return runs now instead of Latte credits. This PR adds this.

image

### DIFF
--- a/packages/core/src/services/claimedRewards/claimNewUserReferrals.ts
+++ b/packages/core/src/services/claimedRewards/claimNewUserReferrals.ts
@@ -78,7 +78,7 @@ export async function claimNewUserReferrals(
 
     const granting = await issueGrant(
       {
-        type: QuotaType.Credits,
+        type: QuotaType.Runs,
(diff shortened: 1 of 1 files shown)


packages/core/src/constants.ts
```diff
@@ -122,13 +122,13 @@ export enum RewardType {
...
 
 export const REWARD_VALUES: Record<RewardType, number> = {
-  [RewardType.XFollow]: 3,
-  [RewardType.LinkedInFollow]: 3,
-  [RewardType.GithubStar]: 3,
-  [RewardType.XPost]: 5,
-  [RewardType.LinkedInPost]: 5,
-  [RewardType.AgentShare]: 10,
-  [RewardType.Referral]: 10,
+  [RewardType.XFollow]: 2_000,
```
learningbizz, line 125: These values are gotten from our past runs rewards (and the ones that didn't exist, invented them with similar amounts of the other existing ones)

PR #1867, opened 2025-10-24
Add jobs to add seats to hobby and team plans

learningbizz:
Adding maintenance jobs to upgrade current hobby and team plans to the desired specification in task LAT-154

### DIFF
--- a/apps/workers/src/workers/schedule.ts
+++ b/apps/workers/src/workers/schedule.ts
@@ -59,17 +59,17 @@ export async function setupSchedules() {
     { opts: { attempts: 1 } },
   )
 
-  // Every day at 11 AM CEST - Upgrade Hobby users to HobbyV3
+  // Every day at 12 PM UTC - Upgrade Hobby users to HobbyV3
   await maintenanceQueue.upsertJobScheduler(
     'upgradeHobbyUsersToV3Job',
-    { pattern: '0 10 * * *' },
+    { pattern: '0 12 * * *' },
     { opts: { attempts: 1 } },
   )
 
-  // Every day at 11:30 AM CEST - Upgrade TeamV2 users to TeamV3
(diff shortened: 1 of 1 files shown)


packages/core/src/jobs/job-definitions/maintenance/upgradeTeamV2UsersToV3Job.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Job that runs daily at 11:30am CEST to schedule individual jobs for upgrading
+ * all users with TeamV2 subscriptions to TeamV3.
+ *
+ * This job:
+ * 1. Finds all workspaces with TeamV2 plans
+ * 2. Enqueues individual upgrade jobs for each workspace
+ * 3. Each individual job will upgrade the workspace to TeamV3 subscription
+ */
+export const upgradeTeamV2UsersToV3Job = async (
```
andresgutgon, line 20: nickpic: Users here is misleading.

andresgutgon: Also not sure we need 2-step jobs for this with the volume of workspaces we have. This update should be quick


I would say all in one job. All workspaces and all plans upgrades

learningbizz: This is only going to run once, so no need to make it perfect

learningbizz: Also, there are 20k hobby plans, which is quite substantial

andresgutgon: My point was about not making perfect xD and put all in one single job.

PR #1872, opened 2025-10-27
Simulation Settings in Document Logs

csansoon:
First step of [Improve Simulation Environment](https://linear.app/latitude/issue/LAT-163/improve-simulation-environment):
---
I'll create a new "Simulation Settings" attribute on Stream Manager. When resolving tools, if simulateToolCalls is enabled, the resolved tools will call our generation service instead of the actual tool. This way, everything still goes through the regular runDocumentAtCommit pipeline.

When storing DocumentLogs, the simulationSettings attribute will be stored into the log, so that the UI will be able to display what tool responses are simulated and why.

An additional simulationDescription attribute is added to these Simulation Settings. This will act as a custom prompt to guide the simulation into a desired output. This can be used to test how the agent reacts in different situations.

When enabled, all tools will be simulated, except for:
- Subagents
- Some Latitude tools (think & todo)

---

- [x] New simulationSettings attribute in DocumentLogs **← You are HERE**
- [ ] Add Simulation Pipeline to runDocumentAtCommit and StreamManager, and replace the old runDocumentAtCommitWithAutoToolResponses
- [ ] Add UI to show what tool calls have been generated by AI instead of the real output
- [ ] Add "Simulation Description" parameter to Experiments

### DIFF
--- a/packages/constants/package.json
+++ b/packages/constants/package.json
@@ -28,7 +28,8 @@
     "./promptl": "./src/promptl.ts",
     "./trigger": "./src/trigger.ts",
     "./documentTriggers": "./src/documentTriggers/schema.ts",
-    "./toolSources": "./src/toolSources.ts"
+    "./toolSources": "./src/toolSources.ts",
+    "./simulation": "./src/simulation.ts"
   },
   "devDependencies": {
     "@latitude-data/eslint-config": "workspace:*",
--- a/packages/constants/src/simulation.ts
+++ b/packages/constants/src/simulation.ts
@@ -1,0 +1,4 @@
+export type SimulationSettings = {
+  simulationDescription?: string // A prompt used to guide and generate the simulation result
+  simulateToolResponses?: boolean
+}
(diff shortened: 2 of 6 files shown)


packages/constants/src/index.ts
```diff
@@ -62,6 +62,11 @@ export type ExperimentScores = {
   [evaluationUuid: string]: ExperimentEvaluationScore
 }
 
+export type SimulationSettings = {
+  simulationDescription?: string // A prompt used to guide and generate the simulation result
+  simulateToolResponses?: boolean
+}
```
andresgutgon: Don't put more things here. Export a new file if there is no one appropiate for this

geclos:
> When storing DocumentLogs, the simulationSettings attribute will be stored into the log, so that the UI will be able to display what tool responses are simulated and why.


Can we avoid adding more stuff to logs model? We will be deprecating logs in favor our spans

PR #1876, opened 2025-10-27
Add role in signup form

learningbizz:
This is the first task of the new onboarding, adding a new role parameter in the signup form called "role".

This will be useful for us to know what users we're getting into our application.

image

### DIFF
--- a/apps/web/src/app/(public)/setup/_components/SetupForm/index.tsx
+++ b/apps/web/src/app/(public)/setup/_components/SetupForm/index.tsx
@@ -13,21 +13,21 @@ import { Text } from '@latitude-data/web-ui/atoms/Text'
 import { useToast } from '@latitude-data/web-ui/atoms/Toast'
 import Link from 'next/link'
 import { Select } from '@latitude-data/web-ui/atoms/Select'
-import { USER_ROLES, UserRole } from '@latitude-data/constants/users'
+import { USER_TITLES, UserTitle } from '@latitude-data/constants/users'
 
-export const humanizeUserRole = (role: UserRole): string => {
-  switch (role) {
-    case UserRole.Engineer:
+export const humanizeUserTitle = (title: UserTitle): string => {
+  switch (title) {
+    case UserTitle.Engineer:
       return 'Engineer'
-    case UserRole.DataAIAndML:
+    case UserTitle.DataAIAndML:
       return 'Data/AI/ML'
-    case UserRole.ProductManager:
+    case UserTitle.ProductManager:
       return 'Product Manager'
-    case UserRole.Designer:
+    case UserTitle.Designer:
       return 'Designer'
-    case UserRole.Founder:
+    case UserTitle.Founder:
       return 'Founder'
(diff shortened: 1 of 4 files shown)


packages/web-ui/src/ds/atoms/Select/Primitives/index.tsx
```diff
@@ -32,7 +32,7 @@ function SelectValueWithIcon({
   return (
     <div className='w-full flex flex-row items-center gap-x-2 min-w-0 truncate'>
       {typeof icon === 'string' ? <Icon name={icon as IconName} /> : icon}
-      {typeof children === 'string' ? <Text.H6>{children}</Text.H6> : children}
+      {typeof children === 'string' ? <Text.H5>{children}</Text.H5> : children}
```
learningbizz, line 35: This was causing a visual error, as all selects had a smaller font size than other regular inputs

packages/core/src/schema/models/users.ts
```diff
@@ -1,17 +1,25 @@
...
+    email: text('email').notNull().unique(),
+    confirmedAt: timestamp('confirmed_at'),
+    admin: boolean('admin').notNull().default(false),
+    lastSuggestionNotifiedAt: timestamp('last_suggestion_notified_at'),
+    devMode: boolean('dev_mode'),
+    role: varchar('role', { length: 128 }).$type<UserRole>(), // can be null as signup with google bypasses the role selection
+    ...timestamps(),
+  },
+  (table) => ({
+    usersRoleIndex: index('users_role_idx').on(table.role),
```
andresgutgon: This syntax is deprecated. Use an array now that you're touching it

apps/web/src/app/(public)/setup/_components/SetupForm/index.tsx
```diff
@@ -12,6 +12,25 @@ import { Separator } from '@latitude-data/web-ui/atoms/Separator'
 import { Text } from '@latitude-data/web-ui/atoms/Text'
 import { useToast } from '@latitude-data/web-ui/atoms/Toast'
 import Link from 'next/link'
+import { Select } from '@latitude-data/web-ui/atoms/Select'
+import { USER_ROLES, UserRole } from '@latitude-data/core/constants'
```
andresgutgon: Put UserRole constant in constants package. Easier to share

packages/core/drizzle/0225_new_user_roles.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "latitude"."users" ADD COLUMN "role" varchar(128);--> statement-breakpoint
```
andresgutgon: Role in user's table usually means permissions. I would named something more specific like `professionRole` or something like that

learningbizz: We can later add another name like `permissionRole` when we add permissions

learningbizz: Decided to change to `title` in the end 👍

PR #1879, opened 2025-10-27
Move point reference to commit from issue to issue histogram

andresgutgon:
# What?
We want to show issues with histograms created in that commit of previous merged commits

### DIFF
--- a/packages/core/src/schema/models/issueHistograms.ts
+++ b/packages/core/src/schema/models/issueHistograms.ts
@@ -12,6 +12,7 @@ import { latitudeSchema } from '../db-schema'
 import { timestamps } from '../schemaHelpers'
 import { workspaces } from './workspaces'
 import { issues } from './issues'
+import { commits } from './commits'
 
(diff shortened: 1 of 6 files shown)


packages/core/src/schema/models/issueHistograms.ts
```diff
@@ -23,13 +24,17 @@ export const issueHistograms = latitudeSchema.table(
...
     date: date('date').notNull(),
     count: integer('count').notNull(),
     ...timestamps(),
   },
   (table) => [
     index('issue_histograms_workspace_id_idx').on(table.workspaceId),
     index('issue_histograms_issue_id_idx').on(table.issueId),
+    index('issue_histograms_commit_id_idx').on(table.commitId),
     index('issue_histograms_date_idx').on(table.date),
     unique('issue_histograms_unique_workspace_issue_date').on(
```
neoxelox, line 39:
```suggestion
    unique('issue_histograms_unique_issue_date').on(
```

PR #1881, opened 2025-10-28
Onboarding v3

learningbizz:
This is the whole PR to add the new onboarding in the application, under a feature flag. We were going to divide it into smaller PRs, but the whole thing is small enough to review well.

### DIFF
--- a/apps/web/src/actions/datasets/generateDataset.ts
+++ b/apps/web/src/actions/datasets/generateDataset.ts
@@ -3,6 +3,7 @@
 import { z } from 'zod'
 import { authProcedure } from '$/actions/procedures'
 import { generateDatasetWithCopilot } from '@latitude-data/core/services/datasets/generateWithCopilot'
+import { createDatasetFromJson } from '@latitude-data/core/services/datasets/createFromJson'
 
 export const generateDatasetAction = authProcedure
   .inputSchema(
@@ -16,12 +17,18 @@ export const generateDatasetAction = authProcedure
   .action(async ({ parsedInput, ctx }) => {
     const { parameters, description, rowCount, name } = parsedInput
 
-    return await generateDatasetWithCopilot({
-      workspace: ctx.workspace,
-      user: ctx.user,
+    const generatedDatasetContent = await generateDatasetWithCopilot({
       parameters,
       description,
       rowCount,
-      name,
+    }).then((r) => r.unwrap())
+
+    return await createDatasetFromJson({
+      author: ctx.user,
+      workspace: ctx.workspace,
+      data: {
+        name,
+        rows: JSON.stringify(generatedDatasetContent.rows),
+      },
     }).then((r) => r.unwrap())
   })
--- a/apps/web/src/actions/datasets/generateOnboardingDataset.ts
+++ b/apps/web/src/actions/datasets/generateOnboardingDataset.ts
@@ -33,12 +33,18 @@ export const generateOnboardingDatasetAction = authProcedure
       }).then((r) => r.unwrap())
     }
 
-    return await generateDatasetWithCopilot({
-      workspace: ctx.workspace,
-      user: ctx.user,
+    const generatedDatasetContent = await generateDatasetWithCopilot({
       parameters,
       prompt,
       rowCount,
-      name,
+    }).then((r) => r.unwrap())
+
+    return await createDatasetFromJson({
+      author: ctx.user,
+      workspace: ctx.workspace,
+      data: {
+        name,
+        rows: JSON.stringify(generatedDatasetContent.rows),
+      },
     }).then((r) => r.unwrap())
   })
--- a/apps/web/src/actions/sdk/generateDatasetPreviewAction.ts
+++ b/apps/web/src/actions/sdk/generateDatasetPreviewAction.ts
@@ -1,11 +1,8 @@
 'use server'
-import { BadRequestError } from '@latitude-data/constants/errors'
-import { env } from '@latitude-data/env'
-import { createSdk } from '$/app/(private)/_lib/createSdk'
 import { generatePreviewRowsFromJson } from '@latitude-data/core/services/datasetRows/generatePreviewRowsFromJson'
 import { authProcedure } from '$/actions/procedures'
 import { z } from 'zod'
-import { CLOUD_MESSAGES, LogSources } from '@latitude-data/core/constants'
+import { generateDatasetWithCopilot } from '@latitude-data/core/services/datasets/generateWithCopilot'
 
 export const generateDatasetPreviewAction = authProcedure
   .inputSchema(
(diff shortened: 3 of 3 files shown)


apps/web/src/actions/datasets/generateDataset.ts
```diff
@@ -3,82 +3,50 @@
...
+      fromCloud: z.boolean(),
     }),
   )
-  .action(async ({ parsedInput }) => {
-    if (!env.LATITUDE_CLOUD) {
+  .action(async ({ parsedInput, ctx }) => {
+    const { parameters, description, prompt, rowCount, name, fromCloud } =
+      parsedInput
+
+    if (fromCloud && !env.LATITUDE_CLOUD) {
```
andresgutgon: Why `fromCloud` is needed? we should not trust the client

apps/web/src/app/(onboarding)/onboarding-dataset/paste-your-prompt/_components/PasteYourPrompt/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+      runGenerateAction,
+      setCurrentOnboardingStep,
+      updateDocumentContent,
+    ],
+  )
+
+  return (
+    <div className='flex flex-row items-center gap-10 h-full w-full'>
+      <div className='flex flex-col items-end w-full h-full'>
+        <div className='relative flex-1 w-full max-h-[350px] max-w-[600px]'>
```
andresgutgon: Adding max-w-[600px] is repeated in all sub components. Is an smell that should be in the parent

apps/web/src/components/BlocksEditor/Editor/state/promptlToLexical/fromAstToBlocks.ts
```diff
@@ -874,4 +874,13 @@ export function fromAstToBlocks({
   } satisfies BlockRootNode
 }
 
+export const emptyRootBlock: BlockRootNode = {
```
andresgutgon: Constants are all UPPER_CASE

apps/web/src/components/MetadataProvider.tsx
```diff
@@ -56,8 +56,9 @@ export function MetadataProvider({ children }: MetadataProviderProps) {
         currentWorker.terminate()
         setWorker(null)
       }
+      reset()
```
learningbizz, line 59: We used to not reset the worker after we closed it, making if we logged in with another account, the prompt from the previous user to be in the new users editor

apps/web/src/actions/datasets/generateDataset.ts
learningbizz: I agree, but just making it more comfortable for them. If they want to say false, it will break for them when creating the dataset

apps/web/src/app/(onboarding)/onboarding-dataset/paste-your-prompt/_components/PasteYourPrompt/index.tsx
learningbizz: I've changed it, adding a max to the root, but I had to add a min width now because we want the editor to be 560px and when it wasnt there, it would distribute equally in both columns taking away space from the editor

apps/web/src/app/(onboarding)/onboarding-dataset/generate-dataset/_components/GenerateDataset/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const { initialValue, documentParameters } = useDatasetOnboarding()
+  const router = useNavigate()
+  const moveNextStep = useCallback(() => {
+    router.push(ROUTES.onboarding.dataset.runExperiment)
+  }, [router])
+
+  return (
+    <div className='flex flex-row items-center gap-10 h-full w-full'>
+      <div className='flex flex-col items-end w-full h-full'>
+        <div className='relative flex-1 w-full max-h-[350px] min-w-[560px]'>
```
andresgutgon, line 24: No need to change but too many of this adhoc -[SOME_RANDOM_NUMBERpx] is a smell that you're not using the full potential of flexbox

apps/web/src/components/MetadataProvider.tsx
andresgutgon: I see this ok. cc @geclos

apps/web/src/stores/datasetOnboarding.ts
```diff
@@ -0,0 +1,39 @@
...
+import {
+  useLocalStorage,
+  AppLocalStorage,
+} from '@latitude-data/web-ui/hooks/useLocalStorage'
+
+/**
+ * Store for managing dataset onboarding state.
+ * Provides access to onboarding step navigation and persisted values (initialValue, documentParameters).
+ */
+export function useDatasetOnboarding() {
```
andresgutgon: This is not a store. It is not backed by SWR. Also, put this code next to where it is used. No need to be globally accessible, no?

apps/web/src/stores/datasets.ts
```diff
@@ -137,6 +139,40 @@ export default function useDatasets(
...
+        description: error.message,
+        variant: 'destructive',
+      })
+    },
+    onSuccess: ({ data: dataset }) => {
+      mutate([...data, dataset])
+    },
+  })
+
+  const {
```
andresgutgon, line 159: Same. This could be in the onboarding folder. No need to make globally available

packages/core/src/constants.ts
```diff
@@ -59,7 +59,7 @@ export const LATITUDE_DOCS_URL = 'https://docs.latitude.so'
 export const LATITUDE_EMAIL = 'hello@latitude.so'
 export const LATITUDE_SLACK_URL =
   'https://join.slack.com/t/trylatitude/shared_invite/zt-35wu2h9es-N419qlptPMhyOeIpj3vjzw'
-export const DEFAULT_PROVIDER_MAX_FREE_RUNS = 100
+export const DEFAULT_PROVIDER_MAX_FREE_RUNS = 400
```
andresgutgon: Is this ok?

packages/core/src/services/onboardingResources/createDatasetOnboarding.ts
```diff
@@ -0,0 +1,36 @@
...
+        user,
+      },
+      transaction,
+    )
+    if (projectResult.error) {
+      return projectResult
+    }
+    const { project, documents, commit } = projectResult.unwrap()
+
+    return Result.ok({
```
andresgutgon, line 30: Is this service creating a dataset?

packages/web-ui/src/lib/hooks/useLocalStorage.ts
```diff
@@ -20,6 +20,10 @@ export enum AppLocalStorage {
   chatDebugMode = 'chatDebugMode',
   latteThread = 'latteThread',
   latteSidebarWidth = 'latteSidebarWidth',
+  datasetOnboardingInitialValue = 'datasetOnboardingInitialValue',
+  datasetOnboardingParameters = 'datasetOnboardingParameters',
+  datasetOnboardingLatestDatasetName = 'datasetOnboardingLatestDatasetName',
+  datasetOnboardingStep = 'datasetOnboardingStep',
```
andresgutgon: Why 4 items for the onboarding? We could store all in one object under the same key

packages/core/src/constants.ts
learningbizz: Nope that was me to be able to test more hehehe, deleting!

packages/core/src/services/onboardingResources/createDatasetOnboarding.ts
learningbizz: This will be better renamed when deleting the previous onboarding, as I want to call this the prompt engineering onboarding as the one we already have

packages/web-ui/src/lib/hooks/useLocalStorage.ts
learningbizz: Good idea

apps/web/src/stores/datasetOnboarding.ts
learningbizz: what would you call this then if not a store? Regarding accessibility, makes sense

apps/web/src/app/(onboarding)/onboarding-dataset/generate-dataset/_components/GenerateDataset/index.tsx
learningbizz: Agree, open for specific feedback if you have any, however, the specific designs that we have require quite strict constraints of size which I believe requires these constraints

apps/web/src/stores/datasets.ts
learningbizz: We need it here because we need to mutate the SWR data with the new dataset created

andresgutgon: There are ways to do that without moving the code here but is fine

PR #1882, opened 2025-10-28
Publish toolExecuted events for Latitude Tools

csansoon:


packages/core/src/lib/streamManager/resolveTools/latitudeTools.ts
```diff
@@ -124,12 +125,15 @@ function instrumentLatitudeTool(
 ) {
   const originalExecute = definition.execute!
   definition.execute = async (...args) => {
+    const user = await findFirstUserInWorkspace(streamManager.workspace)
```
andresgutgon, line 128: Maybe user could be included in the context and avoid doing N times this query

PR #1883, opened 2025-10-29
APIv3 - Add getConversation endpoint

learningbizz:
Adding a new `getConversation` endpoint to be able to extract the entire message history of a conversation given its uuid.

# TODOs
- [x] Create endpoint
- [x] Test it
- [x] Add documentation

### DIFF
--- a/packages/core/src/services/users/createUser.ts
+++ b/packages/core/src/services/users/createUser.ts
@@ -13,9 +13,9 @@ export async function createUser(
     title,
   }: {
     email: string
-    name: string | null
-    confirmedAt?: Date | null
-    title?: UserTitle | null
+    name: string
+    confirmedAt?: Date
+    title?: UserTitle
   },
   transaction = new Transaction(),
 ) {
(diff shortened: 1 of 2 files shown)


packages/core/src/tests/factories/users.ts
```diff
@@ -1,24 +1,17 @@
...
 function makeRandomUserData() {
   return {
     name: faker.person.firstName(),
     email: faker.internet.email(),
-    password: faker.internet.password(),
   }
 }
-export type ICreateUser = {
-  name?: string
-  email?: string
```
learningbizz, line 14: This was old news, updating it to be up-to-date

packages/core/src/services/users/createUser.ts
```diff
@@ -13,9 +13,9 @@ export async function createUser(
     title,
   }: {
     email: string
-    name: string
-    confirmedAt?: Date
-    title?: UserTitle
+    name: string | null
+    confirmedAt?: Date | null
```
learningbizz: Wasn't aligned with the DB schema, now it is

PR #1887, opened 2025-10-30
Experiments in Runs

csansoon:
Run Experiments in the Runs background.

There's no custom Job for Experiments anymore. Instead, it just uses the same `backgroundRunJob`! Just plug in a `simulationSettings`, an optional `experimentId` for updating the experiment when starting and finishing, and all done! Same service for all prompt executions!


packages/core/src/repositories/runsRepository.ts
```diff
@@ -204,7 +222,12 @@ export class RunsRepository {
         if (listing.error) return Result.error(listing.error)
         const active = listing.value
 
-        active[runUuid] = { uuid: runUuid, queuedAt, startedAt: undefined }
+        active[runUuid] = {
+          uuid: runUuid,
+          queuedAt,
+          startedAt: undefined,
+          source,
+        }
```
neoxelox:
```suggestion
        active[runUuid] = { uuid: runUuid, queuedAt, source }
```

packages/core/src/jobs/job-definitions/runs/backgroundRunJob.ts
```diff
@@ -108,6 +177,17 @@ export const backgroundRunJob = async (
...
     await endRun({ workspaceId, projectId, runUuid })
+
+    if (experiment) {
+      await updateExperimentStatus(
+        {
+          workspaceId,
+          experiment,
+        },
+        (progressTracker) =>
+          progressTracker.incrementErrors(experiment!.evaluationUuids.length),
```
neoxelox, line 184: increment errors? this is a finally no? this should be in the catch maybe?

csansoon: oh yeah mb

PR #1892, opened 2025-10-31
Filter runs

csansoon:
New Tab filters in "Runs", to group "Production" runs (API, SDK, Triggers...) and "Playground" runs (Playground, Experiments...)

image

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx
@@ -6,32 +6,106 @@ import {
   useCompletedRuns,
   useCompletedRunsCount,
 } from '$/stores/runs/completedRuns'
-import { ActiveRun, CompletedRun, Run } from '@latitude-data/constants'
+import {
+  ActiveRun,
+  CompletedRun,
+  LogSources,
+  Run,
+  RUN_SOURCES,
+  RunSourceGroup,
+} from '@latitude-data/constants'
 import { Pagination } from '@latitude-data/core/helpers'
 import { ProjectLimitedView } from '@latitude-data/core/schema/models/types/Project'
 import { SplitPane } from '@latitude-data/web-ui/atoms/SplitPane'
 import { Text } from '@latitude-data/web-ui/atoms/Text'
-import { useEffect, useMemo, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState } from 'react'
 import { useDebounce } from 'use-debounce'
 import { RunPanel } from './RunPanel'
 import { RunsList } from './RunsList'
+
+function sumCounts(
+  counts: Record<LogSources, number> | undefined,
+  sourceGroup: RunSourceGroup,
+): number {
+  if (!counts) return 0
+  const sources = RUN_SOURCES[sourceGroup]
+  return sources.reduce((sum, source) => sum + (counts[source] ?? 0), 0)
+}
+
+// Group search state by selected sourceGroup tab
+function useSearchByGroup({
+  defaultSearch,
+  sourceGroup,
+}: {
+  defaultSearch: Pagination
+  sourceGroup: RunSourceGroup
+}): {
+  search: Pagination
+  setSearch: (search: Pagination) => void
+  debouncedSearch: Pagination
+} {
+  const [searchByGroup, setSearchByGroup] = useState<
+    Record<RunSourceGroup, Pagination>
+  >({
+    ...(Object.fromEntries(
+      Object.values(RunSourceGroup).map((group) => [
+        group,
+        { page: 1, pageSize: defaultSearch.pageSize },
+      ]),
+    ) as Record<RunSourceGroup, Pagination>),
+    [sourceGroup]: defaultSearch,
+  })
+
+  const search = useMemo(
+    () => searchByGroup[sourceGroup],
+    [searchByGroup, sourceGroup],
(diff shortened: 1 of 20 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx
```diff
@@ -33,6 +51,10 @@ export function RunsPage({
   const [completedSearch, setCompletedSearch] = useState(serverCompleted.search)
   const [debouncedCompletedSearch] = useDebounce(completedSearch, 100)
 
+  const [sourceGroup, setSourceGroup] =
+    useState<RunSourceGroup>(defaultSourceGroup)
+  const [debouncedSourceGroup] = useDebounce(sourceGroup, 100)
```
Comment (learningbizz) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx:90]: Nitpick, isnt 100ms really little? We use 500 normally

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/page.tsx
```diff
@@ -17,21 +20,30 @@ export default async function RunsPage({
...
   }
   const completedSearch = {
     page: completedPage ? Number(completedPage) : undefined,
     pageSize: completedPageSize ? Number(completedPageSize) : undefined,
+    sourceGroup: (sourceGroup as RunSourceGroup) ?? defaultSourceGroup,
   }
 
-  const activeRuns = await listActiveRunsCached({ projectId, ...activeSearch}) // prettier-ignore
-  const completedRuns = await listCompletedRunsCached({ projectId, ...completedSearch}) // prettier-ignore
+  const activeRuns = await listActiveRunsCached({ projectId, ...activeSearch }) // prettier-ignore
```
Comment (learningbizz) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/page.tsx:45]: why?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/page.tsx
```diff
@@ -17,21 +20,30 @@ export default async function RunsPage({
...
+    completedPageSize,
+  } = await searchParams
+
+  const defaultSourceGroup = RunSourceGroup.Production
 
   const projectId = Number(_projectId)
   const activeSearch = {
     page: activePage ? Number(activePage) : undefined,
     pageSize: activePageSize ? Number(activePageSize) : undefined,
+    sourceGroup: (sourceGroup as RunSourceGroup) ?? defaultSourceGroup,
```
Comment (learningbizz) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/page.tsx:37]: Nitpick: As we only have some source groups, adding validation  here in the backend would be good.

I'd suggest a zod schema for the params and a parse to make sure they are the correct values, and we additionally get the type guarded

apps/web/src/stores/runs/completedRuns.ts
```diff
@@ -96,16 +104,19 @@ export function useCompletedRunsCount(
   opts?: SWRConfiguration,
 ) {
   const route = ROUTES.api.projects.detail(project.id).runs.completed.count
-  const fetcher = useFetcher<number>(disable ? undefined : route, {
-    serializer: (data) => (data as any)?.count,
-    fallback: null,
-  })
+  const fetcher = useFetcher<Record<LogSources, number>>(
+    disable ? undefined : route,
```
learningbizz, line 108: what is this?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/page.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/page.tsx:45]: why... what? the space at the end? It's just for adding padding, easier to read. Prettier does that automatically but this line has been set to ignore

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx:90]: It depends on the use case. waiting for 500 without input feels really slow when loading the page.
Also, this is not something new from this PR. I'm just following the debounce time used in this page for every other parameter

apps/web/src/stores/runs/completedRuns.ts
csansoon: I changed what the route returns, so I had to update the fetcher types.
You can see the changes in the route. It now returns the run count grouped by source instead of as a whole.

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/page.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/page.tsx:37]: It's not a security concern and it will only fail if users try to tweak it manually. No need for a safeguard really

apps/web/src/stores/runs/completedRuns.ts
learningbizz: I'm more confused with the disable argument, sorry for not being specific

csansoon: We do this everywhere. Sometimes you need the hook in the component but you don't want to perform a fetch automatically (for example, if the hook fetches the data for a selected item but there's no item selected). In that case, we use `undefined` as the `useFetcher` route. When that's the case, no fetch is requested.
In this case, there's a `disable` attribute. When it is true, we won't perform the fetch.

However, I did not change that in the PR! I just changed the type from the fetcher response data, and that triggered prettify to add a few line breaks

PR #1899, opened 2025-11-04
Redirect onboarding to playground runs

learningbizz:
With the new addition of the Playground/Production runs, we are redirecting the onboarding users to the runs page and they see no logs, as they are in the Playground tab.

image

This PR redirects users to the playground tab

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx
@@ -1,6 +1,8 @@
 'use client'
 
 import { useCurrentProject } from '$/app/providers/ProjectProvider'
+import { useCurrentCommit } from '$/app/providers/CommitProvider'
+import { ROUTES } from '$/services/routes'
 import { useActiveRuns, useActiveRunsCount } from '$/stores/runs/activeRuns'
 import {
   useCompletedRuns,
@@ -84,6 +86,7 @@ export function RunsPage({
   defaultSourceGroup: RunSourceGroup
 }) {
   const { project } = useCurrentProject()
+  const { commit } = useCurrentCommit()
 
   const [sourceGroup, setSourceGroup] =
     useState<RunSourceGroup>(defaultSourceGroup)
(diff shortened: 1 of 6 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx
```diff
@@ -108,21 +111,28 @@ export function RunsPage({
...
+      .runs.root({
+        activePage: debouncedActiveSearch.page,
+        activePageSize: debouncedActiveSearch.pageSize,
+        completedPage: debouncedCompletedSearch.page,
+        completedPageSize: debouncedCompletedSearch.pageSize,
+        sourceGroup: debouncedSourceGroup,
+      })
 
-    const targetUrl = `${currentUrl}${queryParams ? `?${queryParams}` : ''}`
+    const targetUrl = `${window.location.origin}${runsRoute}`
```
Comment (csansoon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/runs/_components/RunsPage.tsx:125]: cool

apps/web/src/services/routes.ts
```diff
@@ -253,7 +253,26 @@ export const ROUTES = {
...
+                  const searchParams = new URLSearchParams()
+                  if (params.activePage !== undefined) searchParams.set('activePage', String(params.activePage)) // prettier-ignore
+                  if (params.activePageSize !== undefined) searchParams.set('activePageSize', String(params.activePageSize)) // prettier-ignore
+                  if (params.sourceGroup !== undefined) searchParams.set('sourceGroup', params.sourceGroup) // prettier-ignore
+                  if (params.completedPage !== undefined) searchParams.set('completedPage', String(params.completedPage)) // prettier-ignore
+                  if (params.completedPageSize !== undefined) searchParams.set('completedPageSize', String(params.completedPageSize)) // prettier-ignore
+
+                  const query = searchParams.toString()
+                  return query ? `${base}?${query}` : base
+                },
```
csansoon, line 275: nice

csansoon (approved):

apps/web/src/services/routes.ts
andresgutgon: Routes is already big. Already talked offline. No need to change but I think doing query params in routes can be really hard to manage this file. 

Anothe option would be a helper function where `runRoute({ some, conditions })` and it produce the right route from ROUTES + query params

PR #1902, opened 2025-11-04
Fix editor size issue + create HITL evaluation

learningbizz:
We didnt create the HITL evaluationn for the onboarding prompt, and the editor size in step 2 and 3 wasn't showing all the text, which was acting strange for some users.

This PR fixes all of this.


packages/core/src/services/projects/createOnboardingProject.ts
```diff
@@ -35,6 +35,7 @@ export async function createOnboardingProject(
       user,
       commit,
       path: 'onboarding',
+      createDemoEvaluation: true,
```
andresgutgon, line 38: Test

PR #1907, opened 2025-11-05
Add integrate in production banner

learningbizz:
Adding a new banner in the sidebar to try and increase the amount of users integrating our SDK. This will only show if the user hasn't integrated with us with the API, SDK, or has published an agent and used it with triggers

image

### DIFF
--- a/docs/guides/integration/overview.mdx
+++ b/docs/guides/integration/overview.mdx
@@ -22,7 +22,7 @@ You can also interact with Latitude directly using the [HTTP API](/guides/api/ap
 
 Latitude supports two main integration patterns:
 
-### Use Latitude as an AI Gateway
+### Use Latitude as an AI Gateway (SDK and API support)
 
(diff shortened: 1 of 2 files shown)


apps/web/src/components/Sidebar/index.tsx
```diff
@@ -3,13 +3,16 @@ import { ReactNode } from 'react'
...
 }: {
   header: ReactNode
   tree: ReactNode
+  banner?: ReactNode
 }) {
   return (
-    <aside className='flex flex-col gap-y-2 w-full max-h-full overflow-y-hidden'>
-      <div className='p-4 py-6 gap-y-2'>{header}</div>
+    <aside className='flex flex-col gap-y-4 w-full max-h-full overflow-y-hidden p-4'>
+      {banner}
```
andresgutgon, line 14: Just put the banner here. No need to configure Sidebar which is only used once

learningbizz: As talked offline, we can always change the banner in the future for different CTAs we're looking for, this makes it reusable and easier to read

PR #1911, opened 2025-11-06
Experiment Simulation Settings

csansoon:
Added a new section in Experiment creation Modal to tweak the experiment simulations! Decide which tools you want to simulate, and which should be executed. Control how the simulation will run and what's the expected result it should generate.

https://github.com/user-attachments/assets/cac9455a-e7f5-4f88-9c19-06d5d6fa8c75

### DIFF
--- a/apps/web/src/components/RunExperimentModal/ExperimentForm/_components/DatasetSelector.tsx
+++ b/apps/web/src/components/RunExperimentModal/ExperimentForm/_components/DatasetSelector.tsx
@@ -39,11 +39,11 @@ export function DatasetSelector({
   )
 
   if (isLoadingDatasets) {
-    return <Skeleton height='h2' className='w-1/2' />
+    return <Skeleton height='h2' className='w-2/3' />
   }
 
   return (
-    <div className='flex flex-row items-center gap-4 w-1/2'>
+    <div className='flex flex-row items-center gap-4 w-2/3'>
       {datasets.length > 0 && (
         <div className='flex-grow'>
           <Select
@@ -70,7 +70,7 @@ export function DatasetSelector({
           </>
         )}
         {parameters === undefined ? (
-          <Skeleton height='h2' className='w-1/2' />
+          <Skeleton height='h2' className='w-2/3' />
         ) : (
           <Link
             className='flex flex-row items-center gap-1 hover:underline'
--- a/apps/web/src/components/RunExperimentModal/ExperimentForm/_components/SimulationSettings.tsx
+++ b/apps/web/src/components/RunExperimentModal/ExperimentForm/_components/SimulationSettings.tsx
@@ -1,0 +1,325 @@
+import { Text } from '@latitude-data/web-ui/atoms/Text'
+import { ExperimentFormPayload } from '../useExperimentFormPayload'
+import { Icon, IconName } from '@latitude-data/web-ui/atoms/Icons'
+import { SwitchToggle } from '@latitude-data/web-ui/atoms/Switch'
+import { ReactNode, useCallback, useMemo } from 'react'
+import {
+  BackgroundColor,
+  colors,
+  TextColor,
+} from '@latitude-data/web-ui/tokens'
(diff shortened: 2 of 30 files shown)


andresgutgon:
I would add a search box as we have in the editor. Otherwhise are too many tools to scan with the eye
image

packages/web-ui/src/ds/atoms/TextArea/index.tsx
```diff
@@ -74,7 +74,7 @@ const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
               {
                 'border-red-500 focus-visible:ring-red-500': errors,
                 // Account for inner textarea padding
-                '!min-h-[calc(100%-theme(spacing.6))]': autoGrow,
+                '!min-h-[calc(100%-theme(spacing.6))] resize-none': autoGrow,
```
andresgutgon, line 77: Why this?

packages/web-ui/src/ds/atoms/Icons/index.tsx
```diff
@@ -357,6 +358,7 @@ const Icons = {
   search: SearchIcon,
   sentry: Sentry,
   settings: SettingsIcon,
+  settings2: Settings2,
```
andresgutgon: Do we need a different icon? Can't we just swap 1 by 2 if 2 is nicer?

packages/web-ui/src/ds/atoms/TextArea/index.tsx
csansoon: It `autoGrow` is enabled, we want it to grow based on the content, not the user input... right?

packages/web-ui/src/ds/atoms/Icons/index.tsx
csansoon: They're different stuff, I feel like they convey different meanings.
Anyways, it's not actually being used anymore so I'll remove it

apps/web/src/components/RunExperimentModal/ExperimentForm/_components/DatasetSelector.tsx
```diff
@@ -39,11 +39,11 @@ export function DatasetSelector({
   )
 
   if (isLoadingDatasets) {
-    return <Skeleton height='h2' className='w-1/2' />
+    return <Skeleton height='h2' className='w-2/3' />
   }
 
   return (
-    <div className='flex flex-row items-center gap-4 w-1/2'>
+    <div className='flex flex-row items-center gap-4 w-2/3'>
```
learningbizz, line 46: nitpicking here, but I see we're hardcoding a lot this 2/3, substituting this by the 1/2 of before, creating a constant for this would be easier to change in the future if we want to

apps/web/src/components/RunExperimentModal/ExperimentForm/_components/SimulationSettings.tsx
```diff
@@ -0,0 +1,323 @@
...
+}
+
+const getToolLabel = (toolName: string, manifest: ToolManifest): string => {
+  const sourceData = manifest.sourceData as ToolSourceData
+  if (sourceData.source === ToolSource.Integration && sourceData.toolLabel) return sourceData.toolLabel // prettier-ignore
+  if (sourceData.source === ToolSource.Latitude) return capitalize(sourceData.latitudeTool) // prettier-ignore
+  return toolName
+}
+
+const getForcedSimValue = (manifest: ToolManifest): boolean | undefined => {
```
learningbizz: took me a sec to understand this const, words are free in code, add the whole word please!!

apps/web/src/components/RunExperimentModal/ExperimentForm/_components/DatasetSelector.tsx
csansoon: this ui is getting revamped. It's pretty ugly

learningbizz (approved):

PR #1913, opened 2025-11-07
Small background run improvements

learningbizz:
So...

We were getting a few "conversations not found" errors when users tried to attach to a conversation that already ended.
However, we had a miss-alignment between the cleanup time of the Redis stream and the bullMQ job, where the Redis stream takes 10s to remove, while BullMQ removes the job instantaniously once completed. 

This PR aligns these times, and additionally adds a maintenance job to cleanup orphan background jobs that remain in the Redis cache due to possible edge cases.

### DIFF
--- a/packages/core/src/jobs/job-definitions/maintenance/cleanupOrphanedRunsJob.ts
+++ b/packages/core/src/jobs/job-definitions/maintenance/cleanupOrphanedRunsJob.ts
@@ -3,6 +3,8 @@ import { cache } from '../../../cache'
 import { queues } from '../../queues'
 import { Job } from 'bullmq'
 import { endRun } from '../../../services/runs/end'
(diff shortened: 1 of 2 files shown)


packages/core/src/jobs/job-definitions/maintenance/cleanupOrphanedRunsJob.ts
```diff
@@ -0,0 +1,88 @@
...
+        error,
+      )
+    }
+  }
+
+  if (orphanedRuns > 0) {
+    console.log(
+      `Cleaned up ${orphanedRuns} orphaned runs in workspace ${workspaceId}, project ${projectId}`,
+    )
+  }
```
andresgutgon, line 91: Remove this before merge

PR #1915, opened 2025-11-10
Replace JSON value with Hash in background runs Redis mechanism

learningbizz:
We're having an issue with our current background runs mechanism that it doesn't handle well concurrency.

image

At the moment, we're using a JSON object for saving our active runs for a given workspace and project in a Redis table. To create/update/delete this JSON, we lock the Redis key/value pair until the action finished using an exponential backoff to avoid race conditions. 

This is creating a bottleneck, as when a specific workspace/project runs many documents (i.e. experiments that now run on background runs) our system runs one created/updated/deleted at a time.

To solve this issue, I propose a fast, good solution that can sustain better concurrency to up to around 10K active runs per project and workspace. Taking into account our current rate limits for our cloud users (team, pro and hobby), 10K is enough to sustain the possibility of a user spamming runs as much as possible.

This solution is by changing the data type of the Redis value to a [Redis hash](https://redis.io/docs/latest/develop/data-types/hashes/) instead of a JSON. A Redis hash, with field/value pairs like a record, enables us to atomically add, delete and update each field/value without having to lock the entire hash with O(1) complexity. 

In addition, I added an expiry date to EACH KEY for the TTL of 3 hours, like this no runs are saved indefinitely by accident.

Therefore, we can still have our keys by workspace and project, and have our hash by field (runUuid) and value a JSON with the uuid, enquedAt date and source.

### DIFF
--- a/packages/core/src/repositories/runsRepository.ts
+++ b/packages/core/src/repositories/runsRepository.ts
@@ -85,14 +85,14 @@ export class RunsRepository {
     cache = cache ?? (await redis())
 
     try {
-      // Use HGETALL to get all runs from hash at once
+      // Use HGETALL to get all runs from a workspace/project hash at once (O(N) but entire hash expires in 3 hours, so N won't be too large)
       const hashData = await cache.hgetall(key)
 
       if (!hashData || Object.keys(hashData).length === 0) {
         return Result.ok<Record<string, ActiveRun>>({})
       }
 
-      // Parse each hash value (JSON string) to ActiveRun
+      // Parse each hash value (JSON string) to an ActiveRun object
       const active: Record<string, ActiveRun> = {}
       const now = Date.now()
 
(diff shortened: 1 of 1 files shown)


packages/core/src/repositories/runsRepository.ts
```diff
@@ -278,51 +313,64 @@ export class RunsRepository {
...
-      },
-    })
+      return Result.error(
+        new NotFoundError(`Run not found with uuid ${runUuid}`),
+      )
+    } catch (error) {
+      return Result.error(error as Error)
+    }
   }
 }
```
geclos, line 375: good occasion to move the write operations to services and add unit tests

learningbizz: Can do that in a separate PR, don't want to add more complexity here

learningbizz: Doing it rn in another PR

PR #1917, opened 2025-11-10
Chore/separate runs repo

learningbizz:
Separating the runsRepository into different services, together with adding tests to test each individual service.


packages/core/src/services/runs/active/listCached.ts
```diff
@@ -0,0 +1,52 @@
...
+  try {
+    // Use HGETALL to get all runs from a workspace/project hash at once (O(N) but entire hash expires in 3 hours, so N won't be too large)
+    const hashData = await redisCache.hgetall(key)
+
+    if (!hashData || Object.keys(hashData).length === 0) {
+      return Result.ok<Record<string, ActiveRun>>({})
+    }
+
+    // Parse each hash value (JSON string) to an ActiveRun object
+    const active: Record<string, ActiveRun> = {}
```
andresgutgon: Maybe you could just return
```javascript
const activeRuns: ActiveRun[] = []
```

Outside I see you do in `listActive` `Object.values`

PR #1918, opened 2025-11-10
Fix double count in runs page

learningbizz:
We had an issue where if you were in the runs page and you ran documents from somewhere else, the counter would count twice each run. 

This was because we had the sidebar and the runs page using the useActiveRunsCount hook, which would create two listeners of the WS event, and mutate twice the frontend state. 

This fix fixes that by creating a provider that uses this hook, and instantiating it in the layout for the children to use, and avoid duplicate counts.

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/ActiveRunsCountProvider.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/ActiveRunsCountProvider.tsx
@@ -3,7 +3,7 @@
 import { useCurrentProject } from '$/app/providers/ProjectProvider'
 import { useActiveRunsCount } from '$/stores/runs/activeRuns'
 import { LogSources } from '@latitude-data/constants'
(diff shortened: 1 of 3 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/ActiveRunsCountProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <ActiveRunsCountContext.Provider value={value}>
+      {children}
+    </ActiveRunsCountContext.Provider>
+  )
+}
+
+export function useActiveRunsCountContext() {
+  const context = useContext(ActiveRunsCountContext)
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/ActiveRunsCountProvider.tsx]: Now need for custom hook for context in react. 
```javascript
const myThing = use(ActiveRunsCountContext) 
```
Takes care of checking the provider is provided

Comment (learningbizz) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/_components/ActiveRunsCountProvider.tsx]: top tyty for the KNOWLEDGE

PR #1921, opened 2025-11-11
Apply new visuals to issues table

andresgutgon:
# What?
Some design changes on the issues table

## TODO
- [x] Badges for different statuses
- [x] New design for histogram 
- [x] Remove regressed filter
- [x] Actions for resolving or ignoring an issue. Validate before ignoring is not resolved and the other way around.
- [x] Avoid re-fetching issue stats when optimistically the issue is `resolved` or `ignored`. Move all this fetching logic to the issues table and do a local state like `waitingForUserUndoAction` which when is true we return avoid refetching. Merge useMemo and `useEffect` We don't need `useMemo`
- [x] Invalidate the cache of the list of issues where the action goes. This way is immediately re-refeched. (Use mutate)
- [x] Why are the histogram stats refetched if we're not doing it automatically in the store?
- [x] Rebase PR and merge

## Question
~~What consequences does resolving or ignoring an issue have on the vector DB? I think we should remove from the pull of issues~~
We do nothing

# Next
- [ ] Add tests for issues dashboard filtering SQL
- [ ] Add some tests for ignoring and resolving
- [ ] When ignoring an issue remove from Vector DB
- [ ] When ignoring an issue do not use for the `centroid`

## Escalating spec
```bash
Escalating if:
There are events in the last 1 day, and
The 7-day event count is > 2× the previous 7-day average, and
The total count in that 7-day window is above a minimum threshold (say 20 events, to avoid noise).
```


apps/web/src/actions/evaluationsV2/issues/unignoreIssue.ts
```diff
@@ -0,0 +1,25 @@
...
+      .find(parsedInput.issueId)
+      .then((r) => r.unwrap())
+
+    const response = await unignoreIssue({
+      issue,
+      user: ctx.user,
+    }).then((r) => r.unwrap())
+
+    return response.issue
+  })
```
csansoon, line 25: "unignore" lol

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/issues/_components/HistogramCell/index.tsx
```diff
@@ -1,71 +1,183 @@
...
+      case 3:
+        // Random spiky
+        count = Math.round(Math.random() * 12 * noise)
+        break
+    }
+
+    bars.push({ date: `placeholder-${i}`, count: Math.max(count, 1) })
+  }
+  return bars
+}
```
Comment (csansoon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/issues/_components/HistogramCell/index.tsx:55]: wait till i can lay my hands on this 😈

csansoon (approved):

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/issues/_components/HistogramCell/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/issues/_components/HistogramCell/index.tsx:55]: 😂 miedo

PR #1922, opened 2025-11-11
Run Experiments from Logs

csansoon:
Experiments can not only be executed from Datasets.
They can also be manually configured (when no parameters are found), and they can now be executed from past logs!

This means we have to update the metadata payload to define different properties depending on the source of the experiment data.


apps/web/src/components/RunExperimentModal/index.tsx
```diff
@@ -73,20 +73,63 @@ export function RunExperimentModal({
...
-    if (formPayload.parameters.length > 0 && !formPayload.selectedDataset) {
+    const source = formPayload.selectedParametersSource
+
+    // Validation: if using dataset source and have parameters, need dataset selected
+    if (
+      source === 'dataset' &&
+      formPayload.parameters.length > 0 &&
+      !formPayload.selectedDataset
+    ) {
+      return
```
learningbizz, line 91: why not return a failed toast here to tell the user?

packages/constants/src/experiments.ts
```diff
@@ -8,3 +9,77 @@ export const experimentVariantSchema = z.object({
...
+export type ExperimentDatasetSource = IExperimentParametersSource & {
+  source: 'dataset'
+  datasetId: number
+  fromRow: number
+  toRow: number
+  datasetLabels: Record<string, string>
+  parametersMap: Record<string, number>
+}
+
+const experimentDatasetSourceSchema = z.object({
```
learningbizz, line 14: Nitpick: instead of writing this, couldnt we infer the experimentDataSource?

packages/constants/src/experiments.ts
```diff
@@ -8,3 +9,77 @@ export const experimentVariantSchema = z.object({
...
+  parametersMap: z.record(z.string(), z.number()),
+})
+
+// Experiment ran from last logs (from commit and creation time of experiment)
+export type ExperimentLogsSource = IExperimentParametersSource & {
+  source: 'logs'
+  count: number
+}
+
+const experimentLogsSourceSchema = z.object({
```
learningbizz, line 28: same as above

apps/web/src/components/RunExperimentModal/index.tsx
csansoon: sure, good idea!

packages/constants/src/experiments.ts
csansoon: wdym? This is for the 'create' server action, that requires a zod schema for input validation

packages/constants/src/experiments.ts
csansoon: same as above

packages/constants/src/experiments.ts
learningbizz: my bad, thought you could do z.infer in reverse, getting a zod from a type.

As its the other way around, why not part from the schema and do a z.infer to get the type? Like this we have one source of truth and not 2

packages/constants/src/experiments.ts
```diff
@@ -8,3 +9,77 @@ export const experimentVariantSchema = z.object({
 })
 
 export type ExperimentVariant = z.infer<typeof experimentVariantSchema>
+
+interface IExperimentParametersSource {
+  source: string
+}
```
andresgutgon: 🔥

packages/constants/src/experiments.ts
```diff
@@ -8,3 +9,77 @@ export const experimentVariantSchema = z.object({
...
+
+const experimentDatasetSourceSchema = z.object({
+  source: z.literal('dataset'),
+  datasetId: z.number(),
+  fromRow: z.number(),
+  toRow: z.number(),
+  datasetLabels: z.record(z.string(), z.string()),
+  parametersMap: z.record(z.string(), z.number()),
+})
```
andresgutgon, line 21: Infer

PR #1928, opened 2025-11-12
fix: Running prompts with custom tools from SDK and API

csansoon:
Fixes:
 - Running /run or /chat with custom tools and background=true did not fail.
 - Running /run or /chat endpoint with custom tools and stream=false failed with a confusing error message.
 - Running /chat with custom tools and stream=true was not working but did not show an error either.


apps/gateway/src/routes/api/v3/projects/versions/documents/run/run.handler.ts
```diff
@@ -35,6 +35,19 @@ export const runHandler: AppRouteHandler<RunRoute> = async (c) => {
...
+    if (background) {
+      throw new BadRequestError(
+        'Custom tools are not supported in background runs',
+      )
+    }
+
+    if (!useSSE) {
+      throw new BadRequestError('You must enable Stream to use custom tools')
+    }
+  }
```
andresgutgon, line 49: nice

andresgutgon (approved):

PR #1930, opened 2025-11-13
Feat/add connect eval to issues

learningbizz:
Adding preliminary capability for issues to be linked with evaluations in the settings page.

This feature will be completed better over the next few days, but this is enough for the webinar for William to link a created eval to an existing issue.

image

# TODOS
- [x] Integration test 
- [x] Update update eval action
- [x] Update create eval action
- [x] Add select in create/update eval modal for issues without expected output 
- [ ] Investigate why all eval results are not attaching to their issue id
- [ ] Talk if the evaluation results should be logged
- [ ] Fix to-dos evaluation generation (especially Andres request)

### DIFF
--- a/apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/issues/search/route.ts
+++ b/apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/issues/search/route.ts
@@ -15,7 +15,6 @@ export type SearchIssueResponse = Awaited<
 
 const paramsSchema = z.object({
   projectId: z.coerce.number(),
-  commitUuid: z.string(),
   documentUuid: z.string(),
 })
 
@@ -37,7 +36,6 @@ export const GET = errorHandler(
       const query = request.nextUrl.searchParams
       const { projectId, documentUuid } = paramsSchema.parse({
         projectId: params.projectId,
-        commitUuid: params.commitUuid,
         documentUuid: query.get('documentUuid'),
       })
       const title = query.get('query')
--- a/apps/web/src/components/evaluations/ConfigurationForm.tsx
+++ b/apps/web/src/components/evaluations/ConfigurationForm.tsx
@@ -307,6 +307,7 @@ export function ConfigurationAdvancedForm<
         >
           <Select
             searchable={true}
+            // TODO(eval-generation): This will not get the issueId from page 2,3, etc. need to change with Andres feedback
             value={issueId ?? null}
             name='issueId'
             placeholder='Select an existing issue'
@@ -318,7 +319,7 @@ export function ConfigurationAdvancedForm<
               value: issue.id,
             }))}
             onSearch={onSearch}
-            onChange={(value) => setIssueId(value)}
+            onChange={setIssueId}
             errors={errors?.['issueId']}
           />
         </FormFieldGroup>
--- a/apps/web/src/components/evaluations/EvaluationV2Form.tsx
+++ b/apps/web/src/components/evaluations/EvaluationV2Form.tsx
@@ -162,12 +162,6 @@ export default function EvaluationV2Form<
 
   const commitMerged = mode === 'update' && !!commit.mergedAt
 
-  const handleSetIssueId = (value: number | null) => {
-    if (setIssueId) {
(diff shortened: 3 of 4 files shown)


apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/issues/search/route.ts
```diff
@@ -36,26 +35,19 @@ export const GET = errorHandler(
...
-      const commitsRepo = new CommitsRepository(workspace.id)
-      const commit = await commitsRepo
-        .getCommitByUuid({
-          projectId,
-          uuid: commitUuid,
-        })
-        .then((r) => r.unwrap())
       const docsScope = new DocumentVersionsRepository(workspace.id)
       const document = await docsScope
-        .getDocumentByUuid({
```
Comment (learningbizz) [inline apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/issues/search/route.ts:56]: This was changed as there was a bug.

We want to get the issues of a document regardless of its version.

apps/web/src/components/evaluations/ConfigurationForm.tsx
```diff
@@ -272,6 +299,30 @@ export function ConfigurationAdvancedForm<
...
       )}
+      {doesntRequireExpectedOutput && setIssueId && (
+        <FormFieldGroup
+          label='Link issue'
+          description='Link the evaluation to an issue in your agent logs'
+          layout='horizontal'
+        >
+          <Select
+            searchable={true}
+            value={issueId ?? null}
```
andresgutgon, line 311: This will not show the issue if it's not in the first page of the search query

apps/web/src/components/evaluations/EvaluationV2Form.tsx
```diff
@@ -158,6 +162,12 @@ export default function EvaluationV2Form<
 
   const commitMerged = mode === 'update' && !!commit.mergedAt
 
+  const handleSetIssueId = (value: number | null) => {
+    if (setIssueId) {
+      setIssueId(value)
```
andresgutgon: Should they be able to unset an issue? what are the consequences? Ask product

packages/core/src/services/evaluationsV2/create.test.ts
```diff
@@ -133,10 +151,13 @@ describe('createEvaluationV2', () => {
...
         id: evaluation.versionId,
         commitId: commit.id,
+        issueId,
         createdAt: expect.any(Date),
       }),
     ])
-    expect(mocks.publisher).toHaveBeenCalledExactlyOnceWith({
+    // 2 times as the issue and the evaluation are created (issue for testing cases)
+    expect(mocks.publisher).toHaveBeenCalledTimes(2)
+    expect(mocks.publisher).toHaveBeenCalledWith({
```
andresgutgon, line 160: This is wrong. Something is not right. Tests should not account for other events other than the one being tested. `evaluationV2Created`. Change as it was before

apps/web/src/components/evaluations/ConfigurationForm.tsx
andresgutgon: Check what we do on annotations form
https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/components/evaluations/Annotation/Form/IssuesSelector/index.tsx#L83-L102

We get independendently the issue associated with the eval

packages/core/src/services/evaluationsV2/create.test.ts
learningbizz: We don't have a factory yet for issues, so I used this for now :)

apps/web/src/components/evaluations/EvaluationV2Form.tsx
learningbizz: Yes they should. Maybe you don't want that eval to track that issue anymore, and you want another one to, so you remove it.

Consequences are that the failed eval results will not auto categorize under that issue

apps/web/src/components/evaluations/ConfigurationForm.tsx
learningbizz: Lets do it in the following PR, its not critical now for the webinar

PR #1931, opened 2025-11-13
feat: results from draft versions do not modify issue centroid

geclos:
This will keep results assigned to the proper issue but ensure we don't add noise to the issue's centroid.

### DIFF
--- a/packages/core/src/services/issues/results/add.ts
+++ b/packages/core/src/services/issues/results/add.ts
@@ -3,9 +3,9 @@ import {
   EvaluationMetric,
   EvaluationResultSuccessValue,
   EvaluationType,
-  ISSUE_JOBS_GENERATE_DETAILS_DEBOUNCE,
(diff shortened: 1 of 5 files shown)


packages/core/src/services/issues/results/add.ts
```diff
@@ -163,9 +163,34 @@ export async function addResultToIssue<
...
+  {
+    issue,
+    commitId,
+  }: {
+    issue: Issue
+    commitId: number
+  },
+  db = database,
+) {
+  const commitIds = await db
```
andresgutgon, line 183: Use the repository

andresgutgon (dismissed):

packages/core/src/services/issues/results/add.ts
geclos: remember you should not need to use repositories within services as tenancy has already been checked at this point

geclos: also this is a very custom query which I actually think is better to keep local

PR #1934, opened 2025-11-13
Reflect when the last histogram occurred

andresgutgon:
# What?
In the UI we show the "Last occurrence" of the issue in the table of issues. The problem is that we use a date field for the issue occurrence so we can't show if the issue occurred 1 minute ago or 10 hours ago. We need to change this

Last occurrence should reflect the time when was last. Now the column is a `date` and we lost that information.
image

### DIFF
--- a/packages/core/drizzle/0235_change_date_by_datetime_histograms.sql
+++ b/packages/core/drizzle/0235_change_date_by_datetime_histograms.sql
@@ -1,0 +1,9 @@
+-- Add occurred_at column to issue_histograms table and backfill it with date values. With timezone support.
+ALTER TABLE "latitude"."issue_histograms" ADD COLUMN "occurred_at" timestamp with time zone DEFAULT now() NOT NULL;--> statement-breakpoint
+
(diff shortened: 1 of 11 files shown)


packages/core/drizzle/0235_change_date_by_datetime_histograms.sql
```diff
@@ -0,0 +1,9 @@
+-- Add occurred_at column to issue_histograms table and backfill it with date values. With timezone support.
+ALTER TABLE "latitude"."issue_histograms" ADD COLUMN "occurred_at" timestamp with time zone DEFAULT now() NOT NULL;--> statement-breakpoint
+
+-- Backfill occurred_at column with date values (casting date to text first, then appending time)
+UPDATE "latitude"."issue_histograms" SET "occurred_at" = (CAST("date" AS text) || ' 00:00:00')::timestamp with time zone;--> statement-breakpoint
```
andresgutgon, line 5: Important to check. This backfill date column into occurred_at column.

geclos (approved):

PR #1935, opened 2025-11-14
Chore/remove todos connect issue

learningbizz:
Remove to-dos left from yesterdays connect feature for the webinar. 

# TODOs
- [x] Investigate why all eval results are not attaching to their issue id
- [x] Talk if the evaluation results should be logged
- [x]  Fix to-dos evaluation generation (especially Andres request)
- [ ] Add issue substatuses to be able to filter by those instead of only active/inactive

### DIFF
--- a/apps/web/src/components/evaluations/ConfigurationForm.tsx
+++ b/apps/web/src/components/evaluations/ConfigurationForm.tsx
@@ -330,8 +330,8 @@ export function ConfigurationAdvancedForm<
           layout='horizontal'
         >
           <Select
-            searchable={true}
-            removable={true}
+            searchable
+            removable
             value={selectedIssue?.id}
             name='issueId'
             placeholder='Select an existing issue'
--- a/packages/core/src/services/evaluationsV2/create.ts
+++ b/packages/core/src/services/evaluationsV2/create.ts
@@ -11,11 +11,11 @@ import { BadRequestError } from '../../lib/errors'
 import { Result } from '../../lib/Result'
 import Transaction from '../../lib/Transaction'
 import { IssuesRepository } from '../../repositories/issuesRepository'
+import { ProjectsRepository } from '../../repositories/projectsRepository'
 import { evaluationVersions } from '../../schema/models/evaluationVersions'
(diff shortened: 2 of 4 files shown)


apps/web/src/components/evaluations/ConfigurationForm.tsx
```diff
@@ -299,25 +323,22 @@ export function ConfigurationAdvancedForm<
...
         <FormFieldGroup
           label='Link issue'
           description='Link the evaluation to an issue in your agent logs'
           layout='horizontal'
         >
           <Select
             searchable={true}
-            // TODO(eval-generation): This will not get the issueId from page 2,3, etc. need to change with Andres feedback
-            value={issueId ?? null}
+            removable={true}
```
andresgutgon: No need for `removable={true}` when is `true`. Just `removable`

Image

packages/core/src/services/evaluationsV2/create.ts
```diff
@@ -26,22 +30,41 @@ export async function createEvaluationV2<
...
+    project?: Project
   },
   transaction = new Transaction(),
 ) {
   return await transaction.call(async (tx) => {
     if (!options) options = {}
     options = compactObject(options)
 
+    let issue: Issue | null = null
+    if (issueId && project) {
```
andresgutgon: No need to pass project. Find it in `commit.projectId`

packages/core/src/services/evaluationsV2/update.ts
```diff
@@ -27,14 +33,16 @@ export async function updateEvaluationV2<
...
     workspace,
   }: {
     evaluation: EvaluationV2<T, M>
     commit: Commit
+    workspace: Workspace
     settings?: Partial<Omit<EvaluationSettings<T, M>, 'type' | 'metric'>>
     options?: Partial<EvaluationOptions>
     issueId?: number | null
-    workspace: Workspace
+    project?: Project
```
andresgutgon: Same, it's in commit

andresgutgon (dismissed):

PR #1936, opened 2025-11-14
Fix: Experiments Tracking

csansoon:
This pull request fixes a critical bug in the experiment tracking system that caused progress to be reported incorrectly, especially when parts of an experiment failed. The primary accomplishment is a complete overhaul of the tracking mechanism to be "row-aware," ensuring that the progress displayed in the UI is a true and reliable reflection of the
  experiment's state.

   - Accurate Progress: Experiment progress is now tracked on a per-row basis. The UI correctly displays how many input rows have been fully processed, rather than showing a confusing and often incorrect count of individual evaluations.
   - Robust Error Handling: The system now gracefully handles cases where a prompt fails to run for a given data row. It correctly marks the row and all its dependent evaluations as errored, allowing the experiment to proceed to completion.
   - Improved User Experience: The experiment status in the UI is now clear, intuitive, and trustworthy. Users can confidently monitor their experiments, knowing the statistics are accurate.
   - Enhanced Maintainability: The underlying code for creating experiments and tracking their status has been significantly refactored, making it cleaner, more modular, and easier to maintain.

  Why was this change necessary?

  The previous experiment tracking system had a fundamental design flaw: it treated all evaluations as a flat list and was not aware of the relationship between an input data row and its set of evaluations.

  This led to several problems:
   1. Inaccurate Reporting: If a prompt failed to run for a single row, the system did not account for the evaluations that were consequently skipped.
   2. "Stuck" Experiments: This failure mode would prevent the experiment from ever reaching 100% completion, as the overall progress counter would never reach its expected total.
   3. Confusing UI: The progress bar and statistics were misleading, making it difficult for users to understand the true status of their experiment and diagnose issues.

  This PR addresses these issues by re-architecting the progress tracking to be centered around the lifecycle of each individual row, from the initial prompt run to the completion of all its evaluations.

### DIFF
--- a/packages/core/src/jobs/utils/progressTracker.ts
+++ b/packages/core/src/jobs/utils/progressTracker.ts
@@ -3,16 +3,17 @@ import { Redis, RedisOptions } from 'ioredis'
 import { buildRedisConnection, REDIS_KEY_PREFIX } from '../../redis'
 
 export type TrackedProgress = {
-  total: number
-  completed: number
-  errors: number
-  enqueued: number
-  failed: number
+  total: number // Total number of rows to process
+  completed: number // Rows completed
+  passed: number // Evaluations passed
+  failed: number // Evaluations failed
+  errors: number // Evaluations that failed with an error or did not execute due to an error in the document
   totalScore: number
 }
 
 export class ProgressTracker {
   private redis: Redis | null = null
+  private evaluationsPerRow?: number = undefined
 
   constructor(private batchId: string) {}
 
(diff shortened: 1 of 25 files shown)


packages/core/src/jobs/utils/progressTracker.ts
```diff
@@ -47,55 +50,101 @@ export class ProgressTracker {
...
+      .get(this.getKey('evaluationsPerRow'))
+      .then((v) => (v ? parseInt(v, 10) : 0))
+    return this.evaluationsPerRow
   }
 
-  async incrementFailed(count?: number) {
-    const redis = await this.ensureConnection()
-    await redis.incrby(this.getKey('failed'), count ?? 1)
+  private async incrementRow(redis: Redis, uuid: string, count?: number) {
+    const newCount = await redis.hincrby(this.getKey('rows'), uuid, count ?? 1)
```
andresgutgon, line 62: This 1 is a magic constant. Make it explicit

csansoon: It's not a magic constant 😭 

If instead of doing `incr(5)` you do `incr()`, it's assumed it will be incremented by 1

PR #1937, opened 2025-11-14
Issues banner

andresgutgon:
# What?
Implement a banner in runs to make users aware that they need to annotate logs in order to start using issues so they can see where are the biggest problems on their prompts.

## TODO
- [ ] Apply designs


andresgutgon:
Done here
https://github.com/latitude-dev/latitude-llm/pull/1975

PR #1939, opened 2025-11-17
Chore/add issue substatus

learningbizz:
We used to have the status of an issue be active or inactive, when really the statuses of each issue is ignored, merged, resolved... For this reason, as we will want to eventually filter by issue type in an aggregated format, this reframing makes things clean

### DIFF
--- a/packages/constants/src/issues/constants.ts
+++ b/packages/constants/src/issues/constants.ts
@@ -54,7 +54,7 @@ export type QueryParams = { [key: string]: string | string[] | undefined }
 export const DEFAULTS_ISSUE_PARAMS = {
   limit: 25,
   filters: {
-    group: ISSUE_STATUS.active,
+    status: ISSUE_STATUS.active,
   },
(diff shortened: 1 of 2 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/issues/_components/IssueItemActions/useActions.ts
```diff
@@ -48,12 +45,12 @@ export function useIssueActions({ issue }: { issue: SerializedIssue }) {
 
   const onActivation = useCallback(() => {
     // When an issue becomes active (unresolve/unignore), invalidate the active tab cache
-    invalidateOppositeTab(ISSUE_STATUS.active)
+    invalidateOppositeTab('active')
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/issues/_components/IssueItemActions/useActions.ts]: Why?

andresgutgon (approved):

PR #1953, opened 2025-11-19
Tab Links

csansoon:
image

When options contain a `route`, Tab buttons will now be wrapped in links. This avoids having to add custom routing logic, as NextJS' Link routing implementation is far superior than a simple `router.push`


apps/web/src/app/(private)/AppTabs/index.tsx
```diff
@@ -1,23 +1,18 @@
...
 export function AppTabs() {
-  const router = useNavigate()
   const selected = useSelectedPath()
   return (
     <div className='flex flex-row'>
       <TabSelector
         showSelectedOnSubroutes
         options={MAIN_NAV_LINKS}
         selected={`/${selected}` as DocumentRoutes}
-        onSelect={(value) => {
```
learningbizz, line 18: one of the main reasons we used useNavigate was because of the little line that shows when loading the new scree, does this new implementation also add this?

csansoon: Links use NextJS's navigation anyways, so yes

PR #1956, opened 2025-11-19
Added an optional generic type to API routes to enforce the return type

csansoon:
The returned type from an API route and the expected type from a fetcher have nothing to do and could change without warning.

Although this does not completely fix that, it does improve the type safety by adding an optional type enforcement.

Let me explain with an example:

Let's say you have a hook with the following fetcher:
```ts
const fetcher = useFetcher<CompletedRuns[]>(CUSTOM_ROUTE)
```
You are expecting the route to return a response with the `CompletedRuns[]` type.

However, if the route logic changes at any point and you forget to update all hooks that use this route, you will not notice it!

Now, you can add this same type to the route, to enforce the return!
```ts
// apps/web/src/api/.../route.ts

export const GET = errorHandler<CompletedRuns[]>(
  authHandler(
    async(...) => {
      ...
      return NextResponse.json(result, { status: 200 })
    }
  )
)
```

> [!TIP]
> You can add the generic type in any `errorHandler`, `authHandler` or `adminHandler`. You can even add it to all of them at the same time!

Now that you have added the `CompletedRuns[]` type to the handler, it will display a typecheck error if the `result` object does not match this defined type.

So now, you can improve the dev exp by using the same type both on the route definition and useFetcher implementation!

### DIFF
--- a/apps/web/src/app/api/projects/[projectId]/runs/completed/route.ts
+++ b/apps/web/src/app/api/projects/[projectId]/runs/completed/route.ts
@@ -1,12 +1,16 @@
 import { authHandler } from '$/middlewares/authHandler'
 import { errorHandler } from '$/middlewares/errorHandler'
 
-import { RunSourceGroup, SpanType } from '@latitude-data/constants'
+import {
+  CompletedRun,
+  RunSourceGroup,
+  SpanType,
+} from '@latitude-data/constants'
 import { listCompletedRuns } from '@latitude-data/core/services/runs/completed/listCompleted'
 import { Workspace } from '@latitude-data/core/schema/models/types/Workspace'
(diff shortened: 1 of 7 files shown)


apps/web/src/app/api/projects/[projectId]/runs/completed/route.ts
```diff
@@ -43,7 +47,7 @@ export const GET = errorHandler(
         next: runs.next ? JSON.stringify(runs.next) : null,
       }
 
-      return NextResponse.json(response, { status: 200 })
+      return NextResponse.json(response.items, { status: 200 })
```
Comment (andresgutgon) [inline apps/web/src/app/api/projects/[projectId]/runs/completed/route.ts:50]: This was the bug 🐛 . Now that you're here remove `response` constant enterely 🔥

PR #1960, opened 2025-11-20
Send email when an issue is escalating

andresgutgon:
# What?
We want to let users know when an issue is escalating. To do that, we'll send an email to all members of workspaces that have activated notifications for escalating issues.

## TODO
- [x] Send in batches of 100 emails to our email provider (max batch size is 1000)
- [x] Add tests to send escalating email handler
- [x] Add test to increment histogram
- [x] Check email is sent.
- [x] Allow users to skip receiving escalating issues.
- [x] Move user settings to a modal with a URL
- [x] Add footer on email to change notification preferences.
- [x] Design escalating email 
- [x] Fix notifications URL. Now it check by workspace not membership
- [x] Check that not having the settings the email is not sent
- [x] Check with Mailgun to see if batch sending is working. 
- [x] Squash and rebase


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/issues/_components/IssueDetailPanel/index.tsx
```diff
@@ -60,7 +60,7 @@ export function IssuesDetailPanel({
 
   return (
     <DetailsPanel bordered ref={ref}>
-      <div className='relative w-full overflow-hidden'>
+      <div className='relative w-full overflow-hidden custom-scrollbar'>
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/issues/_components/IssueDetailPanel/index.tsx:63]: Fix sliding horizontal overflow ⚠️ (I'll do in another PR)

packages/core/src/services/memberships/updateEscalatingIssuesEmailPreference.ts
```diff
@@ -0,0 +1,45 @@
...
+  return transaction.call(async (tx) => {
+    const updates = await tx
+      .update(memberships)
+      .set({ wantToReceiveEscalatingIssuesEmail: wantToReceive })
+      .where(eq(memberships.id, membership.id))
+      .returning()
+
+    const updated = updates[0]!
+
+    publisher.publishLater({
```
andresgutgon: Move to the callback

packages/core/src/services/memberships/updateWeeklyEmailPreference.ts
```diff
@@ -0,0 +1,45 @@
...
+  return transaction.call(async (tx) => {
+    const updates = await tx
+      .update(memberships)
+      .set({ wantToReceiveWeeklyEmail: wantToReceive })
+      .where(eq(memberships.id, membership.id))
+      .returning()
+
+    const updated = updates[0]!
+
+    publisher.publishLater({
```
andresgutgon: Move to the callback

apps/web/src/components/Notifications/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+    updateEscalatingIssuesEmail,
+    isUpdatingEscalatingIssuesEmail,
+    isLoading: isLoadingUser,
+  } = useCurrentUser()
+
+  if (!user && !isLoadingUser) return null
+
+  return (
+    <div className='flex flex-col gap-4'>
+      <Text.H4B>Email Notifications</Text.H4B>
```
andresgutgon: 🔥 Move to where is used

packages/web-ui/src/ds/atoms/DatePicker/Range/usePresets.ts
```diff
@@ -11,7 +11,7 @@ import {
   subMonths,
   subWeeks,
 } from 'date-fns'
-import type { RelativeDate } from '@latitude-data/core/constants'
```
andresgutgon, line 14: `web-ui` was coupled with `core`, I used sone UI components on `emails` and that gave me cycle dependencies. So I remove `@latitude-data/core` from `web-ui` and moved common parts to constants. 

Web UI should never dependent on core. If a UI depends on core it belongs to `apps/web`

apps/web/src/actions/workspaces/switch.ts
```diff
@@ -37,5 +39,9 @@ export const switchWorkspaceAction = authProcedure
       await cookies(),
     )
 
+    if (parsedInput.redirectTo) {
```
learningbizz, line 42: why?

packages/core/src/mailers/emails/_components/ContainerLayout/index.tsx
andresgutgon, line 1: 🔥 All mails were moved

packages/core/package.json
```diff
@@ -193,7 +193,7 @@
     "rate-limiter-flexible": "5.0.3",
     "react": "catalog:",
     "react-dom": "catalog:",
-    "react-email": "4.0.3",
+    "@latitude-data/emails": "workspace:*",
```
andresgutgon, line 196: Isolated so is easy to update and understand

packages/web-ui/package.json
```diff
@@ -272,7 +272,6 @@
     "@dnd-kit/modifiers": "^9.0.0",
     "@dnd-kit/utilities": "^3.2.2",
     "@latitude-data/constants": "workspace:^",
-    "@latitude-data/core": "workspace:^",
```
andresgutgon, line 275: No dep with core 🔥

apps/web/src/app/(private)/dashboard/notifications/[workspaceId]/_components/WorkspaceSwitcherRedirect.tsx
```diff
@@ -0,0 +1,56 @@
...
+    useLatitudeAction(switchWorkspaceAction)
+
+  useOnce(() => {
+    async function performSwitch() {
+      const [, error] = await switchWorkspace({
+        workspaceId: targetWorkspaceId,
+        redirectTo: ROUTES.dashboard.notifications.root,
+      })
+
+      if (error) {
```
Comment (learningbizz) [inline apps/web/src/app/(private)/dashboard/notifications/[workspaceId]/_components/WorkspaceSwitcherRedirect.tsx:31]: why not add this as the onFailure of the action?

apps/web/src/actions/workspaces/switch.ts
andresgutgon: This is used on the normal switch selector and now on the swich redirect for the notifications

apps/web/src/app/(private)/dashboard/notifications/[workspaceId]/page.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * 1. If this workspace is valid we find if the user in the current session
+ * (if any) has a membership in that workspace. If not current session we
+ * redirect to login page with `redirectTo` param set to this page.
+ *
+ * 2. If the user has a membership in that workspace but is not the current
+ * workspace we render workspace switcher redirect component to switch to that
+ * so frontend can set the right session in cookies.
+ *
+ * 3. If the user has membership to that workspace and it's the current workspace
+ * we redirect to the default notifications page
```
Comment (learningbizz) [inline apps/web/src/app/(private)/dashboard/notifications/[workspaceId]/page.tsx:20]: tremenda logica

apps/web/src/stores/currentUser.ts
```diff
@@ -42,8 +44,80 @@ export function useCurrentUser(opts?: SWRConfiguration) {
     [executeUpdateEditorMode],
   )
 
+  const { execute: updateWeeklyEmail, isPending: isUpdatingWeeklyEmail } =
+    useLatitudeAction(updateWeeklyEmailPreferenceAction, {
```
learningbizz, line 48: nitpick: These two actions dont have to do with the currentUser entity, its more of a new "notification" store

apps/web/src/stores/currentUser.ts
```diff
@@ -9,23 +8,26 @@ import { useToast } from '@latitude-data/web-ui/atoms/Toast'
...
     ...rest
-  } = useSWR<User>(compact(route), fetcher, opts)
+  } = useSWR<CurrentUser>(compact(route), fetcher, opts)
 
   const { execute: executeUpdateEditorMode, isPending: isUpdatingEditorMode } =
     useLatitudeAction(updateUserAction, {
-      onSuccess: async ({ data: user }) => {
-        mutate(user) // Note: silent update
+      onSuccess: async () => {
+        mutate() // Get fresh data
```
learningbizz, line 30: why do we need to do an entire mutate if we have the user data?

apps/web/src/stores/currentUser.ts
andresgutgon: Is totally related with the current user. Only current user can handle notifications. If this growth more we can move to another place

apps/web/src/stores/currentUser.ts
andresgutgon: Because now the source of truth has changed. The notification settings related to the current user are on the backend only. So whenever the user change i refresh the frontend

packages/emails/src/types.ts
```diff
@@ -0,0 +1,14 @@
+import { ReactNode } from 'react'
+
+export type ContainerLayoutProps = {
+  children: ReactNode
+  previewText: string
+  title?: string
+  // FIXME: REMOVE
```
learningbizz: remove?

apps/web/src/stores/currentUser.ts
learningbizz: You cant return them as part of the update user?

andresgutgon: Yes, but also do this fix the problem

PR #1962, opened 2025-11-20
feat: cleanup some potentially leaky connections to cache

geclos:
Noticed spikes on number of connections to the cache. Not convinced this fixes the memory leak but it's a potential optimization.


packages/core/src/services/experiments/updateStatus.ts
```diff
@@ -53,21 +59,29 @@ export async function initializeExperimentStatus({
...
-  })
+    })
 
-  return Result.nil()
+    return Result.nil()
+  } finally {
+    // CRITICAL: Always cleanup Redis connection to prevent memory leaks
+    await progressTracker.cleanup().catch(() => {
+      // Silently ignore cleanup errors to not mask the original error
+    })
```
andresgutgon, line 85: We could abstract this somehow at some point to avoid relying on the dev

PR #1974, opened 2025-11-25
feat/merge issues job

geclos:


packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
-- Remove losing issues from vector db
-- Insert/Update in bulk the winning issue histogram with the historic losing issue histograms. HOWEVER, leave the losing issue histograms as they are to have some traceability in the ui.
-- Mark the other losing issues as "merged". This will filter out the issues in the dashboard.
-- Add a pointer from `merged` issue to `winning` issue for traceability (make a new column issue.mergedToIssueId)
-- Enqueue generateIssueDetails job
-- Publish issueMerged event
-- Evaluations from the merged issues `evaluation.issueId` has to be `ignoredAt`. This data migration is not implemented yet. Also we should update the evaluations UI at some point to reflect this.
-- Add a pointer from `merged` issue to `winning` issue for traceability (make a new column issue.mergedToIssueId)
-*/
+import { env } from '@latitude-data/env'
```
andresgutgon, line 1: Test!

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
+          return Result.error(
+            new UnprocessableEntityError('Issue already merged'),
+          )
+        }
+
+        refreshedMergedIssues.push(refreshed.value)
+      }
+
+      const timestamp = new Date()
+      const mergedCentroid = mergeIssueCentroids({
```
andresgutgon: Is this going to block transaction? Is doing something outside the DB?

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
+    .values(
+      associations.map(({ evaluationResultId }) => ({
+        workspaceId,
+        issueId: winnerId,
+        evaluationResultId,
+        createdAt: timestamp,
+        updatedAt: timestamp,
+      })),
+    )
+    .onConflictDoNothing()
```
andresgutgon: You already have this
https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/services/issueEvaluationResults/add.ts

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
+
+  await tx
+    .update(evaluationResultsV2)
+    .set({ issueId: winnerId, updatedAt: timestamp })
```
andresgutgon: Wait. No need. `issueId` is totally deprecated

andresgutgon: I didn't removed but this column could be gone. I'm 97% secure 😅

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
+  await tx
+    .update(evaluationVersions)
+    .set({ issueId: winnerId, updatedAt: timestamp })
+    .where(
+      and(
+        eq(evaluationVersions.workspaceId, workspaceId),
+        inArray(evaluationVersions.issueId, issueIds),
+      ),
+    )
+}
```
andresgutgon, line 381: I don't think we want to do this. I think we talk about these evaluations of the no winners. 

1. Set `live` to false
2. Add a new column called `ignoredAt` and filter these ignored evaluations from the UI

Talk with @cesr or @learningbizz I could be wrong

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
+        issueHistograms.commitId,
+        issueHistograms.date,
+      ],
+      set: {
+        count: sql<number>`${issueHistograms.count} + excluded.count`,
+        occurredAt: sql<Date>`GREATEST(${issueHistograms.occurredAt}, excluded.occurred_at)`,
+        updatedAt: timestamp,
+      },
+    })
+}
```
andresgutgon, line 448: This is a beast. Move to histogram service and use here

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
+      })),
+    )
+    .onConflictDoUpdate({
+      target: [
+        issueHistograms.issueId,
+        issueHistograms.commitId,
+        issueHistograms.date,
+      ],
+      set: {
+        count: sql<number>`${issueHistograms.count} + excluded.count`,
```
andresgutgon: what is doing this?

packages/core/src/services/issues/merge.ts
learningbizz: Yep, as Andres says ☝️ We are pending adding the ignoredAt column in evaluationVersions

packages/core/src/services/issues/merge.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      .where(eq(evaluationResultsV2.id, result.id))
+      .then((r) => r[0]!)
+    expect(updatedResult.issueId).toBe(winner.id)
+
+    const updatedEvaluation = await database
+      .select()
+      .from(evaluationVersions)
+      .where(eq(evaluationVersions.id, evaluation.versionId))
+      .then((r) => r[0]!)
+    expect(updatedEvaluation.issueId).toBe(winner.id)
```
andresgutgon: Not testing 
1. centroid is updated
2. Eval results are assigned
3. issues are marked as merged
4. Vector DB is called with the right params

What else?

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
-- Recompute the winning issue's centroid embedding with the losing issues centroid embedding.
-- Update winning issue's centroid in vector db
-- Remove losing issues from vector db
-- Insert/Update in bulk the winning issue histogram with the historic losing issue histograms. HOWEVER, leave the losing issue histograms as they are to have some traceability in the ui.
-- Mark the other losing issues as "merged". This will filter out the issues in the dashboard.
-- Add a pointer from `merged` issue to `winning` issue for traceability (make a new column issue.mergedToIssueId)
-- Enqueue generateIssueDetails job
-- Publish issueMerged event
-- Evaluations from the merged issues `evaluation.issueId` has to be `ignoredAt`. This data migration is not implemented yet. Also we should update the evaluations UI at some point to reflect this.
-- Add a pointer from `merged` issue to `winning` issue for traceability (make a new column issue.mergedToIssueId)
```
andresgutgon, line 18: Here there is UI work. Also I think we need also to add a new column to know from where was merged an issue. Also make sure in the issue dashboard merged issues appears as inactive

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
-// TODO(AO): implement this
-
-/*
-- check LATITUDE_CLOUD
-- Search the candidate-related issues by cosine similarity on their centroid embedding. Filter by 80% similarity.
-- Select the biggest issue, the issue with the most all-time events (biggest histogram count). (Note respect the biggest issue state, maybe it is resolved/ignored, that means that it will continue to be ignored or go to escalating because new events are discovered)
-- Relink in bulk all evaluation results tied to the losing issues to the winning issue.
-- Relink all evaluations tied to the losing issues to the winning issue.
```
andresgutgon, line 8: Already talked. @neoxelox though this way but not sure is the most updated though now

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
+
+      const score = similarity(anchorEmbedding, embedding)
+      if (score < ISSUE_DISCOVERY_MIN_SIMILARITY) return null
+
+      return { issue, embedding }
+    })
+    .filter(Boolean) as Array<{ issue: Issue; embedding: number[] }>
+}
+
+async function getHistogramTotals({
```
learningbizz, line 228: This should be in the repository of issueHistogram

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
+
+async function findSimilarIssues({
+  anchorIssue,
+}: {
+  anchorIssue: Issue
+}): Promise<Array<{ issue: Issue; embedding: number[] }>> {
+  const anchorEmbedding = embedCentroid(anchorIssue.centroid)
+  if (anchorEmbedding.length === 0) return []
+
+  const candidates = await database
```
learningbizz: we could extract this query in a repo as well

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
-- Update winning issue's centroid in vector db
-- Remove losing issues from vector db
-- Insert/Update in bulk the winning issue histogram with the historic losing issue histograms. HOWEVER, leave the losing issue histograms as they are to have some traceability in the ui.
-- Mark the other losing issues as "merged". This will filter out the issues in the dashboard.
-- Add a pointer from `merged` issue to `winning` issue for traceability (make a new column issue.mergedToIssueId)
-- Enqueue generateIssueDetails job
-- Publish issueMerged event
-- Evaluations from the merged issues `evaluation.issueId` has to be `ignoredAt`. This data migration is not implemented yet. Also we should update the evaluations UI at some point to reflect this.
-- Add a pointer from `merged` issue to `winning` issue for traceability (make a new column issue.mergedToIssueId)
-*/
```
andresgutgon, line 19: Also question:
Can a merged issue be back to active? If it can we should also have a service to unmerge it and un-ignore their evaluation

packages/core/src/services/issues/merge.ts
```diff
@@ -1,19 +1,482 @@
...
+    )
+    .groupBy(issueHistograms.issueId)
+
+  rows.forEach((row) => totals.set(row.issueId, row.total ?? 0))
+
+  return totals
+}
+
+async function removeVectors(issuesToDelete: Issue[]) {
+  if (!env.WEAVIATE_API_KEY) return
```
learningbizz: do we want to be able to do all the merge if then the vector key is not set? We should do this together with the env.CLOUD no?

packages/core/src/services/issues/merge.ts
geclos: not this method but the updateIssue method is. There isn't much I can do, I need transactionality in this method and the update method is coupled with a vector db operation.

packages/core/src/services/issues/merge.ts
geclos: this is a bulk operation

packages/core/src/services/issues/merge.ts
geclos: updates the histogram of the winning issue to account for the histogram of the merge issue

packages/core/src/services/issues/merge.ts
geclos: not possible

packages/core/src/services/issues/merge.ts
andresgutgon: Then use `addBulk`

packages/core/src/services/issues/merge.ts
andresgutgon: Wait why you resolved this comment? I think this is not necessary

packages/core/src/services/issues/merge.ts
andresgutgon: What was the conclusion here?

packages/core/src/services/issues/merge.ts
geclos: it's an implementation detail of the merge operation i don't want to expose it as a public api in the repo

packages/core/src/services/issues/merge.ts
geclos: i actually like it here, don't want to expose services that make no sense outside the context of the merge operation, it just adds unnecessary indirection

packages/core/src/services/issues/merge.ts
geclos: none of the current solutions are ok we probably need an env var that controls embedding operations, if set we embed, if not we don't

PR #1979, opened 2025-11-26
feat: prevent assigning merged issues in annotation form

geclos:
## Summary
- Add merged issue link display in issue detail panel showing which active issue a merged issue was merged into
- Prevent selecting merged issues in the annotation form's issue selector dropdown
- Update activeWithResolved group filter to exclude merged issues from search results
- Show disabled state with tooltip when viewing spans from merged issues
- Backend validation prevents issue assignment changes for merged issues

### DIFF
--- a/apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/issues/search/route.ts
+++ b/apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/issues/search/route.ts
@@ -17,8 +17,6 @@ export type SearchIssueResponse = Awaited<
 const paramsSchema = z.object({
   projectId: z.coerce.number(),
   documentUuid: z.string(),
-  statuses: z.string().optional(),
-  group: z.string().optional(),
 })
 
 export const GET = errorHandler(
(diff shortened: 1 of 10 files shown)


apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/issues/search/route.ts
```diff
@@ -39,13 +37,13 @@ export const GET = errorHandler(
       },
     ) => {
       const query = request.nextUrl.searchParams
-      const { projectId, documentUuid, statuses, group } = paramsSchema.parse({
```
Comment (geclos) [inline apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/issues/search/route.ts:42]: bugarro

packages/core/src/repositories/issuesRepository.ts
```diff
@@ -439,9 +439,18 @@ export class IssuesRepository extends Repository<Issue> {
...
         groupConditions = and(
-          this.withResolvedIssues(true),
+          or(
+            // Active: not resolved and not ignored
+            and(this.withResolvedIssues(false), this.withIgnoredIssues(false)),
+            // OR Resolved: has resolvedAt timestamp
+            this.withResolvedIssues(true),
+          ),
+          // But exclude ignored and merged issues
           this.withIgnoredIssues(false),
```
geclos, line 452: this was excluding all active issues we just didn't notice because there was another bug masking this bug.

geclos: also very confusing way of doing query composition `with*(boolean)`

andresgutgon: Agree, the with* is confusing. Also could you add a test fix?

PR #1981, opened 2025-11-26
fix: Evaluation Results in Runs page

csansoon:
Fixes: Evaluation Results appear in the Runs page


packages/core/src/services/evaluationsV2/llm/buildStreamEvaluationRun.ts
```diff
@@ -100,6 +100,7 @@ export async function buildStreamEvaluationRun({
     promptUuid: evaluation.uuid,
     template: evaluation.configuration.prompt,
     parameters: parameters,
+    source: LogSources.Evaluation,
```
learningbizz, line 103: A test to make sure this doesnt happen again would be great

csansoon: yup, done!

PR #1983, opened 2025-11-27
Add cases to LLM as judge eval generation

learningbizz:
To improve the generation of LLM-as-judge configurations, we need to pass the annotations of the user to the prompt. This improvement enables that

### DIFF
--- a/packages/core/src/data-access/issues/getSpanMessagesAndEvaluationResultsByIssue.test.ts
+++ b/packages/core/src/data-access/issues/getSpanMessagesAndEvaluationResultsByIssue.test.ts
@@ -202,14 +202,14 @@ describe('getSpanMessagesAndEvaluationResultsByIssue', () => {
     expect(result.ok).toBe(true)
     if (!result.ok) return
 
-    const data = result.value
+    const data = result.value!
     expect(data).toHaveLength(1)
     expect(data[0]!.messages).toHaveLength(2)
-    expect(data[0]!.messages[0]!.role).toBe('user')
+    expect(data[0]!.messages[0]!.role).toBe(MessageRole.user)
     expect(data[0]!.messages[0]!.content).toEqual([
       { type: 'text', text: 'Hello, how are you?' },
     ])
-    expect(data[0]!.messages[1]!.role).toBe('assistant')
+    expect(data[0]!.messages[1]!.role).toBe(MessageRole.assistant)
     expect(data[0]!.messages[1]!.content).toEqual([
       { type: 'text', text: 'I am doing well, thank you!' },
     ])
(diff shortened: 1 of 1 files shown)


packages/core/src/data-access/issues/getSpanMessagesAndEvaluationResultsByIssue.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const evaluationResultsRepository = new EvaluationResultsV2Repository(
+    workspace.id,
+  )
+  const evaluationResultsResult =
+    await evaluationResultsRepository.listByIssueId(issue.id)
+
+  const messagesAndEvaluationResults: SpanMessagesWithEvalResultReason[] = []
+  for (const span of spans.spans) {
+    if (span.type !== SpanType.Prompt) {
```
andresgutgon: Filter by prompt in the sql query

packages/core/src/repositories/evaluationResultsV2Repository.ts
```diff
@@ -499,31 +499,118 @@ export class EvaluationResultsV2Repository extends Repository<EvaluationResultV2
...
+          resultInstances.find((r) => r.joinedIssueId === issueId) ??
+          resultInstances[0]!
+
+        result = activeInstance.result
+      }
+
+      grouped[issueId] = [...(grouped[issueId] ?? []), result]
+    }
+
+    return grouped
```
andresgutgon: Testsss

andresgutgon: Too complicated 

Before `listByIssueIds` we used to get eval result for 1 or N issues. now we have two ways of doing the same. 

Groping should belong to the code that's using this query

packages/core/src/repositories/evaluationResultsV2Repository.ts
```diff
@@ -499,31 +499,118 @@ export class EvaluationResultsV2Repository extends Repository<EvaluationResultV2
...
+      let result: EvaluationResultV2
+      let issueId: number
+
+      if (resultInstances.length === 1) {
+        result = resultInstances[0]!.result
+        issueId = resultInstances[0]!.joinedIssueId!
+      } else {
+        const firstResult = resultInstances[0]!.result
+        const activeAssignment =
+          await issueEvalResultsRepo.findLastActiveAssignedIssue({
```
andresgutgon: Nothing of this is necessary if we ensure issueIds are only from merged issues.

packages/core/src/repositories/evaluationResultsV2Repository.ts
learningbizz: Talked offline, we found an easier method if we make this query return also the joinedIssue from the other table, and we make this `listByIssueIds` only for non-merged issues (our current use case)

packages/core/src/repositories/evaluationResultsV2Repository.ts
learningbizz: Replied above

andresgutgon (dismissed):

PR #1984, opened 2025-11-27
Fix refiner and document suggestor

learningbizz:
These two features were still using the old provider log mechanism to find the documentLog. Therefore, for the newer logs created, the evaluatedLogId wasn't found and it would break. This fix fixes this by adding the new method using spans and traces, and falls back to the evaluatedLogId if they're not found (for old logs).

### DIFF
--- a/packages/core/src/services/documentSuggestions/serialize.ts
+++ b/packages/core/src/services/documentSuggestions/serialize.ts
@@ -10,6 +10,7 @@ import { Result } from '../../lib/Result'
 import {
   DocumentLogsRepository,
   ProviderLogsRepository,
+  SpansRepository,
 } from '../../repositories'
 import { type Workspace } from '../../schema/models/types/Workspace'
 import { serialize as serializeDocumentLog } from '../documentLogs/serialize'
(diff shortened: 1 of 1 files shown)


packages/core/src/services/documentSuggestions/serialize.ts
```diff
@@ -84,14 +85,46 @@ export async function serializeEvaluationResult<
       result as EvaluationResultSuccessValue<T, M>,
     ) || 'No reason reported'
 
-  const providerLogsRepository = new ProviderLogsRepository(workspace.id, db)
-  const providerLog = await providerLogsRepository
-    .find(result.evaluatedLogId)
-    .then((r) => r.unwrap())
+  let documentLogUuid: string
+  if (result.evaluatedSpanId && result.evaluatedTraceId) {
```
andresgutgon, line 89: Tests

packages/core/src/services/documentSuggestions/serialize.ts
```diff
@@ -84,14 +85,46 @@ export async function serializeEvaluationResult<
...
+      return Result.error(new Error(`Span missing documentLogUuid`))
+    }
+
+    documentLogUuid = span.documentLogUuid
+  } else if (result.evaluatedLogId) {
+    // Fallback to legacy providerLog approach for backward compatibility
+    const providerLogsRepository = new ProviderLogsRepository(workspace.id, db)
+    const providerLog = await providerLogsRepository
+      .find(result.evaluatedLogId)
+      .then((r) => r.unwrap())
```
andresgutgon, line 114: Be careful. Who is using this might be waiting for a `Result` not a throw. Check if that's the case

andresgutgon: We're already mixing `.unwrap` and returning result in this method. Fucking confusing pattern

packages/core/src/services/documentSuggestions/serialize.ts
```diff
@@ -84,14 +85,46 @@ export async function serializeEvaluationResult<
...
+
+    if (!span.documentLogUuid) {
+      return Result.error(new Error(`Span missing documentLogUuid`))
+    }
+
+    documentLogUuid = span.documentLogUuid
+  } else if (result.evaluatedLogId) {
+    // Fallback to legacy providerLog approach for backward compatibility
+    const providerLogsRepository = new ProviderLogsRepository(workspace.id, db)
+    const providerLog = await providerLogsRepository
```
andresgutgon, line 112: Should we still be using and reading provider logs @geclos ?

geclos: no

packages/core/src/services/documentSuggestions/serialize.ts
learningbizz: Yeah, the problem is that we shouldn't return Results in repositories, but we didnt apply this rule till later 😄

packages/core/src/services/documentSuggestions/serialize.ts
geclos: don't waste time on this, this feature is gonna get removed!

PR #1989, opened 2025-11-30
Fix issue search API route missing histogram table

learningbizz:
After the commit #1974,  the `useIssuesSearch` API endpoint was breaking when entering a setting of an existing evaluation.

This was due to the `findByTitleAndStatuses` missing the join of the histogram table, useful when searching for the inactive group. 

This PR adds the necessary join

### DIFF
--- a/packages/core/src/repositories/issuesRepository.ts
+++ b/packages/core/src/repositories/issuesRepository.ts
@@ -182,12 +182,14 @@ export class IssuesRepository extends Repository<Issue> {
   async findByTitleAndStatuses({
     project,
     document,
+    commit,
     title,
     statuses,
     group,
   }: {
     project: Project
     document: DocumentVersion
+    commit: Commit
     title: string | null
     statuses?: IssueStatuses[]
     group?: IssueGroup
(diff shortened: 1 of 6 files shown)


packages/core/src/repositories/issuesRepository.ts
```diff
@@ -197,13 +199,21 @@ export class IssuesRepository extends Repository<Issue> {
...
+    })
+
     return this.db
       .select({
         id: issues.id,
         title: issues.title,
         documentUuid: issues.documentUuid,
       })
       .from(issues)
+      .innerJoin(subquery, eq(subquery.issueId, issues.id))
```
andresgutgon, line 216: Wait what? The search never had a histogram

learningbizz: It didnt need to, but when we added the possibility to search by groups, we starting needing it 

We can add it conditionally only when needed

andresgutgon: What you mean search by groups?

PR #1992, opened 2025-12-01
quickfix: issues update

geclos:


packages/core/src/services/issues/update.ts
```diff
@@ -102,27 +102,32 @@ async function upsertVector({
...
+    if (embedding) {
+      payload.vectors = embedding
+    }
+    if (title && description) {
+      payload.properties = { title, description }
+    }
+    if (!payload.vectors && !payload.properties) {
+      return Result.error(
+        new BadRequestError(
+          'Received update issue operation without values to update',
```
andresgutgon, line 119: The error will keep happening. I not sure how we can trace when is happening

PR #1993, opened 2025-12-01
Improved dev tracing P1

csansoon:
The `telemetry.trace(...)` wrapper callback from the Demo Day.

Note: This will NOT be the final callback look. That's why this PR doesn't include any documentation or package bump. ~However~ _Nonetheless_, the next changes involve many files so I'd rather do the PR incrementally.

Part 1/?


packages/core/src/services/tracing/spans/prompt.ts
```diff
@@ -28,17 +34,84 @@ async function process(
...
+      path: promptPath,
+    })
+    if (docResult.error) {
+      return undefined
+    }
+
+    return docResult.value.documentUuid
+  } catch {
+    return undefined
+  }
```
andresgutgon, line 116: `t` please `t`

packages/telemetry/typescript/src/processors/index.ts
```diff
@@ -0,0 +1 @@
+export { NormalizingSpanProcessor } from './normalize'
```
andresgutgon, line 1: Do we need to hace this barrel file?

packages/core/src/services/tracing/spans/prompt.ts
```diff
@@ -28,17 +34,84 @@ async function process(
...
-  return Result.ok({
+  // Get promptUuid from attributes, or try to resolve from promptPath
+  let promptUuid = attributes['latitude.documentUuid'] as string | undefined
+  const promptPath = attributes[ATTR_LATITUDE_PROMPT_PATH] as string | undefined
+  const projectId = attributes['latitude.projectId'] as number | undefined
+  const versionUuid =
+    (attributes['latitude.commitUuid'] as string) || HEAD_COMMIT
+
+  // If promptPath is provided but promptUuid is not, resolve it
+  if (promptPath && !promptUuid && projectId) {
```
learningbizz, line 45: nitpick: instead of a comment and the logic in the if, create a constant with a readable name and add it to the if

packages/core/src/services/tracing/spans/prompt.ts
```diff
@@ -28,17 +34,84 @@ async function process(
...
+  db,
+}: {
+  promptPath: string
+  projectId: number
+  versionUuid: string
+  workspaceId: number
+  db: typeof database
+}): Promise<string | undefined> {
+  try {
+    // Get the commit
```
learningbizz, line 92: no need for this comment, its self explanatory

packages/core/src/services/tracing/spans/prompt.ts
```diff
@@ -28,17 +34,84 @@ async function process(
...
+    const commitResult = await commitsRepo.getCommitByUuid({
+      uuid: versionUuid,
+      projectId,
+    })
+    if (commitResult.error) {
+      return undefined
+    }
+    const commit = commitResult.value
+
+    // Get the document by path
```
learningbizz, line 103: same here

PR #1996, opened 2025-12-01
Rename "Runs" to "Annotations"

andresgutgon:
# What?
The runs section is doing too much. We want to let users know where annotations happen. We'll move background runs visualization under each project in their `/traces` section

## TODO
- [x] Rename the section. That was easy lol
- [x] Move background runs to traces section under each document
- [x] Review traces open state. Looks like data is not fetched correctly 🐛 
- [x] `activeRuns` not generated 🐛 
- [x] Details Active run panel. Review it can be stoped
- [x] Pause realtime as in Runs
- [x] Check active panel works by doing a bit longer running prompt
- [x] Active runs on sidebar indicator
- [x] Fix active runs tests. Adapt to be scoped by document 🔥 

## Another PR
- [ ] Review the Annotations page with the product to filter what we define as annotable traces there


apps/gateway/src/routes/api/v3/conversations/stop/stop.handler.ts
```diff
@@ -21,7 +23,12 @@ export const stopHandler: AppRouteHandler<StopRoute> = async (ctx) => {
     .getProjectById(run.projectId)
     .then((r) => r.unwrap())
 
-  await stopRun({ run, project, workspace }).then((r) => r.unwrap())
+  await stopRun({
+    run,
+    project,
+    workspace,
+    documentUuid: run.documentUuid,
```
geclos, line 30: weird why? the run already has the document uuid

apps/web/src/actions/runs/stopByDocument.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export const stopRunByDocumentAction = withProject
+  .inputSchema(
+    withProjectSchema.extend({
+      documentUuid: z.string(),
+      runUuid: z.string(),
+    }),
+  )
+  .action(async ({ parsedInput, ctx }) => {
+    const run = await getRunByDocument({
```
geclos, line 20: so you've reimplemented the same services but with byDocument* suffix, do we need to keep the old services? also the apis are weird, both receiving a run uuid and a document uuid what is gonna happen? it's gonna find by uuid and then check it matches the document uuid you passed? if you already have the run uuid why do you need to pass the document uuid?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/DocumentTraces.tsx
```diff
@@ -1,58 +1,95 @@
...
+            items: [span, ...prev.items],
+            count: prev.count ? prev.count + 1 : null,
+          }
+        },
+        { revalidate: false },
+      )
+    },
+    [document.documentUuid, currentCursor, mutateSpans],
+  )
+  useSockets({ event: 'spanCreated', onMessage: onSpanCreated })
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/DocumentTraces.tsx:91]: move to hook

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx
```diff
@@ -43,44 +57,95 @@ export function TraceSpanSelectionProvider({
...
+    [selection, onClickTraceRow],
+  )
 
   return (
     <TraceSpanSelectionContext.Provider
-      value={{ selection, selectTraceSpan, clearSelection }}
+      value={{
+        selection,
+        onClickTraceRow,
+        selectSpan,
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx:144]: unexpected public api that exposes both a select span method and onClickTraceRow method

apps/web/src/components/evaluations/Annotation/FormWrapper/index.tsx
```diff
@@ -278,9 +278,10 @@ AnnotationFormWrapper.FailedWithoutReasonWarning =
         <div className='px-3 py-2 bg-background rounded-md flex items-center gap-x-2'>
           <AnnotationsProgressIcon isCompleted />
           <Text.H6 color='foregroundMuted'>
-            Please write feedback on why this annotation did not pass. This is
-            important for improving your prompt and ensuring Latitude can create
-            better issues. Check the{' '}
+            Please explain what went wrong with this AI interaction. Your
```
geclos, line 281: s/explain/write

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx:144]: selectSpan is what we already have and is passed down in the recursive waterfall of spans when you open a trace

apps/web/src/actions/runs/stopByDocument.ts
andresgutgon: No, i removed old services.

packages/core/src/data-access/runs.ts
```diff
@@ -13,11 +13,14 @@ export async function unsafelyFindActiveRun(runUuid: string) {
     )
   }
 
-  const { workspaceId, projectId } = job.data as BackgroundRunJobData
+  const { workspaceId, projectId, documentUuid } =
+    job.data as BackgroundRunJobData
 
-  const getting = await getRun({
+  // Use document-scoped get (with fallback to project-level cache)
```
geclos, line 19: why do we need the fallback? there's no more project-level runs no?

andresgutgon: That comment is legacy. No fallback

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/DocumentTraces.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/DocumentTraces.tsx:91]: This file is 150LOC. Is ok here imo

apps/gateway/src/routes/api/v3/conversations/stop/stop.handler.ts
andresgutgon: True

apps/web/src/components/evaluations/Annotation/FormWrapper/index.tsx
andresgutgon: Will changed (in the future)

PR #1998, opened 2025-12-02
Reasoning Settings UI

csansoon:
image
(Appears only for supported models)


packages/core/src/services/ai/estimateCost/helpers.ts
```diff
@@ -1,8 +1,17 @@
 import type { ModelCost } from './index'
 
+export type ReasoningEffort = 'none' | 'minimal' | 'low' | 'medium' | 'high'
+export type ReasoningSummary = 'auto' | 'detailed'
+
+export type ReasoningCapabilities = {
```
andresgutgon, line 6: Massive conflict with @geclos is in your future

PR #1999, opened 2025-12-02
Include model data in UI

csansoon:
https://github.com/user-attachments/assets/ae33f3d1-0ae9-408e-9a77-9c1d1509de08

Note: UI still in revision


learningbizz (changes requested):
I like the idea, but lets talk to Danil abt the UI as some of us find it too condensed

PR #2003, opened 2025-12-03
Adding confusion matrix to the frontend

learningbizz:
As part of the "Evaluation alignment analytics" epic, we wanted to show a confusion matrix for the more technical users to understand better how we calculate our quality metric.

This PR adds the functionality to save the confusion matrix in database and be able to show it in the frontend. For now we show it in the tooltip, but we can change it in the future once Danil starts raging.

image
image

### DIFF
--- a/apps/web/src/components/ConfusionMatrix/index.tsx
+++ b/apps/web/src/components/ConfusionMatrix/index.tsx
@@ -1,0 +1,69 @@
+import { QualityMetricMetadata } from '@latitude-data/constants'
+import { EVALUATION_QUALITY_EXPLANATION } from '@latitude-data/constants/issues'
+
+export function ConfusionMatrixTooltipContent({
+  confusionMatrix,
+}: {
+  confusionMatrix?: QualityMetricMetadata['confusionMatrix']
+}) {
+  return (
+    <div className='flex flex-col gap-2'>
+      <div>{EVALUATION_QUALITY_EXPLANATION}</div>
+      <ConfusionMatrixTable confusionMatrix={confusionMatrix} />
+    </div>
+  )
+}
+
+function ConfusionMatrixTable({
+  confusionMatrix,
+}: {
+  confusionMatrix?: QualityMetricMetadata['confusionMatrix']
+}) {
+  if (!confusionMatrix) {
+    return null
+  }
+
+  return (
+    <div className='flex flex-col gap-2'>
+      <table className='border-collapse'>
+        <thead>
+          <tr>
+            <th className='p-2 text-center text-xs font-medium'>
+              Confusion Matrix
+            </th>
+            <th className='border-l p-2 text-center text-xs font-medium'>
+              Actually Positive
+            </th>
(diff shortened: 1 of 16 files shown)


packages/core/src/schema/models/evaluationVersions.ts
```diff
@@ -15,6 +15,7 @@ import {
   EvaluationConfiguration,
   EvaluationMetric,
   EvaluationType,
+  QualityMetricMetadata,
```
andresgutgon, line 18: Is this imported from right place?

apps/web/src/components/ConfusionMatrix/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { EVALUATION_QUALITY_EXPLANATION } from '@latitude-data/constants/issues'
+
+export function ConfusionMatrixTooltipContent({
+  confusionMatrix,
+}: {
+  confusionMatrix?: QualityMetricMetadata['confusionMatrix']
+}) {
+  return (
+    <div className='flex flex-col gap-2'>
+      <div>{EVALUATION_QUALITY_EXPLANATION}</div>
```
andresgutgon, line 11: Use Text component

apps/web/src/components/ConfusionMatrix/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    return null
+  }
+
+  return (
+    <div className='flex flex-col gap-2'>
+      <table className='border-collapse'>
+        <thead>
+          <tr>
+            <th className='p-2 text-center text-xs font-medium'>
+              Confusion Matrix
```
andresgutgon, line 32: Text component

apps/web/src/components/ConfusionMatrix/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+            </td>
+            <td className='border-t border-border p-2 text-center text-xs'>
+              {confusionMatrix.trueNegatives} (TN)
+            </td>
+          </tr>
+        </tbody>
+      </table>
+    </div>
+  )
+}
```
andresgutgon, line 69: Text component everywhere

packages/core/src/schema/models/evaluationVersions.ts
learningbizz: Yes! Added it in evaluations/index.ts

apps/web/src/components/ConfusionMatrix/index.tsx
learningbizz: Talked offline, as its for a tooltip, we will leave as is for now

apps/web/src/components/ConfusionMatrix/index.tsx
learningbizz: same as below

apps/web/src/components/ConfusionMatrix/index.tsx
learningbizz: same as below

PR #2006, opened 2025-12-04
Fix generation evaluation logic

learningbizz:
Changed a few things:
- Our previous implementation wanted the evaluation to pass (hasPassed=true) when the issue appeared in the evaluation, but we want the opposite, as the issue discovery works this way, and also the general idea with the evaluations of a document is to work like UT, so if a trace goes through evals and they all pass, it means that the trace is good, else it has an issue.  This PR fixes this logic and makes this logic more clear throughout the application
- As we always want the eval to fail when the issue appears, the reverseScale option is redundant for our LLM that creates the configuration. So simplified its logic and hard coded the reverseScale to always be false.
- Remove 👍 or 👎 if 0 in the need more annotations state
- Change backend queries used in the evaluation generation flow to use only HITL eval results (to feed the eval config generation and create the MCC)
- Fix bug when runEvaluation children fail, the parent calculateQualityMetric job waits indefinitely. Now when a child fails it continues to the parent, and the parent waits for all the other children to process using the job retry system

### DIFF
--- a/packages/core/src/services/evaluationsV2/generateFromIssue/generateFromIssue.ts
+++ b/packages/core/src/services/evaluationsV2/generateFromIssue/generateFromIssue.ts
@@ -91,7 +91,7 @@ export async function generateEvaluationConfigFromIssueWithCopilot(
 
   const existingEvaluationNames = existingEvaluationNamesResult.unwrap()
 
-  // Getting negative examples (failed evaluation results about the issue) to feed the copilot
+  // Getting failed examples (evaluation results with the issue attached) to feed the copilot
   const messagesAndReasonWhyFailedForIssueResult =
     await getSpanMessagesAndEvaluationResultsByIssue({
       workspace: workspace,
@@ -105,19 +105,19 @@ export async function generateEvaluationConfigFromIssueWithCopilot(
   const messagesAndReasonWhyFailedForIssue =
     messagesAndReasonWhyFailedForIssueResult.unwrap()
 
-  // Getting positive examples (passed evaluation results or failed about other issues of the same document) to feed the copilot
-  const positiveExampleMessagesFromIssueDocumentResult =
+  // Getting passed examples (passed evaluation results or failed about other issues of the same document) to feed the copilot
+  const passedExampleMessagesFromIssueDocumentResult =
     await getSpanMessagesByIssueDocument({
       workspace: workspace,
       commit: commit,
       issue: issue,
     })
 
-  if (!Result.isOk(positiveExampleMessagesFromIssueDocumentResult)) {
-    return positiveExampleMessagesFromIssueDocumentResult
+  if (!Result.isOk(passedExampleMessagesFromIssueDocumentResult)) {
+    return passedExampleMessagesFromIssueDocumentResult
   }
-  const positiveExampleMessagesFromIssueDocument =
-    positiveExampleMessagesFromIssueDocumentResult.unwrap()
+  const passedExampleMessagesFromIssueDocument =
+    passedExampleMessagesFromIssueDocumentResult.unwrap()
 
   const evaluationConfigResult = await runCopilot({
     copilot: copilot,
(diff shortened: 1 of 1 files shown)


packages/core/src/data-access/issues/getHITLSpansByDocument.ts
```diff
@@ -0,0 +1,96 @@
...
+    excludeIssueId: number
+    page: number
+    pageSize: number
+  },
+  db = database,
+) {
+  const resultsRepository = new EvaluationResultsV2Repository(workspace.id, db)
+  const { results: paginatedResults, hasNextPage } =
+    await resultsRepository.fetchPaginatedHITLResultsByDocument({
+      workspace,
```
andresgutgon, line 50: No need. Already there

packages/core/src/data-access/issues/getHITLSpansByDocument.ts
```diff
@@ -0,0 +1,96 @@
...
+    (result) => sql`(${result.evaluatedSpanId}, ${result.evaluatedTraceId})`,
+  )
+
+  const fetchedSpans = await db
+    .select()
+    .from(spans)
+    .where(
+      and(
+        eq(spans.workspaceId, workspace.id),
+        sql`(${spans.id}, ${spans.traceId}) IN (${sql.join(spanTraceIdPairs, sql`, `)})`,
```
andresgutgon: Do we do this in other places?

packages/core/src/data-access/issues/getHITLSpansByDocument.ts
```diff
@@ -0,0 +1,96 @@
...
+    const key = `${span.id}:${span.traceId}`
+    spanMap.set(key, span as Span)
+  }
+
+  const orderedSpans = paginatedResults
+    .map((result) => {
+      const key = `${result.evaluatedSpanId}:${result.evaluatedTraceId}`
+      return spanMap.get(key)
+    })
+    .filter((span): span is Span => span !== undefined)
```
andresgutgon: All the span fetching is not already done in the repository?

packages/core/src/data-access/issues/getHITLSpansByDocument.ts
```diff
@@ -0,0 +1,96 @@
...
+      spans: [] as Span[],
+      hasNextPage: false,
+    })
+  }
+
+  const spanTraceIdPairs = paginatedResults.map(
+    (result) => sql`(${result.evaluatedSpanId}, ${result.evaluatedTraceId})`,
+  )
+
+  const fetchedSpans = await db
```
andresgutgon: DRY. This is the same here
https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/data-access/issues/getSpansByIssue.ts#L123-L150

packages/core/src/data-access/issues/getHITLSpansByIssue.ts
```diff
@@ -0,0 +1,92 @@
...
+    .map((result) => {
+      const key = `${result.evaluatedSpanId}:${result.evaluatedTraceId}`
+      return spanMap.get(key)
+    })
+    .filter((span): span is Span => span !== undefined)
+
+  return Result.ok({
+    spans: orderedSpans,
+    hasNextPage,
+  })
```
andresgutgon, line 68: DRY me baby

learningbizz: Dried it sir 🫡

packages/core/src/data-access/issues/getHITLSpansByDocument.ts
learningbizz: Dried it sir 🫡

packages/core/src/data-access/issues/getHITLSpansByDocument.ts
learningbizz: Dried it sir 🫡

packages/core/src/data-access/issues/getHITLSpansByDocument.ts
learningbizz: Dried it sir 🫡

packages/core/src/data-access/issues/getHITLSpansByDocument.ts
learningbizz: Talked offline: it was just the dry what we had to do

PR #2008, opened 2025-12-04
Fix collection operation tenacy

neoxelox:


packages/core/src/services/issues/shared.ts
```diff
@@ -168,8 +168,6 @@ export async function embedReason(reason: string) {
   if (!env.VOYAGE_API_KEY) return Result.ok<number[]>([])
 
   try {
-    if (!env.VOYAGE_API_KEY) return Result.nil()
```
andresgutgon, line 171: Is this ok?

andresgutgon (changes requested):
TESTs

PR #2013, opened 2025-12-05
Add feedback loop to automatic eval generation

learningbizz:
This PR adds the ability to pass the false positives and negatives calculated in the validation phase of an attempt of generating an evaluation, so if the attempt didn't arrive to our desired MCC threshold, it gives context to the new generation of the configuration to be aware of what was missed with the previous configuration.

### DIFF
--- a/packages/core/src/jobs/job-definitions/evaluations/calculateAlignmentMetricJob.test.ts
+++ b/packages/core/src/jobs/job-definitions/evaluations/calculateAlignmentMetricJob.test.ts
@@ -1,33 +1,35 @@
+import {
+  ActiveEvaluation,
+  EvaluationType,
+  EvaluationV2,
+  LlmEvaluationMetric,
+  Providers,
+} from '@latitude-data/constants'
+import { MIN_ALIGNMENT_METRIC_THRESHOLD } from '@latitude-data/constants/issues'
 import { Job } from 'bullmq'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { Result } from '../../../lib/Result'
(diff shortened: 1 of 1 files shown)


packages/core/src/data-access/issues/getSpanMessagesAndEvaluationResultsByIssue.ts
```diff
@@ -41,7 +41,7 @@ export async function getSpanMessagesAndEvaluationResultsByIssue({
     workspace: workspace,
     commit: commit,
     issue: issue,
-    pageSize: 5,
+    pageSize: 3,
```
learningbizz, line 44: Move back to 3, like this we have a 50/50 split when the user tries to generate an eval with 5 pos and 5 neg results.

The todo here in the future is to have an even split regardless of how many we have, not only 3

packages/core/src/jobs/job-definitions/evaluations/calculateAlignmentMetricJob.ts
```diff
@@ -231,6 +266,15 @@ async function deleteEvaluationAndRetryGeneration({
       model: model,
       workflowUuid: workflowUuid,
       generationAttempt: generationAttempt + 1, // retry generating another configuration
+      falsePositivesSpanAndTraceIdPairs:
+        falsePositivesSpanAndTraceIdPairs.slice(0, 3), // Limit the number to avoid too many tokens when generating the evaluation configuration
```
learningbizz, line 269: Added limit of 3 here as when it was unlimited, they added too much tokens.

PR #2017, opened 2025-12-09
Weekly produced logs

andresgutgon:
# What?
We want to send a report for each workspace with some stats. One of those stats is focused on the number of logs produced, tokens, and cost of the tokens. Also, knowing if they connected Latitude to production.


geclos:
Maybe we can reenable the analytics section of the project with these new stats? if they compute fast enough

PR #2030, opened 2025-12-11
Upgrade ts telemetry instrumentation

csansoon:
**Fixed traceloop instrumentation**
Traceloop instrumentation packages were all either outdated or deprecated. All of the packages have been upgraded or removed to use the latests supported versions.

**Removed package-specific types**
The `LatitudeTelemetry` class constructor is expected to be used with an `instrumentations` attribute, containing the class or package from your provider's official SDK. (Almost) every value was typed with the expected class or object, but this lead to type errors when the package version was off by even a patch. They've all been changed to "any" to mitigate this issue and reduce package size. Now, you must head to the documentation to know exactly what is expected.

**Integrations Documentation**
Speaking of docs, there's a new (hidden) tab in the documentation, displaying how to connect Latitude Telemetry to an existing JS project, with all our current supported providers and frameworks. This documentation is hidden for now, because some of these providers still don't work as expected, which will be solved in the next PR.


packages/telemetry/typescript/package.json
```diff
@@ -52,44 +52,34 @@
...
-    "@traceloop/instrumentation-together": "0.13.0",
-    "@traceloop/instrumentation-vertexai": "0.13.0",
+    "@traceloop/instrumentation-anthropic": "0.22.2",
+    "@traceloop/instrumentation-bedrock": "0.22.2",
+    "@traceloop/instrumentation-cohere": "0.22.2",
+    "@traceloop/instrumentation-langchain": "0.22.2",
+    "@traceloop/instrumentation-llamaindex": "0.22.2",
+    "@traceloop/instrumentation-openai": "0.22.2",
+    "@traceloop/instrumentation-together": "0.22.2",
+    "@traceloop/instrumentation-vertexai": "0.22.2",
```
andresgutgon, line 62: Should be optional dependencies. And when you want to install Vertex for example you install this package? Otherwise, are we loading all these packages for everyone?

csansoon: We are loading these packages for everyone. The idea is that integrating latitude telemetry in your app should be as simple and fast as possible

andresgutgon: But it adds more bloat to everyone. Just saying

PR #2034, opened 2025-12-12
feat(web): Add evaluation filtering and api endpoints

geclos:


apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/documents/[documentUuid]/evaluations/route.ts
```diff
@@ -1,58 +0,0 @@
...
-          projectId: projectId,
-          commitUuid: commit.uuid,
-          documentUuid: document.documentUuid,
-        })
-        .then((r) => r.unwrap())
-
-      return NextResponse.json(evaluations, { status: 200 })
-    },
-  ),
-)
```
Comment (geclos) [inline apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/documents/[documentUuid]/evaluations/route.ts:58]: keep REST api endpoints as flat as possible please, filters should not be uri params

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/RunsPage.tsx
```diff
@@ -99,6 +103,18 @@ export function RunsPage({
...
       { keepPreviousData: true },
     )
+  const hitlAnnotations = useMemo(() => {
+    return annotations.filter((a) => {
+      const evaluation = evaluations.find((ev) => ev.uuid === a.evaluationUuid)
+      if (!evaluation) return false
+
+      return (
+        evaluation.type === EvaluationType.Human &&
+        (evaluation as EvaluationV2<EvaluationType.Human>).configuration
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/RunsPage.tsx:113]: This is supper weird. Why we need to filter annotations. Annotations are HITL evaluations

Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/RunsPage.tsx:113]: The question is why annotations are not already filtered

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/RunsPage.tsx:113]: because the store is evaluationresultsbytrace, which can be more things that just annotations, better to keep the api flat and do the filtering client-side otherwise we end with a huge backend api surface that is harder to maintain

PR #2037, opened 2025-12-12
Key performance evaluation and some enhancements

neoxelox:


apps/web/src/components/evaluations/human/index.tsx
```diff
@@ -64,6 +72,33 @@ function ConfigurationAdvancedForm<M extends HumanEvaluationMetric>({
...
+        ),
+    [uuid, evaluations],
+  )
+
+  useEffect(() => {
+    if (mode === 'update') return
+    if (!metricSpecification) return
+    // FIXME: use proper callback setState so that you don't depend on options
+    // in the useEffect hook
+    setConfiguration({ ...configuration, enableControls: defaultControls })
```
andresgutgon: Yep, think if this can be done in a callback and avoid this useEffect enterely

packages/web-ui/src/ds/atoms/Label/index.tsx
```diff
@@ -19,7 +19,7 @@ const labelVariants = cva(
     variants: {
       variant: {
         default: '',
-        destructive: 'text-destructive',
+        destructive: '!text-destructive',
```
andresgutgon, line 22: Why is this needed?

PR #2041, opened 2025-12-15
Better db:fix

csansoon:
Improved `db:fix` script.

**Features**
 - Detect unapplied migrations (even those that drizzle doesn't)
 - Browse all migrations
 - Mark any migration as applied or unappled
 - Run (or re-run) any migration
 - Run specific migration statements

https://github.com/user-attachments/assets/6d35910b-0dad-47d4-9c90-fed472a154ca


geclos:
kinda feels like this should be in tools rather than packages/core

PR #2047, opened 2025-12-15
Eval icon in issue updates IRL

learningbizz:
As the evaluation generated within an issue uses a different store with a different SWR key than the one used when detecting which issues have evaluations in the dashboard page, when we generated an evaluation or deleted one from an issue, the icon would still be cached in the other SWR key. 

This PR fixes that by mutating the other store if these two actions occur

### DIFF
--- a/apps/web/src/stores/activeEvaluations.ts
+++ b/apps/web/src/stores/activeEvaluations.ts
@@ -47,7 +47,10 @@ export function useActiveEvaluations(
               opts.onEvaluationEnded(args.evaluation)
               if (!args.evaluation?.error) {
                 globalMutate(
-                  (key) => Array.isArray(key) && key[0] === 'issueEvaluations',
+                  (key) =>
+                    Array.isArray(key) &&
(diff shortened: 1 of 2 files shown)


apps/web/src/stores/activeEvaluations.ts
```diff
@@ -44,6 +45,11 @@ export function useActiveEvaluations(
           if (args.event === 'evaluationEnded') {
             if (opts?.onEvaluationEnded) {
               opts.onEvaluationEnded(args.evaluation)
+              if (!args.evaluation?.error) {
+                globalMutate(
+                  (key) => Array.isArray(key) && key[0] === 'issueEvaluations',
+                )
+              }
             }
             return prev.filter(
```
csansoon, line 57: This will force a mutate to all stores of `issuesEvaluations`, for all projects, commits and documents. Isn't there a more direct way to do it instead of forcing a mutate on all cached stores?

learningbizz: You're right, I can scope by project and commit

PR #2049, opened 2025-12-16
fix: allow issue updates without vectors when nothing to update

geclos:
## Problem

This PR fixes a production error: `BadRequestError: Received update issue operation without vectors`

The error occurred in the `reassignResultFromIssue` flow when:
1. `addResultToIssue` was called with an existing issue (`issueWasNew = false`)
2. The issue had results from other commits, so `canUpdateCentroid = false`
3. `updateIssue` was called with `{ centroid: undefined, issue: issue }` (no title, description, or centroid)
4. `upsertVector` received all undefined values and threw an error

## Solution

Updated `upsertVector` in `packages/core/src/services/issues/update.ts` to:
- ✅ Allow skipping Weaviate update when there's nothing to update (all three are undefined), while still allowing database update to proceed (for `updatedAt` timestamp)
- ✅ Preserve validation for partial property updates (only title OR only description) which should still error
- ✅ Maintain validation that new vectors require either vectors or properties

## Testing

- ✅ All existing tests pass
- ✅ Fixes the production error while preserving existing test behavior

## Stack Trace Context

The error was happening in:
```
at upsertVector (file:///app/apps/workers/dist/server.js:73584:11)
at updateIssue (file:///app/apps/workers/dist/server.js:73522:20)
at addResultToIssue (file:///app/apps/workers/dist/server.js:74185:26)
at reassignResultFromIssue (file:///app/apps/workers/dist/server.js:74624:10)
at assignEvaluationResultV2ToIssue (file:///app/apps/workers/dist/server.js:74669:10)
at discoverResultIssueJob (file:///app/apps/workers/dist/server.js:74902:35)
```

### DIFF
--- a/packages/core/src/services/issues/results/add.ts
+++ b/packages/core/src/services/issues/results/add.ts
@@ -104,7 +104,7 @@ export async function addResultToIssue<
       )
       if (!Result.isOk(validating)) return validating
 
-      let centroid
+      let centroid = issue.centroid
       if (canUpdateCentroid && embedding) {
         centroid = updateCentroid(
           { ...issue.centroid, updatedAt: issue.updatedAt },
--- a/packages/core/src/services/issues/results/remove.ts
+++ b/packages/core/src/services/issues/results/remove.ts
@@ -118,7 +118,7 @@ export async function removeResultFromIssue<
         return Result.error(validating.error)
       }
 
-      let centroid: IssueCentroid | undefined
+      let centroid: IssueCentroid | undefined = issue.centroid
       if (canUpdateCentroid && embedding) {
         centroid = updateCentroid(
           { ...issue.centroid, updatedAt: issue.updatedAt },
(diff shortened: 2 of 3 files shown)


packages/core/src/services/issues/results/add.ts
```diff
@@ -104,7 +104,7 @@ export async function addResultToIssue<
       )
       if (!Result.isOk(validating)) return validating
 
-      let centroid
+      let centroid = issue.centroid
```
geclos, line 107: i think it's safe here to assign to the existing issue's centroid

packages/core/src/services/issues/results/remove.ts
```diff
@@ -118,7 +118,7 @@ export async function removeResultFromIssue<
         return Result.error(validating.error)
       }
 
-      let centroid: IssueCentroid | undefined
+      let centroid: IssueCentroid | undefined = issue.centroid
```
geclos, line 121: ditto

packages/core/src/services/issues/results/add.ts
geclos: @neoxelox if you can confirm please

PR #2051, opened 2025-12-16
Telemetry capture wrapper

csansoon:
Added several changes to the `LatitudeTelemetry` class

## `telemetry.capture` wrapper
Wrap any instrumented code into the `telemetry.capture` method to automatically generate External spans on Latitude! The only thing you need is the `projectId` and a `path` for your prompt, even if it does not previously exist!

```ts
import { LatitudeTelemetry } from '@latitude-data/telemetry'
import OpenAI from 'openai'

export async function generateSupportReply(input: string) {
  const telemetry = new LatitudeTelemetry(process.env.LATITUDE_API_KEY, {
    instrumentations: { openai: OpenAI },
  })

  return telemetry.capture({
    projectId: 123, // The ID of your project in Latitude.
    path: 'your-prompt-path', // Add a name for your prompt. E.g.: "generate-support-reply".
  }, async () => {

    // Your regular LLM-powered feature code here
    const client = new OpenAI({ apiKey: process.env.OPENAI_API_KEY });
    const completion = await client.chat.completions.create({ ... })
    return completion.choices[0].message.content; // You don't need to return any specific value

  })
}
```

## Recategorized `LatitudeTelemetry` methods

Previously all methods were available right from the class instance. From `telemetry.shutdown` to `telemetry.completion`.

Since we will require users to instrument their own code when using services that we don't support, I reorganized the methods to improve discoverability.

These methods are still the same:
- `telemetry.flush`
- `telemetry.shutdown`
- `telemetry.capture` (well, this is a new one)

All methods that are used just to define spans, are moved to a `.span` group:
- `telemetry.span.completion`
- `telemetry.span.tool`
- `telemetry.span.http`
- `telemetry.span.step`
- ...

Other methods have been reorganized and renamed:
- `telemetry.resume(ctx)` →	`telemetry.context.resume(ctx)`
- `telemetry.instrument()` →	`telemetry.instrumentation.enable()`
- `telemetry.uninstrument()` →	`telemetry.instrumentation.disable()`
- `telemetry.tracerProvider(scope)` →	`telemetry.tracer.provider(scope)`
- `telemetry.tracer(scope)` →	`telemetry.tracer.get(scope)`

## Made context optional for span definition methods

All methods in `telemetry.span.*` will now have the context parameter optional, defaulting to `otel.context.active()`, which automatically obtains the currently active context.

This means that, when users instrument their own code execution and wrap it with the `.capture` method, they don't need to pass down and explicitly define the context.

---

For the `.capture` wrapper, a new "UnresolvedExternal" span type has been added. This type will never be stored in database, and is only used to capture telemetry on spans without having to resolve the documentUuid beforehand. When this span is processed for storage, it will automatically resolve the right values and then be stored as a regular "External" span instead.

### DIFF
--- a/packages/core/src/services/tracing/spans/unresolvedExternal.ts
+++ b/packages/core/src/services/tracing/spans/unresolvedExternal.ts
@@ -1,0 +1,184 @@
+import { database } from '../../../client'
+import {
+  ExternalSpanMetadata,
+  LogSources,
+  SPAN_SPECIFICATIONS,
+  SpanType,
+  BaseSpanMetadata,
+  HEAD_COMMIT,
+  ATTR_LATITUDE_PROJECT_ID,
+  ATTR_LATITUDE_PROMPT_PATH,
+  ATTR_LATITUDE_COMMIT_UUID,
+  ATTR_LATITUDE_DOCUMENT_UUID,
+  ATTR_LATITUDE_DOCUMENT_LOG_UUID,
+  ATTR_LATITUDE_SOURCE,
+} from '../../../constants'
+import { Result, TypedResult } from '../../../lib/Result'
+import {
+  CommitsRepository,
+  DocumentVersionsRepository,
+  ProjectsRepository,
+} from '../../../repositories'
+import { ExternalSpanSpecification } from './external'
+import { SpanBackendSpecification, SpanProcessArgs } from './shared'
+import { DocumentVersion } from '../../../schema/models/types/DocumentVersion'
+import { PromisedResult } from '../../../lib/Transaction'
+import { Workspace } from '../../../schema/models/types/Workspace'
+import { Commit } from '../../../schema/models/types/Commit'
+import { findFirstUserInWorkspace } from '../../../data-access/users'
+import { createNewDocument } from '../../documents'
+import { createCommit, mergeCommit } from '../../commits'
+import { generateUUIDIdentifier } from '../../../lib/generateUUID'
+import { omit } from 'lodash-es'
+
+const specification = SPAN_SPECIFICATIONS[SpanType.UnresolvedExternal]
+
+type ExternalMetadataResult = TypedResult<
+  Omit<ExternalSpanMetadata, keyof BaseSpanMetadata>
+>
+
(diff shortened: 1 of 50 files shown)


packages/core/src/services/tracing/spans/unresolvedExternal.ts
```diff
@@ -0,0 +1,179 @@
...
+  const { documentVersion, commit } = result.unwrap()
+
+  const resolvedAttributes = {
+    ...omit(attributes, [
+      ATTR_LATITUDE_PROMPT_PATH,
+      ATTR_LATITUDE_PROJECT_ID,
+      ATTR_LATITUDE_COMMIT_UUID,
+    ]),
+    [ATTR_LATITUDE_DOCUMENT_UUID]: documentVersion.documentUuid,
+    [ATTR_LATITUDE_DOCUMENT_LOG_UUID]: generateUUIDIdentifier(),
```
geclos: i should be able to pass an identifier from the sdk, so that i can group multiple generations/traces in a chat context

csansoon: There's the `externalId` that we use for prompts too

geclos: that's different, it's intended to be used to link traces to particular identifiers in the client's side. In latitude we only group traces by document loguuid, this is basically to support the same thing other observ tools do to 'group' traces, eg langfuse calls them [sessions](https://langfuse.com/docs/observability/features/sessions).

csansoon: Hmmm that would be the `documentLogUuid`

csansoon: Yeah I could add an optional `conversationUuid` (?)

geclos: 👌🏼 that would be clutch

csansoon: done 👌

PR #2054, opened 2025-12-17
Calculate MCC alignment daily

learningbizz:
Once the evaluation for a specific issue is generated, we don't realign the evaluation with new annotations the user does. We have to re-calculate it so the user is aware of its evolution.

This PR adds a daily maintenance job to keep the alignment of each evaluation attached to an issue up-to-date with the new users annotations.

In addition:
- Remove alignment_metric from DB and calculating the % at runtime (like this we only have one source of truth)
- Rename validation job of the generation of evals to be clearer

Todos:
- Improve daily alignment calculation by saving latest positive and negative eval results used.
- Improve query HITL spans to connect well with evaluationVersions table
- Re-calculate the alignment when updating an eval configuration in the eval details page (service already created to do so, just pending linkning with an action)

### DIFF
--- a/packages/core/src/services/evaluationsV2/generateFromIssue/recalculateAlignmentMetric.ts
+++ b/packages/core/src/services/evaluationsV2/generateFromIssue/recalculateAlignmentMetric.ts
@@ -82,6 +82,8 @@ export async function recalculateAlignmentMetric(
     // Re-balance to have equal amounts
     const targetLength = Math.min(passSpansFromYesterday.length, failSpansFromYesterday.length) // prettier-ignore
 
+    console.log('targetLength', targetLength)
+
     const passSpansFromYesterdaySliced = passSpansFromYesterday.slice(0, targetLength) // prettier-ignore
     const failSpansFromYesterdaySliced = failSpansFromYesterday.slice(0, targetLength) // prettier-ignore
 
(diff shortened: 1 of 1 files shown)


packages/core/src/services/evaluationsV2/generateFromIssue/getEqualAmountsOfPositiveAndNegativeExamples.ts
```diff
@@ -0,0 +1,82 @@
...
+
+IMPORTANT: 
+- The evaluation MUST fail when the issue is present in the span, as this logic is used within the issue discovery and its how we want our end goal to be.
+  We want the evaluations to be like unit tests, where if all of them pass for a given trace of a document, that means that the trace has no issues, that its good!
+- The spans MUST be from HITL evaluation results only, as we want to use ONLY the user's annotations to calculate the MCC, not from other evaluations results.
+
+Thumbs up evalResults of the same document or evalResults of other issues of the same document count as negative evalResults because
+ they are cases in which the new evaluation should return a positive result, as that span doesnt have that issue (its good!).
+*/
+export async function getEqualAmountsOfPositiveAndNegativeExamples(
```
learningbizz, line 24: This is not new, just extracted it to its own service

PR #2056, opened 2025-12-17
Improvement/cutoff dates spans

learningbizz:
Currently, when calculating the daily alignment, we don't have an effective system to use the HITL spans from the user. If a user one day annotates a bad annotation for an issue but no good ones, we won't take it into consideration for the daily re-calculation as we don't have a good one to calculate it with (we always need to have a balance of positive and negative cases to calculate a good confusion matrix).

This PR adds a date system to keep track of what is the latest span used for both positive and negative calculation of the alignment metric, and with this keep track of the pending spans in an elegant way to not waste any.

### DIFF
--- a/packages/core/src/jobs/job-definitions/evaluations/validateGeneratedEvaluationJob.test.ts
+++ b/packages/core/src/jobs/job-definitions/evaluations/validateGeneratedEvaluationJob.test.ts
@@ -247,14 +247,14 @@ describe('validateGeneratedEvaluationJob', () => {
           {
             id: 'span-1',
             traceId: 'trace-1',
-            createdAt: new Date('2024-01-15T10:00:00.000Z'),
+            createdAt: '2024-01-15T10:00:00.000Z',
           },
         ],
         spanAndTraceIdPairsOfExamplesThatShouldFailTheEvaluation: [
           {
             id: 'span-2',
             traceId: 'trace-2',
-            createdAt: new Date('2024-01-15T11:00:00.000Z'),
+            createdAt: '2024-01-15T11:00:00.000Z',
           },
         ],
       })
@@ -410,14 +410,14 @@ describe('validateGeneratedEvaluationJob', () => {
           {
             id: 'span-1',
             traceId: 'trace-1',
-            createdAt: new Date('2024-01-15T10:00:00.000Z'),
+            createdAt: '2024-01-15T10:00:00.000Z',
           },
         ],
         spanAndTraceIdPairsOfExamplesThatShouldFailTheEvaluation: [
           {
             id: 'span-2',
             traceId: 'trace-2',
-            createdAt: new Date('2024-01-15T11:00:00.000Z'),
+            createdAt: '2024-01-15T11:00:00.000Z',
           },
         ],
         evaluationResults: {


packages/core/src/jobs/job-definitions/evaluations/recalculateAlignmentMetricJob.ts
```diff
@@ -96,22 +91,33 @@ export const recalculateAlignmentMetricJob = async (
       alreadyCalculatedAlignmentMetricMetadata = evaluation.alignmentMetricMetadata ?? undefined // prettier-ignore
     }
 
-    const { confusionMatrix } = await evaluateConfiguration({
-      childrenValues,
-      spanAndTraceIdPairsOfExamplesThatShouldPassTheEvaluation,
-      spanAndTraceIdPairsOfExamplesThatShouldFailTheEvaluation,
-      alreadyCalculatedAlignmentMetricMetadata,
-    }).then((r) => r.unwrap())
+    const { confusionMatrix, latestPositiveSpanDate, latestNegativeSpanDate } =
```
learningbizz, line 94: Getting the latest span dates now after all the runEvals are done, as we're making sure now here that we always have the same amount of positive and negative results, regardless if one eval run fails (what I call rebalancing results).

As we're doing this rebalancing here now, we will calculate the cutoff dates directly here and avoid recalculating twice.

packages/constants/src/evaluations/index.ts
```diff
@@ -167,6 +167,9 @@ export type AlignmentMetricMetadata = {
     falsePositives: number
     falseNegatives: number
   }
+  // Cutoff dates for incremental processing - ISO date strings
+  lastProcessedPositiveSpanDate?: string
+  lastProcessedNegativeSpanDate?: string
```
csansoon, line 172: Why use `string` instead of `Date`?

csansoon (dismissed):

packages/constants/src/evaluations/index.ts
learningbizz: To keep consistency with the database, as we're saving the metadata as a JSONB, and it serializes as a string in the database.

Also, we're not needing any Date operation with it, so no need for adding complexity of serializing/deserializing

packages/core/src/jobs/job-definitions/evaluations/recalculateAlignmentMetricJob.ts
```diff
@@ -15,20 +15,15 @@ import {
...
-  spanAndTraceIdPairsOfExamplesThatShouldPassTheEvaluation: {
-    spanId: string
-    traceId: string
-  }[]
-  spanAndTraceIdPairsOfExamplesThatShouldFailTheEvaluation: {
-    spanId: string
-    traceId: string
-  }[]
+  spanAndTraceIdPairsOfExamplesThatShouldPassTheEvaluation: Pick<Span, 'id' | 'traceId' | 'createdAt'>[] // prettier-ignore
+  spanAndTraceIdPairsOfExamplesThatShouldFailTheEvaluation: Pick<Span, 'id' | 'traceId' | 'createdAt'>[] // prettier-ignore
```
andresgutgon: Why this ocd? I find it even prettier (pun intended) multi-line

andresgutgon (dismissed):

packages/core/src/jobs/job-definitions/evaluations/recalculateAlignmentMetricJob.ts
learningbizz: OCD 🤷 It look real weird with the pick 😂

PR #2061, opened 2025-12-18
fix: unoptimized traces

csansoon:
`assembleTrace` is used on several different parts of the code to fetch and build the **structure** of a trace. In some of the cases, we will also get the list of messages from the metadata of the last generation span.

Previously, we were fetching the metadata from every single span when assembling the trace, which is VERY EXPENSIVE since we're accessing S3 N times.

Now, instead, the method has been changed to `assembleTraceStructure`, which does not perform any access to S3 storage, and only THEN accessing the last storage span's metadata to obtain the list of messages

### DIFF
--- a/apps/web/src/app/api/conversations/[conversationId]/route.ts
+++ b/apps/web/src/app/api/conversations/[conversationId]/route.ts
@@ -3,7 +3,7 @@ import { errorHandler } from '$/middlewares/errorHandler'
 import { AssembledTrace } from '@latitude-data/core/constants'
 import { SpansRepository } from '@latitude-data/core/repositories'
 import { Workspace } from '@latitude-data/core/schema/models/types/Workspace'
-import { assembleTrace } from '@latitude-data/core/services/tracing/traces/assemble'
+import { assembleTraceStructure } from '@latitude-data/core/services/tracing/traces/assemble'
 import { NextRequest, NextResponse } from 'next/server'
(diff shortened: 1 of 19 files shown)


apps/web/src/app/api/conversations/[conversationId]/route.ts
```diff
@@ -3,7 +3,7 @@ import { errorHandler } from '$/middlewares/errorHandler'
 import { AssembledTrace } from '@latitude-data/core/constants'
 import { SpansRepository } from '@latitude-data/core/repositories'
 import { Workspace } from '@latitude-data/core/schema/models/types/Workspace'
-import { assembleTrace } from '@latitude-data/core/services/tracing/traces/assemble'
+import { assembleTraceStructure } from '@latitude-data/core/services/tracing/traces/assemble'
```
Comment (geclos) [inline apps/web/src/app/api/conversations/[conversationId]/route.ts:6]: nitpick but i don't think the rename was necessary

packages/core/src/services/tracing/traces/assemble.ts
```diff
@@ -63,44 +73,82 @@ export async function assembleTrace(
   return Result.ok({ trace })
 }
 
-async function assembleSpan({
+/**
+ * Assembles a trace structure and fetches only the completion span's metadata.
```
geclos, line 77: specify which completion span there can be many in a trace

PR #2065, opened 2025-12-22
Telemetry spec upgrade

csansoon:
**TELEMETRY SPEC UPGRADE**
Our telemetry processing is still following deprecated specs, which means that many attributes are not being read correctly.
This PR focuses on updating it to use all current (and old) standards that we support. The supported standards are:
 - OpenTelemetry
 - Openllmetry
 - Openinference
 - Vercel's

---

**Reorganization**
Before simply updating the specs, some reorganization had to be done.

***Attributes and values***: Before, all attribute names were defined as lots of individual constants. Instead, they are now defined in an `ATTRIBTUES` object. This now allows us to scope each attribute depending on the origin spec, giving us more visibility and control over what each one does and how. This also allows us to add new "_deprecated" attributes making the handling of them easier.

***Span services***: `core/services/tracing/spans` was full of different 10+ files, where some of them did very similar stuff. It was hard to know which did which, so I grouped them based on purpose and logic. There are now 4 categories: `specification`, `ingestion`, `export` and `fetching`.

- `✅ done`

---

**Specs update**
Some of the old specs have to be moved to a `_deprecated` domain, and the new ones added both in the attributes definition and everywhere where it could be stored.

- `✅ done`

---

**Messages update**
We are aligning with OpenTelemetry for storing our messages data. Not only the spec we're using is a PITA to use, but it is also deprecated! So this has to be updated too.

- `⌛ pending` — delayed


packages/constants/src/tracing/span.ts
```diff
@@ -12,18 +12,21 @@ export enum SpanKind {
...
+  Prompt = 'prompt', // Running a prompt
   Chat = 'chat', // Continuing a conversation (adding messages)
   External = 'external', // Wrapping external generation code
   UnresolvedExternal = 'unresolved_external', // External span that needs path & potential version resolution
-  Step = 'step',
+
+  // Added a HTTP span to capture raw HTTP requests and responses when running from Latitude
+  Http = 'http', // Note: raw HTTP requests and responses
+
+  // Any known span from supported specifications will be grouped into one of these types
```
learningbizz, line 24: weird comment, isn't this talking about every spanType?

packages/constants/src/tracing/span.ts
```diff
@@ -46,12 +49,30 @@ export type EvaluableSpanType =
...
 
 export const SPAN_SPECIFICATIONS = {
-  [SpanType.Tool]: {
-    name: 'Tool',
-    description: 'A tool call',
-    isGenAI: true,
+  [SpanType.Prompt]: {
+    name: 'Prompt',
+    description: 'A prompt span',
+    isGenAI: false,
```
learningbizz, line 55: `isGenAI` is not the best name as we're coupling our code to this standard (the most official and used rn) that we might want to change in the future (hope not, but possible). Its not a good practice

Maybe calling it something more generalistic, and adding a comment mentioning the standard is now genAI, would be less coupled and more robust to changes.

Calling it `isMetaStandard` or something like that would suffice

packages/core/src/services/tracing/spans/specifications/tool.ts
```diff
@@ -0,0 +1,159 @@
...
+import { Result, TypedResult } from '../../../../lib/Result'
+import { SpanProcessArgs } from '../shared'
+
+const specification = SPAN_SPECIFICATIONS[SpanType.Tool]
+export const ToolSpanSpecification = {
+  ...specification,
+  process: process,
+}
+
+async function process(
```
learningbizz, line 21: tests

packages/core/src/services/tracing/spans/specifications/external.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Result } from '../../../../lib/Result'
+import { SpanProcessArgs } from '../shared'
+
+const specification = SPAN_SPECIFICATIONS[SpanType.External]
+export const ExternalSpanSpecification = {
+  ...specification,
+  process: process,
+}
+
+async function process(
```
learningbizz, line 17: tests

packages/core/src/services/tracing/spans/specifications/prompt.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Result } from '../../../../lib/Result'
+import { SpanProcessArgs } from '../shared'
+
+const specification = SPAN_SPECIFICATIONS[SpanType.Prompt]
+export const PromptSpanSpecification = {
+  ...specification,
+  process: process,
+}
+
+async function process(
```
learningbizz, line 17: tests again

packages/constants/src/tracing/span.ts
csansoon: This is not something new, I am not changing that for now.
Also, the official spec is named "GenAI". This is not just _"is the currently accepted standard"_ or _"is not custom"_, this is ACTUALLY _"is stored with THE GenAI standard"_

packages/constants/src/tracing/span.ts
csansoon: No. `Prompt`, `Chat`, `External`, and `Http` are not types supported for any spec.
What I mean is that, any span processed from a supported spec (OpenTelemetry, OpenLLMetry, OpenInference & Vercel's AI SDK) will always fall into one of those: `Completion`, `Tool` or `Embedding`. Any other is either set as `Unknown` or added by Latitude.

packages/core/src/services/tracing/spans/specifications/tool.ts
csansoon: I did not change the logic in this file. Just reorganization.
This PR has a specific objective: Reorganize the spec definition and update them. Just with this objective, the PR is already too large.

We should not add a new objective to an already complex PR. I agree these methods should be tested, but that's out of scope for this PR

packages/core/src/services/tracing/spans/specifications/external.ts
csansoon: ditto

packages/core/src/services/tracing/spans/specifications/prompt.ts
csansoon: ditto

PR #2075, opened 2026-01-02
Do not update issues details when centroid is not updated

andresgutgon:
# What?
Some updates on issue details re-generation

## TODO
- [x] Results that shouldn't move the centroid also shouldn't trigger issue details re-generation
- [x] results from experiments are excluded for issue details job generation

### DIFF
--- a/packages/core/src/repositories/evaluationResultsV2Repository.ts
+++ b/packages/core/src/repositories/evaluationResultsV2Repository.ts
@@ -812,6 +812,7 @@ export class EvaluationResultsV2Repository extends Repository<EvaluationResultV2
       eq(issueEvaluationResults.issueId, issueId),
       isNotNull(commits.mergedAt),
       isNull(evaluationResultsV2.error),
+      isNull(evaluationResultsV2.experimentId),
       sql`${evaluationResultsV2.hasPassed} IS NOT TRUE`,
     ]
 
--- a/packages/core/src/services/issues/results/add.ts
+++ b/packages/core/src/services/issues/results/add.ts
@@ -145,6 +145,8 @@ export async function addResultToIssue<
       return Result.ok({ issue, histogram, result })
     },
     async ({ issue }) => {
+      if (!shouldUpdateCentroid) return
+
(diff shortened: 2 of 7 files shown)


packages/core/src/repositories/evaluationResultsV2Repository.ts
```diff
@@ -806,12 +806,16 @@ export class EvaluationResultsV2Repository extends Repository<EvaluationResultV2
     return Result.ok<EvaluationResultV2[]>(results as EvaluationResultV2[])
   }
 
+  /**
+   * Results from experiments are excluded.
```
neoxelox: remove this unnecessary comment

packages/core/src/services/issues/results/add.ts
```diff
@@ -148,7 +148,7 @@ export async function addResultToIssue<
       const payload = { workspaceId: workspace.id, issueId: issue.id }
```
neoxelox, line 150: do an early return here of `!shouldUpdateCentroid,` because both tasks should not be enqueued if the centroid was not moved.

packages/core/src/services/issues/results/remove.ts
```diff
@@ -181,13 +181,15 @@ export async function removeResultFromIssue<
         const payload = { workspaceId: workspace.id, issueId: issue.id }
```
neoxelox: do an early return here (above) of `!shouldUpdateCentroid,` because both tasks should not be enqueued if the centroid was not moved.

packages/core/src/repositories/evaluationResultsV2Repository.ts
andresgutgon: Easy to spot this. I bet in the future someone will ask why experiments are not affecting the centroid 😂

neoxelox: eee, i dont like this, like why comment about experiments and not about the other 31313 filters? the code is self explanatory

neoxelox (approved):

PR #2081, opened 2026-01-05
Improve individual weekly email job performance

andresgutgon:
Last week we improved the fetching of all workspaces that are active based on the number of spans in the last 4 weeks here https://github.com/latitude-dev/latitude-llm/pull/2072 but this week individual jobs also failed for some workspaces because we're hitting the same spans table which is huge.

## TODO
- [x] ~~Run migration to add composite index~~ Not doing it for now
- [x] Fix tests if any is failing

### DIFF
--- a/packages/core/src/data-access/weeklyEmail/logs/index.tsx
+++ b/packages/core/src/data-access/weeklyEmail/logs/index.tsx
@@ -1,7 +1,6 @@
 import {
   and,
   between,
-  count,
   countDistinct,
   desc,
   eq,
@@ -23,12 +22,12 @@ import { Workspace } from '../../../schema/models/types/Workspace'
 import { SureDateRange } from '../../../constants'
 import { getDateRangeOrLastWeekRange } from '../utils'
 
-async function getAllTimesSpansProductionCount(
+async function hasProductionSpans(
   { workspace }: { workspace: Workspace },
   db = database,
-) {
-  return db
-    .select({ count: count() })
+): Promise<boolean> {
+  const result = await db
+    .select({ exists: sql<number>`1` })
     .from(spans)
     .where(
       and(
@@ -37,7 +36,9 @@ async function getAllTimesSpansProductionCount(
         inArray(spans.source, RUN_SOURCES[RunSourceGroup.Production]),
       ),
     )
-    .then((r) => r[0].count)
+    .limit(1)
+
+  return result.length > 0
 }
 
 async function getGlobalLogsStats(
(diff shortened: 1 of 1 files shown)


packages/core/src/data-access/weeklyEmail/logs/index.tsx
```diff
@@ -23,12 +22,12 @@ import { Workspace } from '../../../schema/models/types/Workspace'
...
-async function getAllTimesSpansProductionCount(
+async function hasProductionSpans(
   { workspace }: { workspace: Workspace },
   db = database,
-) {
-  return db
-    .select({ count: count() })
+): Promise<boolean> {
+  const result = await db
+    .select({ exists: sql<number>`1` })
```
andresgutgon, line 30: This improves this query. `count` does a full scan and we just need here to know if at least one span in production exists.

packages/core/src/schema/models/spans.ts
```diff
@@ -84,5 +84,10 @@ export const spans = latitudeSchema.table(
       table.id,
     ),
     index('spans_previous_trace_id_idx').on(table.previousTraceId),
+    index('spans_workspace_started_at_type_idx').on(
```
andresgutgon: I didn't find a workaround to fix the rest of the queries in the logs (spans) service. Please let me know what you think

packages/core/drizzle/0252_add_spans_composite_index_for_workspace_started_at_type.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX "spans_workspace_started_at_type_idx" ON "latitude"."spans" USING btree ("workspace_id","started_at","type");
```
geclos: ⚠️ concurrently otherwise you will block this table for several minutes

geclos: also a reality check composite indeces almost never fix the issue

packages/core/src/schema/models/spans.ts
geclos: i don't think this will fix your issue composite indices improve performance but not dramatically if you already have indeces in the columns which you do i assume

andresgutgon: Will see

andresgutgon: The funny thing is that the same code runs from backoffice is working but during weekly is failing. Maybe because DB is hitting multiple times to spans during weekly email generation?

PR #2083, opened 2026-01-05
Fix HITL span queries to avoid duplicates

learningbizz:
We used to have the evalResults table join the evalVersions only with the uuid, but this is wrong as we have versioning that makes us not have unique evals by uuid. 

This fix fixes this issue, and optimizes it by adding the type and metric of the eval to the result, so we can query it directly without linking to the evaluationVersions table

### DIFF
--- a/apps/workers/src/workers/schedule.ts
+++ b/apps/workers/src/workers/schedule.ts
@@ -59,10 +59,10 @@ export async function setupSchedules() {
     { opts: { attempts: 1 } },
   )
 
-  // Every day at 3 AM - Schedules backfill jobs for evaluation results type and metric
+  // Every day at 12 AM - Schedules backfill jobs for evaluation results type and metric
   await maintenanceQueue.upsertJobScheduler(
     'scheduleBackfillEvaluationResultsTypeAndMetricJobs',
-    { pattern: '0 0 3 * * *' },
+    { pattern: '0 0 12 * * *' },
     { opts: { attempts: 1 } },
   )
 }


packages/constants/src/evaluations/index.ts
```diff
@@ -235,6 +235,8 @@ export type EvaluationResultV2<
   workspaceId: number
   commitId: number
   evaluationUuid: string
+  evaluationType?: T | null
```
neoxelox: just type and metric please

packages/core/src/schema/models/evaluationResultsV2.ts
```diff
@@ -36,6 +37,8 @@ export const evaluationResultsV2 = latitudeSchema.table(
       .notNull()
       .references(() => commits.id, { onDelete: 'restrict' }),
     evaluationUuid: uuid('evaluation_uuid').notNull(),
+    evaluationType: varchar('evaluation_type', { length: 32 }).$type<EvaluationType>(),
```
neoxelox: just type and metric pls

packages/core/drizzle/0253_add_evaluation_type_to_results.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE "latitude"."evaluation_results_v2" ADD COLUMN "evaluation_type" varchar(32);--> statement-breakpoint
+ALTER TABLE "latitude"."evaluation_results_v2" ADD COLUMN "evaluation_metric" varchar(64);--> statement-breakpoint
+CREATE INDEX "evaluation_results_v2_type_workspace_idx" ON "latitude"."evaluation_results_v2" USING btree ("evaluation_type","workspace_id");
```
geclos: concurrently, you will need to edit the migration to not run in a transaction, check out how we do other concurrent index creations in other migrations

packages/core/src/schema/models/evaluationResultsV2.ts
```diff
@@ -101,5 +104,9 @@ export const evaluationResultsV2 = latitudeSchema.table(
     uniqueIndex(
       'evaluation_results_v2_unique_evaluated_span_id_evaluation_uuid_idx',
     ).on(table.evaluatedSpanId, table.evaluatedTraceId, table.evaluationUuid),
+    index('evaluation_results_v2_type_workspace_idx').on(
```
neoxelox, line 109: all the enum thingies i will always create composite index with a timestamp at the end, more usefuls

packages/core/drizzle/0254_backfill_evaluation_type_to_results.sql
```diff
@@ -0,0 +1,13 @@
+-- Custom SQL migration file, put your code below! --
+
+-- Backfill evaluationType and evaluationMetric for existing evaluation results
+-- This migration is idempotent - it only updates rows where evaluation_type IS NULL
+UPDATE "latitude"."evaluation_results_v2" r
+SET evaluation_type = v.type, evaluation_metric = v.metric
+FROM (
+  SELECT DISTINCT ON (evaluation_uuid) evaluation_uuid, type, metric
+  FROM "latitude"."evaluation_versions"
+  ORDER BY evaluation_uuid, id DESC
```
geclos: hmm let's check out the N here, if it's too big this will need to be a data migration in code rather than sql

packages/core/src/schema/models/evaluationResultsV2.ts
learningbizz: explain yourself

packages/core/drizzle/0254_backfill_evaluation_type_to_results.sql
learningbizz: Why?

packages/core/src/schema/models/evaluationResultsV2.ts
learningbizz: the type is of the eval, not the result, I think its a good distinction, no?

packages/core/src/schema/models/evaluationResultsV2.ts
neoxelox: ex: `index('spans_type_started_at_idx').on(table.type, table.startedAt),`

packages/core/src/schema/models/evaluationResultsV2.ts
neoxelox: its the type and metric of the result too

packages/core/drizzle/0254_backfill_evaluation_type_to_results.sql
neoxelox: because the query will timeout

packages/core/src/schema/models/evaluationResultsV2.ts
```diff
@@ -36,6 +37,10 @@ export const evaluationResultsV2 = latitudeSchema.table(
       .notNull()
       .references(() => commits.id, { onDelete: 'restrict' }),
     evaluationUuid: uuid('evaluation_uuid').notNull(),
+    evaluationType: varchar('evaluation_type', {
+      length: 32,
+    }).$type<EvaluationType>(),
+    evaluationMetric: varchar('evaluation_metric', { length: 64 }),
```
andresgutgon: These 2 fields should be not NULL no?

packages/core/drizzle/0254_backfill_evaluation_type_to_results.sql
learningbizz: Okay, I understand because we have more control as its a big number of rows (i.e. more observability, ability to restart not from scratch if the code fails, etc.)

packages/core/src/schema/models/evaluationResultsV2.ts
learningbizz: They have to be in the beginning as there needs to be a data migration. I will remove it after

packages/core/drizzle/0253_add_evaluation_type_to_results.sql
andresgutgon:
```sql
COMMIT;
--> statement-breakpoint
CREATE INDEX CONCURRENTLY ...
```

The important thing is letting PostgreSQL know that the index will be run after the transaction.

andresgutgon: Search for examples in migrations

packages/core/src/schema/models/evaluationResultsV2.ts
andresgutgon: I mean after the data migrations is finished. Don't let these fields as nullable

learningbizz: Yes good catch

packages/core/drizzle/0253_add_evaluation_type_to_results.sql
learningbizz: Yeah yeah found some, thanks

packages/core/src/schema/models/evaluationResultsV2.ts
learningbizz: Sorry I didnt write my question clearly, why do you think adding the timestamp is a good idea?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/RunPanel/index.tsx
```diff
@@ -174,7 +174,12 @@ export function RunPanel({
                   HumanEvaluationMetric
                 >
               }
-              result={annotations.bottom.result}
+              result={
+                annotations.bottom.result as EvaluationResultV2<
+                  EvaluationType.Human,
```
Comment (learningbizz) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/RunPanel/index.tsx:179]: Needed now as we added generics in `type` and `metric` of evaluationResults

apps/workers/src/workers/schedule.ts
```diff
@@ -58,4 +58,11 @@ export async function setupSchedules() {
     { pattern: '0 0 1 * * *' },
     { opts: { attempts: 1 } },
   )
+
+  // Every day at 3 AM - Schedules backfill jobs for evaluation results type and metric
+  await maintenanceQueue.upsertJobScheduler(
+    'scheduleBackfillEvaluationResultsTypeAndMetricJobs',
```
learningbizz, line 64: Will be changed once I get an approve

packages/core/src/schema/models/evaluationResultsV2.ts
learningbizz: sold, done

packages/core/drizzle/0254_backfill_evaluation_type_to_results.sql
learningbizz: done

packages/core/src/schema/models/evaluationResultsV2.ts
```diff
@@ -37,10 +37,12 @@ export const evaluationResultsV2 = latitudeSchema.table(
...
     evaluationUuid: uuid('evaluation_uuid').notNull(),
-    evaluationType: varchar('evaluation_type', {
+    type: varchar('type', {
       length: 32,
-    }).$type<EvaluationType>(),
-    evaluationMetric: varchar('evaluation_metric', { length: 64 }),
+    })
+      .$type<EvaluationType>()
+      .notNull(),
+    metric: varchar('metric', { length: 64 }).notNull(),
```
andresgutgon: Wait, but you can not merge this without backfilling no?

learningbizz: I got confused 😂 fixed it already

PR #2087, opened 2026-01-06
Filter issues from deleted documents in issues dashboard

andresgutgon:
# What?
Issues from deleted documents in current commit should not be visible in the issues dashboard.

## TODO
- [x] Filter issues from deleted documents in the current commit the user is seing
- [x] Unassign the ocurrences to that issue from the commit the document is deleted

### DIFF
--- a/packages/core/src/repositories/issuesRepository.ts
+++ b/packages/core/src/repositories/issuesRepository.ts
@@ -15,6 +15,7 @@ import {
   eq,
   getTableColumns,
   ilike,
+  inArray,
   isNotNull,
   isNull,
   or,
@@ -36,6 +37,7 @@ import { type Issue } from '../schema/models/types/Issue'
 import { type Project } from '../schema/models/types/Project'
 import { CommitsRepository } from './commitsRepository'
 import { IssueEvaluationResultsRepository } from './issueEvaluationResultsRepository'
+import { DocumentVersionsRepository } from './documentVersionsRepository'
 import { IssueHistogramsRepository } from './issueHistogramsRepository'
 import Repository from './repositoryV2'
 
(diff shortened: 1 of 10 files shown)


packages/core/src/services/documents/destroyOrSoftDeleteDocuments.ts
```diff
@@ -141,79 +142,106 @@ export async function destroyOrSoftDeleteDocuments(
...
+        updateDocumentsAsSoftDeleted({ toBeUpdated, commit, tx }),
+      ])
+
+      await invalidateDocumentsCacheInCommit(commit.id, tx)
+      await pingProjectUpdate({ projectId: commit.projectId }, transaction)
+
+      return Result.ok(true)
+    },
+    () => {
+      publisher.publishLater({
```
neoxelox, line 234: The github diff is distraous, i guess the only thing you have changed is sending this new event?

packages/core/src/repositories/issuesRepository.ts
```diff
@@ -444,15 +454,23 @@ export class IssuesRepository extends Repository<Issue> {
...
   }) {
     const conditions: SQL[] = [
       this.scopeFilter,
       eq(issues.projectId, project.id),
     ]
 
+    if (activeDocumentUuids.length > 0) {
+      conditions.push(inArray(issues.documentUuid, activeDocumentUuids))
+    } else {
+      conditions.push(sql`FALSE`)
```
neoxelox, line 471: what why?

neoxelox: Ah, you want to return an empty list of issues because there are no active documents in the commit... maybe you should do an early return in the calling function no? so no db hit at all

packages/core/src/services/documents/destroyOrSoftDeleteDocuments.ts
andresgutgon: Correct

packages/core/src/repositories/issuesRepository.ts
andresgutgon: This else won't happen rarely. All commits should have an active document

PR #2089, opened 2026-01-06
Use correct resultReason method when fetching eval reason

learningbizz:
We used to do a quickfix to get the reason, now we implemented the improved version using the correct resultReason method

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,8 @@ __testing__
 # New spans
 apps/web/ingest
 apps/workers/workspaces/*/traces
+provider-logs/*
+workspaces/*
 
 # Helm chart secrets (do not commit)
 charts/latitude/values.secrets.yaml
--- a/packages/core/src/data-access/issues/getSpanMessagesAndEvaluationResultsByIssue.ts
+++ b/packages/core/src/data-access/issues/getSpanMessagesAndEvaluationResultsByIssue.ts
@@ -5,13 +5,18 @@ import {
   EvaluationResultsV2Repository,
 } from '../../repositories'
 import { Result } from '../../lib/Result'
-import { EvaluationResultV2 } from '../../constants'
+import {
+  EvaluationResultSuccessValue,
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -94,6 +94,8 @@ __testing__
 # New spans
 apps/web/ingest
 apps/workers/workspaces/*/traces
+provider-logs/*
```
learningbizz, line 97: The reason I might have gotten the "missing completion metadata" issue because I didnt have the `FILES_STORAGE_PATH`, `FILE_PUBLIC_PATH`, and `PUBLIC_FILES_STORAGE_PATH` env vars. Adding these created me lots of new files in these folders, so adding them in the gitignore

packages/core/src/data-access/issues/getSpanMessagesAndEvaluationResultsByIssue.ts
```diff
@@ -79,30 +86,37 @@ export async function getSpanMessagesAndEvaluationResultsByIssue({
...
-  // LLM, Rule, and Human evaluations all have a reason field (required for LLM, optional for Rule/Human)
-  if ('reason' in result.metadata && result.metadata.reason) {
-    return result.metadata.reason ?? ''
-  }
-  return ''
+  const specification = getEvaluationMetricSpecification(evaluation)
+  const resultReason = specification.resultReason as (
+    result: EvaluationResultSuccessValue,
+  ) => string | undefined
+  return resultReason(result) ?? ''
```
neoxelox, line 121:
```suggestion
  const resultReason = specification.resultReason(result as EvaluationResultSuccessValue)
  return resultReason ?? ''
```

learningbizz: TS type complaint with this: 

rule.ts(78, 5): 'pattern' is declared here.

PR #2090, opened 2026-01-07
Improve composite evaluation creation form validation

andresgutgon:
When existing evaluations don't meet the criteria we need to create a composite evaluation no evaluations are sent and we were returning a toast message which dissapear and is hard to understand what's happening. Now we display the error message inline in the sub evaluations selector.

image

### DIFF
--- a/packages/core/src/services/evaluationsV2/composite/index.ts
+++ b/packages/core/src/services/evaluationsV2/composite/index.ts
@@ -1,6 +1,7 @@
 import { z } from 'zod'
 import { database } from '../../../client'
 import {
+  CompositeEvaluationConfiguration,
   CompositeEvaluationMetric,
   CompositeEvaluationResultError,
   CompositeEvaluationResultMetadata,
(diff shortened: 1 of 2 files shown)


packages/core/src/services/evaluationsV2/composite/index.ts
```diff
@@ -40,64 +41,78 @@ export const CompositeEvaluationSpecification = {
...
+          {
+            code: 'custom',
+            path: ['evaluationUuids'],
+            message: `Sub-evaluation ${subUuid} not found`,
+          },
+        ]),
       )
     }
 
-    // Note: to simplify things, only batch evaluation mode is supported
```
neoxelox, line 85: can you not delete the notes please

packages/core/src/services/evaluationsV2/composite/index.ts
```diff
@@ -40,64 +41,78 @@ export const CompositeEvaluationSpecification = {
...
+          {
+            code: 'custom',
+            path: ['evaluationUuids'],
+            message: `Sub-evaluation ${evaluation.name} does not support batch evaluation`,
+          },
+        ]),
       )
     }
 
-    // Note: to simplify things, currently sub-evaluations cannot require an expected output
```
neoxelox, line 95: can you not delete the notes please

andresgutgon: Done

packages/core/src/services/evaluationsV2/composite/index.ts
andresgutgon: Done

PR #2092, opened 2026-01-07
New delete orpahned eval results maintenance job

learningbizz:
After doing the backfill of the evaluation results with their `type` and `metric`, we still have around 26K eval results with null type and metric.

This is due to the fact their evaluation doesnt exist in the database. This can occur when a user creates an evaluation within a draft, uses it, and then deletes it before committing the version.

To be able to make the type and metric not null, we have to do something about these eval results:

As these eval results don't have an existing evaluation attached to them, they are worthless for the user, so we can delete them.

This PR adds a maintenance job to delete them safely, without clogging the database

### DIFF
--- a/apps/workers/src/workers/worker-definitions/maintenanceWorker.ts
+++ b/apps/workers/src/workers/worker-definitions/maintenanceWorker.ts
@@ -6,15 +6,12 @@ import { WORKER_CONNECTION_CONFIG } from '../utils/connectionConfig'
 
 const jobMappings = {
   autoScaleJob: jobs.autoScaleJob,
-  backfillEvaluationResultsTypeAndMetricJob:
-    jobs.backfillEvaluationResultsTypeAndMetricJob,
+  cleanupOrphanedEvaluationResultsJob: jobs.cleanupOrphanedEvaluationResultsJob,
   checkScheduledDocumentTriggersJob: jobs.checkScheduledDocumentTriggersJob,
   cleanDocumentSuggestionsJob: jobs.cleanDocumentSuggestionsJob,
   cleanupWorkspaceOldLogsJob: jobs.cleanupWorkspaceOldLogsJob,
   requestDocumentSuggestionsJob: jobs.requestDocumentSuggestionsJob,
   scaleDownMcpServerJob: jobs.scaleDownMcpServerJob,
-  scheduleBackfillEvaluationResultsTypeAndMetricJobs:
-    jobs.scheduleBackfillEvaluationResultsTypeAndMetricJobs,
   updateMcpServerLastUsedJob: jobs.updateMcpServerLastUsedJob,
   scheduleWeeklyEmailJobs: jobs.scheduleWeeklyEmailJobs,
   dailyAlignmentMetricUpdateJob: jobs.dailyAlignmentMetricUpdateJob,


packages/core/src/jobs/job-definitions/maintenance/cleanupOrphanedEvaluationResultsJob.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Job that deletes evaluation results that don't have a type or metric attached.
+ *
+ * This job uses batch processing to avoid locking the table for too long:
+ * 1. Selects IDs in batches of 1000
+ * 2. Deletes the selected IDs
+ * 3. Repeats until no more orphaned records are found
+ */
+export async function cleanupOrphanedEvaluationResultsJob(
```
learningbizz, line 16: There are a total of 26K aprox, so it will take 26 rounds which is manageable in one job.

packages/core/src/jobs/job-definitions/maintenance/cleanupOrphanedEvaluationResultsJob.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  do {
+    const idsToDelete = await database
+      .select({ id: evaluationResultsV2.id })
+      .from(evaluationResultsV2)
+      .where(
+        or(
+          isNull(evaluationResultsV2.type),
+          isNull(evaluationResultsV2.metric),
+        ),
```
andresgutgon, line 31: I guess new evaluation results are created with `type` and `metric` are created always with type and metric no? 
But is not enforced at DB level yet. 

I fear this will delete the new evaluation results without these two attributes. Probably not possible but just saying.

andresgutgon: A safety measure would be to filter by date and filter out results created after yesterday just in case. Maybe it's too much.

andresgutgon (dismissed):

packages/core/src/jobs/job-definitions/maintenance/cleanupOrphanedEvaluationResultsJob.ts
neoxelox: too much, the results are always created with type and metric  now`packages/core/src/services/evaluationsV2/results/create.ts`

packages/core/src/jobs/job-definitions/maintenance/cleanupOrphanedEvaluationResultsJob.ts
```diff
@@ -0,0 +1,57 @@
...
+      )
+      .limit(batchSize)
+
+    if (idsToDelete.length === 0) {
```
neoxelox, line 35: do, `Guard Clauses`, specifically an early break instead of this strange if else

apps/workers/src/workers/worker-definitions/eventsWorker.ts
```diff
@@ -55,6 +55,7 @@ const eventHandlersJobMappings = {
   notifyClientOfEvaluationV2AlignmentUpdated:
     jobs.notifyClientOfEvaluationV2AlignmentUpdated,
   notifyClientOfOptimizationStatus: jobs.notifyClientOfOptimizationStatus,
+  unassignIssuesOnDocumentsDeleted: jobs.unassignIssuesOnDocumentsDeleted,
```
andresgutgon, line 58: Bad rebase

learningbizz: todo gucci

learningbizz:
We will do it by query directly in the db

PR #2093, opened 2026-01-07
Simplify default composite evaluation management

andresgutgon:
# What?
Now documents always have one composite evaluation of type `average`. This way, we simplify the UI, and we don't need all the alerts and toast system to notify when this composite evaluation goes out of sync.

### TODO
- [x] Run migration to add `mainEvaluationUuid` in my dev db.
- [x] Test create eval with issue and see how composite is created
- [x] Test assign issue to eval and see how it's added to composite 
- [x] Test un-assign issue from eval and see how is removed from composite
- [x] Test remove eval with assigned issue and see how is removed from composite
- [x] Check that composite is removed after last eval with issue assigned is removed or updated.
- [x] Check that `defaultTarget` is not used in the evals table
- [x] Check that `defaultTarget` is not used in Composite Form issues selector
- [x] Double-check all the alerts and toast related with `sync-composite` are gone

## Default composite evaluation lifecycle 
1. A composite of type `average` is created by Latitude in the document
2. This evaluation composite is pointed to `document.mainEvaluationUuid`
3. This is used in optimizations when selecting what evaluation to pick. By default we select this one for the users to use.
4. If user add a new evaluation with an issue linked for that document we add that evaluation to the composite/average
5. If user remove the issue from one evaluation that was included in the composite evaluation we remove that eval.uuid from the configuration
6. If user removes the evaluation (and had issue linked) we remove it from the composite
7. If the evaluation removed is the last one we do 2 things: Remove the composite evaluation and set to `null` -> `mainEvaluationUuid` in the document

## From `defaultTarget` to `mainEvaluationUuid`
Before we had `defaultTarget` as part of the jsonb config of the composite evalution. Now we do a real link from this evaluation to the document. 
Also we don't allow users to chose what of the composite evaluations they have in the document is the one that is the default. The default is manage by us. The constrain here is that is always of type `average`. The other types like `weighted` demands manual intervention. We let users manage their own evaluations but this one is auto updated and for that we're removing all the `Alerts` and `Toast` we had to let users know the main evaluation was out of sync. Makes UI and understanding more complicated


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/optimizations/_components/OptimizationForm/EvaluationSelector.tsx
```diff
@@ -33,13 +33,11 @@ export function EvaluationSelector({
...
   const defaultEvaluation = useMemo(() => {
-    return evaluations.find(
-      (e) =>
-        e.type === EvaluationType.Composite &&
-        !!(e as EvaluationV2<EvaluationType.Composite>).configuration
-          .defaultTarget,
-    ) as EvaluationV2<EvaluationType.Composite> | undefined
-  }, [evaluations])
+    if (!document.mainEvaluationUuid) return undefined
+    return evaluations.find((e) => e.uuid === document.mainEvaluationUuid) as
```
Comment (neoxelox) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/optimizations/_components/OptimizationForm/EvaluationSelector.tsx]: is this `as` needed? out of curiosity because i think typescript should be able to infer the type of the array.find

packages/core/src/services/evaluationsV2/composite/index.ts
```diff
@@ -128,7 +128,11 @@ async function validate<M extends CompositeEvaluationMetric>(
     return Result.error(parsing.error)
   }
 
-  const validating = await validateSubEvaluations({ uuid, configuration, evaluations }) // prettier-ignore
+  const validating = await validateSubEvaluations({
```
neoxelox, line 131: xd

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/optimizations/_components/OptimizationForm/EvaluationSelector.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/optimizations/_components/OptimizationForm/EvaluationSelector.tsx]: No needed 🔥

packages/core/src/services/evaluationsV2/composite/index.ts
andresgutgon: I let prettier free always I see this
image

packages/core/src/services/evaluationsV2/sync.ts
```diff
@@ -0,0 +1,321 @@
...
+import { EvaluationsV2Repository } from '../../repositories'
+import { documentVersions } from '../../schema/models/documentVersions'
+import { type Commit } from '../../schema/models/types/Commit'
+import { type DocumentVersion } from '../../schema/models/types/DocumentVersion'
+import { type Workspace } from '../../schema/models/types/Workspace'
+import { createEvaluationV2 } from './create'
+import { deleteEvaluationV2 } from './delete'
+import { updateEvaluationV2 } from './update'
+
+type CompositeTarget = EvaluationV2<EvaluationType.Composite>
```
neoxelox:
```suggestion
type CompositeTarget = EvaluationV2<EvaluationType.Composite, CompositeEvaluationMetric.Average>
```

to ensure the dev knows they can update the subevaluations without worrying

packages/core/src/services/evaluationsV2/sync.ts
```diff
@@ -0,0 +1,321 @@
...
+          evaluationUuids: newEvaluationUuids,
+        },
+      },
+    },
+    transaction,
+  )
+
+  if (updating.error) return updating
+
+  return Result.ok(updating.value.evaluation as CompositeTarget)
```
neoxelox: you shouldn't need the `as`, maybe types have broken with the typed results?

packages/core/src/services/evaluationsV2/sync.ts
```diff
@@ -0,0 +1,321 @@
...
+    )
+
+    if (deleting.error) return Result.error(deleting.error)
+
+    await tx
+      .update(documentVersions)
+      .set({ mainEvaluationUuid: null })
+      .where(
+        and(
+          eq(documentVersions.commitId, commit.id),
```
neoxelox: nope, this is not going to work, you cannot update a specific document version by this ids, because it may have not exist!!! you have to use the updateDocument service!

packages/core/src/services/evaluationsV2/sync.ts
```diff
@@ -0,0 +1,321 @@
...
+    )
+
+    if (creating.error) return Result.error(creating.error)
+
+    const newComposite = creating.value.evaluation
+
+    await tx
+      .update(documentVersions)
+      .set({ mainEvaluationUuid: newComposite.uuid })
+      .where(
```
neoxelox: nope, this is not going to work, you cannot update a specific document version by this ids, because it may have not exist!!! you have to use the updateDocument service!

neoxelox (changes requested):
The document cannot be updated with a simple sql query!

packages/core/src/services/evaluationsV2/sync.ts
andresgutgon: Done

packages/core/src/services/evaluationsV2/sync.ts
andresgutgon: Done

packages/core/src/services/evaluationsV2/sync.ts
andresgutgon: done

PR #2096, opened 2026-01-08
Fix optimizing document on a new project or new document

andresgutgon:
# What?
When optimizing a document that does note exists on the live commit or the project is new and does not have a live commit, optimization fails when updating the document because it's not present in the optimized commit. The solution applied here is that before trying to update the document with the optimization is necessary to fork that commit to ensure it will be always be present

### DIFF
--- a/packages/core/src/services/commits/create.ts
+++ b/packages/core/src/services/commits/create.ts
@@ -7,12 +7,24 @@ import Transaction from '../../lib/Transaction'
 import { commits } from '../../schema/models/commits'
 import { pingProjectUpdate } from '../projects'
 import { CommitsRepository } from '../../repositories/commitsRepository'
+import { Database } from '../../client'
+
+async function findCommit(
+  { baseCommit, project }: { baseCommit?: Commit; project: Project },
+  db: Database,
+) {
+  if (baseCommit) return baseCommit
+
+  const commitsScope = new CommitsRepository(project.workspaceId, db)
+  return commitsScope.getHeadCommit(project.id)
+}
 
 export async function createCommit(
   {
     project,
     user,
     data: { title, description, mergedAt, version },
+    baseCommit,
   }: {
     project: Project
     user: User
@@ -22,13 +34,13 @@ export async function createCommit(
       version?: number
       mergedAt?: Date
     }
+    baseCommit?: Commit
   },
   transaction = new Transaction(),
 ) {
   return transaction.call<Commit>(
     async (tx) => {
-      const commitsScope = new CommitsRepository(project.workspaceId, tx)
-      const liveCommit = await commitsScope.getHeadCommit(project.id)
+      const fromCommit = await findCommit({ project, baseCommit }, tx)
 
       const [commit] = await tx
         .insert(commits)
@@ -39,7 +51,7 @@ export async function createCommit(
           description,
           version,
           mergedAt,
-          mainDocumentUuid: liveCommit?.mainDocumentUuid,
+          mainDocumentUuid: fromCommit?.mainDocumentUuid,
         })
         .returning()
 
--- a/packages/core/src/services/commits/fork.ts
+++ b/packages/core/src/services/commits/fork.ts
@@ -1,0 +1,105 @@
+import { omit } from 'lodash-es'
+import { type Commit } from '../../schema/models/types/Commit'
+import { type Project } from '../../schema/models/types/Project'
+import { type User } from '../../schema/models/types/User'
+import { type Workspace } from '../../schema/models/types/Workspace'
+import { Result } from '../../lib/Result'
+import Transaction from '../../lib/Transaction'
+import {
+  DocumentVersionsRepository,
+  EvaluationsV2Repository,
+} from '../../repositories'
+import { documentVersions } from '../../schema/models/documentVersions'
+import { evaluationVersions } from '../../schema/models/evaluationVersions'
+import { inheritDocumentRelations } from '../documents/inheritRelations'
+import { createCommit } from './create'
+
+/**
+ * Creates a new draft commit by forking from an existing commit.
+ * Copies only the changes (documents and evaluations directly in the source commit)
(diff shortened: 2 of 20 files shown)


packages/core/src/services/commits/create.ts
```diff
@@ -22,13 +35,16 @@ export async function createCommit(
       version?: number
       mergedAt?: Date
     }
+    baseCommit?: Commit
+    from?: 'create' | 'fork'
   },
   transaction = new Transaction(),
 ) {
+  const skipSideEffects = from === 'fork'
```
neoxelox: why? the commit is in fact created no?

andresgutgon: Yes it is but I though is not really a voluntary commit so maybe we don't want to trigger the event? Open for discussion

packages/core/src/services/commits/fork.ts
```diff
@@ -0,0 +1,79 @@
...
+      transaction,
+    ).then((r) => r.unwrap())
+
+    const docsRepo = new DocumentVersionsRepository(workspace.id, tx)
+    const documents = await docsRepo
+      .getDocumentsAtCommit(commit)
+      .then((r) => r.unwrap())
+
+    if (documents.length > 0) {
+      const insertedDocs = await tx
```
neoxelox, line 59: mmm im not sure if this will respect the deleted document versions...

andresgutgon: Good catch 💯

packages/core/src/services/commits/create.ts
neoxelox: i will just trigger the event, less code, less logic... actions were creating a commit is a side effect were already sending these events. Maybe you could just add a new property to the event called source ( "create" | "fork" )? in the case we want that info in analytics?

packages/core/src/services/optimizations/validate/start.ts
```diff
@@ -152,14 +152,14 @@ export async function startValidateOptimization(
 
     const baseline = await createExperiment(
       {
+        workspace,
```
neoxelox: broski

packages/core/src/services/commits/fork.ts
neoxelox: still not resolved no?

neoxelox: ahh you are using now another func ok

packages/core/src/services/commits/create.ts
geclos: no please services should not control side effects and boolean parameters [aren't the way to do it](https://www.notion.so/latitudedata/Aphorisms-0ed50a6dccdb4aa2a7dec86812b30cc8?source=copy_link#62adf3452dfc403e9b38a3fdf94079f9) anyway

packages/core/src/repositories/evaluationsV2Repository.ts
```diff
@@ -26,11 +27,28 @@ const tt = {
   versionId: sql<number>`${evaluationVersions.id}::integer`.as('versionId'),
 }
 
+type EvaluationsV2RepositoryOptions = {
```
neoxelox: these changes are not used and not needed i think

packages/core/src/services/commits/create.ts
andresgutgon: This is not a boolean argument. This is a variable

packages/core/src/services/commits/fork.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    const evalsRepo = new EvaluationsV2Repository(workspace.id, tx)
+    const evaluations = await evalsRepo
+      .getChangesInCommit(commit)
+      .then((r) => r.unwrap())
+
+    if (evaluations.length > 0) {
+      await tx.insert(evaluationVersions).values(
+        evaluations.map((e) => ({
```
neoxelox: instead of setting every field, why dont do something similar on what you did for the document versions? the omit thingie

packages/core/src/services/commits/create.ts
andresgutgon: The other thing is how we should handle side effects. In this case is the internal logic of the service that decides in relation to the `from` argument if it should trigger the event or not. 

Interested on how would you do this. 

In this particular case I think I'm overthinking. I'll remove the `from` argument and just do the side effects always

geclos: you implement another service that does the minimum set of required work you need without the side effects and use that, and also use it in the commits/create service.

andresgutgon: Ok removed the side effect here in any case

packages/core/src/services/commits/fork.ts
```diff
@@ -0,0 +1,107 @@
...
+      .getChangesInCommit(commit)
+      .then((r) => r.unwrap())
+
+    if (evaluations.length > 0) {
+      const evalValues = evaluations.map((e) => ({
+        ...omit(e, [
+          'id',
+          'versionId',
+          'uuid',
+          'evaluationUuid',
```
neoxelox: this is not needed no? the eval uuid will be always the same in every version, no need to omit it and putting it later no?

neoxelox (dismissed):

packages/core/src/services/commits/fork.ts
andresgutgon: When we use the repository we have this mapping
https://github.com/latitude-dev/latitude-llm/blob/main/packages/core/src/repositories/evaluationsV2Repository.ts#L25-L26

So here we're doing a bulk create after the map. So i need to map to the real names of those columns

`versionId` -> in DB is `id`
`evaluationUuid` -> in DB is `uuid`

neoxelox: evaluationUuid is evaluationUuid in db, and uuid is evaluationUuid

andresgutgon: Removed `evaluationUuid` from omit 👍

PR #2099, opened 2026-01-11
Add workspace RBAC for annotations

geclos:
## Summary
- Introduce a workspace RBAC model (roles + permissions) with a new membership `role` column and helpers to assert/resolve permissions; seed initial roles Admin/Annotator and map granular permissions (manage workspace, manage members, access/write annotations).
- Propagate membership/permissions through session, auth procedures, middleware, and API handlers; enforce server-side access for annotations-related APIs/actions and member management, while keeping admin-only scope for everything else.
- Update settings UI to assign/update roles (invite form role select, role edit modal, users store wiring) and redirect annotators to the annotations area, hiding non-annotation navigation/tabs for them.

## Testing
- Not run (not requested)

### DIFF
--- a/packages/core/src/repositories/usersRepository.ts
+++ b/packages/core/src/repositories/usersRepository.ts
@@ -17,7 +17,7 @@ export type WorkspaceUser = User & {
 
 const tt = {
   ...getTableColumns(users),
-  membershipId: memberships.id,
+  membershipId: sql<number>`${memberships.id}`.mapWith(Number).as('membershipId'),
   role: memberships.role,
   confirmedAt: memberships.confirmedAt,
 }


geclos:
@claude review the code, fix tests and failing typescript checks

packages/core/src/services/memberships/create.ts
```diff
@@ -22,7 +22,12 @@ export const createMembership = async (
     async (tx) => {
       const result = await tx
         .insert(memberships)
-        .values({ userId: user.id, workspaceId: workspace.id, ...rest })
+        .values({
+          userId: user.id,
+          workspaceId: workspace.id,
+          role: rest.role ?? 'admin',
```
andresgutgon, line 28: By default are admins the users created?

packages/core/src/services/users/invite.ts
```diff
@@ -36,9 +39,10 @@ export async function inviteUser(
         user = result.unwrap()
       }
 
-      await createMembership({ author, user, workspace }, transaction).then(
-        (r) => r.unwrap(),
-      )
+      await createMembership(
+        { author, user, workspace, role: role ?? 'admin' },
```
andresgutgon, line 43: This default could be just inside the service

packages/core/src/services/memberships/create.ts
geclos: yeah

packages/core/src/services/users/invite.ts
geclos: @claude I agree ☝🏼 , do it

PR #2105, opened 2026-01-12
Refactor chat runs to use disk-backed conversation cache

geclos:
Replace provider/document log dependency during runs with a disk-backed conversation cache. We cache all convos in disk storage with a TTL of 30 minutes. When a user tries to chat with an existing convo, we first read from the cache, and otherwise fallback to reading from the stored span. This setup assumes we process spans in less than 30 minutes (should be the case!). Convos are stored in buckets of 5 minutes so to reduce the burden of the cleanup operation.

### DIFF
--- a/apps/gateway/src/common/createRequestAbortSignal.test.ts
+++ b/apps/gateway/src/common/createRequestAbortSignal.test.ts
@@ -13,7 +13,9 @@ function createMockSocket() {
   }
 }
 
-function createMockContext(socket?: ReturnType<typeof createMockSocket>): Context {
+function createMockContext(
+  socket?: ReturnType<typeof createMockSocket>,
+): Context {
   const env = socket
     ? {
         incoming: { socket },
--- a/apps/web/src/hooks/playgroundChat/usePlaygroundChat.ts
+++ b/apps/web/src/hooks/playgroundChat/usePlaygroundChat.ts
@@ -347,7 +347,9 @@ export function usePlaygroundChat({
         return {
           ...msg,
           content: msg.content.map((c) =>
-            'isStreaming' in c && c.isStreaming ? { ...c, isStreaming: false } : c,
+            'isStreaming' in c && c.isStreaming
+              ? { ...c, isStreaming: false }
+              : c,
           ),
         } as Message
       }),
@@ -514,7 +516,14 @@ export function usePlaygroundChat({
     } else {
       await stopRun({ runUuid: documentLogUuid })
     }
-  }, [isLoading, documentLogUuid, stopRun, isStoppingRun, abortCurrentStream, cleanupStreamingState])
+  }, [
+    isLoading,
+    documentLogUuid,
+    stopRun,
+    isStoppingRun,
+    abortCurrentStream,
+    cleanupStreamingState,
+  ])
 
   const reset = useCallback(() => {
     setMode('preview')
--- a/packages/core/src/lib/streamManager/step/recordAbortedCompletion.test.ts
+++ b/packages/core/src/lib/streamManager/step/recordAbortedCompletion.test.ts
@@ -122,7 +122,8 @@ describe('recordAbortedCompletion', () => {
     const provider = createMockProvider(Providers.OpenAI)
     const config = createMockConfig('gpt-4o')
     const messages = createMockMessages('Write a poem')
-    const accumulatedText = 'Roses are red,\nViolets are blue,\nThis poem was cut'
+    const accumulatedText =
+      'Roses are red,\nViolets are blue,\nThis poem was cut'
 
     recordAbortedCompletion({
       context,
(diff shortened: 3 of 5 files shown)


packages/core/src/lib/streamManager/index.ts
```diff
@@ -323,7 +339,7 @@ export abstract class StreamManager {
 
     this.sendEvent({
       type: ChainEventTypes.ProviderCompleted,
-      providerLogUuid: response.providerLog!.uuid,
+      providerLogUuid: response.documentLogUuid ?? this.uuid,
```
geclos: this is wrong it was ok before

packages/core/src/services/commits/foregroundRun.ts
```diff
@@ -104,10 +106,22 @@ export async function runForegroundDocument(
...
+
+      const providersMap = await buildProvidersMap({
+        workspaceId: workspace.id,
+      })
+      const provider = providersMap.get(config.provider)
+      if (!provider) {
+        throw new LatitudeError(
+          `Provider API key not found for ${config.provider}`,
+        )
+      }
```
geclos, line 109: rather than recomputing the provider here ensure the stream manager returns it

packages/core/src/services/evaluationsV2/llm/shared.ts
```diff
@@ -240,18 +243,34 @@ export async function runPrompt<
...
+          }) * TO_MILLICENTS_FACTOR,
+        )
+      : 0
+
+  const stats = {
+    documentLogUuid: resultUuid,
+    tokens: totalTokens,
+    duration: duration ?? 0,
+    costInMillicents,
+  }
```
geclos, line 281: are these stats used anywhere? if not, we can remove them

packages/core/src/jobs/job-definitions/maintenance/clearConversationCacheJob.ts
```diff
@@ -0,0 +1,15 @@
...
+export type ClearConversationCacheJobData = Record<string, never>
+
+/**
+ * Clears the cached conversations stored on disk.
+ */
+export async function clearConversationCacheJob(
+  _: Job<ClearConversationCacheJobData>,
+) {
+  const disk = diskFactory('private')
+  return disk.deleteAll(CONVERSATION_CACHE_PREFIX)
```
geclos: this clears the whole conversation cache but it should clear only value outside the ttl of one hour

packages/constants/src/events/events.ts
```diff
@@ -54,7 +54,6 @@ export interface LatitudeProviderStartedEventData
 export interface LatitudeProviderCompletedEventData
   extends GenericLatitudeEventData {
   type: ChainEventTypes.ProviderCompleted
-  providerLogUuid: string
```
geclos, line 57: i'm removing this prop from the events, see if anyone complains

geclos:
@claude review

packages/core/src/jobs/job-definitions/maintenance/clearConversationCacheJob.ts
```diff
@@ -0,0 +1,74 @@
...
+        await cacheClient.zrem(
+          CONVERSATION_CACHE_BUCKET_INDEX_KEY,
+          ...deletedBuckets,
+        )
+      }
+
+      const errorResult = deleteResults.find(({ result }) => result.error)
+      if (errorResult?.result.error) {
+        return Result.error(errorResult.result.error)
+      }
```
andresgutgon, line 68: No one is getting this error no? Why return?

packages/core/src/lib/streamManager/step/streamAIResponse.test.ts
```diff
@@ -155,40 +154,6 @@ describe('streamAIResponse', () => {
...
-          source: LogSources.API,
-          costInMillicents: 0,
-          documentLogUuid: null,
-          finishReason: 'stop',
-          output: null,
-          generatedAt: new Date(),
-          createdAt: new Date(),
-          updatedAt: new Date(),
-        } as any),
-      )
```
andresgutgon, line 303: Should be here a replacement with the new way?

packages/core/src/services/documentLogs/addMessages/index.test.ts
```diff
@@ -115,7 +114,7 @@ Hello world
 
     expect(mockChat).toHaveBeenCalledWith(
       expect.objectContaining({
-        documentLogUuid: documentLog.uuid,
+        documentLogUuid,
```
andresgutgon, line 119: Tests for the new way no?

packages/core/src/services/conversations/cache.ts
```diff
@@ -0,0 +1,242 @@
...
+}
+
+async function removeConversationCacheLookupKey(documentLogUuid: string) {
+  try {
+    const cacheClient = await cache()
+    await cacheClient.del(buildConversationCacheLookupKey(documentLogUuid))
+  } catch {
+    // do nothing
+  }
+}
```
andresgutgon, line 242: No tests?

packages/core/src/jobs/job-definitions/maintenance/clearConversationCacheJob.ts
geclos: pure habit ngl, but yeah i can change it

PR #2106, opened 2026-01-12
Fix weekly email spans query by adding missing index

andresgutgon:
# What?
I did some changes in this query here:
https://github.com/latitude-dev/latitude-llm/pull/2081, but was not enough and the query failed. I think the next thing to try is to add the missing index and see if this improve the situation. If this does not fix the issue we'll need to think harder or re-do weekly email and see how we include spans information


packages/core/src/data-access/weeklyEmail/logs/index.ts
andresgutgon, line 1: I don't know how but this was working with `.tsx` 🤯

PR #2108, opened 2026-01-12
Fix resize header in datagrid (datasets table)

andresgutgon:
# What?
I updated react-data-grid dependency and put under catalog because is used in `packages/web-ui` and `apps/web`. But this was not the issue. The issue is that generated CSS was not generating logical properties which are used by Data grid package. In our case we are only LTR (Left To Right) so resize handle on data grid header is always on the right side.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -19,6 +19,14 @@
     "datadog:sourcemaps:server": "datadog-ci sourcemaps upload .next/server --service=latitude-llm-web --minified-path-prefix=/app/apps/web/.next/server/ --release-version=${RELEASE_VERSION:-unknown} --disable-git --max-concurrency=20",
     "datadog:sourcemaps": "pnpm datadog:sourcemaps:client && pnpm datadog:sourcemaps:server"
   },
+  "browserslist": [
+    "Chrome >= 89",
+    "Edge >= 89",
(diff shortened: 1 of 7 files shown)


apps/web/package.json
```diff
@@ -19,6 +19,14 @@
...
     "datadog:sourcemaps": "pnpm datadog:sourcemaps:client && pnpm datadog:sourcemaps:server"
   },
+  "browserslist": [
+    "Chrome >= 89",
+    "Edge >= 89",
+    "Firefox >= 66",
+    "Safari >= 15",
+    "iOS >= 15",
+    "Opera >= 76"
+  ],
```
andresgutgon, line 29: Bug with resize header handler in data grid was that the underlying lib we use, `react-data-grid` was using CSS logical properties. I tried first to update the lib to the latest version, but then I realized it was our Next.js config not generating css logic properties because we're supporting ridiculously old browsers. I changed to these versions, which are old enough, and now this CSS is generated

```css
inset-block-start: 0;
inset-inline-end: 0;
inset-block-end: 0;
inline-size: 10px
```
Which is equivalent to:
```css
top: 0;
bottom: 0;
right: 0;
width: 10px;
```
On a LTR (Left To Right) language.

PR #2113, opened 2026-01-13
GEPA optimizer

neoxelox:


apps/engine/app/core/gepa.py
```diff
@@ -0,0 +1,212 @@
...
+            ],
+            GepaProposeResult,
+            batch_size=10,
+        )
+
+        proposed: System = {}
+        for component, result in zip(components_to_update, results, strict=True):
+            proposed[component] = result.prompt
+
+        return proposed
```
geclos, line 213: so this can run for up to 2 hours no? there isn't any kind of checkpointing / state save to recover from in case the job fails / stops for whatever reason no? if we fail after 1 hour of work, we just lost the whole work?

apps/engine/app/rpc/server.py
```diff
@@ -0,0 +1,361 @@
...
+            batch_size: Optional max number of concurrent calls. If None, all calls
+                        are made concurrently. If set, calls are made in chunks of
+                        this size, waiting for each chunk to complete before starting
+                        the next.
+
+        Returns:
+            List of validated response models in the same order as params_list
+
+        Raises:
+            RuntimeError: If any call fails, the first error is raised
```
geclos, line 169: so we consider the whole batch failed if just one request fails? can't we be less strict?

apps/engine/app/main.py
```diff
@@ -0,0 +1,25 @@
+import asyncio
+import sys
+
+from app.core.gepa import register_gepa_handlers
+from app.rpc.server import RpcServer
+
+
+async def main() -> None:
+    """Main entry point for the engine RPC server."""
```
geclos, line 15: why did you decide to run this server in the workers container rather than being independent, do we know more or less how much cpu/memory is this gonna take?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/optimizations/_components/OptimizationForm/BudgetSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const timeValue = value?.time ?? 0
+  const tokensValue = value?.tokens ?? 0
+
+  const timeStep = useMemo(() => {
+    return 5 * 60 // 5 minutes
+  }, [])
+
+  const tokensStep = useMemo(() => {
+    return 1_000_000 // 1M tokens
+  }, [])
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/optimizations/_components/OptimizationForm/BudgetSelector.tsx]: these can justbe constants

apps/workers/docker/Dockerfile
```diff
@@ -121,6 +139,9 @@ COPY --from=builder --chown=latitude:nodejs /app/${PROJECT_PATH} ./${PROJECT_PAT
 COPY --from=builder --chown=latitude:nodejs /app/packages/telemetry/typescript ./packages/telemetry/typescript
 COPY --from=builder --chown=latitude:nodejs /app/packages/core/src/assets/eu-central-1-bundle.pem ./packages/core/src/assets/eu-central-1-bundle.pem
 
+# Copy Python engine with virtual environment
+COPY --from=python-builder --chown=latitude:nodejs /app/apps/engine ./apps/engine
+
 WORKDIR /app/${PROJECT_PATH}
 
 CMD ["node", "dist/server.js"]
```
geclos, line 141: did we test this new image?

packages/core/src/services/engine/client.ts
```diff
@@ -0,0 +1,342 @@
...
+
+    const validatedParams = paramsSchema.parse(params)
+    const request = createRequest(method, validatedParams, null)
+    this.write(serializeMessage(request))
+  }
+
+  /**
+   * Kill the engine process immediately.
+   */
+  stop(): void {
```
geclos: call it kill then rather than stop, this ain't graceful

packages/core/src/services/engine/client.ts
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    this.readline?.close()
+    this.readline = null
+
+    this.rejectAllPending(new Error('Engine process stopped'))
+
+    this.process = null
+    this.isRunning = false
+  }
```
geclos, line 352: the cleanup is not stopping the py server?

packages/core/src/services/optimizations/optimizers/gepa.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    const prompt = result.optimized[document.path] ?? ''
+
+    return Result.ok(prompt)
+  } catch (error) {
+    return Result.error(error as Error)
+  } finally {
+    engine.stop()
+  }
+}
```
geclos, line 187: nitpick but "engine" is very generic to refer to something that is very specific (a python rpc server in charge of running gepa optimizers), could have been a better name for easier grepping in the future

packages/core/src/services/optimizations/optimizers/gepa.ts
```diff
@@ -0,0 +1,187 @@
...
+      }).then((r) => r.unwrap())
+
+      return {
+        prompt: prompt,
+      }
+    },
+  )
+
+  try {
+    await engine.start()
```
geclos, line 162: oh ok so we start a new engine on each optimization and close it immediately after it's done? is the engine shared between optimization requests if say 2 workspaces request an optimization?

packages/core/src/services/optimizations/optimizers/propose.ts
```diff
@@ -0,0 +1,110 @@
...
+function sanitizeTrajectory(trajectory: Trajectory) {
+  return {
+    ...omit(trajectory, 'id', '_tjr'),
+    trace: sanitizeTrace(trajectory.trace),
+    usage: omit(trajectory.usage, 'inputTokens', 'outputTokens'),
+  }
+}
+
+// BONUS(AO/OPT): Implement multi-document optimization
+export async function proposeFn({
```
geclos: why the fn suffix

packages/core/src/services/optimizations/optimizers/propose.ts
```diff
@@ -0,0 +1,110 @@
...
+}) {
+  if (!env.COPILOT_PROMPT_OPTIMIZATION_PROPOSER_PATH) {
+    throw new Error('COPILOT_PROMPT_OPTIMIZATION_PROPOSER_PATH is not set')
+  }
+
+  const copilot = await getCopilot({
+    path: env.COPILOT_PROMPT_OPTIMIZATION_PROPOSER_PATH,
+  }).then((r) => r.unwrap())
+
+  return async function (
```
geclos, line 64: oh ok because it returns a function, yeah better call it factory

packages/core/src/services/optimizations/optimizers/shared.ts
```diff
@@ -0,0 +1,53 @@
...
+function normalizeScore(score: number, lower: number, upper: number) {
+  if (lower === upper) return score === lower ? OPTIMIZATION_SCORE_SCALE : 0
+  else if (lower < upper) score = Math.min(Math.max(score, lower), upper)
+  else score = Math.min(Math.max(score, upper), lower)
+  const range = Math.abs(upper - lower)
+  const value = Math.abs(score - lower)
+  const map = (value * OPTIMIZATION_SCORE_SCALE) / range
+  const fixed = Number(map.toFixed(16))
+  return Math.min(Math.max(fixed, 0), OPTIMIZATION_SCORE_SCALE)
+}
```
geclos: don't we have similar normalization implementations for evaluation results somewhere?

packages/core/src/services/optimizations/execute.ts
```diff
@@ -131,11 +137,27 @@ export async function executeOptimization(
     )
   }
 
+  // BONUS(AO/OPT): Implement checkpointing saving for fault tolerance
   const optimizing = await optimize({
-    evaluate: evaluatePrompt,
+    evaluate: await evaluateFn({
```
geclos: same could have called it factory but it's a nitpick can ignore

packages/core/src/services/optimizations/shared.ts
```diff
@@ -1,4 +1,8 @@
-import { OptimizationConfiguration } from '../../constants'
+import { OptimizationConfiguration, Span, SpanType } from '../../constants'
+
+export function hashSpan(span: Span<SpanType.Prompt>) {
```
geclos: is this shared? it looks like it's only used in one place

.env.example
```diff
@@ -156,3 +156,10 @@ NEXT_SERVER_ACTIONS_ENCRYPTION_KEY=
 # configure its API key. If you don't configure this correctly the onboarding
 # will not work (but you can skip it).
 DEFAULT_PROVIDER_API_KEY=openai-api-key
+
+
+# Engine
+# Our Python engine powers advanced LLM optimizations and distillations
+# It's optional and can be ignored if you don't want to use it
+ENGINE_PATH=/app/apps/engine
+ENGINE_PYTHON=python3 # It is recommended to use the Python from the virtual environment
```
geclos, line 165: again engine being very generic should have been called gepa_engine or something of the sort to understand what it is

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/optimizations/_components/OptimizationForm/BudgetSelector.tsx
Comment (neoxelox) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/optimizations/_components/OptimizationForm/BudgetSelector.tsx]: (vibe coded) donette

apps/engine/app/core/gepa.py
neoxelox: No, I left it as a bonus in the code (and in linear). However, I'm not sure if it will be easy to do this because the gepa state cannot be loaded tho.

apps/engine/app/main.py
neoxelox: 0% cpu (its 99% IO) and 20 megs for each process in my laptop. I decided this way because it was the simplest thing to do. The algorithm does nothing special, so no sense to spend too much time creating a whole new service with the infra required. However, this can be improved in the near future if needed.

apps/engine/app/rpc/server.py
neoxelox: This errors are hard failures, errors from the eval or schema are learnable. Maybe we could filter out hard failures and continue trying but im not sure how would the gepa algorithm take that because it expects outputs for all batched examples.

packages/core/src/services/engine/client.ts
neoxelox: done

packages/core/src/services/optimizations/optimizers/gepa.ts
neoxelox: its an engine per task

packages/core/src/services/optimizations/optimizers/gepa.ts
neoxelox: mmm for me it was something more like "Our AI Engine", because it the future it could be expanded to other optimizers and distillers

packages/core/src/services/optimizations/optimizers/propose.ts
neoxelox: because it returns a function idk

packages/core/src/services/optimizations/optimizers/propose.ts
neoxelox: ok

packages/core/src/services/optimizations/optimizers/shared.ts
neoxelox: yeah, also this is overkill we are already using the normalized score, i just need to / 100 it to conver it to 1 scale xd

packages/core/src/services/optimizations/shared.ts
neoxelox: ok moved

packages/core/src/services/engine/client.ts
neoxelox: changed to graceful then kill after 5 secs

packages/core/src/services/optimizations/execute.ts
neoxelox: changed

.env.example
neoxelox: mmm for me it was something more like "Our AI Engine", because it the future it could be expanded to other optimizers and distillers

apps/workers/docker/Dockerfile
neoxelox: yess with the new changes it works gucci

PR #2115, opened 2026-01-14
feat: Add workspace switching via URL parameter

geclos:
## Summary
- Allow users to switch workspaces by providing a `workspace_id` query parameter in the URL
- When a user accesses a URL with a different workspace_id than their current session, the system validates access and automatically switches to that workspace
- Include `workspaceId` in issue escalation email links to ensure users are directed to the correct workspace when clicking through

## Test plan
- [x] Test accessing a URL with `?workspace_id=X` where X is a workspace the user has access to
- [x] Verify the session switches to the requested workspace
- [x] Test accessing a URL with a workspace_id the user doesn't have access to (should stay on current workspace)
- [x] Verify issue escalation emails contain the correct workspace_id in the link

### DIFF
--- a/apps/web/src/services/auth/getCurrentUser.ts
+++ b/apps/web/src/services/auth/getCurrentUser.ts
@@ -75,24 +75,20 @@ export const getCurrentUserOrRedirect = cache(async () => {
       const requestedWorkspace = requestedWorkspaceResult.unwrap()
 
       try {
-        await replaceSession({
-          oldSession: sessionData.session,
+        const newSession = await replaceSession({
           sessionData: { user, workspace: requestedWorkspace },
         })
-      } catch {
-        // do nothing
-      }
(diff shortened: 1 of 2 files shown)


andresgutgon (dismissed):

packages/emails/src/templates/issues/IssueEscalatingMail.tsx
```diff
@@ -1,5 +1,3 @@
-import React from 'react'
```
andresgutgon, line 1: This will 💥

andresgutgon: This package the way is configured requires `react` to be imported. I didn't took a look on why is this necessary but without it it fail. Run the `pnpm dev` server on this package to check it

PR #2116, opened 2026-01-14
Remove MCP Hosted code

andresgutgon:
# What?
Checking things on subscriptions, I found all this dead code. All related with 💀 MCP Hosted servers


packages/core/src/services/k8s/index.ts
andresgutgon, line 1: I think K8s was only used for HostedMCPs no @geclos ?

geclos: yup

packages/core/src/services/integrations/McpClient/McpClientManager.ts
```diff
@@ -24,9 +23,9 @@ export interface McpClientManager {
 // Public Functions
 export async function getMcpClient(
   integration: IntegrationDto,
-  streamManager?: StreamManager,
+  _streamManager?: StreamManager,
```
geclos: remove it if it's not used no?

andresgutgon: Yes

PR #2117, opened 2026-01-14
Add new onboarding questions to loops and intercom

learningbizz:
Few things
- Adding the new info of the onboarding to loops
- Removing a case for one of the questions after talking offline with Cesar

This was tested locally doing an E2E using Loops

### DIFF
--- a/packages/core/src/services/users/update.ts
+++ b/packages/core/src/services/users/update.ts
@@ -6,8 +6,13 @@ import Transaction from '../../lib/Transaction'
 import { users } from '../../schema/models/users'
 import { publisher } from '../../events/publisher'
 
-function isFinalOnboardingStep(values: Partial<User>): boolean {
-  return 'latitudeGoal' in values || 'latitudeGoalOther' in values
+const FINAL_ONBOARDING_FIELDS: (keyof User)[] = [
+  'latitudeGoal',
+  'latitudeGoalOther',
+]
+
+export function isFinalOnboardingStep(values: Partial<User>): boolean {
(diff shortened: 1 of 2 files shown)


packages/core/src/services/users/update.ts
```diff
@@ -4,18 +4,37 @@ import { type User } from '../../schema/models/types/User'
...
+        .set(values)
+        .where(eq(users.id, user.id))
+        .returning()
 
-    return Result.ok(updates[0]!)
-  })
+      return Result.ok(updates[0]!)
+    },
+    (updatedUser) => {
+      if (isFinalOnboardingStep(values)) {
```
andresgutgon, line 34: Maybe tests

packages/core/src/services/users/update.ts
```diff
@@ -4,18 +4,37 @@ import { type User } from '../../schema/models/types/User'
 import { Result } from '../../lib/Result'
 import Transaction from '../../lib/Transaction'
 import { users } from '../../schema/models/users'
+import { publisher } from '../../events/publisher'
+
+function isFinalOnboardingStep(values: Partial<User>): boolean {
+  return 'latitudeGoal' in values || 'latitudeGoalOther' in values
```
andresgutgon: Use existing enum or constants

andresgutgon (approved):

packages/core/src/services/users/update.ts
learningbizz: true good idea

PR #2118, opened 2026-01-14
Implement workspace deletion button in backoffice

geclos:


packages/core/src/services/workspaces/destroy.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    await tx
+      .delete(integrations)
+      .where(eq(integrations.workspaceId, workspaceId))
+
+    await tx.delete(workspaces).where(eq(workspaces.id, workspaceId))
+
+    await tx
+      .delete(subscriptions)
+      .where(eq(subscriptions.workspaceId, workspaceId))
```
andresgutgon, line 52: What happen with Stripe?

geclos: @claude cancel also the associated subscription in stripe

geclos: hmm so we don't store the subscription id... i don't like the implementation that claude did could lead to issues, let's keep stripe clean up manually.

geclos: @claude i've changed my mind do not remove the stripe subscriptions upon deleting the workspace, someone will do it manually each time.

andresgutgon: What issues? Is deleting the Stripe subscriptions not the correct way? How are we letting the person removing the workspace know they have to deal with Stripe?

geclos: we do not store the subcription idsin the subscription model (this has to change) which means to delete the subscriptions we need to look for and delete all subscriptions associated to the user which can potentially be too much deleting. I prefer we do it manually until we associate 1:1 strip subscription ids and our subscription table

andresgutgon: Oh I see. Makes sense.

PR #2130, opened 2026-01-18
Investigate latitude-dev error in latitude-llm repository

geclos:
Two changes to fix the "Cannot evaluate a log that is already evaluated" error:

1. In evaluateLiveLogJob, use span.source (from database) instead of spanMetadata.source (from disk storage). The span object already has the source field which is more reliable and immediately available. This ensures experiment spans (source=experiment) are correctly filtered out from live evaluation.

2. Add deduplication to runEvaluationForExperimentJob using the same runEvaluationV2JobKey as other evaluation job enqueuers. This prevents duplicate evaluation jobs even if somehow the same evaluation is requested multiple times.


geclos (commented):
Ok

PR #2139, opened 2026-01-19
MCP Integration: Custom Headers UI

csansoon:
Since #2131 , you can run prompts from SDK adding custom headers to MCP Integrations being used by the agent.

This PR focuses on bringing the same feature to the Playground UI. Now, you can configure and store custom Headers to any MCP Integration when running a prompt, which are also stored locally in the browser.

https://github.com/user-attachments/assets/56ea6865-641b-49f9-b9ab-585e081b3c4c


csansoon:
> Does it make sense to merge this if we're gonna store headers in settings anyway?

Yes, this is part of that

PR #2141, opened 2026-01-19
feat: copilot enterprise

geclos:
In enterprise mode we create a reference workspace with copilot prompts. We reference this workspace by name, as well as the copilot project. 

Took the ocasion to centralize how we run copilot prompts.


geclos:
@claude review

PR #2144, opened 2026-01-20
Fix A/B tests using always base system prompt

andresgutgon:
We found that although the A/B Test is configured correctly and conversations are being split as expected, the system prompt used at runtime is always the Live prompt (A).

Specifically:

Traces and data show conversations correctly labeled with prompt A or prompt B identifiers.

Traffic appears to be distributed correctly between variants.

However, when inspecting the actual messages, the system prompt is always the one from prompt A, even for conversations labeled as prompt B.

We are using A/B Tests, not Shadow Tests.

## TODO
- [x] Reproduce bug locally 
- [x] Check this PR fix the issue

### DIFF
--- a/packages/core/src/services/deploymentTests/resolveAbTestRouting.ts
+++ b/packages/core/src/services/deploymentTests/resolveAbTestRouting.ts
@@ -2,14 +2,17 @@ import { LogSources } from '@latitude-data/constants'
 import {
   CommitsRepository,
   DeploymentTestsRepository,
+  DocumentVersionsRepository,
 } from '../../repositories'
 import { routeRequest } from './routeRequest'
 import { Commit } from '../../schema/models/types/Commit'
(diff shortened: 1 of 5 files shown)


packages/core/src/services/deploymentTests/resolveAbTestRouting.ts
```diff
@@ -104,9 +104,18 @@ export async function resolveAbTestRouting({
     .getCommitById(commitIdToUse)
     .then((r) => r.unwrap())
 
+  const docsRepo = new DocumentVersionsRepository(workspaceId)
+  const effectiveDocument = await docsRepo
```
andresgutgon, line 108: The fix was to find the effective document. (The challenger) when running the prompt so we pick the right system prompt. 

Background runs were correct "by accident" because the job always re-fetches the document based on the commit UUID. Foreground runs passed
the document object directly, so when the original (wrong) document was passed, it used that stale version.

PR #2162, opened 2026-01-22
fix(issues): exclude issues with linked evaluations from merge candidates

geclos:
## Summary
- Issues that have an active evaluation linked to them are now excluded from merge candidates
- This prevents unexpected behavior where merging could affect evaluations the user has set up
- Winner/anchor issue can still have linked evaluations (other issues merge into it)

## Test plan
- [x] Added test: does not merge issues when candidate has a linked evaluation
- [x] Added test: allows merge when the winner issue has a linked evaluation  
- [x] Added test: allows merge when linked evaluation is deleted
- [x] Added test: allows merge when linked evaluation is ignored
- [x] All existing merge tests pass

🤖 Generated with [Claude Code](https://claude.ai/code)

---

> [!NOTE]
> Prevents merging issues that have active linked evaluations to avoid disrupting evaluation setups.
> 
> - Filters `mergedIssues` in `mergeIssues` by `getIssuesWithLinkedEvaluations` (checks `evaluationVersions` where `deletedAt` and `ignoredAt` are null); early-returns if none remain
> - Adjusts locking and subsequent merge steps to operate on filtered set only
> - Adds tests covering: blocking merge when any candidate has a linked evaluation; allowing merge when the winner has one; allowing merge when the linked evaluation is deleted or ignored
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 49d4d1ac07c3e94614c4f726615d3b78819476f4. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/core/src/services/issues/merge.ts
+++ b/packages/core/src/services/issues/merge.ts
@@ -80,11 +80,13 @@ export async function mergeIssues(
     workspaceId: workspace.id,
     issueIds: approvedSimilar.map(({ issue }) => issue.id),
   })
-  const { winner, mergedIssues } = pickWinner({
+  const res = pickWinner({
     anchorIssue: issue,
     candidates: approvedSimilar.map(({ issue }) => issue),
     totals,
   })
+  const winner = res.winner
+  let mergedIssues = res.mergedIssues
 
   if (mergedIssues.length === 0) {
     return Result.ok<MergeResult>({ winner, mergedIssues: [] })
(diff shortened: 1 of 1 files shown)


packages/core/src/services/issues/merge.ts
```diff
@@ -283,6 +295,39 @@ function pickWinner({
...
+  issueIds,
+}: {
+  workspaceId: number
+  issueIds: number[]
+}): Promise<Set<number>> {
+  if (issueIds.length === 0) return new Set()
+
+  const linkedEvaluations = await database
+    .select({ issueId: evaluationVersions.issueId })
+    .from(evaluationVersions)
```
andresgutgon: Is this picking all versions of the same evaluation? Should we just pick the one linked to the commit where merge happens? Maybe does not apply here but wanted to make sure this is correct

geclos: not even that i think we should pick the current eval version in main and compare against that?

andresgutgon: Open the question in Slack, not sure

packages/core/src/services/issues/merge.ts
```diff
@@ -298,6 +319,41 @@ function pickWinner({
...
+  issueIds: number[]
+}): Promise<Set<number>> {
+  if (issueIds.length === 0) return new Set()
+
+  const evalsRepo = new EvaluationsV2Repository(workspaceId)
+  const evaluations = await evalsRepo
+    .listAtCommitByDocument({
+      projectId,
+      documentUuid,
+      commitUuid: HEAD_COMMIT,
```
andresgutgon, line 344: What if a merge happens in an issue that is in a draft commit? Is that a problem?

geclos: the commit of the issue is not a relevant factor, if the issue has an associated eval that is actually merged, then we discard it for the merging process

PR #2167, opened 2026-01-22
Improve bullmq lock times to avoid jobs stuck for a long time in active state

geclos:


packages/core/src/jobs/job-definitions/runs/backgroundRunJob.ts
```diff
@@ -99,8 +101,27 @@ export const backgroundRunJob = async (
...
+      }
+    }, LOCK_EXTENSION_INTERVAL_MS)
+  }
+
+  const stopLockExtension = () => {
+    if (lockExtensionInterval) {
+      clearInterval(lockExtensionInterval)
+      lockExtensionInterval = undefined
+    }
+  }
```
andresgutgon, line 121: This could be a helper. Look like we're going to do this on every job we want to extend

packages/core/src/jobs/job-definitions/maintenance/cleanOrphanedActiveJobsJob.ts
```diff
@@ -0,0 +1,69 @@
...
+    const processedAge = job.processedOn ? now - job.processedOn : jobAge
+
+    if (processedAge > maxAgeMs) {
+      try {
+        await job.moveToFailed(
+          new Error(`Job orphaned: stuck in active state for ${Math.round(processedAge / 60000)} minutes`),
+          'orphan-cleanup',
+          false,
+        )
+        console.log(`[cleanOrphanedActiveJobsJob] Moved orphaned job ${job.id} from ${queueName} to failed`)
```
andresgutgon, line 60: 🔥

PR #2171, opened 2026-01-23
Do not notify Latitude web app when a run start on foreground

andresgutgon:
# What?
We are emiting the same event in foreground and background `documentRunStarted` and our UI start showing active runs row in traces section but is not an active run. Active runs only happen when run is run in backgroud

### DIFF
--- a/packages/core/src/events/handlers/notifyClientOfRunStatusByDocument.ts
+++ b/packages/core/src/events/handlers/notifyClientOfRunStatusByDocument.ts
@@ -10,7 +10,16 @@ export const notifyClientOfRunStatusByDocument = async ({
 }: {
   data: DocumentRunStatusEvent
 }) => {
-  const { workspaceId, projectId, documentUuid, commitUuid, run } = event.data
+  const {
+    workspaceId,
+    projectId,
+    documentUuid,
+    commitUuid,
+    run,
+    eventContext,
+  } = event.data
+
+  if (eventContext === 'foreground') return
 
   await WebsocketClient.sendEvent('documentRunStatus', {
     workspaceId,
@@ -21,6 +30,7 @@ export const notifyClientOfRunStatusByDocument = async ({
       documentUuid,
       commitUuid,
       run,
+      eventContext,
     },
   })
 }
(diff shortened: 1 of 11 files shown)


packages/core/src/services/commits/foregroundRun.ts
```diff
@@ -78,6 +78,7 @@ export async function runForegroundDocument(
   await publisher.publishLater({
     type: 'documentRunStarted',
     data: {
+      notifyUIClient: false,
```
geclos: is this even used for anything other than notifying the frontend? because otherwise i would just emit it from where it makes sense and remove this boolean

packages/core/src/events/handlers/notifyClientOfRunStatusByDocument.ts
```diff
@@ -10,12 +10,22 @@ export const notifyClientOfRunStatusByDocument = async ({
...
+  const {
+    workspaceId,
+    projectId,
+    documentUuid,
+    commitUuid,
+    run,
+    notifyUIClient,
+  } = event.data
+
+  if (!notifyUIClient) return
```
andresgutgon: This is the fix. I added this prop as mandatory to the event and decided where and where not notify the client. Only when creating this event for foreground events we don't want to notify

packages/core/src/services/commits/foregroundRun.ts
andresgutgon: But even if it was not used I think is nice that both foreground and background emit the same event. 

```typescript
  documentRunStarted: [
    notifyClientOfRunStatusByDocument,
    enqueueShadowTestChallengerHandler,
  ],
```

geclos: if the event is used for more stuff, instead of a prop to control side effect logic (services should not be aware of what event handlers do!) pass just general information: in this case the context it's running from (foreground or background) and let the event handler decide whether it should notify or not

andresgutgon: Done

PR #2173, opened 2026-01-26
Span data & styles

csansoon:
A major fix:
 - Now that we're obtaining the Completion span data from Vercel itself, messages have to be converted back to PromptL to store as a Completion Span.

Without this, we did see messages but some data like tool call arguments and tool results were lost ☝️ 

Minor fixes:
 - Span colors adjusted: Blue for Latitude main span (Prompt / External / Chat), Green for completion, Yellow for tool calling, Gray for everything else (unimportant)
 - Fixed a few styles of Span Graph Items
 - Double Clicking on a span in the graph will collapse it too.

### DIFF
--- a/packages/core/src/services/ai/telemetryMiddleware.test.ts
+++ b/packages/core/src/services/ai/telemetryMiddleware.test.ts
@@ -82,7 +82,13 @@ describe('createTelemetryMiddleware', () => {
         {
           provider: providerName,
           model,
-          input: [{ role: 'user', content: 'Hi' }],
+          input: [
+            {
+              role: 'user',
+              content: [{ type: 'text', text: 'Hi' }],
+              toolCalls: [],
+            },
+          ],
           configuration: {
             model,
             prompt: [{ role: 'user', content: 'Hi' }],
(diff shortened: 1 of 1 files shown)


packages/core/src/services/ai/telemetryMiddleware.ts
```diff
@@ -6,17 +6,63 @@ import {
...
-  prompt: unknown,
-): Record<string, unknown>[] | undefined {
-  if (!Array.isArray(prompt)) return undefined
+  vercelMessages: LanguageModelV2Prompt,
+): Message[] {
+  return vercelMessages.map((msg) => {
+    const finalMessage = {
+      role: msg.role as MessageRole,
+      content: [],
+      toolCalls: [],
```
csansoon, line 24: Note: This will not be required when we have the "Universal Message Translator"!

packages/core/src/services/ai/telemetryMiddleware.ts
```diff
@@ -6,17 +6,72 @@ import {
...
+      if (c.type === 'reasoning') {
+        content.push({ type: 'reasoning', text: c.text })
+      }
+
+      if (c.type === 'reasoning') {
+        content.push({ type: 'reasoning', text: c.text })
+      }
+
+      return content
+    })
```
andresgutgon, line 62: I feel we have done something super similar somewhere else. Extract into a function `convertVercelMessageToPromptl` for better testing and reusability

csansoon: Not this exactly, I've been looking for it... We transformed from PromptL to Vercel, but not the other way around!

andresgutgon: Or will this middleware disappear with the universal translator? Is alex working on that?

csansoon: We are building a universal message translator anyways, which will be used in the future for all cases like this one!

csansoon: The middleware will not disappear, just the translation method. The middleware is there to capture completion spans from Vercel

PR #2174, opened 2026-01-26
minor: increase mcp tool calling timeout from 1m to 5m

csansoon:
Increased the default timeout for ExternalMCP tool calling from 1 minute (modelcontextprotocol's default) to 5 min

### DIFF
--- a/packages/core/src/services/integrations/McpClient/callTool.ts
+++ b/packages/core/src/services/integrations/McpClient/callTool.ts
@@ -10,6 +10,8 @@ import { PromisedResult } from '../../../lib/Transaction'
 import { runAction } from '../pipedream/components/runAction'
 import { touchIntegration } from '../touch'
 
+const FIVE_MINUTES = 5 * 60 * 1000
+
 type ResultContent =
   | { type: 'text'; text: string }
   | { type: 'image'; data: string; mimeType: string }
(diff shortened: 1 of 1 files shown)


packages/core/src/services/integrations/McpClient/callTool.ts
```diff
@@ -72,10 +75,16 @@ export async function callIntegrationTool({
...
-      arguments: args,
-    })
+    const result = await client.callTool(
+      {
+        name: toolName,
+        arguments: args,
+      },
+      undefined,
+      {
+        timeout: 5 * MINUTES,
```
geclos: weird constant 😂 , just move the 5 to the minutes constant and call it five_minutes

packages/core/src/services/integrations/McpClient/callTool.ts
```diff
@@ -10,6 +10,9 @@ import { PromisedResult } from '../../../lib/Transaction'
 import { runAction } from '../pipedream/components/runAction'
 import { touchIntegration } from '../touch'
 
+const SECONDS = 1000
```
geclos: no need for this hombre, a single constant for the 5 minutes and palantir

packages/core/src/services/integrations/McpClient/callTool.ts
csansoon: I just really like how this makes sense semantically for time variables xd. It is REALLY EASY to see how this is set to "5 minutes"

PR #2175, opened 2026-01-26
Improve traces page re-renders

andresgutgon:
Try to find places where we could improve performance of traces table. Sometimes clicking on rows freeze the page. Looking at the task manager on Chrome dev tools I can see over time memory continully increase also when an active run is running cpu spikes which makes sense but maybe are things to improve. Overall i think we have some kind of memory leak but was not able to find some clear culprit

# Re-renders
The main culprit of the re-renders was [useFetcher](https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/hooks/useFetcher.ts#L180) using [useCurrentUrl](https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/hooks/useCurrentUrl.ts#L8) that was making the fetcher function change all the time. Use fetcher is used in `useCommits` which is used in [spanRow.tsx](https://github.com/latitude-dev/latitude-llm/blob/main/apps/web/src/app/(private)/projects/%5BprojectId%5D/versions/%5BcommitUuid%5D/documents/%5BdocumentUuid%5D/(withTabs)/traces/_components/SpanRow.tsx) so all the rows changed all the time. 

### Before 

https://github.com/user-attachments/assets/47642a21-58b7-43de-a53d-f36e621471c9

### After 

https://github.com/user-attachments/assets/24d45bc7-a2cd-4864-b6af-75c9114fd6a0

# Other changes
1. I split traces context in 2. Actions and State. The reason is to avoid re-rendering the rows because only use the actions in the context.
2. Replaced `useSearchParams` reactive nextjs hook and `useRouter` by simply intializing the traces with the URL state once. From there the updates are state -> URL. URL is changed in a way now that do not provoke a re-render. We just want to change the URL and push this new URL into the browser history. 
3. Wrap with `memo` main trace row component

# Memory leak
Not sure these changes are fixing the memory leak. But at least are improving the way the table is re-rendered when clicking on the rows

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx
@@ -1,9 +1,14 @@
 'use client'
 
-import { createContext, ReactNode, useState, useCallback } from 'react'
-import { ReadonlyURLSearchParams, useSearchParams } from 'next/navigation'
+import {
+  createContext,
+  ReactNode,
+  useState,
+  useCallback,
+  useMemo,
+  useRef,
+} from 'react'
 import { parseSpansFilters } from '$/lib/schemas/filters'
-import { useNavigate } from '$/hooks/useNavigate'
 import { AssembledSpan, Span } from '@latitude-data/constants'
 import { ROUTES } from '$/services/routes'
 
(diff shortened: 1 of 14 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx
```diff
@@ -32,6 +37,31 @@ export function buildTraceUrl({
   )
 }
 
+function syncUrlWithSelection(selection: SelectionState) {
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx:40]: Now update is
1. state -> URL
2. URL is not re-rendering the UI by using [pushState](https://developer.mozilla.org/en-US/docs/Web/API/History/pushState)

`pushState` docs:
> The new history entry's URL. Note that the browser won't attempt to load this URL after a call to pushState(), but it may attempt to load the URL later, for instance, after the user restarts the browser.

apps/web/src/app/providers/CommitProvider.tsx
```diff
@@ -29,8 +29,10 @@ const CommitProvider = ({
     return commits?.find((c) => c.uuid === serverCommit.uuid) ?? serverCommit
   }, [commits, serverCommit])
 
+  const contextValue = useMemo(() => ({ commit, isHead }), [commit, isHead])
```
andresgutgon, line 32: Probably these memo are not necessary in the future. But these are top level contexts so better to avoid passing new objects all the time.

apps/web/src/hooks/useFetcher.ts
```diff
@@ -177,11 +176,12 @@ export default function useFetcher<
 ) {
   const { toast } = useToast()
   const navigate = useNavigate()
-  const currentUrl = useCurrentUrl()
 
   return useCallback(async () => {
     if (!route) return fallback as R
 
+    const currentUrl = window.location.href
```
andresgutgon, line 183: The real fix

csansoon: lol

PR #2177, opened 2026-01-26
Custom UI for the default system-managed composite evaluation

csansoon:
When an evaluation starts tracking issues in the project, a default Composite evaluation named "Performance" is generated. This evaluation is automatically managed by the system and users cannot modify it in any way. This PR adds UI to let the user know.

(Also, I separated the `EvaluationsTable.tsx` component into three different files because it was getting too large, that's most of the PR changes)

image

image


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/evaluations/[evaluationUuid]/_components/EvaluationActions.tsx
```diff
@@ -308,7 +313,12 @@ function EditEvaluation<
           setOptions={setOptions}
           errors={errors}
           commit={commit}
-          disabled={isUpdatingEvaluation}
+          disabled={isUpdatingEvaluation || isMainEvaluation}
```
Comment (neoxelox) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/evaluations/[evaluationUuid]/_components/EvaluationActions.tsx:319]: Maybe the alert can be outside the evaluationv2form component and just after it here, just like the alert when the commit is merged (cleaner)

apps/web/src/components/evaluations/EvaluationV2Form.tsx
```diff
@@ -299,6 +301,13 @@ export default function EvaluationV2Form<
             }
           />
         )}
+        {disabled && disabledMessage && (
```
neoxelox: maybe we should move out of this component the alert?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/evaluations/[evaluationUuid]/_components/EvaluationActions.tsx
```diff
@@ -262,6 +262,11 @@ function EditEvaluation<
     document.documentUuid,
   ])
 
+  const isMainEvaluation = useMemo(
```
Comment (neoxelox) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/evaluations/[evaluationUuid]/_components/EvaluationActions.tsx]: put this before the "onUpdate", and early return on the callback too...?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/evaluations/[evaluationUuid]/_components/EvaluationActions.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/evaluations/[evaluationUuid]/_components/EvaluationActions.tsx:319]: actually you're right

PR #2178, opened 2026-01-26
Improve span row click

andresgutgon:
# What?
Improve performance when clicking on a row by memoizing client calculations and also by making several API requests. 

Also, fix a bug when clicking on `use this span` in the playground

## TODO
- [x] Parallel fetching of data when clicking on a trace
- [x] Fix 🐛. Use trace is not going to the right document

## No parallel fetch (dev 8s)
image

## Parallel fetch (dev 2.5s)
image

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx
@@ -3,7 +3,7 @@ import { Input } from '@latitude-data/web-ui/atoms/Input'
 import { ReactStateDispatch } from '@latitude-data/web-ui/commonTypes'
 import { DatePickerRange } from '@latitude-data/web-ui/atoms/DatePicker'
 import { SpansFilters } from '$/lib/schemas/filters'
-import { useMemo, useCallback } from 'react'
+import { useMemo, useCallback, useState, ChangeEvent } from 'react'
 import { CommitFilterByUuid } from './CommitFilterByUuid'
 import { ExperimentFilterByUuid } from './ExperimentFilterByUuid'
 import { TestDeploymentFilter } from './TestDeploymentFilter'
@@ -29,10 +29,28 @@ export function SpanFilters({
     projectId: project.id,
   })
 
+  const [localDocumentLogUuid, setLocalDocumentLogUuid] = useState(
+    filterOptions.documentLogUuid ?? '',
+  )
   const filters = useProcessSpanFilters({
     onFiltersChanged,
     filterOptions,
   })
+
+  const onDocumentLogUuidChange = filters.onDocumentLogUuidChange
+
+  /**
+   * Local state managed in this component to reflect immediate changes in the input field
+   * while debouncing the actual filter update to avoid excessive re-renders.
+   */
+  const handleDocumentLogUuidChange = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) => {
+      const value = e.target.value
+      setLocalDocumentLogUuid(value)
+      onDocumentLogUuidChange(value)
+    },
+    [onDocumentLogUuidChange],
+  )
 
   // Get selected commit UUIDs - empty array when no filter is set
   const selectedCommitUuids = useMemo(() => {
@@ -126,8 +144,8 @@ export function SpanFilters({
       <div className='max-w-40'>
         <Input
           placeholder='Conversation ID'
-          value={filterOptions.documentLogUuid ?? ''}
-          onChange={(e) => filters.onDocumentLogUuidChange(e.target.value)}
+          value={localDocumentLogUuid}
+          onChange={handleDocumentLogUuidChange}
         />
       </div>
     </>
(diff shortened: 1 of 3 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx
```diff
@@ -139,6 +187,23 @@ export function TraceSpanSelectionProvider({
...
+    commitUuid: commit.uuid,
+    commitId: commit.id,
+    documentUuid: document.documentUuid,
+  })
+  preloadCtxRef.current = {
+    projectId: project.id,
+    commitUuid: commit.uuid,
+    commitId: commit.id,
+    documentUuid: document.documentUuid,
+  }
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx]: Review this. I think is not necessary

Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext.tsx]: hmmm looks weird

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx
```diff
@@ -34,6 +34,24 @@ export function SpanFilters({
...
+
+  const onDocumentLogUuidChange = filters.onDocumentLogUuidChange
+  const handleDocumentLogUuidChange = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) => {
+      const value = e.target.value
+      setLocalDocumentLogUuid(value)
+      onDocumentLogUuidChange(value)
+    },
+    [onDocumentLogUuidChange],
+  )
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx:53]: why do we need to move it to local state?

apps/web/src/lib/schemas/filters.ts
```diff
@@ -2,7 +2,8 @@ import { z } from 'zod'
 
 export const spansFiltersSchema = z
   .object({
-    documentLogUuid: z.string().optional(),
+    // NOTE: documentLogUuid has to be a real `uuid` otherwise  the query fails in Drizzle
```
geclos, line 5: ok to fix the client but we need to fix the backend it should not crash if I pass something random to the endpoint

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx:53]: Yes, managed inputs with a debounce callback need local state to update the UI while the request to the backend is debounced

apps/web/src/lib/schemas/filters.ts
andresgutgon: It does not fail. Just return `undefined`

geclos: but then what was the crash? returning undefined crashed the app? it shouldn't either

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx
Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx:53]: comment it otherwise the next dev will remove this indirection

apps/web/src/lib/schemas/filters.ts
andresgutgon: The db fails. I'll show you tomorrow

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx:53]: It's not indirection. Not having local state does not allow to reflect in the ui what user types having a debounce

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx:53]: it's an indirection, a necessary one in this case. I'm just saying we should add a comment because it's not obvious why this is necessary at first glance and the next dev will remove it otherwise.

Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/(withTabs)/traces/_components/Filters/index.tsx:53]: I'll add the comment but not sure if there's a way of doing this in react without this pattern

apps/web/src/lib/schemas/filters.ts
andresgutgon: Because this filter is changing the URL and the query is in the `page.tsx` that page 💥 when the DB fails to query by a non valid `uuid`
image

PR #2183, opened 2026-01-27
feat: add backup polling for experiments when realtime updates are unreliable

geclos:
WebSocket notifications can be unreliable - messages might be lost or connections might drop temporarily. This caused issues where experiment progress showed stale data (e.g., 0/5 for 1:20 minutes when the experiment had actually finished in 50s).

This change adds automatic backup polling via SWR's refreshInterval when there are running experiments. The polling:
- Activates only when experiments are in a running state (startedAt && !finishedAt)
- Polls every 5 seconds to fetch fresh data from the server
- Automatically stops when all experiments finish
- Works alongside WebSocket updates for redundancy

The polling ensures the frontend state eventually becomes consistent with the backend state, even if WebSocket events are missed.

https://claude.ai/code/session_01VdDnNasCZCufvUadv48h7S

### DIFF
--- a/apps/web/src/stores/experiments.ts
+++ b/apps/web/src/stores/experiments.ts
@@ -3,18 +3,13 @@ import useFetcher from '$/hooks/useFetcher'
 import useLatitudeAction from '$/hooks/useLatitudeAction'
 import { ROUTES } from '$/services/routes'
 import { toast } from '@latitude-data/web-ui/atoms/Toast'
-import { useCallback, useMemo } from 'react'
+import { useMemo } from 'react'
 import useSWR, { SWRConfiguration } from 'swr'
 import { ExperimentDto } from '@latitude-data/core/schema/models/types/Experiment'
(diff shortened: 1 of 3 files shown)


apps/web/src/stores/experiments.ts
```diff
@@ -3,13 +3,19 @@ import useFetcher from '$/hooks/useFetcher'
...
 import { ExperimentDto } from '@latitude-data/core/schema/models/types/Experiment'
 
 import { Experiment } from '@latitude-data/core/schema/models/types/Experiment'
 const EMPTY_ARRAY: [] = []
 
+const POLLING_INTERVAL_MS = 5000
+
+function hasRunningExperiments(experiments: ExperimentDto[]): boolean {
+  return experiments.some((exp) => exp.startedAt && !exp.finishedAt)
+}
```
geclos: dry this method is repeated

PR #2184, opened 2026-01-27
DB Migration for removing issuesUnlocked flag

csansoon:


packages/core/drizzle/0263_remove_issues_unlocked_column.sql
```diff
@@ -0,0 +1,5 @@
+DROP TABLE "latitude"."document_suggestions" CASCADE;--> statement-breakpoint
+ALTER TABLE "latitude"."evaluation_results_v2" DROP COLUMN "used_for_suggestion";--> statement-breakpoint
+ALTER TABLE "latitude"."evaluation_versions" DROP COLUMN "enable_suggestions";--> statement-breakpoint
+ALTER TABLE "latitude"."evaluation_versions" DROP COLUMN "auto_apply_suggestions";--> statement-breakpoint
```
csansoon, line 4: Why aren't these changes migrated? Why are they caught by this migration?

geclos: that's me 👋🏼 I updated the code but didn't run the migrations because in order to be backwards compatible you cannot drop columns in the same PR/commit where you update the code

geclos: So it's ok to merge

csansoon: okayy perfect then. I just did the same with the unlockIssues flag

PR #2188, opened 2026-01-27
Bullmq worker improvements

geclos:


apps/workers/src/workers/utils/deadLetterHandler.ts
```diff
@@ -0,0 +1,339 @@
...
+    }
+  }
+
+  return {
+    total: jobs.length,
+    byQueue,
+    byError,
+    oldestJobAge: oldestTimestamp ? Date.now() - oldestTimestamp : null,
+  }
+}
```
andresgutgon: I know is draft but this PR has a lot of new concepts and refactors. Huge risk to introduce all at once. Maybe split in smaller changes

PR #2196, opened 2026-01-27
Filter out inactive issues from vector DB search results before reranking

csansoon:
## Summary
Issue discovery was limited to 20 candidates from the vector database, but inactive issues (ignored, merged, resolved) were consuming slots in those results. This could cause valid active issues to be excluded from discovery when many inactive issues existed.

The vector search limit is now increased to 1000 to fetch a larger candidate pool, and candidates are filtered against PostgreSQL to exclude inactive issues before limiting to 20 for reranking. This ensures that only active issues compete for the limited reranking slots.

### DIFF
--- a/packages/core/src/services/issues/discover.ts
+++ b/packages/core/src/services/issues/discover.ts
@@ -1,7 +1,8 @@
 import { env } from '@latitude-data/env'
+import { and, inArray, isNull } from 'drizzle-orm'
 import { Bm25Operator } from 'weaviate-client'
 import { cache as getCache } from '../../cache'
-import { database } from '../../client'
+import { database, Database } from '../../client'
 import {
   CLOUD_MESSAGES,
   EvaluationMetric,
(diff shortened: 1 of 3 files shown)


packages/core/src/services/issues/discover.ts
```diff
@@ -141,14 +152,35 @@ async function findCandidates({
...
+
+    const activeIssues = await db
+      .select({ uuid: issues.uuid })
+      .from(issues)
+      .where(
+        and(
+          inArray(issues.uuid, candidateUuids),
+          isNull(issues.ignoredAt),
+          isNull(issues.mergedAt),
+          isNull(issues.resolvedAt),
```
geclos: pretty sure we must have this query filter already built somewhere, can we dry it

csansoon: This query is not being done anywhere else really.
We have this `!issue.ignoreadAt && !issue.mergedAt && !issue.resolvedAt` filter already built as a runtime check, but not as a sql query

PR #2197, opened 2026-01-27
Add new running status to tool messages in message list

andresgutgon:
Now we're showing tools without a response always as pending. This is not right. In an agentic trace a tool can be pending (`waiting` from now on) and have a response on the next response. So what we'll do now is introduce `running` that will be only true on the playgrounds when stream is happening

## Pending tool ⚠️ 
This tool is finish in the sense that is not happening now but the UI is showing a loading indicator. Now we'll show a `waiting` (clock) icon instead in this case
image

### DIFF
--- a/apps/web/src/components/ChatWrapper/Message/Content/ToolCall/LatitudeTools/Search.tsx
+++ b/apps/web/src/components/ChatWrapper/Message/Content/ToolCall/LatitudeTools/Search.tsx
@@ -14,11 +14,13 @@ import {
   ToolCardIcon,
   ToolCardText,
   ToolCardWrapper,
+  ToolCallStatus,
 } from '../_components/ToolCard'
 import { ToolCardHeader } from '../_components/ToolCard/Header'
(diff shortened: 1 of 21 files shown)


apps/web/src/components/ChatWrapper/Message/Content/ToolCall/LatitudeTools/Search.tsx
```diff
@@ -150,15 +153,30 @@ function WebSearchOutput({
...
+            <>
+              <Icon name='loader' color='foregroundMuted' spin />
+              <Text.H5 color='foregroundMuted'>Searching...</Text.H5>
+            </>
+          ) : (
+            <>
+              <Icon name='clock' color='foregroundMuted' />
+              <Text.H5 color='foregroundMuted'>Tool not executed yet</Text.H5>
+            </>
+          )}
```
csansoon: This could be moved to a reusable component

PR #2198, opened 2026-01-28
Add random delay to issue discovery for experiment results

csansoon:
## Summary
Experiment evaluation results that do not have an issue linked to their evaluation now receive a random delay (0-30 seconds) before their discovery job runs. This reduces data races when multiple experiment results are processed in parallel, where each job would independently try to discover or create issues simultaneously—potentially leading to duplicate issues or inconsistent state.

By staggering the discovery jobs, the first job completes and creates/discovers an issue before subsequent jobs run, allowing them to find and use the existing issue instead of creating duplicates.


csansoon:
Duplicated

PR #2199, opened 2026-01-28
Reliable stream cancellation when stopping chat requests

andresgutgon:
# Problem

The "Stop" button in the playground only worked for initial prompt runs (BullMQ background jobs) but not during chat interactions. When a user clicked "Stop" while chatting, several issues occurred:

1. Abort signal not propagating: The abort signal from the client wasn't reliably reaching the AI provider through the stack (Web App → SDK → Gateway → Core → Vercel AI SDK → OpenAI)
2. UI stuck in loading state: After stopping, the UI would show unhandledRejection: undefined and remain in "running mode"
3. Partial responses lost: Any text generated before the abort wasn't saved or visible in telemetry traces
4. Full response appearing later: The complete AI response would appear on subsequent messages, indicating the stream wasn't actually cancelled

## Root Causes

1. SDK used `node-fetch`: The node-fetch polyfill has unreliable abort signal handling compared to native fetch
2. Gateway relied on `c.req.raw.signal`: This Web Fetch API signal doesn't reliably trigger when HTTP/1.1 connections are closed (no RST_STREAM frame like HTTP/2)
3. No telemetry for aborted completions: When a stream was aborted, the partial output wasn't recorded in spans

## Solution

1. SDK: Switch from `node-fetch` to native `fetch`

• Replaced node-fetch with native fetch (available in Node.js 18+)
• Native fetch properly cancels in-flight requests when the abort signal fires
• Added { once: true } to abort event listeners for automatic cleanup

2. Gateway: Detect disconnects via TCP socket close

• Created createRequestAbortSignal(c) utility that listens to socket.once('close')
• The socket 'close' event fires reliably for both HTTP and HTTPS connections
• Works correctly behind AWS load balancers that terminate TLS and proxy via HTTP/1.1
• Defensive coding for test environments where HttpBindings aren't available

3. Core: Record partial output on abort

• Extracted recordAbortedCompletion() to create a completion span with partial text
• Ensures aborted responses are visible in telemetry traces
• Saves partial provider log for billing/analytics purposes

  4. Web App: Simplified abort handling
• Removed custom AbortController wrapper in chat route - now passes req.signal directly to SDK
• Consolidated disabledSubmit logic into playground.isSubmitDisabled property

## TODO
- [x] ⚠️ Check if we really need SSL (IN DEVELOPMENT) to make the cancel stream work. We don't was just a deviation but the solution works on plain dev environment without SSL
- [x] Add CHANGELOG to typescript SDK. Fix stream + remove `node-fetch`. It's a major version.
- [x] Add test in SDK for abort passing signal
- [x] Add tests for `streamAIResponse` telemetry in abort addition.
- [x] Add tests for chat handler abort
- [x] Explain the problem better problem

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -10,7 +10,7 @@ catalog:
   drizzle-orm: 0.44.2
   drizzle-kit: 0.31.1
   dd-trace: 5.69.0
-  promptl-ai: ^0.9.4
+  promptl-ai: ^0.9.5
   zod: 4.1.8
   react: 19.1.4
   react-dom: 19.1.4


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/ChatInputBox/index.tsx
```diff
@@ -37,7 +37,7 @@ export const ChatInputBox = memo(function ChatInputBox({
           playground.isLoading ||
           playground.isStopping ||
           !!playground.error ||
-          !hasActiveStream()
```
Comment (geclos) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/ChatInputBox/index.tsx:40]: you removed a condition

apps/web/src/components/PlaygroundCommon/Chat/index.tsx
```diff
@@ -168,7 +169,7 @@ function ChatInputBox({
           playground.isLoading ||
           playground.isStopping ||
           !!playground.error ||
-          !hasActiveStream()
```
geclos, line 171: condition removed

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/ChatInputBox/index.tsx
Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/ChatInputBox/index.tsx:40]: My Cursor:
> We removed !hasActiveStream() because after aborting a chat, the send button was staying disabled. The `hasActiveStream()` was returning false after abort which, combined with other conditions kept the button disabled.

Playing with it looks like it works like this without this condition.

Comment (geclos) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/ChatInputBox/index.tsx:40]: what happens now if i continue a convo that has previously been aborted? does it pick up where the convo was stopped? make sure that's the case otherwise i would maintain the button disabled.

Comment (andresgutgon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/_components/DocumentEditor/Editor/ChatInputBox/index.tsx:40]: It works ™

apps/web/src/components/PlaygroundCommon/Chat/index.tsx
andresgutgon: Same

apps/gateway/src/routes/api/v3/conversations/attach/attach.handler.ts
```diff
@@ -89,7 +90,7 @@ async function handleNonStreamingMode(
   ctx: Parameters<AppRouteHandler<AttachRoute>>[0],
   args: Parameters<typeof attachRun>[0],
 ) {
-  const abortSignal = ctx.req.raw.signal // FIXME: this is not working
+  const abortSignal = createRequestAbortSignal(ctx)
```
csansoon, line 93: real nice

packages/sdks/typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@latitude-data/sdk",
-  "version": "5.3.1",
+  "version": "5.4.0",
```
andresgutgon, line 3: It's a minor because api remains equal. Although this potentially might break some clients i doubt anyone is using NodeJS < 18. I tried with 18, which is already out of LTS, and it works. 

So good bye `node-fetch`, we thank you for your service 🫡

apps/web/src/hooks/playgroundChat/usePlaygroundChat.ts
```diff
@@ -328,6 +330,24 @@ export function usePlaygroundChat({
...
+        if (!hasStreaming) return msg
+        return {
+          ...msg,
+          content: msg.content.map((c) =>
+            'isStreaming' in c && c.isStreaming ? { ...c, isStreaming: false } : c,
+          ),
+        } as Message
+      }),
+    )
+  }, [setMessages])
```
geclos, line 355: hard to understand what this is doing and why, would leave a comment

apps/web/src/hooks/playgroundChat/usePlaygroundChat.ts
```diff
@@ -473,8 +502,13 @@ export function usePlaygroundChat({
     if (!documentLogUuid) return
     if (isStoppingRun) return
 
-    await stopRun({ runUuid: documentLogUuid })
-  }, [isLoading, documentLogUuid, stopRun, isStoppingRun])
+    if (isChat.current && abortCurrentStream) {
+      abortCurrentStream()
+      cleanupStreamingState()
+    } else {
+      await stopRun({ runUuid: documentLogUuid })
```
geclos, line 515: hmm weird there's two different handlers for stopping a stream since they are conceptually the same, what is stopRun doing that is different than 

```
 abortCurrentStream()
 cleanupStreamingState()
```

apps/web/src/hooks/playgroundChat/usePlaygroundChat.ts
```diff
@@ -489,6 +523,7 @@ export function usePlaygroundChat({
     setProvider(undefined)
     setModel(undefined)
     resetTimer()
+    isChat.current = false
```
geclos, line 532: why do we need to keep track of whether we are in chat mode or in rundocumentatcommit mode?

packages/core/src/lib/streamManager/step/streamAIResponse.ts
```diff
@@ -107,6 +108,20 @@ export async function streamAIResponse({
...
+        config,
+        messages,
+        accumulatedText: accumulatedText.text,
+      })
+
+      try {
+        controller.close()
+      } catch (_) {
+        // Controller may already be closed
+      }
```
geclos: nein

andresgutgon: I'll change

apps/web/src/hooks/playgroundChat/usePlaygroundChat.ts
andresgutgon: This was already here afaik. I think is changed when the run finishes. Don't remember but this was working and keep working

apps/web/src/hooks/playgroundChat/usePlaygroundChat.ts
andresgutgon: `cleanupStreamingState` is pure frontend

andresgutgon: Ah and `stopRun` is stopping the active background run (bullmq) not a http streaming request

geclos: uhm, shouldn't we also stop the active background run as well in chat? or are chats happening in foreground?

PR #2204, opened 2026-01-29
Add multi-turn conversation simulation for experiments

csansoon:
## Summary

Adds multi-turn conversation simulation to the experiments feature, enabling evaluation of agents across full conversations rather than single turns.

When enabled, experiments can now simulate ongoing user interactions with the tested agent. After the agent responds, a simulated user evaluates the response and decides whether to:
- **End the conversation** — when the goal is achieved or the conversation reaches a natural conclusion
- **Continue with a new message** — generating a realistic follow-up based on the conversation context

### Configuration

- **Enable/disable** multi-turn simulation via the experiment settings UI
- **Maximum turns** — configurable limit (2-10 turns) to cap conversation length

### Decision Flow

The simulated user evaluates the assistant's response before deciding to continue or end, ensuring the simulation accurately reflects natural conversation patterns where users react to responses rather than predicting outcomes.

### Feature Flag

This feature is currently hidden behind a `multiTurnSimulations` feature flag.

### Next Steps

This feature is not finished. This PR only introduces the backend logic for multi-turn simulation and a setting for max turns.

In a future PR, an option for adding Behaviour and Goals instructions to the simulated user will be added, being able to define it from the Experiment Modal or from a dataset column!


apps/web/src/components/SimulationSettings/ConversationSimulation.tsx
```diff
@@ -0,0 +1,144 @@
...
+function ConversationTurnsSetting({
+  value,
+  onChange,
+  disabled,
+}: {
+  value: number
+  onChange: (value: number) => void
+  disabled?: boolean
+}) {
+  const values = useMemo(() => [value], [value]) // This is made to avoid generating a new array on every render
```
andresgutgon, line 47: This is weird. Why not pass an array?

apps/web/src/components/SimulationSettings/ToolSimulation.tsx
```diff
@@ -0,0 +1,348 @@
...
+                    ),
+                  )}
+              </div>
+            </div>
+          }
+        />
+      )}
+    </div>
+  )
+}
```
andresgutgon, line 348: This is all pre-existing right?

packages/core/src/jobs/job-definitions/runs/backgroundRunJob.ts
```diff
@@ -157,6 +164,24 @@ export const backgroundRunJob = async (
...
+        abortSignal: abortController.signal,
+        writeStream,
+        workspaceId,
+        projectId,
+        documentUuid,
+        commitUuid,
+        runUuid,
+        initialMessages: await result.conversation.messages,
+      })
+    }
```
andresgutgon, line 183: Tests here

apps/web/src/components/SimulationSettings/ConversationSimulation.tsx
csansoon: Because the array would be redefined on every render

apps/web/src/components/SimulationSettings/ToolSimulation.tsx
csansoon: Yes I just moved it to its own file, no changes here!

csansoon: Well not actually true! I added a condition to not render the Advanced tool settings if tool simulation is disabled

packages/core/src/jobs/job-definitions/runs/helpers/experimentHandler.ts
```diff
@@ -0,0 +1,115 @@
...
+  runUuid,
+}: MarkExperimentFailureArgs) {
+  await updateExperimentStatus(
+    {
+      workspaceId,
+      experiment,
+    },
+    (progressTracker) => progressTracker.documentRunFinished(runUuid, false),
+  )
+}
```
andresgutgon, line 115: Is this helper tested in the multi turn tests?

packages/core/src/jobs/job-definitions/runs/backgroundRunJob.ts
csansoon: There already are!

packages/core/src/jobs/job-definitions/runs/helpers/experimentHandler.ts
csansoon: Nope, there are no tests for these methods (which already existed before and I did not change, I just moved them to its own file to avoid having a 700+ lines service)

andresgutgon: Maybe it's a good time to cover it

apps/web/src/components/SimulationSettings/ConversationSimulation.tsx
andresgutgon: But why we need to allow single and multiple. This component should accept multiple. Also investigate maybe memo this component avoid re render even passing an array

packages/core/src/jobs/job-definitions/runs/helpers/multiTurnSimulation.ts
```diff
@@ -0,0 +1,203 @@
...
+import { RunIdentifiers } from './types'
+
+/**
+ * Determines if multi-turn simulation should be run based on the simulation settings
+ */
+export function shouldRunMultiTurnSimulation(
+  simulationSettings?: SimulationSettings,
+): simulationSettings is SimulationSettings {
+  return Boolean(
+    simulationSettings?.maxTurns && simulationSettings.maxTurns > 1,
```
geclos, line 24: bigger than one? this means if a set up simulation with 1 turn there won't be any user response no? it's just the first experiment run, or i'm misunderstanding?

packages/core/src/jobs/job-definitions/runs/helpers/multiTurnSimulation.ts
```diff
@@ -0,0 +1,203 @@
...
+ * Simulates multi-turn conversations by generating user responses
+ * and continuing the conversation until maxTurns is reached or
+ * the conversation naturally ends.
+ *
+ * Each turn:
+ * 1. Generates a simulated user response based on conversation context
+ * 2. Sends the response to the document and gets the assistant's reply
+ * 3. Forwards the stream events to Redis for client consumption
+ */
+export async function simulateUserResponses({
```
geclos, line 155: i assumed this runs in a single job, would have been a good opportunity to leverage bullmq's parent-child workflow feature and have single job per turns. This frees up the experiments queue for other users and will make the system more resilient if anyone tries to run a 1k row experiment with multiturn simulation that could take hours to execute.

packages/core/src/jobs/job-definitions/runs/helpers/multiTurnSimulation.ts
csansoon: Yes. 1 turn is just the regular execution.

packages/core/src/jobs/job-definitions/runs/helpers/multiTurnSimulation.ts
csansoon: can do for the next PR!

packages/core/src/jobs/job-definitions/runs/helpers/multiTurnSimulation.ts
geclos: hmm i think this will be confusing for users? i would expect 1 turn to mean "1 assistant-user interaction" — but maybe it's just me, talk it with product 🙏🏼

PR #2206, opened 2026-01-29
Fix/events refering to old providerlog response

geclos:
Some parts of the code were still refering to a property that I assumed I had removed from code, but I hadn't for some reason.

### DIFF
--- a/apps/gateway/scripts/test-run-chat.ts
+++ b/apps/gateway/scripts/test-run-chat.ts
@@ -1,0 +1,326 @@
+const baseUrl = process.env.GATEWAY_URL ?? 'http://localhost:8787'
+const apiKey =
+  process.env.LATITUDE_API_KEY ?? '5a1ea68f-e9cf-40fd-b9f7-34763c9f8248' // test api key, have a good day scrappers :wave:
+const debugSse = process.env.DEBUG_SSE === '1'
+
+const promptPath = 'gateway-run-chat-test'
+const promptContent = `
+---
+provider: openai
+model: gpt-4.1-mini
+---
+
+You are a concise assistant. Answer in two sentences.
+
+Topic: {{ topic }}
+`.trim()
+
+type ProjectResponse = {
+  project: { id: number; name: string }
+  version: { uuid: string }
+}
+
+type RunResponse = {
+  uuid: string
+  response?: { text?: string }
+}
+
+type SseEvent = {
+  event?: string
+  data?: unknown
+}
+
+async function requestJson<T>(path: string, body?: unknown): Promise<T> {
+  const response = await fetch(`${baseUrl}${path}`, {
+    method: 'POST',
+    headers: {
+      Authorization: `Bearer ${apiKey}`,
+      'Content-Type': 'application/json',
+    },
+    body: body ? JSON.stringify(body) : undefined,
+  })
+
+  if (!response.ok) {
+    const text = await response.text()
+    throw new Error(`Request failed (${response.status}): ${text}`)
+  }
+
+  return (await response.json()) as T
+}
+
+function parseSseEvent(block: string): SseEvent | null {
+  const lines = block.split('\n')
+  const dataLines: string[] = []
+  let event: string | undefined
+
+  for (const line of lines) {
+    const trimmed = line.trimEnd()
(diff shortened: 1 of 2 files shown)


apps/gateway/scripts/test-run-chat.ts
```diff
@@ -0,0 +1,326 @@
...
+  console.log('Foreground run (stream) OK:', runStreamText)
+  console.log('Background run OK:', runBackground.uuid)
+  console.log('Chat (non-stream) OK:', chatSync.response.text)
+  console.log('Chat (stream) OK:', chatStreamText)
+}
+
+main().catch((error) => {
+  console.error(error)
+  process.exitCode = 1
+})
```
andresgutgon, line 326: is this a disposable?

packages/core/src/services/conversations/cache.ts
```diff
@@ -0,0 +1,242 @@
...
+}
+
+async function removeConversationCacheLookupKey(documentLogUuid: string) {
+  try {
+    const cacheClient = await cache()
+    await cacheClient.del(buildConversationCacheLookupKey(documentLogUuid))
+  } catch {
+    // do nothing
+  }
+}
```
andresgutgon, line 242: Does this file have tests? I'm on mobile maybe missed

packages/core/src/services/documentLogs/addMessages/index.ts
```diff
@@ -156,50 +128,133 @@ async function retrieveData({
   documentLogUuid,
 }: {
   workspace: WorkspaceDto
-  documentLogUuid: string | undefined
+  documentLogUuid: string
 }) {
-  const logsRepo = new DocumentLogsRepository(workspace.id)
-  const logResult = await logsRepo.findByUuid(documentLogUuid)
```
andresgutgon, line 162: this was already wrong no?

packages/core/src/services/documentLogs/addMessages/index.ts
```diff
@@ -156,50 +128,133 @@ async function retrieveData({
...
+}: {
+  workspace: WorkspaceDto
+  providerId?: number
+  config: LatitudePromptConfig
+}) {
+  if (providerId) {
+    const provider = await unsafelyFindProviderApiKey(providerId)
+    if (!provider) {
+      return Result.error(
+        new NotFoundError(
```
andresgutgon, line 217: Should not be an ai error?

packages/sdks/typescript/src/test/run-sync-response.ts
```diff
@@ -1,9 +1,15 @@
...
 export const RUN_TEXT_RESPONSE = {
   uuid: 'a8f2e5d8-4c72-48c7-a6e0-23df3f1cbe2a', // Random
   conversation: [],
   response: {
     streamType: 'text' as const,
     text: 'some-text',
+    input: [],
+    model: 'gpt-4o',
+    provider: Providers.OpenAI,
+    cost: 0,
```
andresgutgon, line 12: Should not return this all cost types? input, output...

packages/core/src/services/conversations/cache.ts
geclos: it does

apps/gateway/scripts/test-run-chat.ts
geclos: what do u mean disposable? i want it to error out if main fails

packages/core/src/services/documentLogs/addMessages/index.ts
geclos: i don't think so, what makes you think that?

packages/sdks/typescript/src/test/run-sync-response.ts
geclos: stream manager returns a number, it is true we should refactor stream manager to return the cost breakdown, but it's outside the scope of my work in this PR.

packages/core/src/services/documentLogs/addMessages/index.ts
geclos: should it? I think errors should be semantically correct and notfounderror makes more sense in this case

PR #2209, opened 2026-01-31
Migrate to ESlint 9 flat config

andresgutgon:
# What?
Next.js latest 16 version requires ESlint 9 so this is good opportunity to move to that version


apps/web/src/stores/datasetRows/position.ts
```diff
@@ -20,7 +20,7 @@ export default function useDatasetRowPosition(
   }: {
     dataset: Dataset
     datasetRow: DatasetRow
-    pageSize?: Number
+    pageSize?: number
```
andresgutgon, line 23: lol

packages/core/src/services/documentTriggers/deleteAllDocumentTriggersFromCommit.test.ts
```diff
@@ -249,7 +249,7 @@ describe('deleting documents...', () => {
         })
         .then((r) => r.unwrap())
       // None should be deleted since the operation should have aborted on error
-      expect(triggers.find((t) => t.uuid === created1.uuid)).toBeTruthy
+      expect(triggers.find((t) => t.uuid === created1.uuid)).toBeTruthy()
```
andresgutgon, line 252: lol

packages/core/src/services/tracing/spans/specifications/unresolvedExternal.ts
```diff
@@ -58,7 +58,7 @@ async function getResolvedData(
 
   // If prompt is found in the commit, return the document version and commit.
   const docsResult = await docsRepo.getDocumentsAtCommit(commit)
-  if (!Result.isOk(docsResult)) docsResult
+  if (!Result.isOk(docsResult)) return docsResult
```
andresgutgon, line 61: Real bug cached in ESLint 9

eslint.config.mjs
andresgutgon, line 1: Now, all ESLint config for all the monorepo is flat centralized here. No more ESLint config spread in packages and apps

PR #2214, opened 2026-02-02
Use rosetta in the ai backend

neoxelox:


packages/core/src/lib/vercelSdkFromV5ToV4/convertResponseMessages.ts
```diff
@@ -142,19 +24,43 @@ export function convertResponseMessages({
...
+
+      toolCalls.push({
+        id: content.toolCallId,
+        name: content.toolName,
+        arguments: content.toolArguments ?? content.args ?? {},
+        _sourceData: resolvedTools?.[content.toolName]?.sourceData,
+      })
+    }
+
+    // Duplicating tool calls in the assistant message for legacy compatibility
```
geclos, line 62: leave a TODO here plz, will revisit to remove this

packages/core/src/services/ai/convertLatitudeMessagesToVercelFormat.test.ts
andresgutgon, line 1: So these tests were moved somewhere else? It has some cases

neoxelox: index.test.ts

PR #2215, opened 2026-02-02
minor: show error in experiment when simulation fails

csansoon:
~Note: Although we persist the error in `runErrors`, it is not being displayed anywhere yet when reloading the page.~
(Not persisting anymore)

### DIFF
--- a/packages/core/src/jobs/job-definitions/runs/backgroundRunJob.ts
+++ b/packages/core/src/jobs/job-definitions/runs/backgroundRunJob.ts
@@ -179,7 +179,7 @@ export const backgroundRunJob = async (
         commitUuid,
         runUuid,
         initialMessages: await result.conversation.messages,
-      })
+      }).then((r) => r.unwrap())
     }
 
     if (experiment) {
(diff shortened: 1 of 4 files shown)


packages/core/src/jobs/job-definitions/runs/backgroundRunJob.ts
```diff
@@ -195,6 +200,14 @@ export const backgroundRunJob = async (
   } catch (error) {
     writeStream.write({ type: ChainEventTypes.ChainError, data: error })
 
+    if (documentLogUuid) {
+      await createChainRunError({
```
andresgutgon: Check if it's used

csansoon: not used, removing

geclos:
> Note: Although we persist the error in runErrors, it is not being displayed anywhere yet when reloading the page.

runErrors will be deprecated, run errors are stored in events attribute of the relevant span. If we require another abstraction to store experiment errors let's discuss it.

PR #2216, opened 2026-02-02
Fix last output not showing correct last completion response

andresgutgon:
# What?
The logic for selecting the correct last span was not correct. Now we use the same logic everywhere and fix the bug

## TODO
- [x] Fix bug described
- [x] Do not allow annotate on spans other than top level span

## BUG 1
The last output in production is the output of another completion span in the trace
image

## Fixed
In this case is an enrichment agent and the output is telling me who is this person
image

## Bug 2
Empty place for annotation UI in sub-spans
image

Fixed
image

### DIFF
--- a/apps/web/src/components/ChatWrapper/Message/Content/ToolCall/_components/ToolCard/index.tsx
+++ b/apps/web/src/components/ChatWrapper/Message/Content/ToolCall/_components/ToolCard/index.tsx
@@ -92,7 +92,13 @@ export function ToolCardWrapper({
   }, [messageIndex, contentBlockIndex, getAnnotationsForBlock, span])
 
   const evaluation = evaluations[0]
-
+  const canAnnotate =
+    !blockAnnotations.length &&
+    evaluation &&
+    messageIndex !== undefined &&
+    contentBlockIndex !== undefined &&
+    span !== undefined &&
+    isMainSpan(span)
   return (
     <div
       className={cn(
@@ -101,7 +107,7 @@ export function ToolCardWrapper({
       )}
     >
       {children}
-      {(blockAnnotations.length > 0 || evaluation) && span && (
+      {canAnnotate ? (
         <div className='flex flex-col gap-y-4 border-t pt-4 px-4 pb-4'>
           {blockAnnotations.map((annotation: AnnotatedTextRange) => (
             <AnnotationForm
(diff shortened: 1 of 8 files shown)


apps/web/src/components/ChatWrapper/Message/Content/ToolCall/_components/ToolCard/index.tsx
```diff
@@ -101,7 +107,7 @@ export function ToolCardWrapper({
       )}
     >
       {children}
-      {(blockAnnotations.length > 0 || evaluation) && span && (
+      {canAnnotate ? (
```
geclos, line 110: you changed the condition here

apps/web/src/components/tracing/spans/Prompt.tsx
```diff
@@ -27,10 +27,17 @@ export default {
...
-  )
+
+
+  const completionSpan = useMemo(() => {
+    if (!trace) return undefined
+
+    return findLastSpanOfType({
+      children: trace.children,
+      spanType: SpanType.Completion,
+      includeMainSpanChildren: true,
```
geclos: why... don't you just include it always? this is iterating over memory that is already stored you are saving nothing by not returning the children

apps/web/src/components/ChatWrapper/Message/Content/ToolCall/_components/ToolCard/index.tsx
andresgutgon: The full annotated block is not visible unless the user can annotate. In the description, I put the current behaviour where a white empty space remains even if you can not annotate.

apps/web/src/components/tracing/spans/Prompt.tsx
andresgutgon:
```typescript
  return findLastSpanOfType({
    children: mainSpan.children,
    spanType: SpanType.Completion,
    includeMainSpanChildren: false,  // Intentionally false
  })
```

`findCompletionSpanForSpan` is using the flag set as `false`. It's used to find the completion span for a specific agent, without descending into subagent prompts. 

It's about search scope, not performance.

andresgutgon: Alternative name proposals:


The current name includeMainSpanChildren is confusing because it sounds like it's about including children in the results, when it's really about to recurse into main spans during the search.

Better names?
• `crossAgentBoundaries` - Clearly describes the use case: do you want to
    search across nested agent/prompt boundaries?
• `recurseIntoNestedAgents` - Explicit about the traversal behavior
• `searchNestedAgents` - Simple and describes what happens when true

Or inverted (default true):
• `stopAtAgentBoundary` - Describes the boundary behavior when true

andresgutgon: I like `searchNestedAgents`. I'll go with that

apps/web/src/components/ChatWrapper/Message/Content/ToolCall/_components/ToolCard/index.tsx
csansoon: fixed* the condition

csansoon (approved):
😌

PR #2217, opened 2026-02-02
Feat/annotate full messages

geclos:


apps/web/src/components/ChatWrapper/Message/Content/Text.tsx
```diff
@@ -281,9 +301,41 @@ const ContentText = memo(
...
+                    selectedContext={{
+                      messageIndex,
+                      contentBlockIndex,
+                      contentType: 'text',
+                    }}
+                    span={span as SpanWithDetails<MainSpanType>}
+                  />
+                )}
+            </div>
+          )}
```
andresgutgon: This component is a monster. Can we give some shape before adding more things? Would be nice to encapsulate text selection logic in a hook. I think there are already parts in hooks but still there is a lot code here

apps/web/src/components/ChatWrapper/Message/Content/Text.tsx
```diff
@@ -281,9 +301,41 @@ const ContentText = memo(
       })
     }, [groups, index, color, TextComponent, renderSegmentWithAnnotations])
 
+    const evaluation = evaluations[0]
+
     return (
       <div className='flex flex-col gap-4'>
         <div className='flex flex-col gap-y-1'>{messagesList}</div>
+        {(fullBlockAnnotations.length > 0 || evaluation) &&
+          span &&
```
andresgutgon: This block is the same in tools. Can we abstract it?

PR #2219, opened 2026-02-03
Conversation traces UI

andresgutgon:
# What?

1. Conversation show all the messages on all the traces
2. Display a new wrapping bar `conversations` when there are more than 1 trace

image

## TODO
- [x] Metadata: Add `versionId`, prompt name, `conversation` (tag), parameters, and last output of last trace. 
- [x] Metadata: Add all evaluations of all traces ordered by execution time. 
- [x] Review trace metadata evaluations if they're correct after grouping by `documentLogUuid`.
- [x] Fix download traces. Select all spans under conversation and download all
- [x] Fix generate dataset. Select all spans under conversation of type `prompt` or `external
- [x] Check if Websocket works
- [x] QA filters
- [x] Review link a log from evaluations
- [x] Review link a log from annotations
- [x] Review if some existing traces API endpoints or data access methods are no longer used. I think they're used in annotations.
- [x] Add tests for conversations data access queries


PLAN_CONVERSATION_TRACES.md
```diff
@@ -0,0 +1,226 @@
...
+
+- Groups spans by `documentLogUuid`
+- Aggregates tokens/duration/cost across all traces
+- Returns version (commitUuid), source from latest span
+- Keyset pagination support
+- Supports filters (commitUuids, experimentUuids, createdAt, etc.)
+
+### 1.3 API Endpoint - List Conversations
+
+**File:** `apps/web/src/app/api/conversations/route.ts`
```
geclos: i feel like we have many endpoints to fetch traces / spans / conversations, make sure to dry routes whenever possible ideally we would only need these endpoints:
- get conversation
- get trace (returns the assembled trace)
- get span (return the span with details)

PLAN_CONVERSATION_TRACES.md
```diff
@@ -0,0 +1,226 @@
...
+Flat endpoint with zod validated query params:
+
+```typescript
+const searchParamsSchema = z.object({
+  projectId: z.string(),
+  commitUuid: z.string(),
+  documentUuid: z.string(),
+  from: z.string().optional(),
+  limit: z.coerce.number().int().min(1).max(100).default(DEFAULT_PAGINATION_SIZE),
+  filters: z.string().optional(),
```
geclos: make sure to keep the filtering we are doing in traces page currently. When a conversation uuid is selected we do not show the table with the selected trace open but rather a table with a single item that is the filtered conversation. This is to avoid issues with pagination (and we also recently broke it and had to refix it, so let's not rebreak it again)

packages/core/src/data-access/conversations/fetchConversations.ts
```diff
@@ -0,0 +1,138 @@
...
+  const pageUuids = hasMore ? paginatedUuids.slice(0, limit) : paginatedUuids
+
+  if (pageUuids.length === 0) {
+    return Result.ok({ items: [], next: null })
+  }
+
+  const uuidList = pageUuids.map((r) => r.documentLogUuid!)
+
+  const result = await db
+    .select(conversationAggregateFields)
```
andresgutgon, line 159: Note here, I do 2 queries. First one to get all documentLogUuids, and now only for those documentUuids, we do all the aggregations of a conversation. 
1. Tokens of all traces
2. Cost of all traces
3. Duration of all traces

Doing it this way the aggregations run on a subset of the `spans` table. 

What do you think?

PLAN_CONVERSATION_TRACES.md
andresgutgon: I think it remains the same

andresgutgon:
@claude review please

packages/core/src/repositories/spansRepository.ts
```diff
@@ -157,6 +157,20 @@ export class SpansRepository extends Repository<Span> {
...
+
+    return this.db
+      .select({ traceId: spans.traceId, spanId: spans.id })
+      .from(spans)
+      .where(
+        and(this.scopeFilter, inArray(spans.documentLogUuid, documentLogUuids)),
+      )
+      .then((r) =>
+        r.map((row) => ({ traceId: row.traceId, spanId: row.spanId })),
+      )
```
geclos, line 256: might as well group them?

andresgutgon: This is used in the dataset preview spans to get a flat list of trace-span

packages/core/src/data-access/conversations/fetchConversations.ts
```diff
@@ -0,0 +1,230 @@
...
+      documentLogUuid: spans.documentLogUuid,
+      latestStartedAt: sql<string>`MAX(${spans.startedAt})`.as(
+        'latest_started_at',
+      ),
+    })
+    .from(spans)
+    .where(and(...allConditions))
+    .groupBy(spans.documentLogUuid)
+    .orderBy(desc(sql`MAX(${spans.startedAt})`), desc(spans.documentLogUuid))
+    .limit(limit + 1)
```
geclos, line 147: @claude think about the performance of this query in a document with 100k spans in a single commit

PR #2222, opened 2026-02-03
Add user goal setting for conversation simulation experiments

csansoon:
## Summary

Adds support for configuring a user goal in multi-turn conversation simulations. The goal describes the simulated user's intended behavior and helps guide the simulation to decide when to stop the conversation.

The goal can be configured in two ways:
- **Global**: A single goal text that applies to all rows in an experiment
- **From dataset column**: A column from the selected dataset, allowing per-row goals for more targeted simulations

When a dataset is selected in the experiment modal, users can toggle between these two modes. Without a dataset, only the global text input is shown.

The goal value is resolved per-row at experiment start time and passed to each background job, where it's used by the simulated user response generation.

https://github.com/user-attachments/assets/e0721ce8-b263-45a7-aa97-67aa87cfba86

### DIFF
--- a/apps/web/src/components/RunExperimentModal/ExperimentForm/_components/SimulationSettings.tsx
+++ b/apps/web/src/components/RunExperimentModal/ExperimentForm/_components/SimulationSettings.tsx
@@ -1,9 +1,17 @@
 import { SimulationSettingsPanel } from '$/components/SimulationSettings'
 import { SimulationSettings } from '@latitude-data/constants/simulation'
-import { useCallback } from 'react'
+import { useCallback, useEffect } from 'react'
 import { ExperimentFormPayload } from '../useExperimentFormPayload'
+import { useLabels } from './useLabels'
 
 export function ExperimentSimulationSettings(payload: ExperimentFormPayload) {
+  const { labels, buildLabels } = useLabels()
+
+  useEffect(() => {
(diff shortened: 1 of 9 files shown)


apps/web/src/components/RunExperimentModal/ExperimentForm/_components/SimulationSettings.tsx
```diff
@@ -1,21 +1,33 @@
...
 import { ExperimentFormPayload } from '../useExperimentFormPayload'
+import { useLabels } from './useLabels'
 
 export function ExperimentSimulationSettings(payload: ExperimentFormPayload) {
+  const { labels, buildLabels } = useLabels()
+
+  useEffect(() => {
+    if (!payload.selectedDataset) return
+    buildLabels(payload.selectedDataset)
+  }, [payload.selectedDataset, buildLabels])
```
andresgutgon, line 13: Can we avoid this `useEffect`?

csansoon: I think here it is fine, not all useEffects are bad :(

andresgutgon: All of them except for subscriptions to external stores.

PR #2225, opened 2026-02-03
Add Temporal.io workflow orchestration

csansoon:
## Summary

Introduces Temporal.io as a workflow orchestration platform alongside the existing BullMQ job queue system. This enables complex, multi-step workflows with native support for long-running processes, automatic retries, and durable execution.

### Infrastructure

**Docker Compose Services:**
- `temporal` - Temporal server using `temporalio/auto-setup` image, configured to use the existing PostgreSQL database
- `temporal-ui` - Web dashboard for workflow visibility and debugging (accessible at port 8080)
- Added `temporal` database to `docker/init-db.sh` for Temporal's event store

**Environment Variables (in `packages/env`):**
- `TEMPORAL_ADDRESS` - Server address (default: `localhost:7233`)
- `TEMPORAL_NAMESPACE` - Namespace for workflow isolation (default: `default`)
- `TEMPORAL_TASK_QUEUE` - Default task queue name
- `TEMPORAL_TLS_*` - TLS configuration for production deployments
- `TEMPORAL_WORKERS_HOST/PORT` - Worker health check configuration

### New Packages and Apps

**`packages/core/src/temporal/`:**
- `client.ts` - Singleton Temporal client with helpers for starting, signaling, and querying workflows
- `connection-client.ts` - Client connection safe for Next.js/browser contexts
- `connection-worker.ts` - Native worker connection (isolated to prevent bundling issues)
- `queues.ts` - Queue definitions with configurable concurrency (runs: 100, evaluations: 100, events: 100, tracing: 25, maintenance: 5)
- `workflows/` - Workflow definitions
- `activities/` - Activity implementations

**`apps/temporal-workers/`:**
- New application dedicated to running Temporal workers
- Creates multiple workers polling different task queues with priority-based concurrency
- Express health check endpoint
- Graceful shutdown handling
- Docker support with `docker/Dockerfile`

### Experiment Execution

As a proof-of-concept, the experiment execution flow has been migrated to use Temporal:
- Single `experimentWorkflow` replaces the previous multi-job BullMQ coordination
- Activities routed to priority queues (`runs` for document execution, `evaluations` for evaluation runs)
- Real-time progress updates via Temporal queries
- Cancellation support via Temporal signals
- BullMQ implementation preserved as comments for reference

### Queue Priority Configuration

Mirrors the existing BullMQ queue structure with activity-level routing:
| Queue | Concurrency | Purpose |
|-------|-------------|---------|
| `main` | 50 | Main workflow orchestration | 
| `runs` | 100 | Document runs (high priority) |
| `evaluations` | 100 | Evaluation execution |
| `events` | 100 | Event processing |
| `tracing` | 25 | Span ingestion |
| `maintenance` | 5 | Cleanup jobs |

### Migration Note

**For developers with existing databases:** The `docker/init-db.sh` script only runs on fresh database initialization. If you already have a running PostgreSQL container, you need to manually create the Temporal database:
```
docker compose exec db psql -U latitude -c "CREATE DATABASE temporal;"
```

### DIFF
--- a/apps/temporal-workers/docker/Dockerfile
+++ b/apps/temporal-workers/docker/Dockerfile
@@ -1,0 +1,100 @@
+ARG PROJECT="@latitude-data/temporal-workers"
+ARG PROJECT_PATH="apps/temporal-workers"
+ARG DD_GIT_REPOSITORY_URL
+ARG DD_GIT_COMMIT_SHA
+
+FROM node:22-alpine AS alpine
+
+RUN apk add --no-cache libc6-compat curl python3 make g++ gcc
+
+FROM alpine AS base
+
+RUN npm install -g corepack@0.31.0 && corepack enable
+
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+
+RUN pnpm i -g turbo
+
+FROM base AS pruner
+
+ARG PROJECT
+
+WORKDIR /app
+
+COPY . .
+
+RUN turbo prune "${PROJECT}" --docker
+
+# BUILDER stage
+# ------------------------------------------------------
+FROM base AS builder
+
+ARG PROJECT
+ARG PROJECT_PATH
+ARG DD_GIT_COMMIT_SHA
+
+ENV RELEASE_VERSION=${DD_GIT_COMMIT_SHA:-unknown}
+
+WORKDIR /app
+
+COPY --from=pruner /app/out/pnpm-lock.yaml ./pnpm-lock.yaml
+COPY --from=pruner /app/out/pnpm-workspace.yaml ./pnpm-workspace.yaml
+COPY --from=pruner /app/out/json/ .
+
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install \
+  --shamefully-hoist \
+  --frozen-lockfile \
+  --ignore-scripts \
+  --filter "${PROJECT}..."
+
+COPY --from=pruner /app/out/full/ .
+
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store \
+  BUILDING_CONTAINER=true \
+  pnpm turbo build --filter="${PROJECT}..."
+
+RUN rm -fr node_modules
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install \
+  --prod \
+  --frozen-lockfile \
+  --shamefully-hoist \
+  --filter "${PROJECT}..."
+
+# RUNNER stage
+# ------------------------------------------------------
+FROM node:22-alpine AS runner
+
+RUN apk add --update --no-cache curl
+
+ARG PROJECT_PATH
+ARG DD_GIT_REPOSITORY_URL
+ARG DD_GIT_COMMIT_SHA
+ARG PORT=8080
+
+ENV PORT=$PORT
+ENV NODE_ENV=production
+ENV DD_GIT_REPOSITORY_URL=${DD_GIT_REPOSITORY_URL:-}
+ENV DD_GIT_COMMIT_SHA=${DD_GIT_COMMIT_SHA:-}
+ENV RELEASE_VERSION=${DD_GIT_COMMIT_SHA:-}
+
+EXPOSE $PORT
+
+RUN addgroup --system --gid 1001 nodejs
+RUN adduser --system --uid 1001 latitude
+RUN set -e; \
+  mkdir -p /app/storage/files; \
+  mkdir -p /app/apps/web/public/files; \
+  chown -R 1001:1001 /app/storage/files /app/apps/web/public/files
+USER latitude
+
+WORKDIR /app
+
+COPY --from=builder --chown=latitude:nodejs /app/node_modules ./node_modules
+COPY --from=builder --chown=latitude:nodejs /app/${PROJECT_PATH} ./${PROJECT_PATH}
+COPY --from=builder --chown=latitude:nodejs /app/packages/telemetry/typescript ./packages/telemetry/typescript
+COPY --from=builder --chown=latitude:nodejs /app/packages/core/src/assets/eu-central-1-bundle.pem ./packages/core/src/assets/eu-central-1-bundle.pem
+
+WORKDIR /app/${PROJECT_PATH}
+
+CMD ["node", "dist/server.js"]
--- a/apps/temporal-workers/tsconfig.json
+++ b/apps/temporal-workers/tsconfig.json
@@ -1,0 +1,13 @@
+{
+  "extends": "@latitude-data/typescript-config/base.json",
+  "compilerOptions": {
+    "moduleResolution": "Bundler",
+    "strict": true,
+    "outDir": "./dist",
+    "baseUrl": ".",
+    "types": ["node"],
+    "jsx": "preserve"
+  },
+  "include": ["src/**/*.ts"],
+  "exclude": ["node_modules"]
+}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,11 @@ services:
     volumes:
       - ./docker/init-db.sh:/docker-entrypoint-initdb.d/init-db.sh
       - db-data:/var/lib/postgresql/data
+    healthcheck:
+      test: ['CMD-SHELL', 'pg_isready -U $${POSTGRES_USER:-latitude}']
+      interval: 5s
+      timeout: 5s
+      retries: 5
 
   redis:
     image: redis:7-alpine
@@ -71,11 +76,41 @@ services:
         soft: 262144
         hard: 262144
     healthcheck:
-      test: ["CMD", "wget", "--spider", "-q", "http://localhost:8123/ping"]
+      test: ['CMD', 'wget', '--spider', '-q', 'http://localhost:8123/ping']
       interval: 10s
       timeout: 5s
       retries: 3
       start_period: 10s
+
+  temporal:
+    image: temporalio/auto-setup:1.25.2
+    restart: always
+    environment:
+      - DB=postgres12
+      - DB_PORT=5432
+      - POSTGRES_USER=${POSTGRES_USER:-latitude}
+      - POSTGRES_PWD=${POSTGRES_PASSWORD:-secret}
+      - POSTGRES_SEEDS=db
+      - DBNAME=temporal
+      - DYNAMIC_CONFIG_FILE_PATH=/etc/temporal/dynamicconfig/development.yaml
+    ports:
+      - '7233:7233'
+    volumes:
+      - ./docker/temporal/dynamicconfig:/etc/temporal/dynamicconfig
+    depends_on:
+      db:
+        condition: service_healthy
+
+  temporal-ui:
+    image: temporalio/ui:2.31.2
+    restart: always
+    environment:
+      - TEMPORAL_ADDRESS=temporal:7233
+      - TEMPORAL_CORS_ORIGINS=http://localhost:3000
+    ports:
+      - '8088:8080'
+    depends_on:
+      - temporal
 
   web:
     image: ghcr.io/latitude-dev/web:latest
@@ -119,6 +154,21 @@ services:
       - db
       - redis
       - weaviate
+
+  temporal-workers:
+    image: ghcr.io/latitude-dev/temporal-workers:latest
+    restart: on-failure
+    env_file:
+      - .env
+    environment:
+      TEMPORAL_ADDRESS: temporal:7233
+    volumes:
+      - shared-storage:/app/storage/files
+      - shared-storage:/app/apps/web/public/files
+    depends_on:
+      - temporal
+      - db
+      - redis
 
   websockets:
     image: ghcr.io/latitude-dev/websockets:latest
--- a/docker/init-db.sh
+++ b/docker/init-db.sh
@@ -5,4 +5,5 @@ psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" <<-EOSQL
     CREATE DATABASE latitude_production;
     CREATE DATABASE latitude_development;
     CREATE DATABASE latitude_test;
+    CREATE DATABASE temporal;
 EOSQL
--- a/packages/core/src/services/experiments/startWithTemporal.ts
+++ b/packages/core/src/services/experiments/startWithTemporal.ts
@@ -1,0 +1,91 @@
+import { eq } from 'drizzle-orm'
+import { LatitudeError } from '../../lib/errors'
+import { Result } from '../../lib/Result'
+import Transaction, { PromisedResult } from '../../lib/Transaction'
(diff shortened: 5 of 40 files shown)


apps/temporal-workers/docker/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+WORKDIR /app
+
+COPY --from=builder --chown=latitude:nodejs /app/node_modules ./node_modules
+COPY --from=builder --chown=latitude:nodejs /app/${PROJECT_PATH} ./${PROJECT_PATH}
+COPY --from=builder --chown=latitude:nodejs /app/packages/telemetry/typescript ./packages/telemetry/typescript
+COPY --from=builder --chown=latitude:nodejs /app/packages/core/src/assets/eu-central-1-bundle.pem ./packages/core/src/assets/eu-central-1-bundle.pem
+
+WORKDIR /app/${PROJECT_PATH}
+
+CMD ["node", "dist/server.js"]
```
geclos, line 100: run this locally to ensure it workz please

apps/temporal-workers/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "@latitude-data/typescript-config/base.json",
+  "compilerOptions": {
+    "moduleResolution": "Bundler",
+    "strict": true,
+    "outDir": "./dist",
+    "baseUrl": ".",
+    "paths": {
+      "$/*": ["./src/*"],
+      "acorn": ["node_modules/@latitude-data/typescript-config/types/acorn"]
```
geclos: do we need this

docker/init-db.sh
```diff
@@ -5,4 +5,5 @@ psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" <<-EOSQL
     CREATE DATABASE latitude_production;
     CREATE DATABASE latitude_development;
     CREATE DATABASE latitude_test;
+    CREATE DATABASE temporal;
```
geclos, line 8: add a custom drizzle migration to create the database, is temporal gonna use the same user/password than the rest of apps to access this db?

packages/core/src/services/experiments/startWithTemporal.ts
```diff
@@ -0,0 +1,96 @@
...
+          failed: 0,
+          errors: 0,
+          totalScore: 0,
+        },
+      },
+    },
+  })
+
+  try {
+    await startWorkflow(EXPERIMENT_WORKFLOW_NAME, {
```
geclos: this is fire and forget i imagine? it won't fail if the workflow fails to execute.

geclos: you are gonna have to revert the transaction operation if this fails ☝🏼

packages/core/src/temporal/activities/document/run.ts
```diff
@@ -0,0 +1,89 @@
...
+  const commit = await commitsRepository
+    .getCommitByUuid({ uuid: params.commitUuid, projectId: params.projectId })
+    .then((r) => r.unwrap())
+
+  const experimentsRepository = new ExperimentsRepository(workspace.id)
+  const experiment = await experimentsRepository
+    .find(params.experimentId)
+    .then((r) => r.unwrap())
+
+  const documentsRepository = await import('../../../repositories').then(
```
geclos: k ase

packages/core/src/temporal/activities/document/run/handler.ts
```diff
@@ -0,0 +1,89 @@
...
+      success: true,
+      conversationUuid: result.uuid,
+      messages,
+    }
+  } catch (error) {
+    captureException(error as Error)
+    return {
+      success: false,
+      error: (error as Error).message,
+    }
```
geclos, line 94: I imagine this is the expected api from temporal to manage failures?

packages/core/src/temporal/activities/experiment/complete.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export async function completeExperimentActivity(
+  params: CompleteExperimentParams,
+): Promise<void> {
+  const experimentsRepository = new ExperimentsRepository(params.workspaceId)
+  const experiment = await experimentsRepository
+    .find(params.experimentId)
+    .then((r) => r.unwrap())
+
+  await completeExperiment({ experiment }).then((r) => r.unwrap())
```
geclos: here we throw? i a bit confused about error handling in activities right now

packages/core/src/temporal/proxies/document.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const defaultRetry = {
+  maximumAttempts: 3,
+  initialInterval: '1s',
+  maximumInterval: '30s',
+  backoffCoefficient: 2,
+} as const
+
+export const { runDocumentActivity, simulateTurnActivity } =
+  proxyActivities<typeof activities>({
```
geclos: aren't you bundling all activity exported from the barrel file here even if you only return. I'd prefer it if we avoid barrel files and instead import what you really need

```
runDocumentActivity, simulateTurnActivity
```

packages/core/src/temporal/proxies/document.ts
```diff
@@ -0,0 +1,17 @@
+import { proxyActivities } from '@temporalio/workflow'
+import type * as activities from '../activities'
```
geclos: i'd prefer if we avoided barrel file exports

packages/core/src/temporal/workflows/experiment.ts
```diff
@@ -0,0 +1,325 @@
...
+            total: progress.total,
+            completed: progress.completed,
+            passed: progress.passed,
+            failed: progress.failed,
+            errors: progress.errors,
+            totalScore: progress.totalScore,
+          },
+        })
+        return
+      }
```
geclos, line 217: crazy how beautiful this is 😂

packages/core/src/temporal/workflows/experiment.ts
```diff
@@ -0,0 +1,325 @@
...
+      errors: progress.errors,
+      totalScore: progress.totalScore,
+    },
+  })
+
+  return {
+    success: !cancelled && progress.errors === 0,
+    progress,
+  }
+}
```
geclos, line 324: a cool feature of temporal is that workflows can be stopped/resumed, is there some kind of state to control where in the workflow iteration we stopped so that when we resume we continue from that iteration point?

docker-compose.yml
```diff
@@ -71,12 +76,42 @@ services:
...
+    image: temporalio/ui:2.31.2
+    restart: always
+    environment:
+      - TEMPORAL_ADDRESS=temporal:7233
+      - TEMPORAL_CORS_ORIGINS=http://localhost:3000
+    ports:
+      - '8088:8080'
+    depends_on:
+      - temporal
+
```
geclos, line 114: what does this do? do we need it to deploy it to production as well?

docker/init-db.sh
andresgutgon: Why? This will happen only once in our infra and in development this script takes care of it no? How drizzle is related with this DB?

docker-compose.yml
andresgutgon: Would be awesome is like bullmq UI so ideally yes

andresgutgon: For sure there is an auth recipe somewhere

docker/init-db.sh
geclos: enterprise

andresgutgon: But with Drizzle? This db is not managed with drizzle

packages/core/src/temporal/workflows/experiment.ts
csansoon: yup, everything is stored in db

packages/core/src/temporal/workflows/experiment.ts
csansoon: i know 😭

apps/temporal-workers/tsconfig.json
csansoon: Actually, nope. Removing them!

docker/init-db.sh
csansoon: 👆

packages/core/src/services/experiments/startWithTemporal.ts
csansoon: Hmmm yeah. I can move the `startWorkflow` inside of the transaction. It should just check that the job has been successfully queued before applying the transaction.

packages/core/src/temporal/activities/document/run.ts
csansoon: okay what the fuck was that.
fixed.

packages/core/src/temporal/activities/document/run/handler.ts
csansoon: In Temporal, "Activities" (jobs) return an object, which are used when the workflow is continued. If the error was thrown, the activity would have been retried as per the retry policy.

Ideally we would use our TypedResult object, but Temporal requires the returned object to be serializable.

packages/core/src/temporal/activities/experiment/complete.ts
csansoon: This error is retryable.
It's not the same when a document run fails (we return the error), than when we try to mark an experiment as completed (here just the db didn't work for some reason, we can just retry before continuing)

packages/core/src/temporal/proxies/document.ts
csansoon: Barrel files are kind of required by Temporal's design. 

There's another option, more similar to what we have in ROUTES.
The actual activity code and the proxy must be defined in separate files. Then, the proxy must import the activity ONLY AS type, and the workflow imports the proxy.

So, instead of having this (what I designed initially)
```
temporal/
├── activities/
│   └── document/
│       └── run.ts
└── proxies/
    └── document.ts
```

we could have this:
```
temporal/
├── activities/
│   └── document/
│       └── run/
│           ├── activity.ts
│           └── proxy.ts

```

And then just import the exact proxy we want from the workflow.

packages/core/src/temporal/proxies/document.ts
csansoon: (talked in a different comment)

packages/core/src/temporal/proxies/document.ts
geclos: too complex, let's go with barrel files then

docker/init-db.sh
csansoon: I could add it anyways ig

packages/core/src/temporal/workflows/index.ts
```diff
@@ -0,0 +1,50 @@
...
+ * - Event-sourced: All state changes are recorded and can be replayed
+ *
+ * ## Determinism Constraints
+ *
+ * Workflows are replayed from history on resume. To ensure consistent replay:
+ *
+ * ❌ FORBIDDEN:
+ * - `Math.random()`, `Date.now()`, `new Date()`
+ * - `setTimeout`, `setInterval`
+ * - Direct I/O (database, network, file system)
```
geclos: you are reading from db in some of your workflows

docker/init-db.sh
geclos: we *need* to add it, enterprise customer with already deployed database need that db to be created and it's just better if we do it for them

packages/core/src/temporal/workflows/experiment.ts
geclos: what happens in this workflow that is doing N iterations if the job fails for whatever reason in iteration 4 out of 10?

apps/temporal-workers/docker/Dockerfile
csansoon: workz 👌

docker/init-db.sh
csansoon: Actually I think it cannot be done in a Drizzle migration:
 - I cannot use `IF NOT EXISTS`, which would make the migration fail for all projects that have already created the database from the `init-db.sh`.
 - `CREATE DATABASE` cannot run inside a postgres transaction, which Drizzle use for migrations.

packages/core/src/temporal/proxies/document.ts
csansoon: I actually thought of this approach first, and I kind of like it. It's like how we have both `route.ts` and `handler.ts` files for the gateway and api. It's similar here: we could add the `handler.ts` and `proxy.ts`. The proxy contains the job configuration: queue, retry config, timeout...

I decided to move everything to different folders so that we could reuse the same proxy for all jobs within the same queue though, just to make it look more like our current approach. But I actually like the other option too!

packages/core/src/temporal/workflows/experiment.ts
csansoon: The workflow function gets executed again from start. The cool thing is that on iterations 1-3, when the code should call the "Activity" handler, it just returns the cached result from before instead. So the orchestration logic just keeps running in the exact same way as the first attempt until it gets to the iteration 4 again.

This is why the workflow cannot have non-deterministic code like `Math.random()` or `Date.now()`. I added comments explaining these constraints on `packages/core/src/temporal/workflows/index.ts` so that we can have them present when coding.

packages/core/src/temporal/workflows/index.ts
csansoon: No there's only 1 workflow defined: `experiment`, and it is using Activities, calling them through Proxies, to fetch data. The workflow itself does not fetch any data.

1. Define Activity to fetch data: `fetchExperimentDataActivity`
2. Create a Proxy for this Activity
3. Import and call the Proxy from the Workflow

This way, when the workflow code gets re-executed (this happens every time an activity is called), it will obtain the results from the cached state instead of re-running the fetching.

geclos: ah ok fetchExperimentData does not have the Activity suffix that's what confused me, add it and we are golden 👌🏼

packages/core/src/temporal/workflows/experiment.ts
geclos: awesome

docker/init-db.sh
geclos: you can escape out of drizzle's transaction with a head COMMIT; statement and use sql to create the database conditionally, sql is turing complete!

```
SELECT format('CREATE DATABASE %I', 'your_db')
WHERE NOT EXISTS (
  SELECT 1 FROM pg_database WHERE datname = 'your_db'
)
```

geclos (dismissed):

packages/core/src/temporal/workflows/index.ts
csansoon: whoops

docker-compose.yml
csansoon: It might not be required if we're using enterprise, idk

PR #2227, opened 2026-02-04
feat: more type improvements

geclos:
Remove vercel dependency, replace enums with union string types.


andresgutgon (commented):
Have you checked prompts with OpenAI tools like search?

csansoon (approved):
much much better 🙏

PR #2231, opened 2026-02-05
feat: span fetching timing out in projects with 100k+ spans

geclos:


packages/core/src/repositories/spansRepository.ts
```diff
@@ -295,16 +295,6 @@ export class SpansRepository extends Repository<Span> {
     experimentUuids?: string[]
     createdAt?: { from?: Date; to?: Date }
   }) {
-    // Fetch commit UUIDs - this should be fast with proper commit indexes
-    // FIXME: This is wrong use CommitRepository.getCommitsHistory
```
andresgutgon, line 299: This is picking up all spans of the project at all time. It does not respect the versioning system I think

geclos: that is fine, we want all spans from the project in this method

PR #2235, opened 2026-02-05
feat: remove offending workspace spans

geclos:


packages/core/drizzle/0266_lovely_talon.sql
```diff
@@ -0,0 +1,24 @@
...
+--> statement-breakpoint
+
+DO $$
+DECLARE
+  deleted_count INTEGER;
+  total_deleted INTEGER := 0;
+BEGIN
+  LOOP
+    WITH del AS (
+      SELECT ctid
```
andresgutgon, line 11: Is this ok `ctid`?

geclos: yes it's just a postgres system column to reference the primary key

PR #2236, opened 2026-02-05
Add trigger settings to control which conversation interactions get evaluated

csansoon:
## Summary
Introduces trigger configuration for evaluations to control which interaction within a conversation gets evaluated. Previously, all interactions were evaluated. Now users can choose to evaluate only the first interaction, every interaction (default), or the last interaction in a conversation.

For live evaluations with "last interaction" selected, a debounce mechanism ensures the evaluation only runs after a configurable period of inactivity (default 120 seconds), preventing premature evaluation while the conversation is still active.

Includes a migration to backfill existing evaluations with the default trigger setting (`target: 'every'`).

### DIFF
--- a/apps/web/src/components/evaluations/EvaluationV2Form.tsx
+++ b/apps/web/src/components/evaluations/EvaluationV2Form.tsx
@@ -2,11 +2,15 @@ import type { ICommitContextType } from '$/app/providers/CommitProvider'
 import { ActionErrors } from '$/hooks/useLatitudeAction'
 import {
   CompositeEvaluationMetric,
+  DEFAULT_LAST_INTERACTION_DEBOUNCE_SECONDS,
   EvaluationMetric,
   EvaluationOptions,
   EvaluationSettings,
+  EvaluationTriggerTarget,
   EvaluationType,
   HumanEvaluationMetric,
+  LAST_INTERACTION_DEBOUNCE_MAX_SECONDS,
+  LAST_INTERACTION_DEBOUNCE_MIN_SECONDS,
   LlmEvaluationMetric,
   RuleEvaluationMetric,
 } from '@latitude-data/constants'
@@ -26,6 +30,7 @@ import {
   ConfigurationSimpleForm,
 } from './ConfigurationForm'
 import { EVALUATION_SPECIFICATIONS } from './index'
+import { IconName } from '@latitude-data/web-ui/atoms/Icons'
 
 /**
  * This can be improved by passing specific schemas per type/metric
@@ -39,6 +44,8 @@ type EvaluationV2FormSchema = StandardSchemaV1<{
   options: string
   settings: string
   evaluateLiveLogs: string
+  'trigger.target': string
+  'trigger.lastInteractionDebounce': string
 }>
 
 export type EvaluationV2FormErrors = ActionErrors<EvaluationV2FormSchema>
@@ -253,6 +260,17 @@ export default function EvaluationV2Form<
             description='You will be able to manually evaluate responses in the runs/logs dashboard or via the API/SDK'
           />
         )}
+
+        <TriggerSettings
+          configuration={settings.configuration}
+          setConfiguration={(value) =>
+            setSettings({ ...settings, configuration: value })
+          }
+          evaluateLiveLogs={!!options.evaluateLiveLogs}
+          errors={errors}
+          disabled={disabled || commitMerged}
+        />
+
         {settings.type !== EvaluationType.Composite && (
           <CollapsibleBox
             title='Advanced configuration'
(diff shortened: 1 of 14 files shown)


juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **2 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>The "first" trigger logic may encounter duplicate evaluations during out-of-order log ingestion</strong>
   - [Location](https://github.com/latitude-dev/latitude-llm/blob/dcb3ddb455c9b78e86d18d89d5e5d0fcb45395d1/packages/core/src/events/handlers/evaluateLiveLog.ts#L133-L140): `packages/core/src/events/handlers/evaluateLiveLog.ts:L133-L140`
   - Impact: If a secondary span arrives before the first, the system might incorrectly trigger multiple evaluations for the same conversation.
   - Suggestion: Consider adding an idempotency check or enforcing strict ordering in the ingestion pipeline to ensure only one evaluation occurs.

2. <strong>The debounce mechanism for the "last" trigger could face race conditions under high load</strong>
   - [Location](https://github.com/latitude-dev/latitude-llm/blob/dcb3ddb455c9b78e86d18d89d5e5d0fcb45395d1/packages/core/src/events/handlers/evaluateLiveLog.ts#L159-L168): `packages/core/src/events/handlers/evaluateLiveLog.ts:L159-L168`
   - Impact: Concurrent events might prevent the latest payload from being scheduled, potentially causing the final response to be missed.
   - Suggestion: You might want to wrap the remove-and-add job sequence in a transaction or implement a retry logic when a job ID conflict is detected.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

apps/web/src/components/evaluations/EvaluationV2Form.tsx
```diff
@@ -295,6 +309,44 @@ export default function EvaluationV2Form<
                     />
                   )}
                 </FormFieldGroup>
+                {metricSpecification?.supportsLiveEvaluation &&
+                  !!options.evaluateLiveLogs &&
+                  settings.configuration.trigger?.target === 'last' && (
+                    <FormFieldGroup
```
neoxelox, line 319: Why is this separated from the TriggerSettings component?

apps/web/src/components/evaluations/EvaluationV2Form.tsx
```diff
@@ -295,6 +309,44 @@ export default function EvaluationV2Form<
...
+                                ...settings.configuration.trigger,
+                                lastInteractionDebounce: isNaN(value)
+                                  ? DEFAULT_LAST_INTERACTION_DEBOUNCE_SECONDS
+                                  : value,
+                              },
+                            },
+                          })
+                        }}
+                        className='w-full'
+                        disabled={disabled}
```
neoxelox, line 350: you are missing the `errors={errors?.["trigger.lastInteractionDebounce"]}`

apps/web/src/components/evaluations/EvaluationV2Form.tsx
csansoon: The `target` is a simple configuration: "what is evaluated". The rest of the configuration is under the "Advanced" section

packages/core/src/events/handlers/evaluateLiveLog.ts
```diff
@@ -28,6 +34,22 @@ const LIVE_EVALUABLE_LOG_SOURCES = Object.values(LogSources).filter(
     source !== LogSources.Optimization,
 ) as LogSources[]
 
+function getTriggerConfig(evaluation: EvaluationV2): TriggerConfiguration {
+  return (
+    evaluation.configuration.trigger ?? {
+      target: 'every',
```
neoxelox: shouldnt it be `last`?

packages/core/src/jobs/job-definitions/evaluations/runEvaluationForExperimentJob.ts
```diff
@@ -42,25 +75,44 @@ export async function runEvaluationForExperimentJob(
...
   const spansRepo = new SpansRepository(workspaceId)
 
-  const traceId = await spansRepo.getLastTraceByLogUuid(conversationUuid)
-  if (!traceId) {
-    if (shouldRetry(job.attemptsStarted)) {
-      const delay = calculateExponentialBackoff(job.attemptsStarted)
-      job.moveToDelayed(Date.now() + delay, token)
-      throw new DelayedError('Waiting for trace to show up')
+  const allSpans = await spansRepo.listByDocumentLogUuid(conversationUuid)
+  const promptSpans = allSpans.filter((span) => span.type === SpanType.Prompt)
```
neoxelox: only prompt spans? what about chat follow ups

packages/core/src/events/handlers/evaluateLiveLog.ts
csansoon: yup

packages/core/src/jobs/job-definitions/evaluations/runEvaluationForExperimentJob.ts
csansoon: good catch!

PR #2240, opened 2026-02-06
Fix experiments run from logs

andresgutgon:
# What?
When picking logs to run an experiment, we were using documentLogs. Now is using spans. 

Also, when generating experiments metadata, we also were using document logs

## TODO
- [x] QA it works
- [x] Add test fixes

### DIFF
--- a/packages/core/src/repositories/experimentsRepository.ts
+++ b/packages/core/src/repositories/experimentsRepository.ts
@@ -1,29 +1,26 @@
 import {
   and,
   count,
+  countDistinct,
   desc,
   eq,
   getTableColumns,
   inArray,
-  isNull,
   sql,
   sum,
 } from 'drizzle-orm'
 
-import { ExperimentScores } from '@latitude-data/constants'
+import { ExperimentScores, SpanType } from '@latitude-data/constants'
 import { omit } from 'lodash-es'
-import { ErrorableEntity } from '../constants'
 import { LatitudeError, NotFoundError } from '../lib/errors'
 import { Result } from '../lib/Result'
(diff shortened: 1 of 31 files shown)


packages/core/src/repositories/experimentsRepository.ts
```diff
@@ -277,107 +273,60 @@ export class ExperimentsRepository extends Repository<Experiment> {
...
-    )
+  ): PromisedResult<ExperimentRunMetadata, LatitudeError> {
+    const promptTraceIds = this.db
+      .selectDistinct({ traceId: spans.traceId })
+      .from(spans)
+      .where(
+        and(
+          eq(spans.workspaceId, this.workspaceId),
+          eq(spans.experimentUuid, uuid),
+          eq(spans.type, SpanType.Prompt),
```
csansoon: Not just of Type Prompt, Chat can also be included in an experiment.
Just remove this type condition, and get the traceId from ANY span with that experimentUuid. Most will be null anyways and you'll just get the Prompt and Chat spans

andresgutgon: Done

packages/core/src/services/documents/tools/resolve/agentsAsTools.ts
```diff
@@ -111,7 +111,9 @@ export async function resolveAgentAsToolDefinition({
         })
 
         const usage = await runUsage
+        const cost = await runCost
         streamManager.incrementRunUsage(usage)
+        streamManager.incrementRunCost(cost)
```
csansoon, line 116: nice!

PR #2242, opened 2026-02-06
Implement trial end in backoffice

andresgutgon:
# What? 
Add ability to modifial trials in backoffice

### DIFF
--- a/apps/web/src/app/(private)/layout.tsx
+++ b/apps/web/src/app/(private)/layout.tsx
@@ -32,10 +32,13 @@ export default async function PrivateLayout({
 }>) {
   const { workspace, user, subscriptionPlan } = await getCurrentUserOrRedirect()
 
-  const completed = await isOnboardingCompleted()
+  const trialEnded = subscriptionPlan.trialInfo?.trialEnded === true
 
-  if (!completed) {
-    redirect(ROUTES.onboarding.root)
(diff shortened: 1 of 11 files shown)


apps/web/src/app/(private)/layout.tsx
```diff
@@ -32,10 +32,13 @@ export default async function PrivateLayout({
 }>) {
   const { workspace, user, subscriptionPlan } = await getCurrentUserOrRedirect()
 
-  const completed = await isOnboardingCompleted()
+  const trialEnded = subscriptionPlan.trialInfo?.trialEnded === true
 
-  if (!completed) {
-    redirect(ROUTES.onboarding.root)
+  if (!trialEnded) {
```
andresgutgon, line 37: This was wrong if trial is ended there is no point on checking onboarding

apps/web/src/stores/availableWorkspaces.ts
```diff
@@ -40,10 +36,9 @@ export default function useAvailableWorkspaces() {
         return
       }
 
-      navigate.push(ROUTES.root)
-      refreshCurrentWorkspace()
+      window.location.href = ROUTES.root
```
andresgutgon, line 39: Redirect enters an infinite loop when the workspace trial ended. Forcing a full page refresh solve it

PR #2244, opened 2026-02-06
Fix traces clicking

andresgutgon:
# What?
Fix issues clicking on traces and sub-traces and closing the row

## TODO
- [x] 🐛 Clicking the sub-trace close the row
- [x] Improve evaluations column query performance.

### DIFF
--- a/packages/core/drizzle/0268_add_workspace_trace_id_index_to_eval_results.sql
+++ b/packages/core/drizzle/0268_add_workspace_trace_id_index_to_eval_results.sql
@@ -1,0 +1,4 @@
+COMMIT;
+--> statement-breakpoint
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "evaluation_results_v2_workspace_trace_id_idx" ON "latitude"."evaluation_results_v2" USING btree ("workspace_id","evaluated_trace_id");--> statement-breakpoint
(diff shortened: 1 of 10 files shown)


packages/core/drizzle/0268_add_workspace_trace_id_index_to_eval_results.sql
```diff
@@ -0,0 +1,4 @@
+COMMIT;
+--> statement-breakpoint
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS "evaluation_results_v2_workspace_trace_id_idx" ON "latitude"."evaluation_results_v2" USING btree ("workspace_id","evaluated_trace_id");--> statement-breakpoint
```
andresgutgon, line 4: This should improve evaluations column performance query on traces table.

packages/core/src/repositories/evaluationResultsV2Repository.ts
```diff
@@ -521,16 +521,11 @@ export class EvaluationResultsV2Repository extends Repository<EvaluationResultV2
     }
 
     const results = await this.db
-      .select({
-        ...tt,
-        commitUuid: commits.uuid,
-      })
+      .select(tt)
       .from(evaluationResultsV2)
-      .innerJoin(commits, eq(commits.id, evaluationResultsV2.commitId))
```
andresgutgon, line 529: Avoid join. At this point N of results is small so we do the query to commits

geclos (approved):

PR #2247, opened 2026-02-09
fix(gateway): rate limit middleware rejects invalid API keys

geclos:
Datadog issue: https://app.datadoghq.eu/error-tracking?query=&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22235fb29a-9bd4-11f0-94ac-da7ad0900005%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&from_ts=1770539419390&to_ts=1770625819390&live=true

Problem
- Gateway rate limit middleware could crash when the Authorization token is malformed / not a UUID, due to a DB query receiving an invalid value.

Fix
- Validate that the Bearer token is a UUID (api_keys.token is a Postgres UUID) before looking it up.
- Return 401 Unauthorized with a clear error message.
- Added unit test to ensure we fail fast and never hit the DB for malformed tokens.

Validation
- pnpm lint
- pnpm tc (after building @latitude-data/sdk and @latitude-data/telemetry to generate dist types)}

### DIFF
--- a/apps/gateway/package.json
+++ b/apps/gateway/package.json
@@ -33,6 +33,7 @@
     "lodash-es": "4.17.21",
     "promptl-ai": "catalog:",
     "rate-limiter-flexible": "5.0.3",
+    "uuid": "11.1.0",
     "zod": "catalog:"
   },
   "devDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -203,6 +203,9 @@ importers:
       rate-limiter-flexible:
         specifier: 5.0.3
         version: 5.0.3
+      uuid:
+        specifier: 11.1.0
+        version: 11.1.0
       zod:
         specifier: 'catalog:'
         version: 4.1.8


apps/gateway/src/middlewares/rateLimit/index.ts
```diff
@@ -51,11 +51,21 @@ async function getTokenRateLimit(token: string): Promise<{
   }
 }
 
+function isUuid(value: string) {
+  // api_keys.token is stored as a Postgres UUID.
+  // Validate before hitting the DB so malformed values (or odd proxy/header bugs)
+  // don't blow up the rate limit middleware.
+  return /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i.test(
```
geclos: @claude can we use an existing library instead of reimplementing uuid identification manually?

geclos: @claude do it

PR #2248, opened 2026-02-09
fix(gateway): handle invalid version uuid in run routes

geclos:
Problem
- Gateway could crash when a malformed version UUID (e.g. `invalid-uuid`) is passed in the /run route path, because we end up querying commits by uuid with an invalid value.

Stack trace (from Datadog)
- CommitsRepository.getCommitByUuid → getProjectByVersionData → getData

Fix
- Validate `commitUuid` before calling CommitsRepository.getCommitByUuid.
- Return BadRequestError (400) when the version uuid is malformed.
- Added unit test to ensure malformed uuid fails fast and does not hit the DB.

Validation
- pnpm lint
- pnpm tc (after building @latitude-data/sdk and @latitude-data/telemetry to generate dist types)}


apps/gateway/src/common/documents/getData.ts
```diff
@@ -11,6 +11,12 @@ import { getDocumentMetadata } from '@latitude-data/core/services/documents/scan
 import { documentPresenterWithProviderAndMetadata } from '$/presenters/documentPresenter'
 import { Workspace } from '@latitude-data/core/schema/models/types/Workspace'
 
+function isUuid(value: string) {
+  return /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i.test(
+    value,
+  )
+}
```
geclos: TODO: use the isValidUuid method implemented in https://github.com/latitude-dev/latitude-llm/pull/2247/changes#r2781337883

PR #2249, opened 2026-02-09
Add sampling rate setting to evaluation triggers

csansoon:
Adds a configurable sampling rate to evaluation trigger settings, allowing users to reduce the percentage of eligible responses that are evaluated. Previously, all matching responses were evaluated unconditionally. Now users can set a rate between 1-100% (defaulting to 100%) to control evaluation frequency, useful for reducing cost and load on high-traffic prompts.

Sampling is applied during live evaluation only — batch experiment evaluations remain deterministic and always evaluate all selected spans. The sampling check runs after trigger target filtering (first/every/last) so it composes naturally with existing trigger logic.

image


apps/web/src/components/evaluations/ActualOutputForm.tsx
```diff
@@ -0,0 +1,311 @@
...
+    setConfiguration({ ...configuration, actualOutput, expectedOutput })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [actualFormatIsAccessible, expectedFormatIsAccessible])
+
+  const [testConfiguration] = useDebounce(configuration.actualOutput, 750)
+  const [showTest, setShowTest] = useState(false)
+
+  return (
+    <>
+      <SelectableSwitch
```
neoxelox: this is not actual output config e

apps/web/src/components/evaluations/ConfigurationForm/OutputConfiguration.tsx
```diff
@@ -0,0 +1,311 @@
...
+          )}
+        </FormFieldGroup>
+      </FormFieldGroup>
+      {showTest && !!configuration.actualOutput && (
+        <div className='w-full flex flex-col gap-2 border-t-2 border-dashed border-border pt-4'>
+          <ActualOutputTest configuration={testConfiguration} />
+        </div>
+      )}
+      {requiresExpectedOutput && (
+        <FormFieldGroup
```
neoxelox, line 228: this is not actual output config either

apps/web/src/components/evaluations/ActualOutputForm.tsx
neoxelox: i think you could bring back this selectable switch to the configuration form component

apps/web/src/components/evaluations/ConfigurationForm/OutputConfiguration.tsx
```diff
@@ -0,0 +1,311 @@
...
+  const expectedFormatIsAccessible = useMemo(
+    () =>
+      !!configuration.expectedOutput?.parsingFormat &&
+      ACCESSIBLE_OUTPUT_FORMATS.includes(
+        configuration.expectedOutput.parsingFormat,
+      ),
+    [configuration.expectedOutput?.parsingFormat],
+  )
+
+  useEffect(() => {
```
andresgutgon, line 74: Beautiful useEffect

apps/web/src/components/evaluations/ConfigurationForm/OutputConfiguration.tsx
neoxelox: rename this component to OutputForm i guess

apps/web/src/components/evaluations/ConfigurationForm/OutputConfiguration.tsx
```diff
@@ -0,0 +1,311 @@
...
+            required
+          />
+        </FormFieldGroup>
+        <FormFieldGroup
+          label='Parsing format'
+          description='How to parse the assistant messages. Stringification is done deterministically'
+          layout='horizontal'
+          tooltip={
+            actualFormatIsAccessible ? (
+              <Text.H6 color='background'>
```
andresgutgon, line 170: Use `<>` we don't need to style text. It's done by the tooltip

apps/web/src/components/evaluations/ConfigurationForm/OutputConfiguration.tsx
```diff
@@ -0,0 +1,311 @@
...
+                onChange={(e) =>
+                  setConfiguration({
+                    ...configuration,
+                    expectedOutput: {
+                      ...(configuration.expectedOutput || {}),
+                      fieldAccessor: e.target.value || undefined,
+                    },
+                  })
+                }
+                errors={errors?.['expectedOutput.fieldAccessor']}
```
andresgutgon, line 285: This is a bit weird

csansoon: @neoxelox

packages/core/drizzle/0269_evaluation_trigger_sample_rate.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE "latitude"."evaluation_versions"
+SET configuration = jsonb_set("configuration", '{trigger,sampleRate}', '100'::jsonb)
```
andresgutgon, line 2: This is a default? Why not just set the default in code if this attribute is not found?

apps/web/src/components/evaluations/EvaluationOptionsForm.tsx
```diff
@@ -0,0 +1,130 @@
...
+  settings: EvaluationSettings<T, M>
+  setSettings: (settings: EvaluationSettings<T, M>) => void
+  supportsLiveEvaluation: boolean
+  errors?: Record<string, string[]>
+  disabled?: boolean
+}) {
+  return (
+    <>
+      <FormFieldGroup label='Options' layout='vertical'>
+        {supportsLiveEvaluation && (
```
neoxelox: move this if outside because evals that do not support live eval will see the group by the label "Options" empty

apps/web/src/components/evaluations/ConfigurationForm/TriggerConfiguration.tsx
```diff
@@ -0,0 +1,130 @@
...
+                  ...settings.configuration.trigger,
+                  sampleRate: value[0],
+                },
+              },
+            })
+          }}
+          min={MIN_EVALUATION_SAMPLE_RATE}
+          max={MAX_EVALUATION_SAMPLE_RATE}
+          step={1}
+          disabled={disabled}
```
neoxelox, line 158: you missing `errors?.["trigger.sampleRate"]`

packages/core/src/events/handlers/evaluateLiveLog.ts
```diff
@@ -139,6 +141,15 @@ export const evaluateLiveLogJob = async ({
       if (!isFirst) continue
     }
 
+    const sampleRate =
+      triggerConfig.sampleRate ?? DEFAULT_EVALUATION_SAMPLE_RATE
+    if (
+      sampleRate < DEFAULT_EVALUATION_SAMPLE_RATE &&
+      Math.random() * 100 >= sampleRate
```
neoxelox: i think it should be only greater than, otherwise 100 is not >= 100 (i know the other condition but it does not make sense also 70 >= 70, because then its not [0, 70] but [0,70) aja 69.99999999% sampling rate instead of 70% sampling rate).

neoxelox (approved):
some comments

apps/web/src/components/evaluations/ConfigurationForm/OutputConfiguration.tsx
neoxelox: this is how it works idk, its to show the zod erros beautifully for every field... otherwise they are not toasted so...

apps/web/src/components/evaluations/ConfigurationForm/TriggerConfiguration.tsx
csansoon: The Slider component doesn't have this attribute. Also, it wouldn't make much sense either, there is no way to input the Slider value incorrectly lol

packages/core/src/events/handlers/evaluateLiveLog.ts
csansoon: Yeah that makes more sense 👌

apps/web/src/components/evaluations/ConfigurationForm/OutputConfiguration.tsx
csansoon: Restructured them better!

apps/web/src/components/evaluations/ActualOutputForm.tsx
csansoon: Restructured them better!

PR #2255, opened 2026-02-09
fix: capture gateway errors consistently

geclos:
## Summary
- ensure gateway error capture creates a span when no active request span exists
- normalize non-Error throwables and capture >=500s in the error handler
- capture SSE stream errors consistently instead of filtering unknowns

## Testing
- not run (not requested)


geclos:
no point in adding unit tests for this

PR #2256, opened 2026-02-09
feat: partition eval results table

geclos:
Eval results is huge and simple index scans with low N (1800 results) take ages to complete. I'm partitioning the table by time and workspace id so that queries are fast again.

Some considerations:
- I'm removing most composite indices from the new table to see if, once partitioned, table queries are performant enough with single-column indices
- uuid uniqueness has to be dropped because postgres requires the partition key to be part of the unique constraint. Regardless, the chance of uuid collisions is so small we can safely remove this constraint.


packages/core/src/jobs/job-definitions/maintenance/manageEvaluationResultsPartitionsJob.ts
```diff
@@ -0,0 +1,136 @@
...
+  for (const row of partitions.rows) {
+    const partitionName = row.partition_name as string
+    const dateStr = partitionName.replace(`${TABLE}_p`, '')
+
+    if (dateStr < formatDate(cutoffDate)) {
+      try {
+        await database.execute(
+          sql.raw(`
+          ALTER TABLE "${SCHEMA}"."${TABLE}"
+          DETACH PARTITION "${SCHEMA}"."${partitionName}"
```
andresgutgon, line 79: Can you explain what's doing this? It's removing the data?

geclos: nope it's detaching it from the parent partition table. When you query evaluation_results_v2, which will be partitioned, postgres will scan only the partitions that are attached to the parent table, detaching it effectively removes the partitions from the "available data to scan".

andresgutgon: So nothing needs to be changed in code, right? We keep consuming original eval results?

andresgutgon: I guess exports will be harder now

geclos: we won't support it for a while, we need to implement a two-tiered storage system with cold + hot storage, ideally using clickhouse rather than postgres, which has automatic backup into s3 cold storage so the dsl does not change

packages/core/drizzle/0271_natural_omega_sentinel.sql
```diff
@@ -0,0 +1,31 @@
+-- Commenting because this migration has already been applied in the previous
```
andresgutgon: So why this migration is needed?

geclos: indices are droped created in a specific order for performance reasons, i do not want the automated migration drizzle gives us in this case.

andresgutgon: Ok but you could squash the migrations in 2

PR #2258, opened 2026-02-10
Enforce trial in gateway

andresgutgon:
Move the check to the gateway boundary

### DIFF
--- a/apps/gateway/src/middlewares/trialCheck.ts
+++ b/apps/gateway/src/middlewares/trialCheck.ts
@@ -1,0 +1,70 @@
+import { Context, Next } from 'hono'
+import { checkPayingOrTrial } from '@latitude-data/core/lib/checkPayingOrTrial'
+import { OpenAPIHono } from '@hono/zod-openapi'
+
+declare module 'hono' {
+  interface ContextVariableMap {
+    skipTrialCheck?: boolean
+  }
+}
+
+const SKIP_TRIAL_CHECK_PATHS = new Set<string>()
+
+/**
+ * Resets the skip paths. Only used for testing.
+ * @internal
+ */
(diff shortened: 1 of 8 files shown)


apps/gateway/src/middlewares/trialCheck.ts
```diff
@@ -0,0 +1,68 @@
...
+import { checkPayingOrTrial } from '@latitude-data/core/lib/checkPayingOrTrial'
+import { OpenAPIHono } from '@hono/zod-openapi'
+
+declare module 'hono' {
+  interface ContextVariableMap {
+    skipTrialCheck?: boolean
+  }
+}
+
+const SKIP_TRIAL_CHECK_PATHS = new Set<string>()
```
geclos, line 11: we are running it on all routes?

geclos: also i'd prefer it if we declare the middleware in the route subset where it's used, rather than maintaining a hardcoded whitelist of routes. @claude check if this is possible, and how it would look if we did it.

geclos: @andresgutgon if you only need the trial check middleware in the run handler put the middleware where appropiate and remove it from top-level routes ☝🏼

andresgutgon: That's the question. Do we want to allow calling the gateway when trial ends? This way is op-out. If we discover in the future we need to allow that is just add that route. But now all gateway is trial aware

andresgutgon: I like option 3

andresgutgon:
@claude review please

PR #2266, opened 2026-02-10
Phase 1: ClickHouse spans table, client, and dual-write ingestion

geclos:
## Summary

- **ClickHouse migration**: Creates `spans` table (both clustered and unclustered variants) with `ReplacingMergeTree(ingested_at)`, partitioned by month, ordered by `(workspace_id, started_at, trace_id, span_id)` with bloom filter indexes
- **Client wrapper**: Lazy-initialized singleton `clickhouseClient()` using `@clickhouse/client` with connection pooling and async inserts, plus `clickhouseHealthCheck()`
- **Feature flags**: Adds `clickhouse-spans-write` and `clickhouse-spans-read` to `GLOBAL_FEATURES` (both default disabled)
- **ClickHouseSpansWriter**: Maps processed spans to ClickHouse rows, extracting `provider` from `CompletionSpanMetadata`
- **Dual-write in `processSpansBulk`**: After Postgres transaction succeeds and events are published, conditionally writes to ClickHouse (fire-and-forget, errors captured but never propagated)

Postgres remains the authoritative source. No read path changes in this PR.

## Test plan

- [x] `pnpm --filter @latitude-data/core ch:up` applies the migration successfully
- [x] `pnpm --filter @latitude-data/core ch:status` shows migration applied
- [x] `pnpm --filter @latitude-data/core ch:connect` → `SHOW CREATE TABLE spans` verifies schema
- [x] `pnpm tc` passes
- [x] `pnpm lint` passes (0 errors)
- [x] Existing span ingestion tests pass without regression
- [x] Enable `clickhouse-spans-write` flag for a workspace and verify rows land in ClickHouse
- [x] Verify ClickHouse write failures don't affect Postgres ingestion

### DIFF
--- a/.github/workflows/run-migrations.yml
+++ b/.github/workflows/run-migrations.yml
@@ -110,11 +110,25 @@ jobs:
           DATABASE_URL: postgresql://${{ secrets.DB_USERNAME }}:${{ secrets.DB_PASSWORD }}@127.0.0.1:${{ secrets.DB_PORT }}/${{ secrets.DB_NAME }}
           NODE_ENV: production
 
-      - name: Install golang-migrate
+      - name: Cache golang-migrate
+        id: cache-migrate
         if: ${{ secrets.CLICKHOUSE_MIGRATION_URL != '' }}
+        uses: actions/cache@v3
+        with:
+          path: ~/bin/migrate
+          key: ${{ runner.os }}-golang-migrate-v4.18.1
+
+      - name: Install golang-migrate
+        if: ${{ secrets.CLICKHOUSE_MIGRATION_URL != '' && steps.cache-migrate.outputs.cache-hit != 'true' }}
         run: |
+          mkdir -p ~/bin
           curl -L https://github.com/golang-migrate/migrate/releases/download/v4.18.1/migrate.linux-amd64.tar.gz | tar xvz
-          sudo mv migrate /usr/local/bin/migrate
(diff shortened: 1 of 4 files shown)


.github/workflows/run-migrations.yml
```diff
@@ -98,6 +110,25 @@ jobs:
           DATABASE_URL: postgresql://${{ secrets.DB_USERNAME }}:${{ secrets.DB_PASSWORD }}@127.0.0.1:${{ secrets.DB_PORT }}/${{ secrets.DB_NAME }}
           NODE_ENV: production
 
+      - name: Install golang-migrate
+        if: ${{ secrets.CLICKHOUSE_MIGRATION_URL != '' }}
+        run: |
+          curl -L https://github.com/golang-migrate/migrate/releases/download/v4.18.1/migrate.linux-amd64.tar.gz | tar xvz
+          sudo mv migrate /usr/local/bin/migrate
```
geclos: @claude can we cache this install so that we don't perform it on every ci run?

geclos: @claude are you there?

geclos:
> Are we not moving away from S3 for metadata, are we?

We are not. Moving the metadata into clickhouse is a bad idea as it would increase query times and skyrocket prices (db storage pricing is much larger than object storage pricing).

PR #2270, opened 2026-02-11
Consolidate read logic: replace ProjectsRepository and UsersRepository with query functions

geclos:
## Summary

- Add `scopedQuery` / `unscopedQuery` wrappers that enforce a consistent `(filters, db?)` API for all read functions, with `workspaceId` inside the filters object for scoped queries
- Migrate all callers of `ProjectsRepository` (~44 files) and `UsersRepository` (~8 files) to use direct query function calls
- Delete `ProjectsRepository` and `UsersRepository` classes entirely
- Create new query functions: `findAllProjects`, `findAllWorkspaceUsers`, `findWorkspaceUserById`
- Update 5 test files to mock query functions instead of repository prototypes

## Test plan

- [x] `pnpm tc` passes across all 11 packages
- [x] `pnpm lint` passes across all 11 packages
- [x] `pnpm test` — full test suite

### DIFF
--- a/rfcs/scoped-queries-migration-rfc.md
+++ b/rfcs/scoped-queries-migration-rfc.md
@@ -1,0 +1,273 @@
+# RFC: Migrate to Scoped/Unscoped Queries and Remove Repositories/Data-Access Layers
+
+Status: Draft
+Owners: Core Platform
+Target: packages/core first, then apps/web and apps/gateway
+
+## Problem
+
+We still have legacy repositories and ad-hoc data-access layers mixed with the new
+query functions. This causes duplicated access patterns, inconsistent tenancy
+handling, and makes changes harder to trace or enforce. We need a unified read
+layer based on scoped/unscoped query functions and to remove repositories and
+data-access layers entirely.
+
+## Goals
+
+- Make query functions the single source of truth for reads.
+- Enforce workspace tenancy at the query boundary via scopedQuery.
+- Replace repositories and data-access layers across the codebase.
+- Standardize query function signatures and naming.
+- Reduce cognitive overhead by consolidating read logic.
+
+## Non-Goals
+
+- Changing the write service layer pattern (Result/Transaction).
+- Overhauling schema models or migrations.
+- Re-architecting services or job orchestration.
+
+## Current State (Recent Commits)
+
+- Introduced function-based query abstraction in packages/core/src/queries with
+  scopedQuery/unscopedQuery, migrated projects/users as proof of concept.
+- Removed ProjectsRepository and UsersRepository and swapped ~65 call sites to
+  query functions.
+- Refactored project queries to return raw rows and aligned call sites.
+- Fixed users query tests and tightened behavior.
+
+## Proposed Architecture
+
+- All read operations live under packages/core/src/queries/.
+- Two entrypoints:
+  - scopedQuery(fn) for workspace-scoped reads ({ workspaceId, ...filters }).
+  - unscopedQuery(fn) for global reads.
+- Query modules are small, single-purpose exported functions (no classes).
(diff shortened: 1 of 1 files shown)


apps/gateway/src/routes/api/v3/projects/getAll/getAll.handler.ts
```diff
@@ -1,11 +1,9 @@
 import { Context } from 'hono'
-import { ProjectsRepository } from '@latitude-data/core/repositories'
+import { findAllActiveProjects } from '@latitude-data/core/queries/projects/findAllActive'
 
 export const getAllHandler = async (c: Context) => {
   const workspace = c.get('workspace')
-  const projectsRepository = new ProjectsRepository(workspace.id)
-  const projectsResult = await projectsRepository.findAllActive()
-  const projects = projectsResult.unwrap()
+  const result = await findAllActiveProjects({ workspaceId: workspace.id })
```
andresgutgon: On one hand is nice in the end points to do `.unwrap` because error handler takes care of it. But when using the queries inside services is annoying to deal with the result. Little feedback from using it. Maybe is worth it the tradeoff

geclos: yeah i agree we can take the opportunity to remove result patterns from read methods

apps/gateway/src/common/documents/getData.ts
```diff
@@ -21,15 +21,14 @@ async function getProjectByVersionData({
...
-  const projectsScope = new ProjectsRepository(workspace.id)
   const commitsScope = new CommitsRepository(workspace.id)
 
   const pid = Number(projectId)
   if (isNaN(pid)) {
     return Result.error(new BadRequestError(`Invalid project id ${projectId}`))
   }
 
-  const projectResult = await projectsScope.getProjectById(pid)
+  const projectResult = await findProjectById({ workspaceId: workspace.id, id: pid })
```
andresgutgon: Allowing workspace Id could be a security concern. Why not enforcing workspace as before?

geclos: why a security concern? repositories already worked with workspace id

andresgutgon: That's true, forget it

packages/core/src/queries/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export { unsafelyFindUserById } from './users/findById'
+export {
+  unsafelyFindUserByEmail,
+  unsafelyFindUserIdByEmail,
+} from './users/findByEmail'
+export { findAllWorkspaceUsers } from './users/findAllInWorkspace'
+export { findWorkspaceUserById } from './users/findInWorkspace'
+export { findFirstUserInWorkspace } from './users/findFirstInWorkspace'
+export { lockUser } from './users/lock'
```
andresgutgon: no barrels no?

packages/core/src/queries/users/findAllInWorkspace.ts
```diff
@@ -0,0 +1,23 @@
...
+export const findAllWorkspaceUsers = scopedQuery(
+  async function findAllWorkspaceUsers(
+    { workspaceId }: { workspaceId: number },
+    db,
+  ): Promise<TypedResult<User[]>> {
+    const result = await db
+      .select(workspaceTt)
+      .from(users)
+      .innerJoin(memberships, eq(memberships.userId, users.id))
+      .where(tenancyFilter(workspaceId))
```
andresgutgon, line 20: How is tenancy enforced?

geclos: with the tenancy filter

PR #2273, opened 2026-02-11
Deprecate old logs gateway endpoint

andresgutgon:
# What?
We discovered a bug on old logs endpoint where the projectId was missing. I fixed this bug but we're deprecating this endpoint in gateway auto generated docs


apps/gateway/src/routes/api/v3/projects/versions/documents/logs/create.handler.ts
```diff
@@ -112,6 +112,7 @@ async function createSpansFromLogData({
       documentLogUuid,
       documentUuid: document.documentUuid,
       commitUuid: commit.uuid,
+      projectId: commit.projectId,
```
andresgutgon, line 115: This was the bug

csansoon: damn

PR #2274, opened 2026-02-11
Send manual traces backfilling

andresgutgon:
# What?
We already have an endpoint to ingest telemetry otp spans. But is not human friendly. The idea is to implement a new endpoint with a signature that's easy to understand so users can build this format from their existing logs and send to us so we can transform this data into telemetry manual spans. 

## TODO
- [ ] Implement example and agree on the format
- [ ] Implement gateway endpoint
- [ ] Implement sdk wrapper method. This is just for sending from SDK. But raw API would be easy enough ideally
- [ ] Allow rosseta to convert messages. 

## IDEA
* Allow to send single turn traces. User ask -> llm respond
* Allow to send multi turn traces (aka conversations). User ask -> llm respond -> user ask (chat) -> llm respond -> ...

### DIFF
--- a/examples/src/sdk/ingest-trace/example.ts
+++ b/examples/src/sdk/ingest-trace/example.ts
@@ -18,11 +18,11 @@ type BackfillToolCall = {
 
 type BackfillLog = {
   conversationId: string
-  previousTraceId?: string
   path: string
   provider: string
   model: string
-  input: Array<{ role: string; content: string }>
+  system?: string                                    // Optional system prompt (for Anthropic, etc.)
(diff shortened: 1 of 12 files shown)


examples/src/sdk/ingest-trace/example.ts
```diff
@@ -0,0 +1,246 @@
...
+  completedAt: Date
+}
+
+type BackfillLog = {
+  conversationId: string
+  previousTraceId?: string
+  path: string
+  provider: string
+  model: string
+  input: Array<{ role: string; content: string }>
```
andresgutgon: Better type this

PR #2277, opened 2026-02-12
Fix annotation failing for traces without a document UUID

csansoon:
## Summary
Annotating a trace would fail with `Invalid data: {"documentUuid": ["Invalid input: expected string, received null"]}` when the span being annotated had no `documentUuid` (e.g., traces sent via the SDK/API without a document context).

The annotation flow was sourcing `documentUuid` from the span, which is optional and can be null. It now sources it from the evaluation, which is always scoped to a document and always has a valid `documentUuid`.


geclos:
this does not fix the issue, chat/external/prompt span types all have document uuid parameters so if they don't there's an underlying issue here and letting users try to annotate that span will just fail

PR #2279, opened 2026-02-12
clickhouse migration 2: migrate evaluation results

geclos:
## Summary
- Add ClickHouse migrations and dual-write plumbing for spans/evaluation results with feature flags and CI migration wiring


andresgutgon (commented):
I know all is a prototype but we're building the fundamentals without any tests

PR #2282, opened 2026-02-12
fix: key integration tools by toolId to allow same tool name across i…

learningbizz:
Tools were keyed only by tool name, so multiple integrations exposing the same tool (e.g. two Gmail MCPs with create_draft) overwrote each other in the merged tool dict and only one was usable.

Config already uses toolIds like integrationName/toolName. This change keys integration tools by that full toolId in lookup and resolve, and passes only the short tool name to the MCP when executing. The model now sees distinct tools (e.g. gmail_personal/create_draft and gmail_work/create_draft) and both can be used in the same prompt.

Proof of working:
image

### DIFF
--- a/packages/core/src/lib/streamManager/ChainStreamConsumer/consumeStream.test.ts
+++ b/packages/core/src/lib/streamManager/ChainStreamConsumer/consumeStream.test.ts
@@ -473,6 +473,7 @@ describe('consumeStream', () => {
         sourceData: {
           source: ToolSource.Integration,
           integrationId: 42,
+          toolName: 'integrationTool',
           toolLabel: 'My Integration',
           imageUrl: 'https://example.com/image.png',
         },
@@ -508,6 +509,7 @@ describe('consumeStream', () => {
         _sourceData: {
           source: ToolSource.Integration,
           integrationId: 42,
+          toolName: 'integrationTool',
           toolLabel: 'My Integration',
           imageUrl: 'https://example.com/image.png',
         },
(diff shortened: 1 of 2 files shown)


geclos (dismissed):

packages/core/src/services/documents/tools/resolve/integrationTools.ts
```diff
@@ -28,6 +28,10 @@ export async function resolveIntegrationToolDefinition({
   if (integrationResult.error) return integrationResult
   const integration = integrationResult.unwrap()
 
+  const shortToolName = toolName.includes('/')
+    ? toolName.split('/').slice(1).join('/')
+    : toolName
+
```
csansoon: tool names cannot contain '/' afaik

packages/core/src/services/documents/tools/lookup/integrationTools.ts
```diff
@@ -135,7 +135,7 @@ export async function lookupIntegrationTools({
         )
       }
 
-      resolvedTools[toolName] = tool
+      resolvedTools[toolId] = tool
```
csansoon: tool names cannot contain '/' afaik

packages/constants/src/toolSources.ts
```diff
@@ -29,6 +29,7 @@ type AgentAsToolSourceData = BaseToolSourceData<ToolSource.Agent> & {
 
 type IntegrationToolSourceData = BaseToolSourceData<ToolSource.Integration> & {
   integrationId: number
+  toolName: string
   toolLabel?: string // Human-readable tool label
   imageUrl?: string // For Pipedream integrations
 }
```
csansoon, line 35: Pretty nice, I like itt

PR #2284, opened 2026-02-12
Allow access to backoffice in enterprise mode

andresgutgon:
# What?
☝️ That

### DIFF
--- a/packages/core/src/services/users/setupService.ts
+++ b/packages/core/src/services/users/setupService.ts
@@ -1,6 +1,7 @@
 import { type User } from '../../schema/models/types/User'
-import { type Workspace } from '../../schema/models/types/Workspace'
+import { type WorkspaceDto } from '../../schema/models/types/Workspace'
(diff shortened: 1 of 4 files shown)


packages/core/src/services/users/setupService.ts
```diff
@@ -36,10 +37,16 @@ export default async function setupService(
...
+): PromisedResult<{ user: User; workspace: WorkspaceDto }> {
   return transaction.call(async () => {
     const user = await createUser(
-      { email, name, confirmedAt: new Date(), title },
+      {
+        email,
+        name,
+        confirmedAt: new Date(),
+        title,
+        admin: env.LATITUDE_ENTERPRISE_MODE,
```
geclos: is this envvar coerced to boolean?

andresgutgon: Yes https://github.com/latitude-dev/latitude-llm/blob/main/packages/env/src/index.ts#L332

PR #2290, opened 2026-02-13
Fix evaluation results leaking across drafts

geclos:
Fixes https://linear.app/latitude/issue/LAT-392/evaluation-results-from-other-drafts-seen-in-current-draft

## What
- Scope evaluation results list/count/pagination API routes to the commit referenced in the route params, unless an explicit `commitIds` filter is provided.
- Add a regression test covering two drafts with the same evaluation.
- Remove the hard compile-time dependency on `@latitude-data/sdk` types from `@latitude-data/telemetry` (use structural typing), and exclude telemetry tests from `pnpm tc`.

## Test plan
- `pnpm lint`
- `pnpm --filter @latitude-data/sdk build` (needed for workspace type resolution)
- `pnpm tc`
- `pnpm --filter @latitude-data/web test -- src/app/api/projects/[projectId]/commits/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationUuid]/results/route.test.ts`


geclos:
@claude you have permission to commit and push

PR #2304, opened 2026-02-16
Fix last result in traces panel showing always last result for trace even for sub-agents

andresgutgon:
# What?

When a trace has multiple calls to sub-agents if you click on the sub-agent, the field for the last result still shows the last result for the full traces. This was happening for all types. Prompt, Chat, and External. We centralized this logic into a hook that extracts the information from the children passed to the panel, which can be the top span or any sub-span, like a sub-agent.

### DIFF
--- a/apps/web/src/components/tracing/spans/useSpanCompletionData.ts
+++ b/apps/web/src/components/tracing/spans/useSpanCompletionData.ts
@@ -1,0 +1,90 @@
+import { useMemo } from 'react'
+import { useTrace } from '$/stores/traces'
+import {
+  CompletionSpanMetadata,
+  SpanType,
+} from '@latitude-data/constants'
+import { findAllSpansOfType } from '@latitude-data/core/services/tracing/spans/fetching/findAllSpansOfType'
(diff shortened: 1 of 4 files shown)


apps/web/src/components/tracing/spans/useSpanCompletionData.ts
```diff
@@ -0,0 +1,93 @@
...
+}: {
+  traceId: string
+  spanId: string
+}) {
+  const { data: trace } = useTrace({ traceId })
+
+  const assembledSpan = useMemo(() => {
+    if (!trace) return undefined
+    return findSpanById(trace.children, spanId)
+  }, [trace, spanId])
```
andresgutgon, line 34: The fix is to pick the actual span assembled.

apps/web/src/components/tracing/spans/useSpanCompletionData.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const completionSpan = useMemo(() => {
+    if (!assembledSpan) return undefined
+
+    return findLastSpanOfType({
+      children: assembledSpan.children,
```
andresgutgon, line 40: Then pick the completion span from the children of that span

PR #2309, opened 2026-02-16
Change properties free trial event

learningbizz:


csansoon (approved):

packages/core/src/queries/users/findWorkspaceCreator.ts
```diff
@@ -0,0 +1,25 @@
...
+  ): Promise<User | null> {
+    if (!workspace.creatorId) return null
+    const rows = await db
+      .select()
+      .from(users)
+      .where(eq(users.id, workspace.creatorId))
+      .limit(1)
+    return (rows[0] as User) ?? null
+  },
+)
```
geclos, line 25: :feelsgood:

learningbizz: what is that emoji lol

learningbizz: is it a face bleeding xdd

PR #2323, opened 2026-02-17
add clickhouse rebuild migrations for evaluation results and spans keys

geclos:
## Summary

Rebuilding the clickhouse tables with better primary keys and indeces to optimize for our query patterns. We can add projections later on.

### DIFF
--- a/packages/core/clickhouse/migrations/clustered/0005_rebuild_evaluation_results_sort_key.up.sql
+++ b/packages/core/clickhouse/migrations/clustered/0005_rebuild_evaluation_results_sort_key.up.sql
@@ -1,6 +1,5 @@
 CREATE TABLE IF NOT EXISTS evaluation_results_rebuild_0005 ON CLUSTER default (
   id UInt64,
-  uuid UUID,
   workspace_id UInt64,
 
   project_id UInt64,
@@ -49,7 +48,6 @@ SETTINGS index_granularity = 8192;
 
 INSERT INTO evaluation_results_rebuild_0005 (
   id,
-  uuid,
   workspace_id,
   project_id,
   commit_uuid,
@@ -77,7 +75,6 @@ INSERT INTO evaluation_results_rebuild_0005 (
 )
 SELECT
   id,
-  uuid,
   workspace_id,
   project_id,
   commit_uuid,
--- a/packages/core/clickhouse/migrations/unclustered/0005_rebuild_evaluation_results_sort_key.up.sql
+++ b/packages/core/clickhouse/migrations/unclustered/0005_rebuild_evaluation_results_sort_key.up.sql
@@ -1,6 +1,5 @@
 CREATE TABLE IF NOT EXISTS evaluation_results_rebuild_0005 (
   id UInt64,
-  uuid UUID,
   workspace_id UInt64,
 
   project_id UInt64,
@@ -49,7 +48,6 @@ SETTINGS index_granularity = 8192;
 
 INSERT INTO evaluation_results_rebuild_0005 (
   id,
-  uuid,
   workspace_id,
   project_id,
   commit_uuid,
@@ -77,7 +75,6 @@ INSERT INTO evaluation_results_rebuild_0005 (
 )
 SELECT
   id,
-  uuid,
   workspace_id,
   project_id,
   commit_uuid,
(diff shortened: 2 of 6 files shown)


packages/core/clickhouse/migrations/unclustered/0006_rebuild_spans_sort_key_by_project.up.sql
```diff
@@ -0,0 +1,133 @@
...
+  trace_id FixedString(32),
+  span_id FixedString(16),
+  parent_id Nullable(FixedString(16)),
+  previous_trace_id Nullable(FixedString(32)),
+
+  api_key_id UInt64,
+  document_log_uuid Nullable(UUID),
+  document_uuid Nullable(UUID),
+  commit_uuid Nullable(UUID),
+  commit_uuid_key UUID MATERIALIZED ifNull(commit_uuid, toUUID('00000000-0000-0000-0000-000000000000')),
```
andresgutgon, line 12: So how works this? Why is needed both `commit_uuid` and `commit_uuid_key`?

geclos: nullable values are bad for primary key columns, specially if there are many of them

packages/core/clickhouse/migrations/clustered/0005_rebuild_evaluation_results_sort_key.up.sql
```diff
@@ -0,0 +1,107 @@
...
+  created_date Date MATERIALIZED toDate(created_at),
+  has_error UInt8 MATERIALIZED if(isNull(error), 0, 1),
+
+  INDEX idx_evaluated_span_trace (evaluated_span_id, evaluated_trace_id)
+    TYPE bloom_filter(0.001)
+    GRANULARITY 1
+)
+ENGINE = ReplicatedReplacingMergeTree(updated_at)
+PARTITION BY toYYYYMM(created_at)
+PRIMARY KEY (workspace_id, project_id, evaluation_uuid, created_at)
```
geclos, line 45: this is the important fix, in olap dbs is very important in what order you  storage your data (order by and pk governed), we make the product decision to optimize for queries that filter results by workpsace, project and evaluation.

Indeces in OLAP are left-biased which means you cannot simply query by eval uuid and expect the same performance than if you query by workspace, project and eval uuid. Always include tenancy queries. If you must non include it, that is ok for now but we might need to implement projections in the future which trades disk space for query performance.

A similar fix is applied to the spans table.

packages/core/clickhouse/migrations/clustered/0006_rebuild_spans_sort_key_by_project.up.sql
```diff
@@ -0,0 +1,134 @@
...
+
+  ingested_at DateTime64(6, 'UTC') DEFAULT now64(6),
+  retention_expires_at DateTime64(6, 'UTC') DEFAULT toDateTime64('2100-01-01 00:00:00', 6, 'UTC'),
+
+  INDEX idx_trace_id trace_id TYPE bloom_filter(0.001) GRANULARITY 1,
+  INDEX idx_span_trace_ids (span_id, trace_id) TYPE bloom_filter(0.001) GRANULARITY 1,
+  INDEX idx_document_log_uuid document_log_uuid TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_experiment_uuid experiment_uuid TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_parent_id parent_id TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_test_deployment_id test_deployment_id TYPE bloom_filter(0.01) GRANULARITY 1
```
geclos, line 45: indeces are important for point-lookups (fetch small-n by Xid) they do not help for large query aggregations, use PK and projections for large data reads

packages/core/src/queries/clickhouse/spans/getExperimentPromptSpansBefore.ts
```diff
@@ -24,7 +24,9 @@ export const getExperimentPromptSpansBefore = scopedQuery(
       SELECT span_id, trace_id
       FROM ${SPANS_TABLE}
       WHERE workspace_id = {workspaceId: UInt64}
+        -- TODO(clickhouse): remove non-_key predicate after key-column rollout.
         AND document_uuid = {documentUuid: UUID}
```
geclos, line 28: when we are done with the migrations i will remove the non-key columns from queries, always use surrogate columns to query spans and eval results by the relevant columns.

packages/core/clickhouse/migrations/unclustered/0006_rebuild_spans_sort_key_by_project.up.sql
andresgutgon: But why not do the `ifNull` in existing `commit_uuid`?

PR #2324, opened 2026-02-17
fix: experiments real-time updates not working

geclos:
## Problem

The issue was that WebSocket events for experiment status updates were sending raw experiment data with progress tracker results, but the frontend expects `ExperimentDto` data as transformed by the repository's `experimentDtoPresenter` function.

This caused a mismatch in the data structure, particularly in the `completed` field calculation. The repository calculates `completed` based on completed evaluation cycles divided by total evaluations, while the progress tracker returns raw counts from Redis.

## Solution

- Create `notifyClientOfExperimentStatus` handler to centralize experiment status notifications  
- Update `updateStatus.ts` to fetch `ExperimentDto` from repository before sending WebSocket events
- Update `stop.ts` to use the same pattern
- Update `stop.test.ts` to mock the repository and notification handler

This ensures that the WebSocket events send data in the same format as the API endpoints, providing consistent data to the frontend and fixing the real-time update issues.

## Related Issue

Fixes LAT-403

## Testing

- [x] pnpm lint passes
- [x] pnpm tc passes
- [ ] Tests updated to reflect new implementation

### DIFF
--- a/packages/core/src/services/experiments/stop.ts
+++ b/packages/core/src/services/experiments/stop.ts
@@ -4,8 +4,9 @@ import { ProgressTracker } from '../../jobs/utils/progressTracker'
 import { LatitudeError } from '../../lib/errors'
 import { Result } from '../../lib/Result'
 import { PromisedResult } from '../../lib/Transaction'
+import { ExperimentsRepository } from '../../repositories'
 import { type Experiment } from '../../schema/models/types/Experiment'
-import { WebsocketClient } from '../../websockets/workers'
+import { notifyClientOfExperimentStatus } from '../../events/handlers/notifyClientOfExperimentStatus'
 import { JOB_FINISHED_STATES } from '../runs/shared'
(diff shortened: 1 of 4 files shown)


packages/core/src/services/experiments/stop.ts
```diff
@@ -68,14 +69,15 @@ export async function stopExperiment({
 
   await progressTracker.cleanup()
 
-  WebsocketClient.sendEvent('experimentStatus', {
+  // Fetch the experiment DTO from the repository to ensure consistent data format
+  const experimentsRepository = new ExperimentsRepository(workspaceId)
+  const experimentDto = await experimentsRepository
+    .findByUuid(updatedExperiment.uuid)
+    .then((r) => r.unwrap())
```
andresgutgon, line 76: This logic of getting the DTO could be inside `notifyClientOfExperimentStatus`

PR #2328, opened 2026-02-18
Fix sub-agent trace URL navigation in annotations

andresgutgon:
## Summary

Fixed an issue where clicking "See complete trace" for a sub-agent span in the Annotations page would navigate to the parent document's traces page instead of the sub-agent's document.

The trace URL now correctly uses the selected span's own document UUID, ensuring sub-agent spans navigate to their respective document's traces page with the correct trace row expanded.

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/AnnotationPanel/index.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/AnnotationPanel/index.tsx
@@ -5,7 +5,6 @@ import { AnnotationsProvider, MessageList } from '$/components/ChatWrapper'
 import {
   HumanEvaluationMetric,
   EvaluationResultV2,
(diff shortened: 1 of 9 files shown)


apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/AnnotationPanel/index.tsx
```diff
@@ -86,13 +73,13 @@ export function AnnotationsPanel({
...
         )}
-        {documentUuid && (
+        {span.documentUuid && (
           <div className='w-full flex justify-center items-center'>
             <Link
               href={buildTraceUrl({
                 projectId: project.id,
                 commitUuid: commit.uuid,
-                documentUuid,
+                documentUuid: span.documentUuid,
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/annotations/_components/AnnotationPanel/index.tsx:82]: This was wrong, was always pointing to the document of the main span. Not the span shown in the panel

PR #2331, opened 2026-02-18
fix: handle span not found race condition with retries

geclos:
## Problem
The 'Span not found' error was caused by a race condition. Spans were written to PostgreSQL in a transaction, but the ClickHouse write was fire-and-forget (not awaited). The spanCreated events were published immediately after the transaction committed, before ClickHouse had the data.

When event handlers tried to read spans from ClickHouse, they failed because the async write hadn't completed yet.

## Solution
Await the ClickHouse write before returning from the transaction callback. The after-commit callback (which publishes events) only runs after the transaction completes, so events are now only published after spans are confirmed in ClickHouse.

## Changes
- processBulk.ts: Changed void bulkCreateClickHouseSpans() to await
- Reverted handler retry changes (wrong approach)

## Verification
- Type check passes

Fixes: Datadog Error Tracking issue 738d1d2e-0728-11f1-b670-da7ad0900005

### DIFF
--- a/packages/core/src/services/tracing/spans/ingestion/processBulk.ts
+++ b/packages/core/src/services/tracing/spans/ingestion/processBulk.ts
@@ -433,40 +433,24 @@ export async function processSpansBulk(
           retentionExpiresAt,
         }))
 
-        void bulkCreateClickHouseSpans(spansForClickHouse)
-          .then((clickhouseResult) => {
-            if (clickhouseResult.error) {
-              captureException(
-                new LatitudeError('ClickHouse bulk span insertion failed'),
-                {
-                  workspaceId: workspace.id,
-                  apiKeyId: apiKey.id,
-                  spansCount: spansForClickHouse.length,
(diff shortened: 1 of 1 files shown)


packages/core/src/services/tracing/spans/ingestion/processBulk.ts
```diff
@@ -433,40 +433,24 @@ export async function processSpansBulk(
...
-                  apiKeyId: apiKey.id,
-                  spansCount: spansForClickHouse.length,
-                  error: String(clickhouseResult.error),
-                },
-              )
-              return
-            }
-
-            captureMessage('ClickHouse bulk span insertion succeeded', 'info', {
+        const clickhouseResult = await bulkCreateClickHouseSpans(spansForClickHouse)
```
andresgutgon, line 436: I guess this is fast enough no?

geclos: ? we are awaiting the op

PR #2336, opened 2026-02-18
fix(gateway): prevent socket listener memory leak on keep-alive connections

cballou:
## Summary

- **Fix memory leak** in `createRequestAbortSignal()` caused by unbounded socket listener accumulation on HTTP keep-alive connections
- Under high throughput (hundreds of req/s), each request added a `socket.once('close', ...)` listener. With keep-alive, the same socket serves many requests, so listeners and their `AbortController` closures accumulated without being GC'd — causing `MaxListenersExceededWarning` and unbounded memory growth
- Uses a `WeakMap<Socket, AbortSignal>` to deduplicate: one listener per socket, shared across all requests on the same connection. The WeakMap entry is automatically cleaned up when the socket is GC'd

## Test plan

- [x] All existing tests pass
- [x] Added tests verifying signal reuse on same socket and distinct signals for different sockets
- [ ] Deploy and verify `MaxListenersExceededWarning` no longer appears under load
- [ ] Monitor gateway memory usage over 24h to confirm leak is resolved

### DIFF
--- a/apps/gateway/src/common/createRequestAbortSignal.test.ts
+++ b/apps/gateway/src/common/createRequestAbortSignal.test.ts
@@ -5,12 +5,12 @@ import { createRequestAbortSignal } from './createRequestAbortSignal'
 
 function createMockSocket() {
   const emitter = new EventEmitter()
-  return {
-    once: vi.fn((event: string, callback: () => void) => {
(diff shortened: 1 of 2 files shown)


cballou:

cballou:
recheck

PR #2343, opened 2026-02-20
Add project-level evaluations for agent builder mode

andresgutgon:
## TODO
- [x] Review empty space in sidebar on the top

image

## Summary
Adds a project-level evaluations page accessible from the sidebar when prompt management is disabled (agent builder mode). This allows users to manage evaluations at the project level without navigating through the document hierarchy.

The evaluations are tied to a synthetic document (`PROJECT_MAIN_DOCUMENT`) that is automatically created when accessing the evaluations page. This reuses the existing evaluations UI components, showing the same blank slate with tutorial video and add functionality when no evaluations exist.

## Bug Fix: Evaluation Results Flash

Fixed a bug where evaluation results would briefly appear and then disappear when loading the evaluation page.

**Root cause:** The server-side page query and client-side API query were using different commit filters:
- Server-side: Queried with `commitIds: undefined`
- API route: Applied `resolveCommitFilterFromUrl()` which adds `commitIds` based on commit history

This caused the initial server-rendered results to include data from all commits, but when the client made the API call, it filtered to only the current commit's history — causing results to disappear.

**Fix:** Apply `resolveCommitFilterFromUrl()` on the server side before querying, and pass the resolved search (with `commitIds` populated) to the client. This ensures both queries use the same filter.

The fix was applied to both:
- Document-level evaluation page (`/projects/[projectId]/versions/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationUuid]`)
- Project-level evaluation page (`/projects/[projectId]/evaluations/[evaluationUuid]`)

## Test plan
- [x] Verify project-level evaluations page loads without flash
- [x] Verify document-level evaluations page loads without flash
- [x] Verify evaluation results persist after initial load
- [x] Verify pagination and filtering work correctly


packages/core/src/services/dualScope/findOrCreateMainDocument.ts
```diff
@@ -0,0 +1,81 @@
...
+) {
+  return transaction.call(async (tx) => {
+    const commitsScope = new CommitsRepository(project.workspaceId, tx)
+    const commits = await commitsScope
+      .filterByProject(project.id)
+      .then((r) => r.unwrap())
+
+    const commit = commits[0]
+
+    if (!commit || commit.mergedAt) {
```
andresgutgon: App without prompt manager assumes no other commit was created, and the one created is in draft. This should be true until migrate old workspaces with commits created. But for new workspaces this should be true. What do you think?

packages/core/src/services/dualScope/findOrCreateMainDocument.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (!Result.isOk(docsResult)) return docsResult
+    const document = docsResult.value.find(
+      (d) => d.path === PROJECT_MAIN_DOCUMENT,
+    )
+
+    if (document) return Result.ok({ document, commit })
+
+    // This will ensure they are in a draft coomit. This will fail if
+    // this project has a `live` commit.
+    const documentResult = await createNewDocument(
```
andresgutgon: This service is a read service but it also takes care of creating the main document if not present

neoxelox (approved):

apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationUuid]/stats/route.ts
```diff
@@ -9,7 +9,7 @@ import { NextRequest, NextResponse } from 'next/server'
 import { evaluationResultsV2SearchFromQueryParams } from '@latitude-data/core/helpers'
 
 import { Workspace } from '@latitude-data/core/schema/models/types/Workspace'
-import { resolveCommitFilterFromUrl } from '../resolveCommitFilterFromUrl'
+import { resolveCommitFilterFromUrl } from '../../../../../../../../../../../lib/resolveCommitFilterFromUrl'
```
Comment (csansoon) [inline apps/web/src/app/api/projects/[projectId]/commits/[commitUuid]/documents/[documentUuid]/evaluations/[evaluationUuid]/stats/route.ts]: There's gotta be a better way to import this

PR #2344, opened 2026-02-20
[WIP] Feat/create leads email system

learningbizz:
Creating a few side effects and improvements to:
- create leads in instantly after a user registers
- create leads in instantly for users that finish the free trial
- send ICPs to slack that are finishing the free trial


packages/core/src/queries/clickhouse/spans/hasAtLeastOneExternalSpan.ts
```diff
@@ -0,0 +1,32 @@
...
+export async function hasAtLeastOneExternalSpan(
+  workspaceId: number,
+  db = database,
+): Promise<boolean> {
+  const useClickHouse = await isClickHouseSpansReadEnabled(workspaceId, db)
+  if (useClickHouse) {
+    const n = await countExternalSpans({ workspaceId }, db)
+    return n >= 1
+  }
+  const result = await db
```
andresgutgon: Why is this on clickhouse?

learningbizz: We save spans on clickhouse no?

andresgutgon: Yes but this is reading from postgresql

PR #2348, opened 2026-02-23
Project Traces page

csansoon:
## Summary

Introduces a project-level Traces page accessible from the sidebar that displays all traces belonging to a project, independently of `documentUuid` or `commitUuid`. Traces are grouped by `traceId` rather than `documentLogUuid`, and no span type filtering is applied, so all trace types (including external telemetry traces without a parent Prompt/Chat span) are shown.

This new page only appears for users with "promptManager" disabled (for now).


packages/core/src/queries/clickhouse/spans/fetchProjectConversations.ts
```diff
@@ -0,0 +1,167 @@
...
+    const params: Record<string, unknown> = {
+      workspaceId,
+      projectId,
+      limit: limit + 1,
+    }
+
+    const conditions = [
+      `workspace_id = {workspaceId: UInt64}`,
+      `project_id = {projectId: UInt64}`,
+      `project_id_key = {projectId: UInt64}`,
```
geclos: commit uuid and document uuid should be in the pk since they are still required in the data model if i ain't mistaken

apps/web/src/stores/projectConversations.ts
```diff
@@ -0,0 +1,179 @@
...
+      handleGoToNextPage,
+      goToPrevPage,
+      reset,
+      mutate,
+      currentCursor,
+    ],
+  )
+}
+
+export type { Conversation }
```
geclos: hmm how is it we need a new store for this isn't it the same than traces page within document but with document uuid being optional?

packages/core/src/data-access/conversations/fetchProjectConversations.ts
```diff
@@ -0,0 +1,322 @@
...
+  const allTime = await fetchProjectConversationsQuery(
+    { ...queryParams, createdAt: undefined },
+    db,
+  )
+
+  return Result.ok<FetchProjectConversationsResult>({
+    ...allTime,
+    didFallbackToAllTime: true,
+  })
+}
```
geclos: move to queries folder, also a good opportunity to move the rest of conversation queries to the queries folder and remove the data-access folder for conversations

packages/core/src/queries/clickhouse/spans/fetchProjectConversations.ts
csansoon: The thing is that here we want to list traces in a project, without filtering by documentUuid (or commitUuid really)

apps/web/src/stores/projectConversations.ts
csansoon: hmmm actually, yeah I'll merge them into s ingle store and api route

packages/core/src/data-access/conversations/fetchProjectConversations.ts
csansoon: sure!

packages/core/src/queries/conversations/fetchConversations.test.ts
```diff
@@ -1311,3 +1348,154 @@ describe('fetchConversations', () => {
...
+      projectId: project.id,
+      filters: {},
+    })
+
+    expect(result.ok).toBe(true)
+    const { items } = result.unwrap()
+    expect(items).toHaveLength(1)
+    expect(items[0]!.documentLogUuid).toBe(documentLogUuid)
+  })
+})
```
andresgutgon: Wait, i'm doing a big change here. We're going to conflict hard

csansoon: 😭

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/traces/_components/ProjectConversationRow.tsx
```diff
@@ -0,0 +1,174 @@
...
+        </TableCell>
+        <TableCell>
+          <EvaluationsColumn conversationId={span.documentLogUuid ?? null} />
+        </TableCell>
+      </TableRow>
+      {isSelected && span.documentLogUuid && (
+        <TableRow hoverable={false}>
+          <TableCell
+            colSpan={999}
+            className='max-w-full w-full h-full !p-0'
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/traces/_components/ProjectConversationRow.tsx:102]: Why these importants?

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/traces/_components/ProjectConversationRow.tsx
```diff
@@ -0,0 +1,174 @@
...
+  ProjectTraceSpanSelectionActionsContext,
+  ProjectTraceSpanSelectionStateContext,
+} from './ProjectTraceSpanSelectionContext'
+import { Span } from '@latitude-data/constants'
+import { Skeleton } from '@latitude-data/web-ui/atoms/Skeleton'
+import { ConversationTimeline } from '../../documents/[documentUuid]/(withTabs)/traces/_components/ConversationTimeline'
+import {
+  TraceSpanSelectionActionsContext,
+  TraceSpanSelectionStateContext,
+} from '../../documents/[documentUuid]/(withTabs)/traces/_components/TraceSpanSelectionContext'
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/traces/_components/ProjectConversationRow.tsx]: Move to `components/Traces` now that this is common

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/traces/_components/ProjectTraceSpanSelectionContext.tsx
```diff
@@ -0,0 +1,91 @@
...
+  )
+
+  return (
+    <ProjectTraceSpanSelectionActionsContext.Provider value={actionsValue}>
+      <ProjectTraceSpanSelectionStateContext.Provider value={stateValue}>
+        {children}
+      </ProjectTraceSpanSelectionStateContext.Provider>
+    </ProjectTraceSpanSelectionActionsContext.Provider>
+  )
+}
```
Comment (andresgutgon) [inline apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/traces/_components/ProjectTraceSpanSelectionContext.tsx]: Maybe DRY context selection if it's the same

apps/web/src/hooks/spanFilters/useProcessSpanFilters.ts
```diff
@@ -212,6 +212,38 @@ export function useProcessSpanFilters({
...
+
+      onFiltersChanged(updatedFilters)
+      setSearchParams(
+        'filters',
+        JSON.stringify(updatedFilters),
+        TRACE_SPAN_SELECTION_PARAMS,
+      )
+    },
+    [filterOptions, onFiltersChanged, setSearchParams],
+  )
```
geclos: what's this

apps/web/src/app/api/conversations/[conversationId]/route.ts
```diff
@@ -36,10 +36,16 @@ export const GET = errorHandler(
...
-        request.nextUrl.searchParams.get('commitUuid') ?? undefined
+        rawCommitUuid && rawCommitUuid !== 'undefined'
+          ? rawCommitUuid
+          : undefined
+      const rawDocumentUuid = request.nextUrl.searchParams.get('documentUuid')
       const documentUuid =
-        request.nextUrl.searchParams.get('documentUuid') ?? undefined
+        rawDocumentUuid && rawDocumentUuid !== 'undefined'
+          ? rawDocumentUuid
+          : undefined
```
Comment (andresgutgon) [inline apps/web/src/app/api/conversations/[conversationId]/route.ts]: Was this causing a bug?

apps/web/src/hooks/spanFilters/useProcessSpanFilters.ts
```diff
@@ -212,6 +212,38 @@ export function useProcessSpanFilters({
     [filterOptions, onFiltersChanged, setSearchParams],
   )
 
+  const onTraceIdChange = useCallback(
```
andresgutgon: Is this specific for project level traces?

packages/core/src/queries/clickhouse/spans/fetchConversation.ts
```diff
@@ -53,28 +54,46 @@ export const fetchConversation = scopedQuery(async function fetchConversation({
...
+          min(started_at) AS trace_started_at,
+          max(ended_at) AS trace_ended_at,
+          anyLast(document_uuid) AS latest_document_uuid,
+          anyLast(source) AS latest_source,
+          anyLast(commit_uuid) AS latest_commit_uuid,
+          anyLast(experiment_uuid) AS latest_experiment_uuid
+        FROM ${TABLE_NAME}
+        WHERE ${whereClause}
+        GROUP BY document_log_uuid, trace_id
+      ) AS traces
```
geclos, line 96: why do you need this? do not merge this pr until we've discussed this

packages/core/src/queries/conversations/fetchConversation.ts
```diff
@@ -61,16 +64,25 @@ export async function fetchConversation(
...
     .from(spans)
     .where(
       and(
         eq(spans.workspaceId, workspace.id),
         inArray(spans.traceId, traceIdsSubquery),
       ),
     )
-    .groupBy(spans.documentLogUuid)
+    .groupBy(spans.documentLogUuid, spans.traceId)
+    .as('traces')
```
geclos, line 77: ditto

apps/web/src/services/routes/api.ts
```diff
@@ -360,6 +360,9 @@ export const API_ROUTES = {
         traceIds: {
           root: `${conversationRoot}/trace-ids`,
         },
+        evaluations: {
+          root: `${conversationRoot}/evaluations`,
```
andresgutgon, line 364: Review i think we're not doing this. Review traces document

apps/web/src/stores/conversations/useConversation.ts
```diff
@@ -71,3 +74,36 @@ export function useConversation(
     [data, mutate, isLoading],
   )
 }
+
+export function useConversationEvaluations({
```
andresgutgon, line 78: Ditto. No need I think

packages/core/src/queries/conversations/shared.ts
```diff
@@ -0,0 +1,86 @@
...
+    string | null
+  >`(array_agg(${spans.experimentUuid} ORDER BY ${spans.startedAt} DESC) FILTER (WHERE ${spans.experimentUuid} IS NOT NULL))[1]`.as(
+    'experiment_uuid',
+  ),
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type TracesSubquery = Record<string, any>
+
+export function buildConversationAggregateFields(traces: TracesSubquery) {
```
andresgutgon, line 44: This type is a bit lol but ok

apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/traces/_components/ProjectConversationRow.tsx
Comment (csansoon) [reply apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/traces/_components/ProjectConversationRow.tsx:102]: Taken from `ProjectConversation`, these make the "row" of the table that contains the conversation timeline take the full width of the table.

apps/web/src/hooks/spanFilters/useProcessSpanFilters.ts
csansoon: This was made through the assumption that some traces would not have a `documentLogUuid` at all, so it depended on `traceId` to mark selection.

apps/web/src/app/api/conversations/[conversationId]/route.ts
Comment (csansoon) [reply apps/web/src/app/api/conversations/[conversationId]/route.ts]: Oh actually it's not happening anymore, this happened while I was doing some tests. No need to add it

apps/web/src/hooks/spanFilters/useProcessSpanFilters.ts
csansoon: This change depended on the assumption that not all rows in the "traces" table would have a `documentLogUuid`, and thus we depended on `traceId` for some. This has been removed since

packages/core/src/queries/clickhouse/spans/fetchConversation.ts
csansoon: Since we now cannot depend on "Main" spans, getting the duration of a session is more tricky.

I have changed it into a two-level aggregation:
 1. Inner query, where it groups spans by `trace_id`. Here, the timing is aggregated as a wall-clock time (dateDiff between min started_at and max ended_at)
 2. Outer query. Here, these are aggregated by `document_log_uuid`, and the past aggregation is added together.

I made this because none of the aggregation methods work by themselves: I can't just use the first `started_at` and the last `ended_at` from a session, because there can be time between traces of the same session that must not be accounted for.

I cannot just SUM durations either, because some spans from the same trace happen in parallel.

So this is the approach I ended with

PR #2351, opened 2026-02-23
Improve legacy POST /logs

neoxelox:


apps/gateway/src/routes/api/v3/projects/versions/documents/logs/create.handler.ts
```diff
@@ -352,7 +339,7 @@ async function createSpansFromLogData({
   await publishSpanCreated({
     spanId: promptSpanId,
     commitUuid: commit.uuid,
-    traceId,
+    traceId: traceId,
```
andresgutgon, line 342: Super necessary change

andresgutgon (approved):

PR #2352, opened 2026-02-23
fix: prevent duplicate document creation under concurrent requests

andresgutgon:
## Summary

Concurrent requests could create duplicate documents with the same path and commit because PostgreSQL treats NULL values as distinct in unique indexes. When `deletedAt` is NULL (non-deleted documents), the existing unique constraint on `(path, commit_id, deleted_at)` did not prevent duplicates.

This change:
- Adds `NULLS NOT DISTINCT` to the unique constraint, ensuring non-deleted documents with the same path and commit are properly rejected
- Includes a data migration that renames existing duplicate documents by appending `_1`, `_2`, etc. to preserve data integrity before applying the stricter constraint

### DIFF
--- a/packages/core/drizzle/0272_fix_duplicated_documents_in_commit.sql
+++ b/packages/core/drizzle/0272_fix_duplicated_documents_in_commit.sql
@@ -1,0 +1,30 @@
+ALTER TABLE "latitude"."published_documents" DISABLE ROW LEVEL SECURITY;--> statement-breakpoint
+DROP TABLE "latitude"."published_documents" CASCADE;--> statement-breakpoint
+
+-- Step 1: Fix duplicate documents by renaming them
+-- For each set of duplicates (same path, commit_id, deleted_at IS NULL),
+-- keep the first one (by id) with original path, rename others with _1, _2, etc.
+WITH duplicates AS (
+  SELECT
+    id,
+    path,
+    commit_id,
+    ROW_NUMBER() OVER (PARTITION BY path, commit_id ORDER BY id) as rn
+  FROM latitude.document_versions
+  WHERE deleted_at IS NULL
+),
+to_rename AS (
+  SELECT id, path, rn
+  FROM duplicates
+  WHERE rn > 1
+)
+UPDATE latitude.document_versions dv
+SET path = tr.path || '_' || (tr.rn - 1)::text
+FROM to_rename tr
+WHERE dv.id = tr.id;--> statement-breakpoint
+
+-- Step 2: Drop the old index
+DROP INDEX IF EXISTS latitude.document_versions_unique_path_commit_id_deleted_at;--> statement-breakpoint
+
+-- Step 3: Create the new constraint with NULLS NOT DISTINCT
+ALTER TABLE "latitude"."document_versions" ADD CONSTRAINT "document_versions_unique_path_commit_id_deleted_at" UNIQUE NULLS NOT DISTINCT("path","commit_id","deleted_at");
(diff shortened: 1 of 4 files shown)


packages/core/drizzle/0272_fix_duplicated_documents_in_commit.sql
```diff
@@ -0,0 +1,30 @@
...
+),
+to_rename AS (
+  SELECT id, path, rn
+  FROM duplicates
+  WHERE rn > 1
+)
+UPDATE latitude.document_versions dv
+SET path = tr.path || '_' || (tr.rn - 1)::text
+FROM to_rename tr
+WHERE dv.id = tr.id;--> statement-breakpoint
```
andresgutgon, line 24: @geclos careful review of this data migration please

packages/core/src/services/documents/createUnsafe.ts
```diff
@@ -125,17 +124,6 @@ export async function createNewDocumentUnsafe(
...
-    const docsScope = new DocumentVersionsRepository(workspace!.id, tx)
-    const currentDocs = await docsScope
-      .getDocumentsAtCommit(commit)
-      .then((r) => r.unwrap())
-
-    if (currentDocs.find((d) => d.path === path)) {
-      return Result.error(
-        new BadRequestError('A document with the same path already exists'),
-      )
-    }
```
andresgutgon, line 137: Now this is not needed. DB fails so this extra query can be removed

andresgutgon:
@claude please review

packages/core/src/services/documents/createUnsafe.ts
```diff
@@ -146,19 +134,31 @@ export async function createNewDocumentUnsafe(
...
+          promptlVersion,
+          documentType,
+        })
+        .returning()
+      document = inserted
+    } catch (e) {
+      // If it's a conflicting error path we want to return a user-friendly
+      // message instead of a generic DB error.
+      const result = Transaction.toResultError(e, {
+        customConflictMessage: 'A document with the same path already exists',
```
andresgutgon: This is necessary to keep the previous app level error message. Better than DB index message.

packages/core/src/queries/clickhouse/spans/getByDocument.ts
```diff
@@ -11,8 +11,8 @@ import { scopedQuery } from '../../scope'
 import { buildExperimentExclusionCondition } from './buildExperimentExclusionCondition'
 import { paginateSpanRows } from './paginateSpanRows'
 
-export const getSpansByDocument = scopedQuery(async function getSpansByDocument(
-  {
+export const getSpansByDocument = scopedQuery(
+  async function getSpansByDocument({
```
andresgutgon, line 15: Lot's prettier changed files. Are we not running prettier check on CI?

packages/core/drizzle/0272_fix_duplicated_documents_in_commit.sql
geclos: run this in a separate commit, merge it before this PR. also i think the duplicates a very few? if that is the case just do a hardcoded list of ids instead of this

andresgutgon: This migration works for cloud and non-cloud setups. This way we don't need to hardcode the ids. Also this table is not as big as others.

neoxelox (dismissed):

packages/core/drizzle/0272_fix_duplicated_documents_in_commit.sql
geclos: merge this in a separate pr please

andresgutgon: Done. This PR only has the migration

PR #2355, opened 2026-02-24
Revert main source code to telemetry v2 AND fix tokens/cost in ingestion tokens

andresgutgon:
# What?
We implemented telemetry v3 and published SDKs for Python and Typescript in beta. But having it  in the main repo do not allow us to fix bugs in V2 SDKs


geclos (approved):

packages/core/src/services/tracing/spans/specifications/completion.ts
```diff
@@ -567,10 +575,14 @@ export function extractInput(
 export function extractOutput(
   attributes: Record<string, SpanAttribute>,
 ): TypedResult<Required<CompletionSpanMetadata>['output']> {
-  for (const /* prettier-ignore */ [ key, structure, provider ] of /* prettier-ignore */ [
+  for (const [key, structure, provider] of [
```
neoxelox: bro why, keep it simple to read, keep it compact bro

packages/core/src/services/tracing/spans/specifications/completion.ts
```diff
@@ -517,11 +521,15 @@ export function extractInput(
     }
   }
 
-  for (const /* prettier-ignore */ [ key, structure, provider ] of /* prettier-ignore */ [
+  for (const [key, structure, provider] of [
```
neoxelox: bro why, keep it simple to read, keep it compact bro

pnpm-workspace.yaml
```diff
@@ -24,6 +24,7 @@ catalog:
   '@opentelemetry/resources': 1.30.1
   '@opentelemetry/instrumentation': 0.57.2
   '@opentelemetry/sdk-trace-node': 1.30.1
+  '@latitude-data/telemetry': 3.0.0-beta.0
```
andresgutgon, line 27: We're already using telemetry 3.0.0-beta.0 in the project.

packages/core/src/services/ai/estimateCost/index.ts
```diff
@@ -211,13 +195,29 @@ export function estimateCost({
...
+  usage,
+  provider,
+  model,
+}: {
+  usage: LanguageModelUsage
+  provider: string
+  model: string
+}): number {
+  return estimateCost({ usage, provider, model })
+}
```
geclos: unnecessary indirection

packages/core/src/services/tracing/spans/ingestion/propagateMetadataFromParentSpans.ts
```diff
@@ -0,0 +1,77 @@
...
+        const fromParent = getDocumentMetadataFromParent(parent.metadata)
+        if (Object.keys(fromParent).length > 0) {
+          Object.assign(span.metadata, fromParent)
+        }
+        break
+      }
+      current = parent
+    }
+  }
+}
```
geclos: fyi this is probably just best-effort, because order of arrival is not guaranteed, so a child could arrive in a separate request from its parent containing the metadata you care about.

packages/core/src/queries/clickhouse/spans/computeDocumentTracesAggregations.ts
```diff
@@ -4,6 +4,11 @@ import { TABLE_NAME } from '../../../schema/models/clickhouse/spans'
 import { TracesAggregations } from '../../../schema/models/types/Span'
 import { scopedQuery } from '../../scope'
 
+/**
+ * Aggregations are 0 when no spans match the filter. Spans must have
+ * document_uuid_key and project_id_key set (e.g. completion spans get these
+ * via propagation from parent external spans at ingestion).
+ */
```
geclos: hmm, i think the fix in read query is much simpler, instead of at write time. Just fallback to the default _key column values. e.g:

```
(document_uuid_key = {document_uuid: UUID} OR document_uuid_key =  '00000000-0000-0000-0000-000000000000')
```

packages/core/src/services/tracing/spans/ingestion/propagateMetadataFromParentSpans.ts
andresgutgon: True, this is hard. Maybe we need to refetch all spans related with this when they're are coming to be fully secure?

geclos: this also adds quite a bit of cpu overhead to an operation that is performed 100s of times per second currently

packages/core/src/queries/clickhouse/spans/computeDocumentTracesAggregations.ts
andresgutgon: I did not change this query. What should I do?

geclos: I can take a look

PR #2357, opened 2026-02-24
Use safeTranslate everywhere

neoxelox:


packages/core/src/services/ai/providers/rules/all.ts
```diff
@@ -3,56 +3,14 @@ import type { Message } from '@latitude-data/constants/messages'
...
-import { applyGoogleRules } from './google'
-import { applyOpenAiRules } from './openai'
-import { applyPerplexityRules } from './perplexity'
+import { applyProviderRules, Props } from './provider'
 import {
   extractMessageMetadata,
   getProviderMetadataKey,
 } from './providerMetadata'
 import { AppliedRules } from './types'
-import { applyVertexAnthropicRules } from './vertexAnthropic'
```
neoxelox, line 17: I had to separate index.ts in all.ts and provider.ts because applyProviderRules is used in frontend and "captureException" does not work in frontend

PR #2367, opened 2026-02-25
Fix broken document version migration

andresgutgon:
# What?
We ran a migration that did not use CONCURRENTLY when DROPPING and ADDING an index. Also, that migration performed a data migration and dropped an unrelated table. Now this migration was split into 3 and is used concurrently

### DIFF
--- a/packages/core/drizzle/0273_drop-legacy-published-document-table.sql
+++ b/packages/core/drizzle/0273_drop-legacy-published-document-table.sql
@@ -1,0 +1,1 @@
+ DROP TABLE IF EXISTS "latitude"."published_documents" CASCADE;--> statement-breakpoint
--- a/packages/core/drizzle/0275_fix_document_unique_index_commit_path_deleted_at.sql
+++ b/packages/core/drizzle/0275_fix_document_unique_index_commit_path_deleted_at.sql
@@ -1,0 +1,24 @@
+COMMIT;--> statement-breakpoint
+
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS
+  "document_versions_unique_path_commit_id_deleted_at__new"
+ON "latitude"."document_versions" ("path","commit_id","deleted_at") NULLS NOT DISTINCT;--> statement-breakpoint
+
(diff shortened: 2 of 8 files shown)


packages/core/drizzle/0273_drop-legacy-published-document-table.sql
```diff
@@ -0,0 +1 @@
+ DROP TABLE IF EXISTS "latitude"."published_documents" CASCADE;--> statement-breakpoint
```
andresgutgon, line 1: The code associated with this table was already removed 🔥

packages/core/drizzle/0275_fix_document_unique_index_commit_path_deleted_at.sql
```diff
@@ -0,0 +1,14 @@
+COMMIT;--> statement-breakpoint
+
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS
+  "document_versions_unique_path_commit_id_deleted_at__new"
```
andresgutgon, line 4: Index with different name so we can concurrently remove existing index

packages/core/drizzle/0275_fix_document_unique_index_commit_path_deleted_at.sql
```diff
@@ -0,0 +1,14 @@
...
+ON "latitude"."document_versions" ("path","commit_id","deleted_at") NULLS NOT DISTINCT;--> statement-breakpoint
+
+BEGIN;--> statement-breakpoint
+
+ALTER TABLE "latitude"."document_versions"
+  DROP CONSTRAINT IF EXISTS "document_versions_unique_path_commit_id_deleted_at";--> statement-breakpoint
+
+ALTER TABLE "latitude"."document_versions"
```
andresgutgon, line 22: Add the constraint to this index. The constraint is what's using this code
```typescript
uniquePathCommitIdDeletedAt: unique(
  'document_versions_unique_path_commit_id_deleted_at',
)
  .on(table.path, table.commitId, table.deletedAt)
  .nullsNotDistinct()
```

PR #2384, opened 2026-02-27
Remove onError callback from stream handlers and fix error status codes

geclos:
## Summary
This PR refactors the stream handling error management by removing the `onError` callback parameter and fixing incorrect HTTP status codes used in error responses.

## Key Changes
- **Removed `onError` callback**: Eliminated the `onError` parameter from `handleStream` and all stream utility functions (`streamRun`, `streamChat`, `streamAttach`). Errors are now thrown directly instead of being passed to a callback.
- **Fixed HTTP status codes**: Changed incorrect `402` status codes to `500` (Internal Server Error) in three locations:
  - Invalid JSON parsing error
  - Chain error handling
  - Generic error handling in catch block
- **Simplified error handling**: Refactored the catch block in `handleStream` to be more concise - errors that are already `LatitudeApiError` instances are re-thrown directly, while other errors are wrapped with the correct status code.
- **Updated type definitions**: Removed `StreamResponseCallbacks` import and replaced the callback-based parameter type with an inline object type definition that only includes `onEvent` and `onToolCall`.
- **Updated tests**: Removed `onError` mock calls from test cases and updated expected error status codes from `402` to `500`.

## Notable Implementation Details
- The change simplifies the error handling contract by making errors propagate through exceptions rather than callbacks, which is more idiomatic for async/await patterns.
- The fix for duplicate error reporting (noted in a FIXME comment) is a side effect of removing the `onError` callback - errors are now only thrown once instead of being reported through the callback and then thrown.

https://claude.ai/code/session_015XPsHtCVt5JHiRtAiSpfUS

### DIFF
--- a/packages/sdks/typescript/CHANGELOG.md
+++ b/packages/sdks/typescript/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to the TypeScript SDK will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [5.7.1] - 2026-02-27
+
+### Fixed
+
+- Fixed misleading "Unexpected API Error: 402" errors during stream processing. Stream errors (connection terminated, provider errors, invalid JSON) now correctly use status 500 instead of 402 (Payment Required).
+- Fixed `onError` callback being called twice for the same error when stream processing fails.
+
 ## [5.7.0] - 2026-02-25
 
 ### Added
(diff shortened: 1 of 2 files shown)


packages/sdks/typescript/src/utils/handleStream.ts
```diff
@@ -20,11 +20,11 @@ import { EventSourceParserStream } from 'eventsource-parser/stream'
 export async function handleStream<S extends AssertedStreamType = 'text'>({
   body,
   onEvent,
-  onError,
```
andresgutgon, line 23: Is this a breaking change?

geclos: no

PR #2390, opened 2026-03-02
Datasets data model

andresgutgon:
# What?
Define datasets data model for v2

## TODO
- [ ] Data model
- [ ] migrations


packages/platform/db-postgres/src/schema/datasets.ts
```diff
@@ -0,0 +1,26 @@
...
+    id: uuid("id").primaryKey(), // UUID, consistent with Better Auth
+    organizationId: text("organization_id").notNull(),
+    projectId: text("project_id").notNull(),
+    authorId: text("author_id"),
+    name: varchar("name", { length: 256 }).notNull(),
+    deletedAt: timestamp("deleted_at", { withTimezone: true }),
+    ...timestamps(),
+  },
+  () => [organizationRLSPolicy()],
+)
```
andresgutgon, line 32: I did a couple of helpers that we should use always

1. `LATITUDE_SCHEMA` to avoid repeating
2. `timestamps()` classic create and update columns
3. `organizationRLSPolicy` RLS policy for the table

PR #2394, opened 2026-03-03
Migrate Drizzle to v1 beta, cuid for everyone, RSL for everyone and named Env errors

andresgutgon:
# What?
Drizzle v1 migration and other things.

- [x] This migrates code to use Drizzle v1
- [x] Use `cuid` drizzle helper for all IDs
- [x] Instruct AGENTS.md with Drizzle model best practices. timestamp with timezone, use org RSL helper, common timestamp helper and use `cuid` for new tables.
- [x] Nicer environment errors
- [x] Setup postgresql seed data.


packages/platform/db-postgres/drizzle/20260303105419_setup-rls-auth-basic-models/migration.sql
```diff
@@ -1,9 +1,9 @@
-CREATE TYPE "public"."member_role" AS ENUM('owner', 'admin', 'member');--> statement-breakpoint
-CREATE TYPE "public"."user_role" AS ENUM('user', 'admin');--> statement-breakpoint
-CREATE TYPE "public"."grant_source" AS ENUM('subscription', 'purchase', 'promocode');--> statement-breakpoint
-CREATE TYPE "public"."quota_type" AS ENUM('seats', 'runs', 'credits');--> statement-breakpoint
+CREATE TYPE "latitude"."member_role" AS ENUM('owner', 'admin', 'member');--> statement-breakpoint
+CREATE TYPE "latitude"."user_role" AS ENUM('user', 'admin');--> statement-breakpoint
+CREATE TYPE "latitude"."grant_source" AS ENUM('subscription', 'purchase', 'promocode');--> statement-breakpoint
+CREATE TYPE "latitude"."quota_type" AS ENUM('seats', 'runs', 'credits');--> statement-breakpoint
```
geclos: why are there enums, remove them

packages/platform/db-postgres/drizzle/20260303113550_setup-tables-and-rls-policies/migration.sql
```diff
@@ -120,53 +119,53 @@ CREATE TABLE "latitude"."projects" (
...
+CREATE POLICY "member_organization_policy" ON "latitude"."member" AS PERMISSIVE FOR ALL TO public USING (organization_id = get_current_organization_id()) WITH CHECK (organization_id = get_current_organization_id());--> statement-breakpoint
+CREATE POLICY "subscription_organization_policy" ON "latitude"."subscription" AS PERMISSIVE FOR ALL TO public USING (reference_id = get_current_organization_id()) WITH CHECK (reference_id = get_current_organization_id());--> statement-breakpoint
 CREATE POLICY "projects_organization_policy" ON "latitude"."projects" AS PERMISSIVE FOR ALL TO public USING (organization_id = get_current_organization_id()) WITH CHECK (organization_id = get_current_organization_id());--> statement-breakpoint
 CREATE POLICY "api_keys_organization_policy" ON "latitude"."api_keys" AS PERMISSIVE FOR ALL TO public USING (organization_id = get_current_organization_id()) WITH CHECK (organization_id = get_current_organization_id());--> statement-breakpoint
 CREATE POLICY "grants_organization_policy" ON "latitude"."grants" AS PERMISSIVE FOR ALL TO public USING (organization_id = get_current_organization_id()) WITH CHECK (organization_id = get_current_organization_id());
```
geclos, line 167: why has this file moved places

geclos: ah i guess because drizzle version has changed

packages/platform/db-postgres/src/schema/better-auth.ts
```diff
@@ -21,17 +28,15 @@ import { cuid } from "../schemaHelpers.js"
...
-export const userRoleEnum = pgEnum("user_role", ["user", "admin"])
+export const userRoleEnum = latitudeEnum("user_role", ["user", "admin"])
 
 /**
  * User table - stores user accounts
  *
  * This is the core Better Auth user table.
  */
-export const user = latitudeSchema.table("user", {
+export const user = LATITUDE_SCHEMA.table("user", {
```
geclos: why is this suddently in caps

packages/platform/db-postgres/src/relations.ts
```diff
@@ -0,0 +1,109 @@
...
+    organization: r.one.organization({
+      from: r.grants.organizationId,
+      to: r.organization.id,
+    }),
+    subscription: r.one.subscription({
+      from: r.grants.subscriptionId,
+      to: r.subscription.id,
+    }),
+  },
+}))
```
geclos, line 109: is this automatically generated? what benefit does it bring? do not want to maintain it unless it's high roi

packages/platform/db-postgres/drizzle/20260303113550_setup-tables-and-rls-policies/migration.sql
andresgutgon: I'm dropping all the DB and running again.

packages/platform/db-postgres/src/schema/better-auth.ts
andresgutgon: Is a global constant imported from helpers

packages/platform/db-postgres/src/relations.ts
andresgutgon: Ideally can be done using `drizzle-kit pull` but i had to gen manually because i broke everything. But i can try to generate a new table with a relation and run the command.

packages/platform/db-postgres/src/schemaHelpers.ts
```diff
@@ -4,8 +4,8 @@ import { pgPolicy, pgSchema, timestamp, varchar } from "drizzle-orm/pg-core"
 
 export const LATITUDE_SCHEMA = pgSchema("latitude")
```
geclos: no need make it upper case tbh all other tables are also constants yet we aren't upercasing them I think uppercase constants only make sense for primitive types

packages/platform/db-postgres/src/schemaHelpers.ts
```diff
@@ -31,6 +31,10 @@ export function timestamps() {
  * Generate a unique ID using CUID2.
  * CUID2 provides 24-25 character URL-safe unique identifiers.
  */
+export function latitudeEnum<T extends string>(name: string, values: readonly [T, ...T[]]) {
+  return LATITUDE_SCHEMA.enum(name, values)
+}
```
geclos: we do not want pg enums, replace for varchar with ts union types please

packages/platform/db-postgres/src/relations.ts
geclos: if it's automated then ok, otherwise i would remove it (unless there some clear benefit)

packages/platform/db-postgres/src/schemaHelpers.ts
andresgutgon: constants are uppercase. This is a constant

geclos: then constant all the table names?

packages/platform/db-postgres/drizzle/20260303113550_setup-tables-and-rls-policies/migration.sql
```diff
@@ -120,53 +115,53 @@ CREATE TABLE "latitude"."projects" (
 --> statement-breakpoint
 ALTER TABLE "latitude"."projects" ENABLE ROW LEVEL SECURITY;--> statement-breakpoint
 CREATE TABLE "latitude"."api_keys" (
-	"id" text PRIMARY KEY NOT NULL,
```
geclos, line 123: you removed the non null?

packages/platform/db-postgres/src/relations.ts
andresgutgon: I checked and is not generated. This is not based in foreign keys. Is an application abstraction
https://orm.drizzle.team/docs/relations-v2#foreign-keys 

So it can't infer from DB

andresgutgon: >   The relations.ts file enables Drizzle's relational query API
  (db.query.*.findFirst / db.query.*.findMany). That's the only thing it
  does.
  Without it, you can only use the SQL-like query builder
  (db.select().from().where().innerJoin()). With it, you get:
  1. Nested `with` clauses — fetch related data in a single call:
```typescript
  db.query.organization.findFirst({
    where: { id: orgId },
    with: {
      members: { with: { user: true } },
      projects: true,
    },
  })
```

  vs. writing manual joins or multiple queries.
  2. Type-safe nested results — the return type automatically includes the
  nested shape, so result.members[0].user.email is fully typed.
  That's it. If you're not using db.query.* with with clauses anywhere, the
  file provides zero value.
  Let me check if we actually use with anywhere.

packages/platform/db-postgres/drizzle/20260303113550_setup-tables-and-rls-policies/migration.sql
andresgutgon: >   The id columns in the migration are varchar(128) PRIMARY KEY — not 
  nullable. In PostgreSQL, PRIMARY KEY implicitly means NOT NULL, so writing
   NOT NULL explicitly is redundant when PRIMARY KEY is present.

AGENTS.md
```diff
@@ -131,6 +133,10 @@ pnpm --filter @app/api test -- -t "health endpoint"
 pnpm --filter @app/api test -- src/some-file.test.ts -t "returns 200"
 ```
 
+### Test Environment Variables
```
neoxelox: no need for this

packages/platform/email-nodemailer/package.json
```diff
@@ -10,7 +10,8 @@
     "check": "biome check src",
     "format": "biome format --write src",
     "typecheck": "tsc -p tsconfig.json --noEmit",
-    "test": "vitest run --passWithNoTests"
+    "test": "vitest run --passWithNoTests",
+    "check:fix": "biome check --fix --unsafe src"
```
geclos: yoyo, fix unsafe no eh

packages/platform/db-weaviate/package.json
```diff
@@ -11,7 +11,8 @@
     "check": "biome check src",
     "format": "biome format --write src",
     "typecheck": "tsc -p tsconfig.json --noEmit",
-    "test": "vitest run --passWithNoTests"
+    "test": "vitest run --passWithNoTests",
+    "check:fix": "biome check --fix --unsafe src"
```
geclos: fix unsafe can break the app

packages/platform/db-postgres/drizzle.config.ts
```diff
@@ -5,14 +5,14 @@ import { config as loadDotenv } from "dotenv"
...
-const nodeEnv = Effect.runSync(parseEnv(process.env.NODE_ENV, "string", "development"))
+const nodeEnv = Effect.runSync(parseEnv("NODE_ENV", "string", "development"))
 const envFilePath = fileURLToPath(new URL(`../../../.env.${nodeEnv}`, import.meta.url))
 
 if (existsSync(envFilePath)) {
   loadDotenv({ path: envFilePath, quiet: true })
 }
 
-const url = Effect.runSync(parseEnv(process.env.LAT_DATABASE_URL, "string"))
+const url = Effect.runSync(parseEnv("LAT_DATABASE_URL", "string", "postgres://localhost:5432/latitude"))
```
geclos: i prefer if we don't hardcode dev default in code instead let's just commit .env.development and .env.test to the repo and that's it

packages/platform/db-postgres/src/relations.ts
geclos: what do you think? i think no one will maintain this

packages/platform/db-postgres/drizzle.config.ts
andresgutgon: I'll remove the default but commiting `.env.*` is a no go

packages/platform/db-postgres/src/relations.ts
andresgutgon: The question is, do we want to use that syntax or no? I think it's nice but not sure

PR #2395, opened 2026-03-03
Spans domain model

csansoon:
## Summary

- Redesigns the ClickHouse `spans` table from scratch around OpenTelemetry and GenAI semantic conventions. Everything lives in one table — metadata, attributes, and LLM content payloads — with no separate object storage layer for span data.
- A new `@domain/spans` package defines two domain shapes: `Span` (listing/query, excludes large payloads) and `SpanDetail` (point lookup, parses `input_messages`/`output_messages` as `GenAIMessage[]` from `rosetta-ai`).
- Platform adapter in `@platform/db-clickhouse` implements the repository port with explicit column selection for list queries and `SELECT *` for detail lookups.
- Adds a temporary `/spans` debug page to the web app for visual verification of seeded data.

## Spans ClickHouse Schema

Single `spans` table stores everything. List queries select only the lightweight columns; detail queries fetch the full row including content payloads.

### Tenancy & scope

| Column | Type | Notes |
|---|---|---|
| `organization_id` | LowCardinality(String) | Tenant partition key |
| `project_id` | LowCardinality(String) | Project within the organization |

### Identity & hierarchy

| Column | Type | Notes |
|---|---|---|
| `session_id` | String DEFAULT '' | `gen_ai.conversation.id`. Empty string = no session. |
| `trace_id` | FixedString(32) | OTEL trace ID (32 hex chars) |
| `span_id` | FixedString(16) | OTEL span ID (16 hex chars) |
| `parent_span_id` | String DEFAULT '' | Empty string for root spans |
| `api_key_id` | LowCardinality(String) DEFAULT '' | CUID2 of the ingesting API key |

### Timing

| Column | Type | Notes |
|---|---|---|
| `start_time` | DateTime64(9, 'UTC') | Nanosecond precision |
| `end_time` | DateTime64(9, 'UTC') | Nanosecond precision |
| `duration_ns` | ALIAS (computed) | Read-time computation, not stored |

### Span metadata

| Column | Type | Notes |
|---|---|---|
| `name` | LowCardinality(String) | Span name |
| `service_name` | LowCardinality(String) | From OTEL resource `service.name` |
| `kind` | Int8 | OTel SpanKind: 0=UNSPECIFIED 1=INTERNAL 2=SERVER 3=CLIENT 4=PRODUCER 5=CONSUMER |
| `status_code` | Int16 DEFAULT 0 | OTel status: 0=UNSET 1=OK 2=ERROR |
| `status_message` | String DEFAULT '' | Error description |
| `trace_flags` | UInt32 DEFAULT 0 | OTEL trace flags |
| `trace_state` | String DEFAULT '' | OTEL trace state |
| `error_type` | LowCardinality(String) DEFAULT '' | From `error.type` attribute |
| `tags` | Array(LowCardinality(String)) DEFAULT [] | User-defined labels |
| `events_json` | String DEFAULT '' | JSON array of span events |
| `links_json` | String DEFAULT '' | JSON array of span links |

### GenAI fields

| Column | OTEL mapping | Notes |
|---|---|---|
| `operation` | `gen_ai.operation.name` | chat / embeddings / etc. |
| `provider` | `gen_ai.system` | openai / anthropic / etc. |
| `model` | `gen_ai.request.model` | Requested model |
| `response_model` | `gen_ai.response.model` | Model actually used |

### Tokens

| Column | Type | Notes |
|---|---|---|
| `tokens_input` | UInt32 DEFAULT 0 | Input/prompt tokens |
| `tokens_output` | UInt32 DEFAULT 0 | Output/completion tokens |
| `tokens_cache_read` | UInt32 DEFAULT 0 | Tokens served from cache |
| `tokens_cache_create` | UInt32 DEFAULT 0 | Tokens written to cache |
| `tokens_reasoning` | UInt32 DEFAULT 0 | Reasoning tokens (o1/o3, etc.) |
| `tokens_total` | UInt32 MATERIALIZED | Sum of all token columns, computed at insert |

### Cost (microcents — 1 USD = 100,000,000)

| Column | Type | Notes |
|---|---|---|
| `cost_input_microcents` | UInt64 DEFAULT 0 | Input token cost |
| `cost_output_microcents` | UInt64 DEFAULT 0 | Output token cost |
| `cost_total_microcents` | UInt64 DEFAULT 0 | Total cost (regular column, not materialized) |
| `cost_is_estimated` | UInt8 DEFAULT 0 | 1 if estimated by Latitude, 0 if provider-reported |

### GenAI response metadata

| Column | Type | Notes |
|---|---|---|
| `response_id` | String DEFAULT '' | `gen_ai.response.id` |
| `finish_reasons` | Array(LowCardinality(String)) DEFAULT [] | `gen_ai.finish_reasons` |

### Dynamic attributes (typed maps)

| Column | Type | Notes |
|---|---|---|
| `attr_string` | Map(LowCardinality(String), String) | String attributes |
| `attr_int` | Map(LowCardinality(String), Int64) | Integer attributes |
| `attr_float` | Map(LowCardinality(String), Float64) | Float attributes |
| `attr_bool` | Map(LowCardinality(String), UInt8) | Boolean attributes |
| `resource_string` | Map(LowCardinality(String), String) | OTEL resource attributes |
| `scope_name` | LowCardinality(String) DEFAULT '' | Instrumentation scope name |
| `scope_version` | LowCardinality(String) DEFAULT '' | Instrumentation scope version |

### LLM content payloads (JSON strings, ZSTD(3))

| Column | Notes |
|---|---|
| `input_messages` | Prompt messages |
| `output_messages` | Completion messages |
| `system_instructions` | System prompt |
| `tool_definitions` | Tool/function schemas |

### Platform metadata

| Column | Type | Notes |
|---|---|---|
| `ingested_at` | DateTime64(3, 'UTC') DEFAULT now64(3) | ReplacingMergeTree version column |

## Test plan

- [ ] Run `ch:up` and `ch:seed` — verify spans insert without errors
- [ ] Visit `/spans` in the web app — verify table shows seeded spans
- [ ] Click a span row — verify detail view renders all fields
- [ ] Run `pnpm check` and `pnpm typecheck` — verify no lint/type errors


packages/domain/spans/src/entities/span-payload.ts
```diff
@@ -0,0 +1,128 @@
+import type { OrganizationId, ProjectId, SpanId, TraceId } from "@domain/shared-kernel"
+
+/**
+ * A single message part in the OTEL GenAI message format.
+ */
+export interface MessagePart {
+  readonly type: string
+  readonly [key: string]: unknown
```
andresgutgon: We should start typing messages as rosseta

csansoon: Actually yes! I changed it to Rosetta's `GenAI` messages

packages/domain/spans/src/entities/span.ts
```diff
@@ -0,0 +1,113 @@
...
+  readonly reasoningTokens: number | null // gen_ai.usage.reasoning_tokens
+
+  readonly inputCost: number | null // gen_ai.usage.input_cost
+  readonly outputCost: number | null // gen_ai.usage.output_cost
+  readonly totalCost: number | null // gen_ai.usage.total_cost
+  readonly costEstimated: boolean
+
+  readonly ingestedAt: Date
+  readonly retentionExpiresAt: Date
+}
```
geclos, line 58: still missing input and output

packages/domain/spans/src/ports/ingest-buffer-repository.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { Effect } from "effect"
+
+/**
+ * Repository port for ephemeral ingest payloads (object storage).
+ *
+ * Raw OTLP span batches are buffered here so the API can return
+ * immediately. A background worker retrieves and deletes them
+ * after processing.
+ *
+ * Key scheme: ingest/spans/{contentHash}
```
geclos: always scope object storage by tenant

packages/domain/spans/src/ports/span-payload-repository.ts
```diff
@@ -0,0 +1,21 @@
...
+  store(payload: SpanPayload): Effect.Effect<void, RepositoryError>
+
+  storeBatch(payloads: readonly SpanPayload[]): Effect.Effect<void, RepositoryError>
+
+  retrieve(
+    organizationId: OrganizationId,
+    traceId: TraceId,
+    spanId: SpanId,
+  ): Effect.Effect<SpanPayload | null, RepositoryError>
+}
```
geclos: i don't undersatnd the diff between ingest buff repo and span payload repo

packages/platform/db-clickhouse/clickhouse/migrations/clustered/0002_recreate_spans.down.sql
```diff
@@ -0,0 +1,38 @@
...
+  attributes String,
+
+  ingested_at DateTime64(6, 'UTC') DEFAULT now64(6),
+
+  INDEX idx_trace_id trace_id TYPE bloom_filter(0.001) GRANULARITY 1,
+  INDEX idx_api_key_id api_key_id TYPE bloom_filter(0.01) GRANULARITY 1
+)
+ENGINE = ReplicatedReplacingMergeTree(ingested_at)
+PARTITION BY toYYYYMM(started_at)
+PRIMARY KEY (workspace_id, started_at)
```
geclos: hmm this isn't what we discussed no? project id required in pk because most common queries will query by project

packages/platform/db-clickhouse/clickhouse/migrations/clustered/0002_recreate_spans.down.sql
```diff
@@ -0,0 +1,38 @@
...
+ORDER BY (workspace_id, started_at, trace_id, span_id);
```
geclos: ditto missing project id ordering

packages/platform/db-clickhouse/clickhouse/migrations/clustered/0002_recreate_spans.down.sql
```diff
@@ -0,0 +1,38 @@
+DROP TABLE IF EXISTS spans ON CLUSTER default;
+
+CREATE TABLE IF NOT EXISTS spans ON CLUSTER default (
```
geclos: why is down migration creating a table? just start migrations from scratch in clickhouse

packages/platform/db-clickhouse/clickhouse/migrations/clustered/0002_recreate_spans.up.sql
```diff
@@ -0,0 +1,52 @@
...
+  retention_expires_at DateTime64(6, 'UTC') DEFAULT toDateTime64('2100-01-01 00:00:00', 6, 'UTC'),
+
+  INDEX idx_trace_id trace_id TYPE bloom_filter(0.001) GRANULARITY 1,
+  INDEX idx_api_key_id api_key_id TYPE bloom_filter(0.01) GRANULARITY 1,
+  INDEX idx_session_id session_id TYPE bloom_filter(0.01) GRANULARITY 1
+)
+ENGINE = ReplicatedReplacingMergeTree(ingested_at)
+PARTITION BY toYYYYMM(start_time)
+PRIMARY KEY (organization_id, project_id, start_time)
+ORDER BY (organization_id, project_id, start_time, trace_id, span_id);
```
geclos: aah ok i was confused by the down migration

packages/platform/db-clickhouse/clickhouse/migrations/registry.json
```diff
@@ -1,5 +1,6 @@
 {
   "migrations": {
-    "0001": "create_spans"
+    "0001": "create_spans",
+    "0002": "recreate_spans"
```
geclos: make it a single migration

packages/domain/spans/src/entities/span.ts
csansoon: do we want input and output in the `span` table? why?

packages/domain/spans/src/ports/span-payload-repository.ts
csansoon: Hmm I see. It was a pattern recommended by Claude, where span raw payloads are immediately added to a "buffer" when ingested, so the API returns a fast response, and then a background wourker would retrieve them from the buffer and process them.

However, we actually use BullMQ job payload directly as this "buffer" anyways, so this is not really needed.

packages/platform/db-clickhouse/clickhouse/migrations/clustered/0002_recreate_spans.down.sql
csansoon: Yeah this is the DOWN query, to revert changes from this migration

packages/platform/db-clickhouse/clickhouse/migrations/clustered/0002_recreate_spans.down.sql
csansoon: ditto

packages/platform/db-clickhouse/clickhouse/migrations/clustered/0002_recreate_spans.down.sql
csansoon: Oh I did not want to remove the original first migration to avoid having db issues with other devs from the start

packages/platform/db-clickhouse/clickhouse/migrations/clustered/0002_recreate_spans.up.sql
csansoon: 😂

packages/platform/db-clickhouse/clickhouse/migrations/registry.json
csansoon: oki

andresgutgon: Please introduce https://github.com/pressly/goose as migration system. Maybe they have some kind of registry

packages/domain/spans/src/ports/span-payload-repository.ts
geclos: no no, it is, we do need an object storage buffer for batch spans you cannot simply put span batches in redis you will overflow the queue very quickly. We already did this in v1 btw, nothing new here.

packages/domain/spans/src/entities/span.ts
geclos: yes — talked offline

packages/platform/storage-object/src/repositories/span-payload-repository.ts
```diff
@@ -0,0 +1,62 @@
...
+  `workspaces/${workspaceId}/traces/${traceId}/${spanId}`
+
+const compress = (data: string): Uint8Array => gzipSync(Buffer.from(data, "utf-8"))
+
+const decompress = (bytes: Uint8Array): string => {
+  const buf = Buffer.from(bytes)
+  if (buf.length >= 2 && buf[0] === GZIP_MAGIC[0] && buf[1] === GZIP_MAGIC[1]) {
+    return gunzipSync(buf).toString("utf-8")
+  }
+  return buf.toString("utf-8")
```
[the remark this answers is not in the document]
csansoon: span-payload-repository removed

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
```diff
@@ -1,42 +1,189 @@
...
+
+    -- Server-side ingestion timestamp. Used as the version column
+    -- for ReplacingMergeTree: later ingestion wins on deduplication.
+    ingested_at              DateTime64(3, 'UTC') DEFAULT now64(3)     CODEC(Delta(8), LZ4),
+
+    -- ═══════════════════════════════════════════════════════
+    -- SKIP INDEXES
+    -- ═══════════════════════════════════════════════════════
+
+    INDEX idx_trace_id       trace_id    TYPE bloom_filter(0.01) GRANULARITY 4,
```
geclos: why granurality 4?

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
```diff
@@ -1,42 +1,189 @@
...
+    INDEX idx_session_id     session_id  TYPE bloom_filter(0.01) GRANULARITY 4,
+    INDEX idx_model          model       TYPE set(200)           GRANULARITY 4,
+    INDEX idx_provider       provider    TYPE set(50)            GRANULARITY 4,
```
geclos: what's type set and the number?

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
```diff
@@ -1,42 +1,189 @@
...
 ENGINE = ReplicatedReplacingMergeTree(ingested_at)
-PARTITION BY toYYYYMM(started_at)
-PRIMARY KEY (workspace_id, started_at)
-ORDER BY (workspace_id, started_at, trace_id, span_id);
+PARTITION BY toYYYYMM(start_time)
+PRIMARY KEY (organization_id, project_id, toStartOfHour(start_time))
+ORDER BY (
+    organization_id,
+    project_id,
+    toStartOfHour(start_time),
```
geclos: why make the order hourly? i might have thousands of spans per hour that i need ordered as well

.env.example
```diff
@@ -61,6 +61,15 @@ LAT_WEAVIATE_GRPC_PORT=50051
...
 LAT_REDIS_PORT=6379
 
+# Object Storage
+LAT_STORAGE_DRIVER=fs
+LAT_STORAGE_FS_ROOT=./storage
+# LAT_STORAGE_S3_BUCKET=latitude-spans
+# LAT_STORAGE_S3_REGION=us-east-1
+# LAT_STORAGE_S3_ACCESS_KEY_ID=your-access-key
+# LAT_STORAGE_S3_SECRET_ACCESS_KEY=your-secret-key
+# LAT_STORAGE_S3_ENDPOINT=https://s3.amazonaws.com
```
geclos, line 71: no need to comment it

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
```diff
@@ -1,42 +1,189 @@
...
+    session_id,
+    trace_id,
+    span_id,
+    start_time
+)
+TTL toDateTime(start_time) + INTERVAL 7 DAY TO VOLUME 'cold'
```
geclos: 7 days is super low, needs to be much larger probably a few months?

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
```diff
@@ -1,42 +1,189 @@
...
+SETTINGS
+    index_granularity       = 8192,
+    ttl_only_drop_parts     = 1,
+    min_bytes_for_wide_part = 10485760,
+    storage_policy          = 'tiered';
```
geclos: what's this?

apps/web/src/domains/spans/spans.functions.ts
```diff
@@ -0,0 +1,113 @@
...
+  WHERE trace_id = {traceId:FixedString(32)}
+    AND span_id = {spanId:FixedString(16)}
+  LIMIT 1
+`
+
+export const listAllSpans = createServerFn({ method: "GET" })
+  .middleware([errorHandler])
+  .handler(async () => {
+    const client = getClickhouseClient()
+    const result = await client.query({ query: LIST_QUERY, format: "JSONEachRow" })
```
andresgutgon: Move all query logic to span domain. Not need to do here but let's not keep logic, in this SQL logic in apps.

andresgutgon: Also apps should not need about Clickhouse at all. It's an implementation detail

apps/web/src/routes/spans/$traceId.$spanId.tsx
```diff
@@ -0,0 +1,201 @@
...
+  )
+}
+
+function SpanDetailPage() {
+  const { traceId, spanId } = Route.useParams()
+  const [span, setSpan] = useState<SpanDetailRecord | null>(null)
+  const [error, setError] = useState<string | null>(null)
+
+  useEffect(() => {
+    getSpanDetail({ data: { traceId, spanId } })
```
andresgutgon: `getSpanDetail` is a server function no? Also we this is not using our store which is tanstackDB

apps/web/src/routes/spans/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+function SpansPage() {
+  const [spans, setSpans] = useState<SpanListRecord[] | null>(null)
+  const [error, setError] = useState<string | null>(null)
+
+  useEffect(() => {
+    listAllSpans()
+      .then(setSpans)
+      .catch((e) => setError(e instanceof Error ? e.message : String(e)))
+  }, [])
```
andresgutgon: Same. This has to be ported to tanstackDB

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
csansoon: It's a common default, one index entry per 32,768 rows .
Not really deeply calculated. For small scales, it's not really appreciable, and for larger scales it is best to benchmark it somehow. I'll adjust them better but we might have to benchmark with larger datasets to improve performance over a larger scale

docker/clickhouse/storage.xml
```diff
@@ -0,0 +1,22 @@
...
+            <disk>default</disk>
+          </hot>
+          <cold>
+            <disk>cold</disk>
+          </cold>
+        </volumes>
+      </tiered>
+    </policies>
+  </storage_configuration>
+</clickhouse>
```
andresgutgon, line 22: What's this?

packages/domain/spans/src/ports/span-repository.ts
```diff
@@ -0,0 +1,31 @@
...
+    projectId: ProjectId,
+    options: SpanListOptions,
+  ): Effect.Effect<readonly Span[], RepositoryError>
+
+  findBySpanId(
+    organizationId: OrganizationId,
+    traceId: TraceId,
+    spanId: SpanId,
+  ): Effect.Effect<SpanDetail | null, RepositoryError>
+}
```
andresgutgon, line 24: Is this used?

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
```diff
@@ -1,42 +1,189 @@
...
-PARTITION BY toYYYYMM(started_at)
-PRIMARY KEY (workspace_id, started_at)
-ORDER BY (workspace_id, started_at, trace_id, span_id);
+PARTITION BY toYYYYMM(start_time)
+PRIMARY KEY (organization_id, project_id, toStartOfHour(start_time))
+ORDER BY (
+    organization_id,
+    project_id,
+    toStartOfHour(start_time),
+    session_id,
```
andresgutgon, line 176: Primary key includes `session_id` before `trace_id`, but `session_id` is often empty. The ORDER BY puts session_id before trace_id. Since most non-LLM spans will have session_id = '', you'll get massive runs of empty strings that don't help narrow granules. More importantly, queries that filter by trace_id without session_id won't benefit from the sort order — they'll fall through to the bloom filter index. If trace-level lookups are a common access pattern (they almost certainly are), swapping trace_id before session_id in the ORDER BY would likely be a meaningful improvement.

-- Claude review

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
andresgutgon: So you propose something smaller like `toStartOfFiveMinutes`?

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00001_create_spans.sql
```diff
@@ -1,42 +1,189 @@
...
+    -- TIMING
+    -- ═══════════════════════════════════════════════════════
+
+    start_time               DateTime64(9, 'UTC')                      CODEC(Delta(8), ZSTD(1)),
+    end_time                 DateTime64(9, 'UTC')                      CODEC(Delta(8), ZSTD(1)),
+
+    -- ALIAS: computed at read time, not stored.
+    -- Use reinterpretAsInt64 directly in aggregations.
+    duration_ns              ALIAS reinterpretAsInt64(end_time)
+                                 - reinterpretAsInt64(start_time),
```
andresgutgon, line 44: The duration_ns ALIAS uses `reinterpretAsInt64` on `DateTime64(9)`. This works, but it's a subtle trick that future maintainers might not understand. More importantly, since it's an ALIAS (computed at read time), any query that filters or sorts by duration will compute it for every row scanned. If duration-based queries are common (e.g., "find slow spans"), a MATERIALIZED column would be faster at the cost of storage.

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
```diff
@@ -1,42 +1,189 @@
...
+                                 tokens_input + tokens_output + tokens_cache_read
+                                 + tokens_cache_create + tokens_reasoning
+                                                                       CODEC(T64, ZSTD(1)),
+
+    -- Cost in microcents (1 USD = 100,000,000 microcents).
+    -- cost_total is a regular column because some providers return a
+    -- total that does not equal input + output exactly.
+    cost_input_microcents    UInt64   DEFAULT 0                        CODEC(T64, ZSTD(1)),
+    cost_output_microcents   UInt64   DEFAULT 0                        CODEC(T64, ZSTD(1)),
+    cost_total_microcents    UInt64   DEFAULT 0                        CODEC(T64, ZSTD(1)),
```
andresgutgon, line 108: cost_total_microcents is a regular column rather than MATERIALIZED. The comment explains why (some providers return totals that don't equal input + output), which is a valid reason. However, this means the ingestion layer is responsible for always populating it correctly. If it's ever omitted, you get 0 instead of a computed fallback. A defensive DEFAULT cost_input_microcents + cost_output_microcents might be safer, with explicit overrides when the provider supplies a different total.

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
```diff
@@ -1,42 +1,189 @@
...
+    toStartOfHour(start_time),
+    session_id,
+    trace_id,
+    span_id,
+    start_time
+)
+TTL toDateTime(start_time) + INTERVAL 7 DAY TO VOLUME 'cold'
+SETTINGS
+    index_granularity       = 8192,
+    ttl_only_drop_parts     = 1,
```
andresgutgon: Monthly partitioning with a 7-day cold TTL. The TTL moves data to cold storage after 7 days, but partitions are monthly. This means the TTL operates at the part level (within a partition), not the partition level — which is correct since you have ttl_only_drop_parts = 1. But this also means that a part won't be moved to cold until all rows in that part are past the 7-day mark. Depending on how parts merge, you might see data lingering on hot storage slightly longer than expected. This isn't a bug, just a subtlety to be aware of.

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
csansoon: `set` is for attributes with limited unique values, where it is way more optimized than `bloom_filter`, but it handles higher-cardinality columns with distinct values.
It makes sense for something like `status_code`, where there are only 3 possible values (can be set to `set(3)`), or maybe `operation`. `set(50)` on `provider` could work, because there might not be more than 50 providers, but I would actually not risk it and move it to `bloom_filter`

packages/platform/db-clickhouse/src/seeds/spans/index.ts
```diff
@@ -1,132 +1,223 @@
...
+    output_messages: JSON.stringify([{ role: "assistant", content: "This document covers..." }]),
+    system_instructions: "",
+    tool_definitions: "",
+    attr_string: {},
+    attr_int: {},
+    attr_float: { "gen_ai.request.temperature": 0.7 },
+    attr_bool: {},
+    resource_string: { "service.name": "my-app" },
+    scope_name: "openai-instrumentation",
+    scope_version: "1.0.0",
```
andresgutgon, line 51: Nice

packages/platform/storage-object/src/index.ts
```diff
@@ -1,12 +1,2 @@
...
-  {
-    readonly type: "object-storage"
-  }
->()("ObjectStorageAdapterTag") {}
-
-export const objectStorageAdapter = {
-  type: "object-storage" as const,
-}
+export type { StorageDriver, StorageDisk } from "./driver.ts"
+export { createStorageDisk, createStorageDiskEffect } from "./driver.ts"
```
andresgutgon, line 2: Is this used?

.env.example
```diff
@@ -61,6 +61,15 @@ LAT_WEAVIATE_GRPC_PORT=50051
 LAT_REDIS_HOST=localhost
 LAT_REDIS_PORT=6379
 
+# Object Storage
+LAT_STORAGE_DRIVER=fs
+LAT_STORAGE_FS_ROOT=./storage
```
andresgutgon, line 66: Add to `gitignore`

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
csansoon: They're still ordered correctly.

If we use `start_time`, down to the nanosecond, most likely every span will have a unique timestamp, and the index will be too sparsed, making the queries less efficient. Using the hour as frouping will compress better in memory.

However, apparently, it is negligible and sorting is better using `start_time` instead, so I'll change it.

.env.example
csansoon: Well, it depends on the `LAT_STORAGE_DRIVER` value. You don't need these env vars if you have "fs" selected, so why requiring them?

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
csansoon: `ttl_only_drop_parts = 1`

When data expires (after 3 months), ClickHouse needs to move it to cold storage. By default, it opens each data part, checks every row, copies the non-expired rows to a new part, and discards the rest. That's expensive — lots of disk I/O for a table ingesting thousands of spans per second.

With this setting, ClickHouse waits until every row in a part has expired, then moves the entire part as-is. No rewriting, just a move. The trade-off is that some data stays on the hot disk slightly longer than 3 months (until the whole part expires), but you avoid expensive rewrite operations.

`storage_policy = 'tiered'`

Tells ClickHouse to use the two-disk layout defined in docker/clickhouse/storage.xml (hot + cold volumes) instead of just the default single disk. Without this, the TTL line TO VOLUME 'cold' would fail because ClickHouse wouldn't know what "cold" means.

If you removed both settings and the TTL line, everything would just live on one disk forever. These two settings exist purely to enable the "move old data to cheaper storage" feature.

---

Other settings are just the Clickhouse default anyways, so I'll remove them

packages/platform/storage-object/src/index.ts
csansoon: Not for now, will be used when implementing ingestion

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
csansoon: TTL now is 3 months

docker/clickhouse/storage.xml
csansoon: Local clickhouse configuration, required for the hot/cold storage to work locally

apps/web/src/routes/_authenticated/projects/$projectId/spans/$traceId/$spanId/index.tsx
andresgutgon, line 1: Do we need to use these `$` symbols in the folders? In the route is defined in `createFileRoute`

PR #2402, opened 2026-03-04
Fix migration without snapshot.json

andresgutgon:
# What?
Last migration was created without snapshot.json. Not sure how that happened. Also added init-db.sh script for postgresql container and a `pnpm pg:reset_db` to make it easy start clean.


packages/platform/db-postgres/drizzle/20260304105916_add-api-key-token-hash-and-encryption/migration.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "latitude"."api_keys" DROP CONSTRAINT "api_keys_token_key";--> statement-breakpoint
+ALTER TABLE "latitude"."api_keys" ADD COLUMN "token_hash" text NOT NULL;--> statement-breakpoint
+ALTER TABLE "latitude"."api_keys" ADD CONSTRAINT "api_keys_token_hash_key" UNIQUE("token_hash");
```
andresgutgon, line 3: Because I started fresh just did not do backfilling

PR #2405, opened 2026-03-04
feat: enforce RLS at boundaries with simplified tenancy model

geclos:
## Summary

Implements simplified multi-tenancy enforcement at application boundaries with RLS policies on org-scoped tables. runCommand accepts an optional org id parameter to enforce tenancy to all its children queries, use at boundaries.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -27,7 +27,6 @@
     "@repo/observability": "workspace:*",
     "@repo/utils": "workspace:*",
     "@types/pg": "catalog:",
-    "better-auth": "catalog:",
     "drizzle-orm": "catalog:",
     "dotenv": "catalog:",
     "effect": "catalog:",
--- a/apps/api/src/middleware/auth.ts
+++ b/apps/api/src/middleware/auth.ts
@@ -29,6 +29,22 @@ const withTimeout = <T>(operation: Promise<T>, fallback: T): Promise<T> => {
 
 const getApiKeyCacheKey = (tokenHash: string): string => `apikey:${tokenHash}`
 
+const isCachedApiKeyResult = (value: unknown): value is { organizationId: string; keyId: string } | null => {
+  if (value === null) {
+    return true
+  }
+
+  if (typeof value !== "object" || value === null) {
+    return false
+  }
+
+  if (!("organizationId" in value) || !("keyId" in value)) {
+    return false
+  }
+
+  return typeof value.organizationId === "string" && typeof value.keyId === "string"
+}
+
 /**
  * Get cached API key result from Redis (keyed by token hash).
  */
@@ -42,7 +58,7 @@ const getCachedApiKey = (
       const cached = await withTimeout(redis.get(cacheKey), null)
       if (!cached) return undefined
       const parsed = JSON.parse(cached)
-      return parsed as { organizationId: string; keyId: string } | null
+      return isCachedApiKeyResult(parsed) ? parsed : undefined
     },
     catch: () => undefined,
   }).pipe(Effect.orDie)
@@ -110,8 +126,8 @@ const validateApiKey = (
     const apiKey = apiKeyOption.value
 
     const result = {
-      organizationId: apiKey.organizationId as string,
-      keyId: apiKey.id as string,
+      organizationId: apiKey.organizationId,
+      keyId: apiKey.id,
     }
 
     // Cache successful validation for 5 minutes (keyed by hash)
@@ -119,7 +135,7 @@ const validateApiKey = (
 
     // Use TouchBuffer for batched updates instead of fire-and-forget
     // This reduces database writes by 90%+ by batching updates
-    touchBuffer.touch(apiKey.id as string)
+    touchBuffer.touch(apiKey.id)
 
     // Enforce minimum time before returning
     yield* enforceMinimumTime(startTime, MIN_VALIDATION_TIME_MS)
@@ -180,11 +196,13 @@ const authenticateWithApiKey = (c: Context, token: string): Effect.Effect<AuthCo
     const result = yield* validateApiKey(c, token)
 
     if (result) {
-      return {
+      const authContext: AuthContext = {
         userId: UserId(`api-key:${result.keyId}`),
         organizationId: OrganizationId(result.organizationId),
-        method: "api-key" as const,
+        method: "api-key",
       }
+
+      return authContext
     }
 
     return null
@@ -219,11 +237,13 @@ const authenticateWithJwt = (c: Context, token: string): Effect.Effect<AuthConte
       return null
     }
 
-    return {
+    const authContext: AuthContext = {
       userId: UserId(session.user.id),
       organizationId: OrganizationId(orgId),
-      method: "jwt" as const,
+      method: "jwt",
     }
+
+    return authContext
   }).pipe(Effect.orDie)
 }
 
--- a/apps/api/src/middleware/organization-context.ts
+++ b/apps/api/src/middleware/organization-context.ts
@@ -1,10 +1,8 @@
-import type { Organization } from "@domain/organizations"
-import { type OrganizationId, PermissionError } from "@domain/shared"
+import { OrganizationId, PermissionError } from "@domain/shared"
 import { BadRequestError } from "@domain/shared"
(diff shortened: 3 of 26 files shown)


apps/api/src/routes/organizations.ts
```diff
@@ -29,12 +29,12 @@ export const createOrganizationsRoutes = () => {
     }
 
     const auth = c.get("auth") as AuthContext
-
-    const organization = await runCommand(c.get("db"), async (txDb) => {
+    const organizationId = OrganizationId(generateId())
```
andresgutgon: This is not responsability of the client. Is the final `createOrganizationUseCase` case the one responsible of generating a valid ID

andresgutgon: I would say more. Is this not done at DB level by drizzle being a primary key? I guess not with correct cuid format no?

docs/tenancy.md
```diff
@@ -0,0 +1,190 @@
...
+```sql
+-- Disable RLS on non-org tables
+ALTER TABLE "latitude"."organization" DISABLE ROW LEVEL SECURITY;
+ALTER TABLE "latitude"."member" DISABLE ROW LEVEL SECURITY;
+ALTER TABLE "latitude"."user" DISABLE ROW LEVEL SECURITY;
+
+-- Enable and force RLS on org-scoped tables
+ALTER TABLE "latitude"."projects" ENABLE ROW LEVEL SECURITY;
+ALTER TABLE "latitude"."projects" FORCE ROW LEVEL SECURITY;
+-- ... (api_keys, invitation, grants, subscription)
```
andresgutgon: This is for this PR clean this doc

docs/tenancy.md
```diff
@@ -0,0 +1,190 @@
+# Tenancy Enforcement
```
andresgutgon: We can remove this https://github.com/latitude-dev/latitude-llm/blob/latitude-v2/packages/platform/db-postgres/drizzle/20260303113537_create-latitude-schema-and-rls-functions/migration.sql#L8-L13

I don't see it used in this PR

docs/tenancy.md
geclos: actually no, i'm building a set of docs for everyone and agents mostly

docs/tenancy.md
geclos: yes done

apps/web/src/domains/organizations/organizations.collection.ts
```diff
@@ -12,18 +12,6 @@ const organizationsCollection = createCollection(
     queryKey: ["organizations"],
     queryFn: () => listOrganizations(),
     getKey: (item: OrganizationRecord) => item.id,
-    onInsert: async ({ transaction }: { transaction: { mutations: Array<{ modified: OrganizationRecord }> } }) => {
```
andresgutgon, line 15: How is created now?

apps/web/src/domains/organizations/organizations.functions.ts
```diff
@@ -1,80 +1,44 @@
...
+    const role = roleByOrganizationId.get(organization.id)
+    if (!role) {
+      return []
+    }
+
+    return {
+      id: organization.id,
+      name: organization.name,
+      slug: organization.slug,
+      role,
```
andresgutgon, line 41: Is this really helpful? Why i need to know my role in other orgs when I'm in one org?

apps/web/src/domains/projects/projects.functions.ts
```diff
@@ -32,11 +32,14 @@ export const listProjects = createServerFn({ method: "GET" })
...
 
-    const projects = (await Effect.runPromise(
-      projectsRepo.findByOrganizationId(OrganizationId(data.organizationId)),
-    )) as readonly Project[]
+    const projects = await runCommand(
+      db,
+      data.organizationId,
+    )(async (txDb) => {
+      const projectsRepo = createProjectPostgresRepository(txDb)
+      return Effect.runPromise(projectsRepo.findByOrganizationId(OrganizationId(data.organizationId)))
```
andresgutgon: This org.id should not be passed from the frontend. This should be extracted from the request context. Is always the current organizatin

docker/reset-postgres.sh
```diff
@@ -5,18 +5,34 @@ REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
...
+if [ -z "$POSTGRES_VOLUME_NAME" ]; then
+	PROJECT_NAME="${COMPOSE_PROJECT_NAME:-$(basename "$REPO_ROOT")}"
+	POSTGRES_VOLUME_NAME="${PROJECT_NAME}_postgres_data"
+fi
+
+if docker volume inspect "$POSTGRES_VOLUME_NAME" >/dev/null 2>&1; then
+	docker volume rm "$POSTGRES_VOLUME_NAME"
+else
+	echo "Volume '$POSTGRES_VOLUME_NAME' not found, skipping"
+fi
```
andresgutgon, line 28: Was this failing before with `docker volume rm data-llm_postgres_data`?

geclos: yes

apps/web/src/domains/projects/projects.functions.ts
geclos: true

apps/web/src/domains/organizations/organizations.functions.ts
geclos: i don't think it hurts to have it 🤷🏼

apps/web/src/domains/organizations/organizations.collection.ts
geclos: this was never used, just placeholder, org creation happens on auth intent completion

packages/platform/db-postgres/drizzle/20260304150928_dusty_lester/migration.sql
andresgutgon, line 1: I really think migration names are really important to understand history. 

Can you create the migrations with meaningful names? It's a benefit in the long term

Also Can you add this to scripts: 
```typescript
// scripts/migration.ts
// scripts/migration.ts
import { execSync } from "node:child_process";
import readline from "node:readline";

function ask(question: string) {
  const rl = readline.createInterface({ input: process.stdin, output: process.stdout });
  return new Promise<string>((resolve) =>
    rl.question(question, (answer) => {
      rl.close();
      resolve(answer.trim());
    }),
  );
}

function slugify(input: string) {
  return input
    .trim()
    .toLowerCase()
    .replace(/['"]/g, "")
    .replace(/[^a-z0-9]+/g, "-")
    .replace(/^-+|-+$/g, "");
}

async function main() {
  // Allow non-interactive usage:
  // pnpm db:migrate:auto add-users
  // pnpm db:migrate:custom backfill-users
  const modeArg = process.argv[2]; // "auto" | "custom" | undefined
  const nameArg = process.argv[3]; // optional name

  let mode = modeArg;
  if (mode !== "auto" && mode !== "custom") {
    const m = await ask("Migration type ([a]uto / [c]ustom): ");
    mode = m.toLowerCase().startsWith("c") ? "custom" : "auto";
  }

  let name = nameArg ?? (await ask("Migration name: "));
  if (!name) {
    console.error("Migration name required");
    process.exit(1);
  }

  // Normalize name to a safe drizzle-friendly slug
  name = slugify(name);

  const customFlag = mode === "custom" ? "--custom " : "";
  const cmd = `pnpm drizzle-kit generate ${customFlag}--name ${name}`;

  execSync(cmd, { stdio: "inherit" });
}

main().catch((err) => {
  console.error(err);
  process.exit(1);
});
```

And then is
```json
"db:generate": "tsx scripts/migration.ts"
```

andresgutgon: This will force developer to decide between generate auto or custom. and choose a a name

geclos: ehm can you do it? 😅 i'm all for what you say but it's gotten to a point that we are holding hostage important baseline abstractions to improve migration ergonomics that can be done in a separate pr

andresgutgon: sure

PR #2407, opened 2026-03-04
feat: Datasets domain — data model, RFC, and Zod 4 migration

andresgutgon:
## Summary

This PR introduces the **Datasets** bounded context — a new domain for managing versioned evaluation datasets used in LLM testing, prompt engineering, and annotation workflows. It also includes an opportunistic migration from **Zod 3 → Zod 4** across `apps/web`.

### Data Model

The data model splits storage across **Postgres** (control-plane metadata) and **ClickHouse** (high-volume row data), following the same hybrid pattern the platform already uses for spans/telemetry.

#### Postgres (`latitude` schema)

| Table | Purpose |
|---|---|
| `datasets` | Dataset identity, org ownership, soft-delete, monotonic `current_version` counter |
| `dataset_versions` | Immutable audit log — one row per version bump, tracking rows inserted/updated/deleted, source, and actor |

Both tables follow project conventions: `cuid` PKs, `organizationRLSPolicy`, `timestamps()` spread, `tzTimestamp` for nullable dates. Versions are FK'd to datasets with `ON DELETE CASCADE`.

#### ClickHouse (`dataset_rows`)

| Column | Purpose |
|---|---|
| `row_id` | Stable row identity across versions |
| `xact_id` | Monotonic version number (matches Postgres `current_version`) |
| `_object_delete` | Tombstone flag for soft-deletes at the row level |
| `input` / `output` / `metadata` | Schemaless JSON payloads |

**Why this design?**

- **Temporal versioning via `xact_id`**: Every mutation appends a new row with an incremented `xact_id` rather than updating in place. This makes datasets fully versioned — you can query the state of a dataset "as of version N" by filtering `xact_id <= N` and collapsing with `argMax(col, xact_id)`. This is critical for reproducible evaluations: if a dataset changes between two eval runs, you can still compare results against the exact snapshot each run used.
- **ClickHouse for row storage**: Datasets can grow to millions of rows (think production traffic samples or large benchmark suites). ClickHouse's columnar storage and `MergeTree` engine handle this volume efficiently while Postgres would struggle with table bloat from the append-only versioning pattern.
- **Postgres for metadata**: Dataset identity, version audit log, and org ownership stay in Postgres where they benefit from transactions, foreign keys, and RLS — things ClickHouse doesn't support.
- **Partition by `(organization_id, toYYYYMM(created_at))`**: Ensures queries are always org-scoped (multi-tenancy) and time-bounded partitions allow efficient pruning.
- **`ORDER BY (organization_id, dataset_id, row_id, xact_id)`**: Optimised for the primary query pattern: "give me the latest state of all rows in dataset X for org Y".

#### Domain Layer (`@domain/datasets`)

New package with clean ports-and-adapters structure:
- **Entities**: `Dataset`, `DatasetVersion`, `DatasetRow` with branded IDs (`DatasetId`, `DatasetRowId`, `DatasetVersionId`)
- **Ports**: `DatasetRepository` (Postgres) and `DatasetRowRepository` (ClickHouse) as Effect `ServiceMap` services
- **Use-cases**: `createDataset`, `listDatasets`, `insertRows`, `listRows`, `getRowDetail`

#### Shared IDs (`@domain/shared`)

Added `DatasetId`, `DatasetRowId`, and `DatasetVersionId` branded types with factory functions.

### Zod 4 Migration

Upgraded `apps/web` from **Zod 3.24** to **Zod 4.3.6** (added to pnpm catalog). Key changes:

- **Removed `@tanstack/zod-adapter`**: TanStack Start 1.166+ natively supports Zod 4 schemas in `.inputValidator()` without a wrapper adapter. All server functions across `auth`, `api-keys`, `members`, `projects`, `spans`, and the new `datasets` functions now pass Zod schemas directly.
- **No schema syntax changes needed**: Zod 4 is backwards-compatible for the schema patterns used here (`z.object`, `z.string`, `z.number`, `.optional()`, `.default()`, `.nullable()`).

### RFC

Includes `rfcs/filters.md` — a design document for the shared filter grammar that will be used by datasets, annotation queues, and the traces UI to query spans in ClickHouse.

## Test plan

- [x] Run `pnpm check` and `pnpm typecheck` — both pass
- [x] Verify Postgres migration applies cleanly: `pnpm --filter @platform/db-postgres pg:migrate`
- [x] Verify ClickHouse migration applies cleanly: `pnpm --filter @platform/db-clickhouse ch:up`
- [x] Seed data works: `pnpm --filter @platform/db-postgres pg:seed && pnpm --filter @platform/db-clickhouse ch:seed`
- [x] Web app starts and datasets UI renders at `/projects/:id/datasets`
- [x] Confirm `@tanstack/zod-adapter` is fully removed and no import references remain

---

## Next
- [x] Upload datasets from CSV
- [ ] Edit fields in row details and make sure version is updated
- [ ] Delete rows and make sure version is updated
- [ ] Delete datasets
- [ ] Rename datasets
- [ ] Upload dataset from traces UI
- [ ] Dynamic datasets based on filters
- [ ] Download datasets as CSVs
- [ ] Keyset pagination of rows and pattern in our stack. Infinite pagination maybe. 
- [ ] Dataset filters? We need to talk

### DIFF
--- a/apps/web/src/domains/api-keys/api-keys.functions.ts
+++ b/apps/web/src/domains/api-keys/api-keys.functions.ts
@@ -3,7 +3,6 @@ import type { ApiKey } from "@domain/api-keys"
 import { ApiKeyId, OrganizationId } from "@domain/shared"
 import { createApiKeyPostgresRepository, runCommand } from "@platform/db-postgres"
 import { createServerFn } from "@tanstack/react-start"
-import { zodValidator } from "@tanstack/zod-adapter"
 import { Effect } from "effect"
 import { z } from "zod"
 import { requireSession } from "../../server/auth.ts"
@@ -53,7 +52,7 @@ export const listApiKeys = createServerFn({ method: "GET" })
 
 export const createApiKey = createServerFn({ method: "POST" })
   .middleware([errorHandler])
-  .inputValidator(zodValidator(z.object({ name: z.string().min(1).max(256) })))
+  .inputValidator(z.object({ name: z.string().min(1).max(256) }))
   .handler(async ({ data }): Promise<ApiKeyRecord> => {
     const { organizationId } = await requireSession()
     const { db } = getPostgresClient()
@@ -75,7 +74,7 @@ export const createApiKey = createServerFn({ method: "POST" })
 
 export const updateApiKey = createServerFn({ method: "POST" })
   .middleware([errorHandler])
-  .inputValidator(zodValidator(z.object({ id: z.string(), name: z.string().min(1).max(256) })))
+  .inputValidator(z.object({ id: z.string(), name: z.string().min(1).max(256) }))
   .handler(async ({ data }): Promise<ApiKeyRecord> => {
     const { organizationId } = await requireSession()
     const { db } = getPostgresClient()
@@ -97,7 +96,7 @@ export const updateApiKey = createServerFn({ method: "POST" })
 
 export const deleteApiKey = createServerFn({ method: "POST" })
   .middleware([errorHandler])
-  .inputValidator(zodValidator(z.object({ id: z.string() })))
+  .inputValidator(z.object({ id: z.string() }))
   .handler(async ({ data }): Promise<void> => {
     const { organizationId } = await requireSession()
     const { db } = getPostgresClient()
--- a/apps/web/src/domains/datasets/datasets.collection.ts
+++ b/apps/web/src/domains/datasets/datasets.collection.ts
@@ -1,0 +1,107 @@
+import { queryCollectionOptions } from "@tanstack/query-db-collection"
+import type { Context, QueryBuilder, SchemaFromSource } from "@tanstack/react-db"
+import { createCollection, useLiveQuery } from "@tanstack/react-db"
+import { getQueryClient } from "../../lib/data/query-client.tsx"
+import type { DatasetRecord, DatasetRowRecord } from "./datasets.functions.ts"
+import { listDatasetsQuery, listRowsQuery } from "./datasets.functions.ts"
+
+const queryClient = getQueryClient()
+
+const makeDatasetsCollection = (projectId: string) =>
+  createCollection(
+    queryCollectionOptions({
+      queryClient,
+      queryKey: ["datasets", projectId],
+      queryFn: async () => {
+        const result = await listDatasetsQuery({ data: { projectId } })
+        return result.datasets
+      },
+      getKey: (item: DatasetRecord) => item.id,
+    }),
+  )
+
+type DatasetsCollection = ReturnType<typeof makeDatasetsCollection>
+const datasetsCollectionsCache = new Map<string, DatasetsCollection>()
+
+const getDatasetsCollection = (projectId: string): DatasetsCollection => {
+  const cached = datasetsCollectionsCache.get(projectId)
+  if (cached) return cached
+  const collection = makeDatasetsCollection(projectId)
+  datasetsCollectionsCache.set(projectId, collection)
+  return collection
+}
+
+type DatasetsSource = { dataset: DatasetsCollection }
+type DatasetsContext = {
+  baseSchema: SchemaFromSource<DatasetsSource>
+  schema: SchemaFromSource<DatasetsSource>
+  fromSourceName: "dataset"
+  hasJoins: false
+}
+
+export const useDatasetsCollection = <TContext extends Context = DatasetsContext>(
+  projectId: string,
+  queryFn?: (datasets: QueryBuilder<DatasetsContext>) => QueryBuilder<TContext>,
+  deps?: Array<unknown>,
+) => {
+  const collection = getDatasetsCollection(projectId)
+  return useLiveQuery<TContext>(
+    (q) => {
+      const datasets = q.from({ dataset: collection })
+      if (queryFn) return queryFn(datasets)
+      return datasets as unknown as QueryBuilder<TContext>
+    },
+    [projectId, ...(deps ?? [])],
+  )
+}
+
+const makeDatasetRowsCollection = (datasetId: string, search?: string) =>
+  createCollection(
+    queryCollectionOptions({
+      queryClient,
+      queryKey: ["datasetRows", datasetId, search ?? ""],
+      queryFn: async () => {
+        const result = await listRowsQuery({
+          data: {
+            datasetId,
+            ...(search ? { search } : {}),
+            limit: 50,
+            offset: 0,
+          },
+        })
+        return result.rows as DatasetRowRecord[]
+      },
+      getKey: (item: DatasetRowRecord) => item.rowId,
+    }),
+  )
+
+type DatasetRowsCollection = ReturnType<typeof makeDatasetRowsCollection>
+const MAX_ROW_COLLECTIONS = 10
+const rowsCollectionsCache = new Map<string, DatasetRowsCollection>()
+
+const getDatasetRowsCollection = (datasetId: string, search?: string): DatasetRowsCollection => {
+  const cacheKey = `${datasetId}:${search ?? ""}`
+  const cached = rowsCollectionsCache.get(cacheKey)
+  if (cached) {
+    rowsCollectionsCache.delete(cacheKey)
+    rowsCollectionsCache.set(cacheKey, cached)
+    return cached
+  }
+
+  const collection = makeDatasetRowsCollection(datasetId, search)
+  rowsCollectionsCache.set(cacheKey, collection)
+
+  if (rowsCollectionsCache.size > MAX_ROW_COLLECTIONS) {
+    const oldest = rowsCollectionsCache.keys().next().value
+    if (oldest) {
+      rowsCollectionsCache.delete(oldest)
+    }
+  }
+
+  return collection
+}
+
+export const useDatasetRowsCollection = (datasetId: string, search?: string) => {
+  const collection = getDatasetRowsCollection(datasetId, search)
+  return useLiveQuery((q) => q.from({ row: collection }), [datasetId, search])
+}
--- a/apps/web/src/domains/datasets/datasets.functions.ts
+++ b/apps/web/src/domains/datasets/datasets.functions.ts
@@ -1,0 +1,116 @@
+import type { Dataset, DatasetRow } from "@domain/datasets"
+import { DatasetRepository, DatasetRowRepository, listDatasets, listRows } from "@domain/datasets"
+import { DatasetId, DatasetVersionId, OrganizationId, ProjectId } from "@domain/shared"
+import { createDatasetRowClickHouseRepository } from "@platform/db-clickhouse"
+import { createDatasetPostgresRepository, runCommand } from "@platform/db-postgres"
+import { createServerFn } from "@tanstack/react-start"
+import { Effect } from "effect"
+import { z } from "zod"
+import { requireSession } from "../../server/auth.ts"
+import { getClickhouseClient, getPostgresClient } from "../../server/clients.ts"
+import { errorHandler } from "../../server/middlewares.ts"
+
+export interface DatasetRecord {
+  readonly id: string
+  readonly organizationId: string
+  readonly projectId: string
+  readonly name: string
+  readonly description: string | null
+  readonly currentVersion: number
+  readonly latestVersionId: string | null
+  readonly createdAt: string
+  readonly updatedAt: string
+}
+
+type JsonValue = string | number | boolean | null | JsonValue[] | { [key: string]: JsonValue }
+
+export interface DatasetRowRecord {
+  readonly rowId: string
+  readonly datasetId: string
+  readonly input: Record<string, JsonValue>
+  readonly output: Record<string, JsonValue>
+  readonly metadata: Record<string, JsonValue>
+  readonly createdAt: string
+  readonly version: number
+}
+
+const toDatasetRecord = (d: Dataset): DatasetRecord => ({
+  id: d.id,
+  organizationId: d.organizationId,
+  projectId: d.projectId,
+  name: d.name,
+  description: d.description,
+  currentVersion: d.currentVersion,
+  latestVersionId: d.latestVersionId,
+  createdAt: d.createdAt.toISOString(),
+  updatedAt: d.updatedAt.toISOString(),
+})
+
+const toRowRecord = (r: DatasetRow): DatasetRowRecord => ({
+  rowId: r.rowId,
+  datasetId: r.datasetId,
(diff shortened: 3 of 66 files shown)


apps/web/src/domains/api-keys/api-keys.functions.ts
```diff
@@ -3,7 +3,6 @@ import type { ApiKey } from "@domain/api-keys"
 import { ApiKeyId, OrganizationId } from "@domain/shared"
 import { createApiKeyPostgresRepository, runCommand } from "@platform/db-postgres"
 import { createServerFn } from "@tanstack/react-start"
-import { zodValidator } from "@tanstack/zod-adapter"
```
andresgutgon, line 6: No need anymore with Zod v4

apps/web/src/domains/datasets/datasets.functions.ts
```diff
@@ -0,0 +1,120 @@
...
+  })
+
+export const listRowsQuery = createServerFn({ method: "GET" })
+  .middleware([errorHandler])
+  .inputValidator(
+    z.object({
+      datasetId: z.string(),
+      versionId: z.string().optional(),
+      search: z.string().optional(),
+      limit: z.number().default(50),
```
csansoon: [ISSUE] `limit` is fully user-controlled here with no upper bound. Since this ultimately fans out to ClickHouse scans/grouping, a large value can become an expensive query and an easy abuse vector. Can we clamp this in the validator (e.g. `z.number().int().min(1).max(500).default(50)`) and mirror the same guard server-side?

apps/web/src/routes/_authenticated/projects/$projectId/datasets/components/dataset-table.tsx
```diff
@@ -0,0 +1,56 @@
...
+            </TableCell>
+            <TableCell>
+              <Text.H6 className="font-mono truncate max-w-48">{truncateJson(row.output)}</Text.H6>
+            </TableCell>
+          </TableRow>
+        ))}
+      </TableBody>
+    </Table>
+  )
+}
```
geclos, line 56: apps/web/domains are intended for state manegement (read stores and function writes), move components within routes close to where they are used. Also i noticed you are using capitalized camel case for components here. I have no beef with what naming convention we use but let's just one rather than 2, i think components in repo ui use lowercase camelcase but review it and make it consistent please 🙏🏼

apps/web/src/routes/_authenticated/projects/$projectId/datasets/components/row-detail-panel.tsx
```diff
@@ -0,0 +1,53 @@
...
+}: {
+  title: string
+  data: Record<string, unknown>
+  defaultOpen?: boolean
+}) {
+  const [open, setOpen] = useState(defaultOpen)
+
+  return (
+    <div className="flex flex-col gap-2">
+      <button type="button" className="flex items-center gap-1.5 cursor-pointer" onClick={() => setOpen(!open)}>
```
geclos, line 37: aren't using Button from design system is that ok?

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00003_create_dataset_rows.sql
```diff
@@ -0,0 +1,21 @@
+-- +goose NO TRANSACTION
+-- +goose Up
+CREATE TABLE IF NOT EXISTS dataset_rows ON CLUSTER default
```
geclos, line 3: datasets aren't scoped by project?

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00003_create_dataset_rows.sql
```diff
@@ -0,0 +1,21 @@
...
+(
+    organization_id   String,
+    dataset_id        String,
+    row_id            String,
+    xact_id           UInt64,
+    created_at        DateTime64(3, 'UTC') DEFAULT now64(3),
+    _object_delete    Bool DEFAULT false,
+    input             String DEFAULT '{}',
```
geclos: compress it, output and metadata as well

apps/web/src/routes/_authenticated/projects/$projectId/datasets/components/row-detail-panel.tsx
andresgutgon: This ui is not definitive

apps/web/src/domains/datasets/datasets.functions.ts
```diff
@@ -0,0 +1,120 @@
...
+  output: r.output as Record<string, JsonValue>,
+  metadata: r.metadata as Record<string, JsonValue>,
+  createdAt: r.createdAt.toISOString(),
+  version: r.version,
+})
+
+let rowRepoInstance: ReturnType<typeof createDatasetRowClickHouseRepository> | undefined
+const getRowRepo = () => {
+  if (!rowRepoInstance) {
+    rowRepoInstance = createDatasetRowClickHouseRepository(createClickhouseClient())
```
geclos: careful you are creating a new clickhouse client on each request?

apps/web/src/domains/datasets/datasets.functions.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+  return rowRepoInstance
+}
```
geclos: no need for this method the repo is instantiated once

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00003_create_dataset_rows.sql
andresgutgon: This is the rows. The schema is 

```bash
datasets (Postgresql)
  |-> N rows (Clickhouse)
```
Rows are scoped by dataset. 

But yes. I forgot scoping dataset by project. Changing that 🙏

packages/platform/db-postgres/drizzle/20260311134756_dataset-tables/migration.sql
```diff
@@ -0,0 +1,34 @@
...
+	"organization_id" varchar(24) NOT NULL,
+	"project_id" varchar(24) NOT NULL,
+	"name" varchar(256) NOT NULL,
+	"description" text,
+	"file_key" text,
+	"current_version" bigint DEFAULT 0 NOT NULL,
+	"deleted_at" timestamp with time zone,
+	"created_at" timestamp with time zone DEFAULT now() NOT NULL,
+	"updated_at" timestamp with time zone DEFAULT now() NOT NULL,
+	CONSTRAINT "datasets_unique_name_per_project_idx" UNIQUE NULLS NOT DISTINCT("organization_id","project_id","name","deleted_at")
```
andresgutgon, line 12: Unique name by project for datasets (not deleted)

apps/web/src/domains/datasets/datasets.collection.ts
```diff
@@ -0,0 +1,107 @@
...
+const rowsCollectionsCache = new Map<string, DatasetRowsCollection>()
+
+const getDatasetRowsCollection = (datasetId: string, search?: string): DatasetRowsCollection => {
+  const cacheKey = `${datasetId}:${search ?? ""}`
+  const cached = rowsCollectionsCache.get(cacheKey)
+  if (cached) {
+    rowsCollectionsCache.delete(cacheKey)
+    rowsCollectionsCache.set(cacheKey, cached)
+    return cached
+  }
```
geclos, line 89: is it necessary to memoize the stores client-side? i'd just let it recompute

apps/web/src/domains/datasets/datasets.functions.ts
geclos: just call createDatasetRowClickHouseRepository(getClickhouseClient()) in line 118 and remove this indirection

apps/web/src/domains/api-keys/api-keys.functions.ts
geclos: thanks, would have been cool in a separate pr though

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00003_create_dataset_rows.sql
```diff
@@ -0,0 +1,52 @@
+-- +goose NO TRANSACTION
+-- +goose Up
+CREATE TABLE IF NOT EXISTS dataset_rows ON CLUSTER default
+(
+    -- ═══════════════════════════════════════════════════════
+    -- TENANCY & SCOPE
+    -- ═══════════════════════════════════════════════════════
+
+    organization_id          LowCardinality(String)                    CODEC(ZSTD(1)),
+    dataset_id               LowCardinality(String)                    CODEC(ZSTD(1)),
```
geclos: no need to compress all the columns it has a computation cost just compress the big ones, input output metadata

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00003_create_dataset_rows.sql
```diff
@@ -0,0 +1,52 @@
...
+    -- SKIP INDEXES
+    -- ═══════════════════════════════════════════════════════
+
+    INDEX idx_row_id         row_id       TYPE bloom_filter(0.01) GRANULARITY 1
+)
+ENGINE = ReplicatedMergeTree()
+PARTITION BY (organization_id, toYYYYMM(created_at))
+PRIMARY KEY (organization_id, dataset_id)
+ORDER BY (organization_id, dataset_id, row_id, xact_id)
+SETTINGS index_granularity = 8192;
```
geclos: i do not think this is necessary

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00003_create_dataset_rows.sql
```diff
@@ -0,0 +1,52 @@
...
+
+    input                    String DEFAULT ''                        CODEC(ZSTD(3)),
+    output                   String DEFAULT ''                        CODEC(ZSTD(3)),
+    metadata                 String DEFAULT ''                        CODEC(ZSTD(3)),
+
+    -- ═══════════════════════════════════════════════════════
+    -- SKIP INDEXES
+    -- ═══════════════════════════════════════════════════════
+
+    INDEX idx_row_id         row_id       TYPE bloom_filter(0.01) GRANULARITY 1
```
geclos: i do not think we need to specify granularity and the filter percentage, let's just use defaults

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00003_create_dataset_rows.sql
```diff
@@ -0,0 +1,52 @@
...
+    -- ═══════════════════════════════════════════════════════
+    -- SKIP INDEXES
+    -- ═══════════════════════════════════════════════════════
+
+    INDEX idx_row_id         row_id       TYPE bloom_filter(0.01) GRANULARITY 1
+)
+ENGINE = MergeTree()
+PARTITION BY (organization_id, toYYYYMM(created_at))
+PRIMARY KEY (organization_id, dataset_id)
+ORDER BY (organization_id, dataset_id, row_id, xact_id)
```
geclos: missing the project id

packages/platform/db-clickhouse/src/repositories/dataset-row-repository.ts
```diff
@@ -0,0 +1,191 @@
...
+            argMax(output, xact_id) AS output
+          FROM dataset_rows
+          WHERE organization_id = {organizationId:String}
+            AND dataset_id = {datasetId:String}
+            ${versionClause}
+          GROUP BY row_id
+          HAVING argMax(_object_delete, xact_id) = false
+            ${searchClause}
+        )
+      `
```
geclos, line 129: parametrize the queries do not interpolate a pelo, even if we control the client

packages/platform/db-clickhouse/src/repositories/dataset-row-repository.ts
```diff
@@ -0,0 +1,191 @@
...
+          min(created_at) AS created_at,
+          max(xact_id) AS latest_xact_id
+        FROM dataset_rows
+        WHERE organization_id = {organizationId:String}
+          AND dataset_id = {datasetId:String}
+          AND row_id = {rowId:String}
+          ${versionClause}
+        GROUP BY row_id
+        HAVING argMax(_object_delete, xact_id) = false
+        LIMIT 1
```
geclos, line 178: ditto

apps/web/src/domains/datasets/datasets.collection.ts
andresgutgon: Avoid memory leak where many "search" are typed. This way this is capped

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00003_create_dataset_rows.sql
andresgutgon: Rows belongs to dataset. Do we need need to point to project too?

apps/web/src/domains/datasets/datasets.collection.ts
geclos: i don't get what's the memory leak here?

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00003_create_dataset_rows.sql
geclos: hm yeah true

apps/web/src/domains/datasets/datasets.collection.ts
andresgutgon: The issue is specifically with the search dimension in the cache key
`(${datasetId}:${search})`. Each unique search string creates a new createCollection with its own query subscription. A user typing "hello" produces 5 collections: "h", "he", "hel", "hell", "hello" — all living forever in a module-level Map with no GC path since the Map itself is a root reference.

The `MAX_ROW_COLLECTIONS` = 10 cap + LRU reordering (delete-then-set on hit) prevents that unbounded growth. Without it, a user who searches a lot in a single session accumulates orphaned collections indefinitely. The datasetsCollectionsCache on line 24 doesn't have this problem because its key is just projectId — naturally bounded by the number of projects visited.

packages/platform/db-clickhouse/src/repositories/dataset-row-repository.ts
andresgutgon: Already are parameterized

PR #2410, opened 2026-03-05
Implement Clickhouse migration system with Goose

andresgutgon:
# What?
Sets up a complete ClickHouse migration and development data system, mirroring the existing Postgres patterns.

### ClickHouse migration system (@platform/db-clickhouse)
- Adopt hybrid versioning: `ch:create` produces timestamp-named files; `ch:fix`
  converts them to sequential before merging. ch:up applies migrations as-is,
  enforcing that fix happened in a `CI` 
- Add `ch:fix` script (goose fix on both unclustered/ and clustered/ dirs)
- Add `ch:seed` script with 5 sample spans (llm/chain/tool kinds, ok/error
  statuses, with/without token data)
- `ch:reset` now runs seed after migrations

### Safety guards
- docker/reset-postgres.sh and docker/reset-clickhouse.sh now fail immediately
   if NODE_ENV=production
- New CI workflow (ch-migrations.yml) blocks PRs that contain timestamp-named
  migration files (enforces that `ch:fix` was run before merging)

### Postgres migration DX (@platform/db-postgres)
- Replace bare drizzle-kit generate commands with an interactive
  scripts/migration.ts script
- `pg:generate` and `pg:generate:custom` now prompt for a migration name (or
  accept it as a quoted argument); name is slugified automatically


packages/platform/db-clickhouse/clickhouse/migrations/clustered/00001_create_spans.sql
```diff
@@ -35,3 +37,6 @@ ENGINE = ReplicatedReplacingMergeTree(ingested_at)
 PARTITION BY toYYYYMM(started_at)
 PRIMARY KEY (workspace_id, started_at)
 ORDER BY (workspace_id, started_at, trace_id, span_id);
+
+-- +goose Down
```
andresgutgon, line 41: This is how Goose mark the UP/Down SQL

packages/platform/db-clickhouse/src/seeds/spans/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    model: "gpt-4o-mini",
+    cost: null,
+    tokens_prompt: null,
+    tokens_cached: null,
+    tokens_reasoning: null,
+    tokens_completion: null,
+    input: JSON.stringify([{ role: "user", content: "Hello!" }]),
+    output: null,
+    attributes: "{}",
+  },
```
andresgutgon, line 130: This is wrong will be fixed when we have the correct schema

PR #2418, opened 2026-03-10
Add trace-level aggregation

csansoon:
## Summary

Introduces a materialized `traces` table in ClickHouse that continuously aggregates span-level telemetry into trace-level summaries, and replaces the flat span listing in the web UI with a traces-first navigation hierarchy (project → traces → spans → span detail).

### How it works

The `traces` table is an `AggregatingMergeTree` destination table populated by a `traces_mv` materialized view. Every time rows are inserted into `spans`, the MV fires against the newly inserted block, groups by `(organization_id, project_id, trace_id)`, and writes partial aggregate rows into `traces`. ClickHouse merges these partial rows in the background.

Because background merges are asynchronous, multiple partial rows for the same trace may exist at any point. **All read queries must `GROUP BY` and re-aggregate** — using `sum`/`min`/`max` for `SimpleAggregateFunction` columns and `-Merge` combinators for `AggregateFunction` columns.

### Schema

**Group keys** (plain columns):
- `organization_id`, `project_id`, `trace_id`

**SimpleAggregateFunction columns** (plain values, re-aggregate at read):

| Column | Aggregate | Purpose |
|---|---|---|
| `span_count` | `sum(UInt64)` | Number of spans in the trace |
| `error_count` | `sum(UInt64)` | Spans with `status_code = 2` |
| `min_start_time` | `min(DateTime64(9))` | Trace start (also used for partitioning) |
| `max_end_time` | `max(DateTime64(9))` | Trace end |
| `duration_ns` | ALIAS from above | Computed at read time, zero storage cost |
| `overall_status` | `max(Int16)` | Worst status: error > ok > unset |
| `tokens_input/output/cache_read/cache_create/reasoning/total` | `sum(UInt64)` | Aggregated token usage |
| `cost_input/output/total_microcents` | `sum(UInt64)` | Aggregated cost (1 USD = 100M microcents) |
| `tags` | `groupUniqArrayArray(Array(String))` | Union of all span tags |

**AggregateFunction columns** (binary state, finalized with `-Merge` at read):

| Column | Aggregate | Purpose |
|---|---|---|
| `models` | `groupUniqArrayIf(model, model != '')` | Distinct LLM models used |
| `providers` | `groupUniqArrayIf(provider, provider != '')` | Distinct providers used |
| `service_names` | `groupUniqArrayIf(service_name, ...)` | Distinct services |
| `root_span_id` | `argMinIf(span_id, start_time, parent = '')` | ID of the root span |
| `root_span_name` | `argMinIf(name, start_time, parent = '')` | Name of the root span (trace label) |
| `input_messages` | `argMinIf(input_messages, start_time, != '')` | First span's input messages |
| `output_messages` | `argMaxIf(output_messages, end_time, != '')` | Last span's output messages |

### Table configuration

- **Engine**: `AggregatingMergeTree` (unclustered) / `ReplicatedAggregatingMergeTree` (clustered)
- **Partition**: `toYYYYMM(min_start_time)` — monthly, aligned with spans
- **Order**: `(organization_id, project_id, trace_id)`
- **Primary key**: `(organization_id, project_id)` — prefix of ORDER BY for granule pruning
- **Skip index**: `minmax` on `min_start_time` for time-range filtering
- **TTL**: 3 months to cold volume, matching spans lifecycle
- **Codecs**: Delta+ZSTD for timestamps, T64+ZSTD for integers, ZSTD(3) for message blobs

### Example read query

```sql
SELECT
    trace_id,
    sum(span_count)                    AS span_count,
    min(min_start_time)                AS start_time,
    max(max_end_time)                  AS end_time,
    max(overall_status)                AS overall_status,
    sum(tokens_total)                  AS tokens_total,
    sum(cost_total_microcents)         AS cost_total_microcents,
    groupUniqArrayIfMerge(models)      AS models,
    argMinIfMerge(root_span_name)      AS root_span_name
FROM traces
WHERE organization_id = {org:String}
  AND project_id = {proj:String}
GROUP BY organization_id, project_id, trace_id
ORDER BY start_time DESC
LIMIT 50

### CODE REVIEW

File: packages/domain/spans/src/entities/trace.ts [hunk 1]
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * TraceDetail — the point-lookup shape returned by single-trace queries.
+ *
+ * Extends Trace with the first input and last output messages.
+ */
+export interface TraceDetail extends Trace {
+  readonly inputMessages: readonly GenAIMessage[]
+  readonly outputMessages: readonly GenAIMessage[]
+}
```
geclos, line 54: isn't there a spans domain already? i'd just centralize all observability related entities into a single domain

csansoon: They're different tables, requiring different queries. Yes, the traces table is populated automatically from data from `spans`, but it still requires custom aggregated queries.  I think it’s useful to keep this in its own domain, since it needs custom logic.

geclos: i just feel we are gonna end up representing 1:1 table schemas <> domains and it shouldn't be the case? Not sure why traces having custom querying logic justifies having their own domain, any table will have custom querying logic

csansoon: Yeah you're right. I'll move it into the spans domain

PR #2423, opened 2026-03-12
Add OTEL-compatible span ingestion endpoint

csansoon:
## Summary

Adds a `POST /v1/traces` endpoint to the ingestion server that accepts OTLP/JSON and OTLP/Protobuf payloads, authenticates via `Authorization: Bearer <api-key>`, resolves the target project from an `X-Latitude-Project` header, and inserts the resulting spans into ClickHouse.

The endpoint extracts GenAI semantic convention attributes (`gen_ai.system`, `gen_ai.request.model`, token usage, etc.) into dedicated span fields during ingestion. Traces are derived automatically via the existing ClickHouse materialized view — no separate trace write path is needed.

> [!NOTE]
> It is only extracting actual OTEL GenAI attributes. There is no "processing" of alternative standards yet. If the OTEL GenAI attribute is not there exactly as we expect, it is not extracted.
> Custom standard processing will be added to a future PR

Also changed api and ingest server auth to an Authorization header, instead of JWT.

### DIFF
--- a/apps/ingest/src/middleware/project.ts
+++ b/apps/ingest/src/middleware/project.ts
@@ -1,0 +1,38 @@
+import { ProjectId, isNotFoundError } from "@domain/shared"
+import { createProjectPostgresRepository, runCommand } from "@platform/db-postgres"
+import { Effect } from "effect"
+import type { MiddlewareHandler } from "hono"
+import { getPostgresClient } from "../clients.ts"
+import type { IngestEnv } from "../types.ts"
+
+/**
+ * Resolves the project from the X-Latitude-Project header (project ID).
+ * Must run after auth middleware (requires organizationId on context).
+ */
+export const projectMiddleware: MiddlewareHandler<IngestEnv> = async (c, next) => {
+  const projectId = c.req.header("X-Latitude-Project")
+  if (!projectId) {
+    return c.json({ error: "X-Latitude-Project header is required" }, 400)
+  }
+
+  const organizationId = c.get("organizationId")
+  const { db } = getPostgresClient()
+
+  try {
+    const project = await runCommand(
+      db,
+      organizationId,
+    )(async (txDb) => {
+      const repo = createProjectPostgresRepository(txDb)
+      return Effect.runPromise(repo.findById(ProjectId(projectId)))
+    })
+
+    c.set("projectId", project.id as string)
+    await next()
+  } catch (error) {
+    if (isNotFoundError(error)) {
+      return c.json({ error: "Project not found" }, 404)
+    }
+    throw error
+  }
+}
--- a/apps/ingest/src/otlp/proto.ts
+++ b/apps/ingest/src/otlp/proto.ts
@@ -1,0 +1,175 @@
+import protobuf from "protobufjs"
+import type { OtlpExportTraceServiceRequest } from "./types.ts"
+
+/**
+ * Protobuf schema for OTLP ExportTraceServiceRequest.
+ * Field IDs match the official opentelemetry-proto definitions exactly.
+ */
+const root = protobuf.Root.fromJSON({
+  nested: {
+    ExportTraceServiceRequest: {
+      fields: {
+        resourceSpans: { rule: "repeated", type: "ResourceSpans", id: 1 },
+      },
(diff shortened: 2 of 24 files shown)


apps/ingest/src/middleware/project.ts
```diff
@@ -0,0 +1,46 @@
...
+  const projectHeader = c.req.header("X-Latitude-Project")
+  if (!projectHeader) {
+    return c.json({ error: "X-Latitude-Project header is required" }, 400)
+  }
+
+  const organizationId = c.get("organizationId")
+  const { db } = getAdminPostgresClient()
+
+  const result = await Effect.runPromise(
+    Effect.tryPromise({
```
andresgutgon: Do not write custom queries in apps. Use repo

apps/ingest/src/otlp/proto.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (typeof value === "object" && "low" in value && "high" in value) {
+    const long = value as { low: number; high: number; unsigned: boolean }
+    const bigint = BigInt(long.high >>> 0) * BigInt(2 ** 32) + BigInt(long.low >>> 0)
+    return bigint.toString()
+  }
+
+  if (typeof value === "object") {
+    const obj = value as Record<string, unknown>
+    const normalized: Record<string, unknown> = {}
+    for (const key of Object.keys(obj)) {
```
csansoon, line 153: [ISSUE] `protobufjs` Long objects can represent signed `int64` values, but this conversion always treats `low/high` as unsigned and ignores the `unsigned` flag. For negative OTLP `intValue` attributes this will decode to a huge positive integer, corrupting attribute data. Could we switch to `long.toString()` (or `Long.fromValue(value).toString()`) so signedness is preserved?

apps/ingest/src/routes/traces.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export const registerTracesRoute = ({ app }: TracesRouteContext) => {
+  app.post("/v1/traces", authMiddleware, projectMiddleware, async (c) => {
+    const contentType = c.req.header("Content-Type") ?? "application/json"
+
+    let request: OtlpExportTraceServiceRequest
+    try {
+      if (contentType.includes("application/x-protobuf")) {
+        const body = await c.req.arrayBuffer()
```
csansoon, line 23: [ISSUE] This reads and decodes the full request body with no payload-size guard. A large OTLP batch (or malformed oversized payload) can force high memory usage in `arrayBuffer()/json()` and in the subsequent transform/insert path. Can we enforce a max body size (or span count) and return `413` when exceeded?

apps/ingest/src/routes/traces.ts
```diff
@@ -0,0 +1,45 @@
...
+import { transformOtlpToSpans } from "../otlp/transform.ts"
+import type { OtlpExportTraceServiceRequest } from "../otlp/types.ts"
+import type { IngestEnv } from "../types.ts"
+
+interface TracesRouteContext {
+  app: Hono<IngestEnv>
+}
+
+export const registerTracesRoute = ({ app }: TracesRouteContext) => {
+  app.post("/v1/traces", authMiddleware, projectMiddleware, async (c) => {
```
csansoon, line 17: [SUGGESTION] Given this endpoint introduces auth + project resolution + OTLP JSON/Protobuf decoding, could we add an integration test matrix for the main paths (valid JSON, valid protobuf, invalid payload, missing headers, invalid key)? Right now regressions here would be easy to miss because this PR doesn’t add route-level tests.

csansoon (commented):
## CI Status
- ❌ Latest run failed (`Knip`): `pnpm-lock.yaml` is out of sync with `apps/ingest/package.json` (frozen lockfile install fails).

Nice work on wiring OTLP/JSON + OTLP/Protobuf ingestion, auth/project scoping, and ClickHouse insertion end-to-end. I left inline comments for two must-fix risks before merge: signed `int64` protobuf normalization currently loses signedness, and `/v1/traces` has no payload-size guard (DoS surface). I also added a test coverage suggestion for the new endpoint path matrix.

PR #2425, opened 2026-03-12
Defer span processing to background worker via object storage

csansoon:
## Summary

The OTEL ingest endpoint previously parsed, transformed, and wrote spans to ClickHouse synchronously within the HTTP request. As the span processing logic grows more complex, this coupling increases response latency and risks timeouts from OTEL exporters.

This change splits ingestion into two phases: the endpoint now validates the payload, buffers it to object storage, and enqueues a BullMQ job — returning immediately. A new background worker picks up the job, runs the transform, persists to ClickHouse, and cleans up the buffered payload. This keeps the ingest endpoint lightweight and allows processing to retry independently on failure.

The OTLP parsing and transform logic was relocated from `apps/ingest` to `@domain/spans` so both the ingest app and workers app can share it without cross-app imports.


apps/ingest/src/clients.ts
```diff
@@ -44,3 +50,24 @@ export const getRedisClient = (): RedisClient => {
...
+const getRedisConnection = (): RedisConnection => {
+  if (!redisConnectionInstance) {
+    redisConnectionInstance = createRedisConnection()
+  }
+  return redisConnectionInstance
+}
+
+export const getStorageDisk = (): StorageDisk => {
+  if (!storageDiskInstance) {
+    storageDiskInstance = createStorageDisk()
```
andresgutgon, line 64: When using this note we have this to centralize how to store things
https://github.com/latitude-dev/latitude-llm/blob/latitude-v2/packages/domain/shared/src/storage.ts

Add a new namespace for your case

apps/workers/src/workers/span-ingestion.ts
```diff
@@ -0,0 +1,57 @@
...
+    const { request, context, ingestedAt } = JSON.parse(raw) as StoredPayload
+
+    const spans = transformOtlpToSpans(request, {
+      ...context,
+      ingestedAt: ingestedAt ? new Date(ingestedAt) : new Date(),
+    })
+
+    if (spans.length > 0) {
+      const spanRepository = createSpanClickhouseRepository(clickhouseClient)
+      await Effect.runPromise(spanRepository.insert(spans))
```
geclos: this can hard crash, handle it?

PR #2426, opened 2026-03-12
feat: transactions and rls rules

geclos:
## Summary

Implements Row-Level Security (RLS) policies and SQL transaction management across the entire stack, from database schema to API boundaries.

## SQL Transactions

All sql operations are handled by the SqlClient service. SqlClient abstracts the underlying sql platform as well as the RLS enforcement. 

Platforms provide SqlService implementations and app boundaries inject them in the use cases that need them.

1. **Domain Layer** (`@domain/shared`): `SqlClient` interface—platform-agnostic contract
2. **Platform Layer** (`@platform/db-postgres`): `SqlClientLive` implementation with RLS
3. **App Layer** (`apps/*`): Provides `SqlClientLive` layer with organization context

### Usage in Code

**Repositories** use `sqlClient.query()` for single operations:

```typescript
findById: (id) =>
  sqlClient
    .query((db) => db.select().from(projects).where(eq(projects.id, id)))
    .pipe(Effect.flatMap(...))
```

.query checks if in transaction, if not it opens one and enforces RLS, if yes it's a pass-through proxy.

**Use cases** wrap multi-step operations in `sqlClient.transaction()`:

```typescript
export const completeAuthIntentUseCase = (input) =>
  Effect.gen(function* () {
    const sqlClient = yield* SqlClient
    yield* sqlClient.transaction(handleIntentByType(intent, input.session))
  })
```

.transaction checks if in transaction, if not it opens one with RLS enforced and updates the context so that any underlying effects pull from this transaction rather than the global sql client. If in transaction, it's a pass-through proxy.

**Boundary routes** provide the `SqlClient` layer with organization context:

```typescript
// apps/api
const project = await Effect.runPromise(
  createProjectUseCase(input).pipe(
    Effect.provide(ProjectRepositoryLive),
    Effect.provide(SqlClientLive(c.var.postgresClient, c.var.organization.id)),
  )
)

// apps/web
const project = await Effect.runPromise(
  createProjectUseCase({...}).pipe(
    Effect.provide(ProjectRepositoryLive),
    Effect.provide(SqlClientLive(client, organizationId)),
  )
)
```

Boundaries provide the sql client implementation with proper context. Most of the times this means a postgres client + a tenant id (the ord id). For non-rls ops simply pass the admin postgres client. If you pass the regular postgres client without org id, rls will fail on any rls-protected underlying operation.

## Key Behaviors

- Every transaction automatically sets `app.current_organization_id` session variable
- RLS policies filter all queries at the database level
- No nested transactions (pass-through proxy)
- Domain errors propagate through Effect error channel
- Any failure triggers automatic rollback

## Files Changed

- `packages/domain/shared/src/sql-client.ts` — Domain SqlClient interface
- `packages/platform/db-postgres/src/sql-client.ts` — RLS-enabled transaction implementation
- `packages/platform/db-postgres/src/schemaHelpers.ts` — RLS policy helper
- `packages/platform/db-postgres/drizzle/*` — Migration files for RLS function
- `apps/*/routes/*.ts` — Updated routes to provide SqlClient layer
- `apps/web/src/domains/*/*.functions.ts` — Server functions with RLS context
- `packages/**/repositories/*.ts` — Repositories using sqlClient.query()
- `packages/**/use-cases/*.ts` — Use cases using sqlClient.transaction()


apps/api/src/middleware/auth.ts
```diff
@@ -94,51 +97,53 @@ const validateApiKey = (
   c: Context,
   token: string,
   options?: AuthMiddlewareOptions,
-): Effect.Effect<{ organizationId: string; keyId: string } | null, never> => {
+): Promise<{ organizationId: string; keyId: string } | null> => {
```
geclos: review if we can have this return an effect as before

apps/api/src/routes/api-keys.test.ts
```diff
@@ -25,18 +23,16 @@ interface ApiKeysRoutesTestContext extends TestContext {
...
+  const encryptedToken = await Effect.runPromise(encrypt(token, TEST_ENCRYPTION_KEY))
 
-  await db.insert(postgresSchema.apiKeys).values({
-    id,
-    organizationId,
-    token: await Effect.runPromise(encrypt(token, TEST_ENCRYPTION_KEY)),
-    tokenHash,
-    name,
-  })
+  await database.client.query(
```
geclos: use the drizzle client intead of raw sql, even better if you can use postgres package repositories

apps/api/src/routes/api-keys.test.ts
```diff
@@ -108,13 +104,13 @@ describe("API Keys Routes Integration", () => {
...
 
-    const [stillActive] = await database.db
-      .select({ deletedAt: postgresSchema.apiKeys.deletedAt })
-      .from(postgresSchema.apiKeys)
-      .where(eq(postgresSchema.apiKeys.id, tenantBKey.id))
-      .limit(1)
+    // Verify the key still exists and is not soft-deleted using parameterized query
+    const result = await database.client.query("SELECT deleted_at FROM latitude.api_keys WHERE id = $1", [
+      tenantBKey.id,
+    ])
```
geclos: ditto, drizzle client instead of raw sql

apps/api/src/test-utils/create-test-app.ts
```diff
@@ -44,40 +44,38 @@ interface TenantSetup {
...
-    tokenHash: await Effect.runPromise(hashToken(apiKeyToken)),
-    name: "auth-key",
-  })
+  // Insert API key using parameterized query
+  const encryptedToken = await Effect.runPromise(encrypt(apiKeyToken, TEST_ENCRYPTION_KEY))
+  const tokenHash = await Effect.runPromise(hashToken(apiKeyToken))
+  await database.client.query(
+    "INSERT INTO latitude.api_keys (id, organization_id, token, token_hash, name, created_at, updated_at) VALUES ($1, $2, $3, $4, $5, NOW(), NOW())",
+    [authApiKeyId, organizationId, encryptedToken, tokenHash, "auth-key"],
+  )
```
geclos: ditto drizzle client instead of raw sql in tests

apps/api/src/test-utils/in-memory-postgres.ts
```diff
@@ -10,10 +11,29 @@ export interface InMemoryPostgres {
...
 
 const unsafeCast = <T>(value: unknown): T => value as T
 
+const createInMemoryPostgresClient = (postgresDb: PostgresDb): PostgresClient => {
+  const transaction = <T>(fn: (txDb: PostgresDb) => Promise<T>): Promise<T> =>
+    (postgresDb as unknown as { transaction: (fn: (tx: unknown) => Promise<T>) => Promise<T> }).transaction(
+      async (tx) => fn(tx as PostgresDb),
+    )
+
+  const withRls = <T>(organizationId: OrganizationId, fn: (txDb: PostgresDb) => Promise<T>): Promise<T> =>
```
geclos: this method no longer exists, you should be able to remove it

packages/platform/db-postgres/src/sql-client.ts
```diff
@@ -0,0 +1,100 @@
...
+            if (currentTx) {
+              return yield* Effect.tryPromise({
+                try: () => fn(currentTx, organizationId),
+                catch: (error) => toRepositoryError(error, "query"),
+              })
+            }
+
+            return yield* Effect.tryPromise({
+              try: () =>
+                client.transaction(async (tx) => {
```
andresgutgon, line 119: Oh so there is a transaction always when there is a query

geclos: Yes. Otherwise rls checks would fail.

PR #2430, opened 2026-03-13
Reliability system specification, documentation and full task list

neoxelox:
https://linear.app/latitude/project/reliability-d4f1379fa6ce/issues

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,24 +11,25 @@ At a glance: **`apps/*`** own HTTP boundaries (validation, authz, routing to use
 ## How to use this guide
 
 1. Skim the **skill glossary** below and open the skill that matches your task.
-2. Read that skill’s `SKILL.md` in full before editing code in that area.
+2. Read that skill's `SKILL.md` in full before editing code in that area.
 
 Detailed policies, command examples, and code samples live under **`.agents/skills/<skill-name>/SKILL.md`**. Load narrow skills instead of memorizing the entire monorepo at once.
 
-**Index coverage:** The glossary lists **every** skill in `.agents/skills/` (one row per `*/SKILL.md`, **11** total), ordered **alphabetically by folder name**. When you add or remove a skill folder, update this table in the same change.
+**Index coverage:** The glossary lists **every** skill in `.agents/skills/` (one row per `*/SKILL.md`, **12** total), ordered **alphabetically by folder name**. When you add or remove a skill folder, update this table in the same change.
 
 ## Skill glossary
 
 | Skill | Path | Use when |
 | --- | --- | --- |
-| **Architecture and boundaries** | [.agents/skills/architecture-boundaries/SKILL.md](.agents/skills/architecture-boundaries/SKILL.md) | Layering, web vs public API, **app layout** (clients, routes, logging), ports/adapters, **web-standard APIs in domain/shared/utils**, multi-tenancy, DDD layout, anti-patterns |
-| **Background jobs and events** | [.agents/skills/async-jobs-and-events/SKILL.md](.agents/skills/async-jobs-and-events/SKILL.md) | **Queues/workers**, **domain events**, side effects **outside** HTTP handlers, **new platform `*-core` / `*-provider`** dependencies |
+| **Architecture and boundaries** | [.agents/skills/architecture-boundaries/SKILL.md](.agents/skills/architecture-boundaries/SKILL.md) | Layering, web vs public API, **app layout** (clients, routes, logging), ports/adapters, **web-standard APIs in domain/shared/utils**, multi-tenancy, DDD layout, anti-patterns, **machine-facing MCP/API product surfaces** |
+| **Background jobs and events** | [.agents/skills/async-jobs-and-events/SKILL.md](.agents/skills/async-jobs-and-events/SKILL.md) | **Queues/workers**, **domain events**, side effects **outside** HTTP handlers, task payload design, debounce/dedupe, delayed job semantics |
 | **Authentication** | [.agents/skills/authentication/SKILL.md](.agents/skills/authentication/SKILL.md) | **Better Auth**, sessions, web session helpers, org context on session, **`@domain/auth`** flows |
-| **Code style and TypeScript** | [.agents/skills/code-style/SKILL.md](.agents/skills/code-style/SKILL.md) | Biome, imports, strict TS, naming, React file names, generated files |
-| **ClickHouse and Weaviate** | [.agents/skills/database-clickhouse-weaviate/SKILL.md](.agents/skills/database-clickhouse-weaviate/SKILL.md) | Parameterized CH queries, Goose migrations, Weaviate collections/migrations |
-| **Postgres and SqlClient** | [.agents/skills/database-postgres/SKILL.md](.agents/skills/database-postgres/SKILL.md) | Drizzle schema, RLS, SqlClient, migrations (Drizzle Kit), repository mappers |
+| **Code style and TypeScript** | [.agents/skills/code-style/SKILL.md](.agents/skills/code-style/SKILL.md) | Biome, imports, strict TS, naming, **Zod-first shared contracts**, literal-union enums, named constants, generated files |
+| **ClickHouse and Weaviate** | [.agents/skills/database-clickhouse-weaviate/SKILL.md](.agents/skills/database-clickhouse-weaviate/SKILL.md) | Parameterized CH queries, Goose migrations, append-only migration rules, Weaviate collections/migrations |
+| **Postgres and SqlClient** | [.agents/skills/database-postgres/SKILL.md](.agents/skills/database-postgres/SKILL.md) | Drizzle schema, RLS, SqlClient, migrations (Drizzle Kit), no-FK rules, repository mappers |
+| **Documentation and specs** | [.agents/skills/docs/SKILL.md](.agents/skills/docs/SKILL.md) | **`docs/*.md`**, **`specs/*.md`**, durable documentation sync, spec structure, promoting stable knowledge into docs |
 | **Effect and errors** | [.agents/skills/effect-and-errors/SKILL.md](.agents/skills/effect-and-errors/SKILL.md) | `Effect` composition, `Data.TaggedError`, `HttpError`, boundary error handling |
 | **Environment configuration** | [.agents/skills/env-configuration/SKILL.md](.agents/skills/env-configuration/SKILL.md) | **`LAT_*` / `VITE_LAT_*`**, `.env.example`, **`parseEnv` / `parseEnvOptional`** |
 | **Testing** | [.agents/skills/testing/SKILL.md](.agents/skills/testing/SKILL.md) | Vitest layers, PGlite/chdb testkit, **`/testing` package exports**, avoiding `vi.mock` for repositories |
 | **Toolchain and commands** | [.agents/skills/toolchain-commands/SKILL.md](.agents/skills/toolchain-commands/SKILL.md) | Node/pnpm/Turbo/Vitest/Biome, scripts, filters, CI, `.env.*` setup, **Docker Compose, dev servers, Mailpit** |
-| **Web frontend** | [.agents/skills/web-frontend/SKILL.md](.agents/skills/web-frontend/SKILL.md) | `apps/web` UI, TanStack Start, collections, `@repo/ui`, layout, **`useMountEffect` policy** |
+| **Web frontend** | [.agents/skills/web-frontend/SKILL.md](.agents/skills/web-frontend/SKILL.md) | `apps/web` UI, TanStack Start, collections, `@repo/ui`, layout, **`-components/`**, legacy UI reference, **`useMountEffect` policy** |
--- a/docs/annotation-queues.md
+++ b/docs/annotation-queues.md
@@ -1,0 +1,306 @@
+# Annotation Queues
+
+Annotation queues are the managed workflow surface for fast human review over traces.
+
+They exist to aggregate telemetry into a focused annotation UX with as little distraction as possible.
+
+## Queue Types
+
+Queue concepts:
+
+- a queue is conceptually `manual` when it has no filter configured and queue membership is created by explicit insertion rather than by stored filter materialization, always materialized as a trace id
+- a queue is conceptually `dynamic` / `live` when it has a filter configured and is populated incrementally over time from that filter plus optional sampling
+
+The filter field intentionally reuses the same future grammar as `EvaluationTrigger.filter`, but until that grammar is finalized it is stored as a plain string inside queue settings.
+
+Creation sources:
+
+- user-created queues are the mainline queue-management flow
+- every project also gets a default set of system-created manual queues that provide immediate review value before the team has curated its own queues
+
+Assignment:
+
+- a queue may have zero, one, or many assignees
+- assignees must be existing Latitude users from the same organization
+
+## Background Tasks
+
+Queue population uses queue tasks in `@domain/queue`, `@platform/queue-bullmq`, and `apps/workers`.
+
+The main contracts are:
+
+- domain events: `SpanIngested`, `TraceEnded`
+- topic tasks: `live-traces:end`, `live-annotation-queues:curate`, `system-annotation-queues:flag`, `system-annotation-queues:annotate`
+
+Rules:
+
+- queue topics are lower-kebab-case and are the durable routing identity
+- each topic may define several lower-kebab-case task names, and the topic worker dispatches by task name
+- payloads carry ids only; workers re-fetch queue definitions and trace context before acting
+- the `domain-events` worker is a dispatcher only: it publishes downstream queue tasks and never performs queue curation or annotation writes inline
+- dynamic queue materialization stays batched per trace inside `live-annotation-queues:curate`, so the system does not need one queue task per matching dynamic queue
+- system-created queue validation/draft creation is expensive enough to live in `system-annotation-queues:annotate`
+
+## System-Created Default Queues
+
+Every project starts with these system-created manual queues:
+
+### Jailbreaking
+
+- description: attempts to bypass system or safety constraints
+- instructions: use this queue for prompt injection, instruction hierarchy attacks, policy-evasion attempts, tool abuse intended to bypass guardrails, role or identity escape attempts, or assistant behavior that actually follows those bypass attempts. Do not use it for harmless roleplay or ordinary unsafe requests that the assistant correctly refuses.
+
+### Refusal
+
+- description: the assistant refuses a request it should handle
+- instructions: use this queue when the assistant declines, deflects, or over-restricts even though the request is allowed and answerable within product policy and system capabilities. Do not use it when the refusal is correct because the request is unsafe, unsupported, or missing required context or permissions.
+
+### Frustration
+
+- description: the conversation shows clear user frustration or dissatisfaction
+- instructions: use this queue when the user expresses annoyance, disappointment, repeated dissatisfaction, loss of trust, or has to restate/correct themselves because the assistant is not helping. Do not use it for neutral clarifications or isolated terse replies without real evidence of frustration.
+
+### Forgetting
+
+- description: the assistant forgets earlier conversation context or instructions
+- instructions: use this queue when the assistant loses relevant session memory, repeats already-settled questions, contradicts previously established facts, or ignores earlier constraints/preferences from the same conversation. Do not use it for ambiguity that was never resolved or context that the user never provided.
+
+### Laziness
+
+- description: the assistant avoids doing the requested work
+- instructions: use this queue when the assistant gives a shallow partial answer, stops early without justification, refuses to inspect provided context, or pushes work back onto the user that the assistant should have done itself. Do not use it when the task is genuinely blocked by missing access, missing context, or policy constraints.
+
+### NSFW
+
+- description: sexual or otherwise not-safe-for-work content appears
+- instructions: use this queue when the trace contains sexual content, explicit erotic material, or other clearly NSFW content that should be reviewed. Do not use it for benign anatomy or health discussion, mild romance, or safety-oriented policy discussion that is not itself NSFW.
+
+### Tool Call Errors
+
+- description: a tool call failed or returned an error state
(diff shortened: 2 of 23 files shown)


AGENTS.md
```diff
@@ -106,6 +108,47 @@ When a Node-specific API is genuinely required outside these scopes, document th
 - Introducing application env vars without the `LAT_` prefix (see Environment Variables)
 - Using `"use client"` or `"use server"` directives — these are Next.js-specific; the web app uses TanStack Start
 
+### Documentation and Specs
+
+- `docs/` contains precise and exhaustive Markdown documentation for each domain or system.
+- Each docs file should cover how the domain or system is architected, built, designed, and structured.
+- Keep docs organized by domain/system (for example: `docs/reliability.md`, `docs/evaluations.md`, `docs/annotations.md`, `docs/scores.md`, `docs/issues.md`, `docs/simulations.md`, `docs/organizations.md`, `docs/projects.md`, `docs/users.md`, `docs/settings.md`, `docs/spans.md`).
+- `specs/` is a temporary folder used while a feature or system is under construction.
+- Each spec/PRD in `specs/` should have its own Markdown file and usually include a task list to track progress (for example: `specs/reliability.md`).
```
geclos: this reference is gonna get stale

RELIABILITY.md
```diff
@@ -0,0 +1,1043 @@
...
+- The dataset that was used to run this simulation. Either the dataset name (with a link to the its specific dashboard), the full dynamic query (highlighted as sql code) or "Custom" if it was custom function.
+- Metadata of the simulation run like threshold, file, sdk...
+- If spans/traces/sessions were captured then: the table of spans/traces/sessions that are related to this simulation run, with the following columns: Timestamp, Score (Passed (green) / Failed (red) / Error (yellow) duotone tag, the right part of the tag will be the value with max 2 decimal places.), Preview (input/output truncated short preview) and Duration. This table will have keyset pagination, and will be sorted by timestamp descending.
+- Otherwise, if there was no instrumentation and spans/traces/sessions were not captured then we will just show directly the list of Scores that are related to this simulation run, with the following columns: Timestamp, Score (Passed (green) / Failed (red) / Error (yellow) duotone tag, the right part of the tag will be the value with max 2 decimal places.) and Duration. This table will have keyset pagination, and will be sorted by timestamp descending.
```
geclos: why is this file in root, also shouldb't this be in docs?

neoxelox: this was my original query, i kept it temporarily meanwhile I was checking the spec with my initial proposal, its now deleted, the spec is verified now

AGENTS.md
neoxelox: currently i think its very well synced, lets see what happens, we can delete all docs at any time... but in the docs appears some info that cannot be understood by the code alone i think they are important?... maybe not

docs/annotation-queues.md
```diff
@@ -0,0 +1,179 @@
...
+
+- a queue is conceptually `manual` when it has no filter configured and is populated by explicitly selected traces
+- a queue is conceptually `dynamic` when it has a filter configured and is populated over time from that filter plus optional sampling
+
+The filter field intentionally reuses the same future grammar as `EvaluationTrigger.filter`, but until that grammar is finalized it is stored as a plain string inside queue settings.
+
+Creation sources:
+
+- user-created queues are in MVP
+- system-created queues, including default sentiment-based queues, are deferred until post-MVP
```
geclos: I think this is in the mvp

geclos: also a bit weird that docs refer to timeline events such as "post mvp" aren't they supposed to be snapshot documentation of what's actually implemented?

docs/annotation-queues.md
```diff
@@ -0,0 +1,179 @@
...
+- do not add GIN/array indexes on `assignees`, do not add GIN/JSONB indexes on `settings`, and do not add text indexes on `description`, `instructions`, or `settings.filter` until queue filtering/search workloads are clearer
+
+### Queue Items
+
+`annotation_queue_items` materialize the actual review backlog for both manual and dynamic queues.
+
+Fields:
+
+- `queue_id`
+- `trace_id`: queued trace id
```
geclos: can't queue items also be sessions rather than traces?

docs/scores.md
```diff
@@ -0,0 +1,158 @@
...
+- annotation scores store raw feedback plus the minimal GenAI anchor fields needed to reopen the selection
+- custom scores store arbitrary user-defined metadata
+
+The metadata field is not intended for heavy analytical querying.
+
+## ClickHouse Layout
+
+Recommended initial physical layout:
+
+- `ReplacingMergeTree(updated_at)`
```
geclos: I'd just use merge tree we can safely assume we won't produce score duplicates and replacing merge tree has a cpu cost/price overhead we want to avoid if possible

docs/scores.md
```diff
@@ -0,0 +1,158 @@
...
+- evaluation dashboards
+- annotations
+- simulations
+- long-term analytics
+
+## Canonical Store
+
+Canonical score history lives in ClickHouse.
+
+Postgres owns configuration and lifecycle state, not the score fact stream itself.
```
geclos: i don't understand you mention postgres but i do not see any reference to any postgres table in this doc

docs/settings.md
```diff
@@ -0,0 +1,93 @@
...
+- `false`: archive the issue-linked evaluation
+- when `ProjectSettings.keepMonitoring` is undefined, it falls back to `OrganizationSettings.keepMonitoring`
+- when a user manually resolves an issue, this setting supplies the default state for the confirmation-modal toggle that decides whether linked evaluations keep monitoring that issue for future regressions
+
+Manual ignore behavior is separate:
+
+- ignoring an issue always archives linked evaluations immediately
+- `keepMonitoring` does not affect the manual ignore path
+
+The user may still override the toggle for that specific manual resolve action.
```
geclos: don't understand what these settings are

docs/issues.md
```diff
@@ -0,0 +1,356 @@
...
+
+`Seen at` should combine recency and age, for example `11d ago / 3y old`.
+
+Issue details include:
+
+- full name and description
+- linked evaluations with derived alignment (MCC)
+- last seen / first seen / occurrence counts
+- 30-day trend chart
+- trace/session drilldown table
```
geclos, line 391: huge spec can't read not enough ram in my head we can discuss it when it comes time to implement issues

docs/simulations.md
```diff
@@ -0,0 +1,143 @@
...
+- dataset reference
+- spans/traces/sessions when instrumentation exists
+- direct scores when instrumentation does not exist
+
+## Post-MVP
+
+Deferred after MVP:
+
+- query-backed datasets
+- additional simulation SDK languages such as Python, Ruby, PHP, and Go
```
geclos: hmm why are we documenting simulations aren't it fully post-mvp?

specs/reliability.md
```diff
@@ -0,0 +1,2336 @@
...
+- support simulation-driven validation before shipping changes
+- remain performant enough for users that evaluate every trace/session
+
+The system uses:
+
+- PostgreSQL for configuration, lifecycle, and relational ownership data
+- ClickHouse for canonical scores and analytical queries
+- Weaviate for issue search and clustering projections
+
+Implementation constraints for this repository:
```
geclos, line 32: we shouldn't be repeating repo-wide patterns in each spec

specs/reliability.md
```diff
@@ -0,0 +1,2336 @@
...
+      .$type<ProjectSettings>()
+      .notNull()
+      .default(sql`'{}'::jsonb`),
+    deletedAt: tzTimestamp("deleted_at"), // existing project soft-delete field
+    lastEditedAt: tzTimestamp("last_edited_at").notNull().defaultNow(), // existing project metadata field
+    ...timestamps(),
+  },
+  () => [organizationRLSPolicy("projects")],
+);
+```
```
geclos, line 275: seems unnecessary to declare the table details here

specs/reliability.md
```diff
@@ -0,0 +1,2336 @@
...
+
+This includes provider/model configuration and provider credentials, which stay embedded in owner `settings` JSONB rather than moving into separate relational tables.
+
+Important repository-specific notes:
+
+- `projects.settings` lives on an organization-scoped table and should use existing RLS protections.
+- `organization.settings` and `user.settings` live on Better Auth-owned tables, so they are protected by boundary auth/membership rules rather than `organizationRLSPolicy`.
+- Provider API keys must be encrypted at rest using the repository crypto helpers. The spec does not lock a new JSON envelope format; it only requires encrypted-at-rest behavior.
```
geclos: nitpick this ain't encryption at rest this is app-level encryption all data is already encrypted at rest by db

docs/annotation-queues.md
neoxelox: yeah, docs shouldn't talked about MVP, they should document everything.

About system-created queues being in the MVP, im not sure, because cesar gave that "task" to will and im not sure if they were thinking to do it now

specs/reliability.md
```diff
@@ -0,0 +1,2336 @@
...
+
+Evaluations generated from issues are the mainline flow of the reliability system.
+
+To generate or realign an evaluation for an issue, collect enough:
+
+1. positive examples where human annotations indicate the issue being aligned is present (that means failed annotation scores linked to that specific issue, not just any failed annotation score)
+2. negative examples where the issue is absent, using this priority order:
+   1. conversations with no failed scores and at least one passed annotation
+   2. conversations with no failed scores
+   3. conversations with scores, either passed or failed, but unrelated to the issue we are trying to align for
```
geclos: we discussed this offline these requirements are just too much we should be able to generate an eval with a single manual annotation from the user, be it positive or not.

specs/reliability.md
```diff
@@ -0,0 +1,2336 @@
...
+- TypeScript should orchestrate the domain-specific pipeline
+- the Python RPC optimizer process can be reused/adapted for GEPA
+- the evaluator and proposer feedback loop from v1 should be studied and adapted for scripts instead of prompts
+- the transport can remain JSON-RPC over stdio if that remains the cleanest fit
+
+Concrete v1 architecture notes worth carrying forward:
+
+- v1 was a queued lifecycle, not a one-shot function call: `start -> prepare -> execute -> validate -> end`, with explicit timestamps, job boundaries, status events, and cancellation across phases
+- TypeScript owned dataset/example curation, candidate execution, candidate evaluation, proposer prompting, persistence, and cancellation; Python only ran the GEPA search loop
+- the workers runtime bundled the Python engine inside the same container image, and TypeScript spawned it as a child process with `python -m app.main`
```
geclos: idea:  we could implement proper abstractions to support multiple script execution runtimes and use it for this rather than bundling python. For example in our case we could add support for lambda execution and we use it to both execute scripts and gepa optimizations

docs/issues.md
neoxelox: buah pues preparate para el spec de verdad, esto solo eran los docs 😂

specs/reliability.md
```diff
@@ -0,0 +1,2336 @@
...
+
+They allow users to test LLM application changes before shipping them by:
+
+- running the agent locally or in CI
+- capturing instrumented traces when available
+- running Latitude evaluations locally
+- running user-defined custom code evaluations
+- pushing results back to Latitude for inspection
+
+For MVP, the simulation runner and SDK are in scope for JavaScript/TypeScript. Additional languages are post-MVP.
```
geclos: i don't think that's the case actually

docs/scores.md
neoxelox: Probably the LLM was talking about the reliability system as a whole. Full tables, including postgres with drizzle and clickhouse with sql migs are in the spec hehe

docs/scores.md
neoxelox: we will produce score duplicates, because issue discovery is gonna be async no?

specs/reliability.md
```diff
@@ -0,0 +1,2336 @@
...
+- [ ] **P0-6**: Document the score/telemetry analytics query requirements that later ClickHouse materializations must satisfy, but defer the exact materialized-table definitions until a later phase.
+- [ ] **P0-7**: Add the Weaviate `Issues` collection definition with the required BM25, vector, and multi-tenancy settings.
+- [ ] **P0-8**: Introduce platform/domain interfaces for text generation, embeddings, reranking, and optimizer orchestration without yet shipping the full sandbox runtime.
+- [ ] **P0-9**: Define the configurable constants for discovery thresholds, centroid decay/weights, evaluation cadence, denoising visibility, simulation reporting defaults, and other magic values inside the owning domain packages instead of scattering inline literals across the codebase.
+- [ ] **P0-10**: Document migration impacts on existing spans/traces storage and ensure the spec/docs are authoritative enough for parallel implementation agents.
```
geclos: i don't think this approach is correct we are generating the full data model for all features instead of implementing vertical slices what we should do:

- build the scores vertical, including data model, implementation, wiring with traces, UI
- build the annotations vertical, including ... 
- build the annotation queues vertical, including ...
- build the issues vertical, including ...


This helps code review massively.

AGENTS.md
geclos: yeah no i mean the explicit example reference to specs/reliability.md

docs/settings.md
neoxelox: The code comments along the settings field did not help? Full details are in the spec, I will prompt the LLM to be more in depth for the settings for the docs

docs/simulations.md
neoxelox: are they fully post mvp?

docs/scores.md
geclos: ok but make the docs make sense also

docs/annotation-queues.md
geclos: cc @cesr que pasa con esto ☝🏼

specs/reliability.md
neoxelox: I can move them to AGENTS.md

specs/reliability.md
neoxelox: why? anyway, all these will be implemented in phase 0 in my PR. I wanted to specifically define the tables. Types and data structures are everything in programming in my opinion. Business logic is just moving data around.

specs/reliability.md
neoxelox: app level encryption, gotcha

specs/reliability.md
neoxelox: ? and it will be like that. this list is just enumerating the levels of "quality" the system will search for. If it does not have the "target" amount it will search for the next "quality".

specs/reliability.md
neoxelox: mmmm this "engine" is not to execute the evaluations, is to execute the optimization... i'm not sure the advantage to complicate the optimization engine to be able to be executed in a lambda func.... (the optimization takes lots of time).

If you were talking about creating an abstract evaluation scripts execuntion runtime so it can be executed in local or lambda then we can do it post mpv i guess (pre MVP we are not gonna even have scripts runtime).

specs/reliability.md
neoxelox: i will check

specs/reliability.md
neoxelox: mmmm ok, lets do phases per domain. However, I would still create all data models in phase 0. Because everything is interconnected and it would be simpler to use mock data (for example for evaluations with issues), than then in later phases adding the changes necessary on top of the evaluations (for example) to add the concept of "issues". I think the all "concepts" should be from day 0

docs/annotation-queues.md
cesr: confirmed it's in the mvp 👍

specs/reliability.md
cesr: hmm generating an eval with a single manual annotation? it won't be very accurate though, no?

docs/annotation-queues.md
```diff
@@ -0,0 +1,304 @@
...
+They exist to aggregate telemetry into a focused annotation UX with as little distraction as possible.
+
+## Queue Types
+
+Queue concepts:
+
+- a queue is conceptually `manual` when it has no filter configured and queue membership is created by explicit insertion rather than by stored filter materialization, always materialized as a trace id
+- a queue is conceptually `dynamic` / `live` when it has a filter configured and is populated incrementally over time from that filter plus optional sampling
+
+The filter field intentionally reuses the same future grammar as `EvaluationTrigger.filter`, but until that grammar is finalized it is stored as a plain string inside queue settings.
```
andresgutgon, line 14: Not sure what this means. Do not answer. I'll delete if I understand after reading all the specs

docs/annotation-queues.md
```diff
@@ -0,0 +1,304 @@
...
+The queue model is backed by two Postgres tables:
+
+- `annotation_queues`: queue definition and review instructions
+- `annotation_queue_items`: materialized queue membership plus completion state
+
+### Queue Definition
+
+`annotation_queues` fields:
+
+- `system`: boolean marker for system-created queues
```
andresgutgon, line 154: `is_system`. Booleans convention

docs/annotation-queues.md
```diff
@@ -0,0 +1,304 @@
...
+- `name`: unique within the project among non-deleted queues
+- `description`: short summary for the queue list
+- `instructions`: reviewer guidance shown in the focused annotation screen and reused as classifier/validator context for system-created queues
+- `settings.filter`: plain-string filter expression for dynamic queues; it stays absent and read-only for system queues
```
andresgutgon, line 158: Not sure we should store filters in plain text. Maybe would be better to have a serializable JSON that can be converted in SQL or in URL string or any other format

docs/annotation-queues.md
```diff
@@ -0,0 +1,304 @@
...
+- `settings.sampling`: optional queue sampling percentage `[0, 100]`; dynamic queues use it after filter matching, and system queues use it before any deterministic or LLM flagging work. When omitted on queue creation/provisioning, initialize it from named constants with initial defaults of `10`
+- `assignees`: array of existing Latitude user ids from the same organization
```
andresgutgon, line 160: We should model this with an intermedary table. `member <- member_id, queue_id > queue`

docs/annotation-queues.md
```diff
@@ -0,0 +1,304 @@
...
+- the create modal edits `name`, `description`, `instructions`, `assignees`, and the optional `settings.filter` / `settings.sampling` fields for user-created queues
+- the edit modal keeps `name`, `description`, `instructions`, and `settings.filter` read-only for `system = true` queues while still allowing `assignees` and `settings.sampling` updates
+- when a queue is created with `settings.filter` and no explicit sampling, the UI/server path initializes `settings.sampling` from the named default constant
+- when a system queue is provisioned for a project, the UI/server path initializes `settings.sampling` from the named default constant and later lets users tune that sampling per queue
+- manual queue insertion entry points live in both the trace dashboard table and the sessions dashboard table
+- the list includes both user-created queues and the project's default system-created queues
+
+Pagination:
+
+- limit/offset
```
andresgutgon, line 242: Can we do infinite keyset to unify tables?

docs/evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+The stored `script` field contains the body of a host-controlled JavaScript-like evaluation function.
+
+The contract should stay aligned with the proposal:
+
+- `Passed(score?, feedback)` and `Failed(score?, feedback)` always require feedback
+- if present, the score value is passed before the feedback
+- `llm(prompt, options?)` accepts an optional configuration object
+- `parse(value, schema)` validates an unknown value against a schema
+- the stored script body evaluates a conversation and returns a `Score`
+- `zod` is available inside the host-controlled runtime
```
andresgutgon, line 28: Be careful with making DB persisted code dependent on external libs. If we were migrating zod from v3 with v4 would be a little pain. Let's abstract validations in our own surface

docs/evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+## Evaluation Model
+
+```typescript
+type EvaluationSettings = {
+  provider?: string; // if not provided, resolution falls back through project settings and then organization settings
+  model?: string; // if not provided, resolution falls back through project settings and then organization settings
+};
+
+type EvaluationTrigger = {
+  filter: string; // runs on traces that match this filter
```
andresgutgon, line 58: Not happy with this. Let's review

docs/evaluations.md
```diff
@@ -0,0 +1,327 @@
...
+  turn: "first" | "every" | "last"; // runs on the first, every, or last ingested trace/turn
+  debounce: number; // debounce time in seconds
```
andresgutgon, line 60: `debounce_in_seconds` or `debounce_ms`

AGENTS.md
```diff
@@ -793,15 +885,20 @@ Mocking repositories with `vi.fn()` tests the wiring, not the queries. In-memory
...
-4. Keep domain unchanged unless business behavior changes
+1. Add concrete provider package in `packages/platform/*-<provider>`
+2. Wire via app composition root and environment-driven config
+3. Keep domain unchanged unless business behavior changes
 
 ## Frontend Guidelines (Web App)
 
 When working on `apps/web` or any frontend code:
 
+### Legacy UI Reference
```
andresgutgon: Not sure this is a good idea. Just move the UI component you need and let your LLm it can use it

specs/reliability.md
```diff
@@ -0,0 +1,2578 @@
...
+  keepMonitoring?: boolean // if true, issue-linked evaluations keep running after resolution; if false they are archived
+}
+
+type OrganizationSettings = {
+  providers: {
+    [name: string]: {
+      apiKey: string // persisted using application-level encryption via repository crypto helpers
+      ...
+    }
+  }
```
andresgutgon, line 227: This should be a schema `llmProviders` in each own. Why a json schema inside organization?

Each provider has their own way of setting it. We can just pick what we already had in v1 no?

specs/reliability.md
```diff
@@ -0,0 +1,2578 @@
...
+- organization-wide
+- project-wide
+- user-wide
+
+The settings shapes should stay very close to the original proposal:
+
+```typescript
+type UserSettings = {
+  ... // User-wide settings. Still pending precise definition.
+}
```
andresgutgon, line 213: Do we really need user settings?

PR #2434, opened 2026-03-14
feat: add Helm chart and Dockerfile for deploying Latitude to Kubernetes

geclos:
Adds a production-ready Helm chart under charts/latitude/ that deploys all four services: web (TanStack Start SSR), api (Hono HTTP), ingest (telemetry ingestion), and workers (Redpanda background jobs).

Includes a multi-stage Dockerfile with per-service build targets (api, web, ingest, workers, migrations) and a health endpoint for the web app.

Key design decisions:
- All four app services deployed with health probes, optional HPA/PDB/Ingress
- Secret management supports both inline values (dev/test) and an external pre-existing Kubernetes Secret (production) via existingSecret
- ConfigMap and Secret cover all LAT_* env vars including Kafka/Redpanda, object storage (S3), Weaviate, ClickHouse migration vars, and admin DB URL
- Pre-install/pre-upgrade migration Job runs Postgres (drizzle-kit), ClickHouse (goose), and Weaviate migrations before app pods roll out
- Workers deployment uses extended terminationGracePeriodSeconds for in-flight job completion
- Pod annotations include config/secret checksums to trigger rolling restarts on configuration changes
- VITE_LAT_* vars documented as build-time (client bundle) with SSR runtime fallback

Made-with: Cursor

### DIFF
--- a/apps/workers/package.json
+++ b/apps/workers/package.json
@@ -17,6 +17,7 @@
     "@domain/spans": "workspace:*",
     "@platform/db-clickhouse": "workspace:*",
     "@platform/db-postgres": "workspace:*",
+    "@platform/env": "workspace:*",
     "@platform/events-outbox": "workspace:*",
     "@platform/queue-redpanda": "workspace:*",
     "@repo/observability": "workspace:*",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -410,6 +410,9 @@ importers:
       '@platform/db-postgres':
         specifier: workspace:*
         version: link:../../packages/platform/db-postgres
+      '@platform/env':
+        specifier: workspace:*
+        version: link:../../packages/platform/env
       '@platform/events-outbox':
         specifier: workspace:*
         version: link:../../packages/platform/events-outbox


.env.example
```diff
@@ -66,6 +65,12 @@ LAT_WEAVIATE_GRPC_PORT=50051
 LAT_REDIS_HOST=localhost
 LAT_REDIS_PORT=6379
 
+# ClickHouse
+LAT_CLICKHOUSE_URL=http://localhost:8123
+LAT_CLICKHOUSE_USER=latitude
+LAT_CLICKHOUSE_PASSWORD=secret
+LAT_CLICKHOUSE_DB=latitude_development
```
csansoon, line 72: We are repeating `CLICKHOUSE_USER`, `CLICKHOUSE_PASSWORD` and `CLICKHOUSE_DB`, both with and without the `LAT_` prefix. Could we avoid having repeated env vars?

apps/web/package.json
```diff
@@ -41,6 +44,7 @@
     "better-auth": "catalog:",
     "effect": "catalog:",
     "lucide-react": "^0.481.0",
+    "nitro": "3.0.260311-beta",
```
andresgutgon, line 47: Is this needed?

geclos: yes it's the recommended way for production builds of tanstack

.env.example
geclos: we either break the LAT convention of env vars or there is no other way

PR #2440, opened 2026-03-16
Download datasets

andresgutgon:
## TODO
- [x] Implement scheduled download
- [x] Check email preview
- [x] Add tests


apps/web/src/domains/datasets/datasets.functions.ts
```diff
@@ -149,9 +155,91 @@ export const listRowsQuery = createServerFn({ method: "GET" })
...
+
+    const orgId = OrganizationId(organizationId)
+    const dataset = await Effect.runPromise(
+      Effect.gen(function* () {
+        const repo = yield* DatasetRepository
+        return yield* repo.findById(DatasetId(data.datasetId))
+      }).pipe(withPostgres(DatasetRepositoryLive, getPostgresClient(), orgId)),
+    )
+    const limit = DATASET_DOWNLOAD_DIRECT_THRESHOLD + 1
+    const result = await Effect.runPromise(
```
andresgutgon: For fully scalable solution do a count check and if bellow threshold do the list

apps/web/src/domains/datasets/datasets.functions.ts
```diff
@@ -149,9 +155,91 @@ export const listRowsQuery = createServerFn({ method: "GET" })
...
+              "project-id": payload.projectId,
+            },
+          },
+        ],
+      })
+      return { type: "enqueued" }
+    }
+
+    const { csv, filename } = buildDatasetCsvExport(dataset.name, result.rows)
+    return { type: "direct", csv, filename }
```
andresgutgon, line 241: Move all of this to domain and add tests

packages/domain/datasets/src/export-csv.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Returns CSV lines for the given rows without a header. Concatenate with
+ * csvExportHeader() + rowsToCsvFragment(batch1) + rowsToCsvFragment(batch2) + ...
+ * for streaming export without holding all rows in memory.
+ */
+export function rowsToCsvFragment(rows: readonly DatasetRow[]): string {
+  if (rows.length === 0) return ""
+  const csvData = rowsToCsvData(rows)
+  return Papa.unparse(csvData, { columns: [...CSV_EXPORT_COLUMNS], header: true })
```
csansoon, line 56: This breaks the worker-side CSV assembly: `dataset-export.ts` starts with `csvExportHeader()` and then appends `rowsToCsvFragment(...)` for each batch, but `rowsToCsvFragment` still emits its own header (`header: true`). That gives repeated headers for every batch, and because `csvExportHeader()` does not include a trailing newline the first chunk becomes `input,output,metadatainput,output,metadata...`. We should either make `rowsToCsvFragment` emit body rows only (`header: false`) or stop prepending `csvExportHeader()`, and add a test that covers a multi-batch export.

andresgutgon: Changed

PR #2441, opened 2026-03-16
Extract sessionId, userId, tags, metadata from OTLP spans into traces

csansoon:
## Summary

Spans now have two additional columns:
 - `userId`: Used by several platforms. We support values added for Langfuse, OpenInference, Arize Phoenix, Traceloop,  OpenLLMetry, LangSmith, OpenLIT.
 - `metadata`: Supports attributes set by Braintrust, OpenInference, HoneyHive, Langfuse, Traceloop, OpenLLMetry, LangSmith, Datadog.

Additionally, traces now have these additional columns, aggregated from its spans:
- `sessionId`
- `userId`
- `metadata`
- `systemInstructions`
- `allMessages`

---

These values have also been added to the dataset's metadata when importing traces to a dataset.


packages/domain/spans/src/entities/span.ts
```diff
@@ -15,6 +15,7 @@ export interface Span {
   readonly organizationId: OrganizationId
   readonly projectId: ProjectId
   readonly sessionId: SessionId
+  readonly userId: UserId
```
andresgutgon: This is not our user Id no?

ATTRIBUTES.md
```diff
@@ -0,0 +1,101 @@
+Here's the exact mapping for all four across every platform that defines them:
```
andresgutgon: 🔥 no?

packages/domain/spans/src/entities/span.ts
csansoon: nono, this is an arbitrary userid that can be found in the span. I'll change the type to avoid using the same one lol

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/20260316145228_user_id_and_metadata.sql
```diff
@@ -0,0 +1,124 @@
...
+    sum(tokens_total)                                               AS tokens_total,
+
+    sum(cost_input_microcents)                                      AS cost_input_microcents,
+    sum(cost_output_microcents)                                     AS cost_output_microcents,
+    sum(cost_total_microcents)                                      AS cost_total_microcents,
+
+    argMaxIfState(session_id, start_time, session_id != '')          AS session_id,
+    argMaxIfState(user_id, start_time, user_id != '')                AS user_id,
+    groupUniqArrayArray(tags)                                       AS tags,
+    maxMap(metadata)                                                 AS metadata,
```
csansoon, line 59: [ISSUE] `maxMap(metadata)` does not implement the "last span wins" behavior described in the PR. It merges keys across *all* spans and picks the max value per key, so trace metadata can end up as a synthetic combination that never existed on any single span. We probably need to store metadata as an `argMax` state keyed by span time (and merge it the same way in the repository) if the intent is to expose the latest span's metadata object.

csansoon: oh that's okay. It's not "last span wins" anymore. Most of the times, all spans have the same metadata anyways. If something is different, it might not look like the user expects it, but that's a user problem

packages/domain/datasets/src/use-cases/add-traces-to-dataset.ts
```diff
@@ -23,6 +21,10 @@ function mapTraceToRow(t: TraceDetail) {
       tokensInput: t.tokensInput,
       tokensOutput: t.tokensOutput,
       costTotalMicrocents: t.costTotalMicrocents,
+      sessionId: t.sessionId,
+      userId: t.userId,
+      systemInstructions: t.systemInstructions,
+      allMessages: t.allMessages,
```
andresgutgon, line 27: Thanks

PR #2443, opened 2026-03-16
Extract LLM attributes from any OTEL convention, not just GenAI

csansoon:
Same as #2431 but pointing to main
(because I cannot do it manually, repo requires to do it through a PR)

### DIFF
--- a/packages/domain/spans/src/otlp/content/openinference.ts
+++ b/packages/domain/spans/src/otlp/content/openinference.ts
@@ -1,0 +1,237 @@
+/**
+ * Content parser for OpenInference (Arize Phoenix).
+ *
+ * OpenInference explodes messages into flattened indexed span attributes:
+ *   llm.input_messages.{i}.message.role
+ *   llm.input_messages.{i}.message.content
+ *   llm.input_messages.{i}.message.tool_calls.{j}.tool_call.function.name
+ *   llm.input_messages.{i}.message.tool_calls.{j}.tool_call.function.arguments
+ *   llm.input_messages.{i}.message.contents.{j}.message_content.type
+ *   llm.input_messages.{i}.message.contents.{j}.message_content.text
+ *   llm.input_messages.{i}.message.contents.{j}.message_content.image.image.url
+ *
+ * Output messages follow the same pattern with llm.output_messages.{i}.
+ *
+ * Tool definitions use:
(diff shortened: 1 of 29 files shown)


packages/domain/spans/src/otlp/content/openinference.ts
```diff
@@ -0,0 +1,215 @@
...
+        type: "function" as const,
+        function: { name: tc.name, arguments: tc.arguments },
+      }))
+    }
+
+    messages.push(msg)
+  }
+
+  return messages
+}
```
andresgutgon, line 150: Monster method. Split

apps/ingest/src/routes/traces.ts
```diff
@@ -8,6 +10,18 @@ import type { IngestEnv } from "../types.ts"
...
+  try {
+    if (contentType.includes("application/x-protobuf")) {
+      return decodeOtlpProtobuf(new Uint8Array(body))
+    }
+    const text = new TextDecoder().decode(body)
+    return JSON.parse(text) as OtlpExportTraceServiceRequest
+  } catch {
+    return null
+  }
+}
```
geclos: wait no, decoding should happen in the background. I already changed that in a previous commit. Keep ingest api very thin: it only checks for authorization and forwards to the background, returning 202 always.

apps/ingest/src/routes/traces.ts
```diff
@@ -18,6 +32,16 @@ export const registerTracesRoute = ({ app }: TracesRouteContext) => {
...
 
+    const request = decodeRequest(body, contentType)
+    if (!request) {
+      return c.json({ error: "Failed to decode trace payload" }, 400)
+    }
+
+    const validationError = validateOtlpCompliance(request)
+    if (validationError) {
+      return c.json({ error: validationError }, 400)
+    }
```
geclos: nope, explained in previous comment

apps/ingest/src/routes/traces.ts
csansoon: But that way you will get silent errors. Spans that have been requested, with a successful response, which never appear in the platform

geclos: that is fine 202 means content accepted but not processed. It can still fail.

PR #2450, opened 2026-03-17
Replace paginated traces table with infinite-scroll smart table

csansoon:
## Summary

The main project view now displays a beautiful infinite table with ALL traces in the project.

Traces are loaded in batches of 50, with lazy loading on scrolling.

The UI is all managed by a new `InfiniteTable` component, which manages:
 - Loading state
 - Row virtualization for optimization
 - Column resizing
 - Column sorting

https://github.com/user-attachments/assets/e3b760ff-14da-47b1-a9fd-1e0cf5152922

### DIFF
--- a/apps/web/src/domains/traces/traces.collection.ts
+++ b/apps/web/src/domains/traces/traces.collection.ts
@@ -1,0 +1,62 @@
+import type { InfiniteTableInfiniteScroll, InfiniteTableSorting } from "@repo/ui"
+import { useInfiniteQuery, useQuery } from "@tanstack/react-query"
+import { useMemo } from "react"
+import { countTracesByProject, listTracesByProject, type TraceRecord } from "./traces.functions.ts"
+
+const BATCH_SIZE = 50
+
+export function useTracesInfiniteScroll({
+  projectId,
+  sorting,
+}: {
+  readonly projectId: string
+  readonly sorting: InfiniteTableSorting
+}) {
+  const {
+    data: paginatedData,
+    isLoading,
+    fetchNextPage,
+    hasNextPage,
+    isFetchingNextPage,
+  } = useInfiniteQuery({
+    queryKey: ["traces", projectId, sorting],
+    queryFn: ({ pageParam }) =>
+      listTracesByProject({
+        data: {
+          projectId,
+          limit: BATCH_SIZE,
+          cursor: pageParam,
+          sortBy: sorting.column,
+          sortDirection: sorting.direction,
+        },
+      }),
+    initialPageParam: undefined as { sortValue: string; traceId: string } | undefined,
+    getNextPageParam: (lastPage) => lastPage.nextCursor,
+  })
+
+  const infiniteScroll: InfiniteTableInfiniteScroll = useMemo(
+    () => ({
+      hasMore: hasNextPage,
+      isLoadingMore: isFetchingNextPage,
+      onLoadMore: fetchNextPage,
+    }),
+    [hasNextPage, isFetchingNextPage, fetchNextPage],
+  )
+
+  const data: readonly TraceRecord[] = useMemo(
+    () => paginatedData?.pages.flatMap((p) => p.traces) ?? [],
+    [paginatedData],
+  )
+
+  return { data, isLoading, infiniteScroll }
+}
+
+export function useTracesCount({ projectId }: { readonly projectId: string }) {
+  const { data: totalCount = 0, isLoading } = useQuery({
+    queryKey: ["traces-count", projectId],
+    queryFn: () => countTracesByProject({ data: { projectId } }),
+    staleTime: 30_000,
+  })
+
+  return { totalCount, isLoading }
+}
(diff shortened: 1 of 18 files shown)


packages/platform/db-clickhouse/src/repositories/trace-repository.ts
andresgutgon: Maybe we can start doing keyset pagination

apps/web/src/routes/_authenticated/projects/$projectId/index.tsx
```diff
@@ -30,102 +26,164 @@ function StatusBadge({ status }: { status: string }) {
...
+
+  const traces = useMemo(() => data?.pages.flatMap((p) => p.traces) ?? [], [data])
+  const totalCount = data?.pages[0]?.totalCount ?? 0
 
-  const traceIds = traces?.map((t) => t.traceId) ?? []
+  const traceIds = useMemo(() => traces.map((t) => t.traceId), [traces])
   const selection = useSelectableRows({
     rowIds: traceIds,
-    totalRowCount: traceIds.length,
+    totalRowCount: totalCount,
```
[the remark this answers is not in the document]
csansoon: not touching this

packages/ui/src/components/infinite-table/types.ts
```diff
@@ -0,0 +1,45 @@
...
+  isLoadingMore: boolean
+  onLoadMore: () => void
+}
+
+export type SortDirection = "asc" | "desc"
+
+export interface InfiniteTableSorting {
+  column: string
+  direction: SortDirection
+}
```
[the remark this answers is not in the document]
csansoon: not doing this

packages/platform/db-clickhouse/src/repositories/trace-repository.ts
csansoon: Looking at it, it wouldn't really change the performance that much. But I think it makes sense to avoid repeated traces when live consuming. I'll give it a try!

apps/web/src/domains/traces/traces.collection.ts
```diff
@@ -0,0 +1,52 @@
...
+    [hasNextPage, isFetchingNextPage, fetchNextPage],
+  )
+
+  const data: readonly TraceRecord[] = useMemo(
+    () => paginatedData?.pages.flatMap((p) => p.traces) ?? [],
+    [paginatedData],
+  )
+
+  return { data, isLoading, infiniteScroll }
+}
```
geclos, line 52: why aren't we using collections for this

apps/web/src/lib/hooks/traces/useTracesSummary.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export function useTracesSummary({ projectId }: { readonly projectId: string }) {
+  const { data: totalCount = 0, isLoading } = useQuery({
+    queryKey: ["traces-count", projectId],
+    queryFn: () => countTracesByProject({ data: { projectId } }),
+    staleTime: 30_000,
+  })
+
+  return { totalCount, isLoading }
+}
```
geclos: ditto why not collections, this introduces a new fetching pattern

packages/ui/src/components/infinite-table/infinite-table.tsx
```diff
@@ -0,0 +1,209 @@
...
+    getScrollElement: () => scrollContainerRef.current,
+    estimateSize: () => ROW_HEIGHT,
+    overscan: 10,
+  })
+
+  const virtualRows = virtualizer.getVirtualItems()
+
+  // TODO(frontend-use-effect-policy): virtualRows is derived inside the virtualizer after
+  // scroll events; the component only has access to the computed items post-render, so a
+  // dependency-tracking effect is the only way to detect the scroll-to-end threshold.
```
geclos: @cursoragent can we do it in another way? how is it we need to manually check if we are at the end to load more data?

geclos: @cursoragent doesn't tansatck/react-virtual provide a way to load more data so that we do not have to sniff the scrolling event? I'm just asking, no need to rush to an implementation

apps/web/src/domains/traces/traces.collection.ts
csansoon: tanstack collections works for `useQuery`, but not for `useInfiniteQuery`, where it adds more values to a previously fetched array. The collection is made to load a full dataset (or page of a dataset).

apps/web/src/lib/hooks/traces/useTracesSummary.ts
csansoon: This is not a collection, it's a single scalar data point. `useQuery` is more than enough, right?
Collections are supposed to be lists. Finite, fully-loadable lists, like with projects, api keys, members, datasets, spans in a trace...

csansoon: I removed these "hooks" and moved them to `web/domains/traces/traces.collection.ts` for consistency, but I'm not really using tanstack collections...

PR #2461, opened 2026-03-18
refactor(ingest): surface storage and queue failures as proper HTTP errors

geclos:
## Problem

`apps/ingest/src/routes/traces.ts` swallows all errors and always returns `202 Accepted`, even when data is lost:

```ts
// Previous behavior — both failures are silent
await putInDisk(...).catch((error) => { logger.error(...); return undefined })
await publisher.publish(...).catch((error) => { logger.error(...) })
return c.json({}, 202)  // ← always 202, even if nothing was stored or queued
```

This means:
- **Storage fails** → data lost, caller gets 202, no retry signal
- **Queue publish fails** → file orphaned in storage, never processed, caller gets 202
- **Both fail** → data lost entirely, caller unaware

## Solution

Extract the storage + queue coordination into a use-case (`ingestSpansUseCase`) that propagates typed errors through Effect's error channel. The route handler becomes a thin composition root. `honoErrorHandler` handles HTTP status mapping automatically via `httpStatus`/`httpMessage` on the error types.

### What changed

- **`QueuePublishError`** (`@domain/queue`): added `httpStatus = 500` and `httpMessage` getter (same pattern as `StorageError`)
- **`StorageDisk`** (`@domain/shared`): new Effect service tag wrapping `StorageDiskPort`
- **`QueuePublisher`** (`@domain/queue`): converted from plain interface to Effect service tag (interface renamed to `QueuePublisherShape`)
- **`StorageDiskLive`** (`@platform/storage-object`): new layer that provides `StorageDisk` from a `StorageDiskPort` instance
- **`QueuePublisherLive`** (`@platform/queue-bullmq`): new layer that provides `QueuePublisher` from a `QueuePublisherShape` instance
- **`ingestSpansUseCase`** (`@domain/spans`): new use-case composing storage write + queue publish with sequential error propagation
- **`apps/ingest/src/routes/traces.ts`**: replaced ~45 lines of nested `.catch` with ~15 lines composing the use-case

### Delivery guarantee

Storage write and queue publish are sequential. If storage fails, publish is never attempted. If publish fails after a successful storage write, the orphaned file is cleaned up from disk via `tapError` + `deleteFromDisk` (delete failure is ignored so the original `QueuePublishError` surfaces). The error propagates to the caller who can retry the full request. No data is silently dropped.

### Tests

6 unit tests for `ingestSpansUseCase`:
- Stores payload to disk and publishes queue message (happy path)
- Uses protobuf extension for protobuf content type
- Fails with `StorageError` when disk write fails
- Fails with `QueuePublishError` and cleans up file when publish fails
- Does not publish when storage fails (sequential guarantee)
- Passes correct headers in queue message

All tests use fake in-memory implementations of `StorageDisk` and `QueuePublisher` — no infrastructure dependencies.

### Verification

- `pnpm check` — all packages pass lint/format
- `pnpm typecheck` — all 32 packages pass
- `pnpm test` — all tests pass (1 pre-existing failure in `@domain/auth` unrelated to this PR)

<div><a href="https://cursor.com/agents/bc-b70d96e1-2d77-4245-abbc-31f7805dab93"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b70d96e1-2d77-4245-abbc-31f7805dab93"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/domain/spans/src/use-cases/ingest-spans.test.ts
+++ b/packages/domain/spans/src/use-cases/ingest-spans.test.ts
@@ -136,7 +136,7 @@ describe("ingestSpansUseCase", () => {
       expect(res.failure._tag).toBe("QueuePublishError")
       expect(res.failure).toBeInstanceOf(QueuePublishError)
       const err = res.failure as QueuePublishError
-      expect(err.httpStatus).toBe(500)
+      expect(err.httpStatus).toBe(502)
       expect(err.httpMessage).toContain("span-ingestion")
     }
   })


apps/ingest/src/routes/traces.ts
```diff
@@ -20,49 +19,18 @@ export const registerTracesRoute = ({ app }: TracesRouteContext) => {
...
+    const publisher = await getQueuePublisher()
+
+    await Effect.runPromise(
+      ingestSpansUseCase({
+        organizationId: OrganizationId(c.get("organizationId")),
+        projectId: ProjectId(c.get("projectId")),
+        apiKeyId: c.get("apiKeyId"),
+        payload: new Uint8Array(body),
+        contentType,
+      }).pipe(Effect.provide(StorageDiskLive(disk)), Effect.provide(QueuePublisherLive(publisher))),
```
andresgutgon, line 32: Super clean

packages/domain/spans/src/use-cases/ingest-spans.test.ts
```diff
@@ -0,0 +1,196 @@
...
+
+const validInput = {
+  organizationId: OrganizationId("org-1"),
+  projectId: ProjectId("proj-1"),
+  apiKeyId: "key-1",
+  payload: new TextEncoder().encode('{"spans":[]}'),
+  contentType: "application/json",
+}
+
+const createFakeDisk = () => {
```
andresgutgon, line 16: These mock helpers should go to the domain it belongs

packages/domain/spans/src/use-cases/ingest-spans.test.ts
```diff
@@ -0,0 +1,196 @@
...
+  const published: { queue: string; message: { body: Uint8Array; key: string | null } }[] = []
+  const publisher: QueuePublisherShape = {
+    publish: (queue, message) => {
+      published.push({ queue, message: { body: message.body, key: message.key } })
+      return Effect.void
+    },
+    close: () => Effect.void,
+  }
+  return { publisher, published }
+}
```
andresgutgon, line 45: Same

packages/domain/spans/src/use-cases/ingest-spans.test.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    expect(published).toHaveLength(1)
+    expect(published[0]?.queue).toBe("span-ingestion")
+    const body = new TextDecoder().decode(published[0]?.message.body)
+    expect(body).toBe(written[0]?.key)
+  })
+
+  it("uses protobuf extension for protobuf content type", async () => {
+    const { disk, written } = createFakeDisk()
+    const { publisher } = createFakePublisher()
```
andresgutgon, line 72: These can be globals. No need to create in every test no?

packages/domain/spans/src/use-cases/ingest-spans.test.ts
```diff
@@ -0,0 +1,196 @@
...
+        Effect.provide(Layer.succeed(QueuePublisher, publisher)),
+      ),
+    )
+
+    expect(written[0]?.key).toMatch(/\.protobuf$/)
+  })
+
+  it("fails with StorageError when disk write fails", async () => {
+    const failingDisk: StorageDiskPort = {
+      put: async () => {
```
andresgutgon, line 86: I think we can do something with vitest mocking so is easy to implement specific states like failures

geclos: @cursoragent create a github issue to improve how we implement specific states like failures in mocks

packages/domain/spans/src/use-cases/ingest-spans.test.ts
geclos: true. cc @cursoragent

packages/domain/spans/src/use-cases/ingest-spans.test.ts
geclos: @andresgutgon we will implement it in another pr

PR #2492, opened 2026-03-24
Llm providers and settings at organization and project level

andresgutgon:
> Read first: Before implementing this phase, first read the full `specs/reliability.md` reliability specification, the full reliability documentation set under `docs/`, and the local markdown task list for this phase in `specs/reliability.md`. Treat the spec/docs plus the dependencies, subtasks, and exit gate below as the authoritative implementation plan.

**Depends on**: none

**Parallelization notes**: can run in parallel with phases 2 through 7.

- [X] Define the canonical shared Zod schemas for the MVP reliability owner settings (`OrganizationSettings.keepMonitoring` and `ProjectSettings.keepMonitoring`) and infer TypeScript types; use literal-string unions rather than TypeScript enums for enum-like contracts.
- [X] Extend `organization` and `projects` with `settings` JSONB columns for `keepMonitoring`, without adding speculative secondary JSONB indexes.
- [X] Include `keepMonitoring` in organization/project settings shapes from the start so downstream settings `apps/web` and public/machine-facing settings work can rely on it, while deferring `user.settings` until a concrete user preference exists.
- [X] Keep settings reads on existing owner-row primary/unique paths and explicitly avoid speculative JSONB secondary indexes in the settings foundation phase.

**Exit gate**: organization/project settings schema and migrations are complete; `keepMonitoring` is settled before downstream phases; user/provider-model settings remain deferred.


packages/platform/db-postgres/src/schema/better-auth.ts
```diff
@@ -103,6 +103,7 @@ export const organization = latitudeSchema.table("organization", {
   creatorId: cuid("creator_id").references(() => user.id),
   currentSubscriptionId: cuid("current_subscription_id"),
   stripeCustomerId: varchar("stripe_customer_id", { length: 256 }),
+  settings: jsonb("settings").$type<{ keepMonitoring?: boolean | undefined }>(),
```
neoxelox, line 106: shouldnt this use the OrganizationSettings type?

packages/platform/db-postgres/src/schema/projects.ts
```diff
@@ -8,6 +8,7 @@ export const projects = latitudeSchema.table(
     organizationId: cuid("organization_id").notNull(),
     name: varchar("name", { length: 256 }).notNull(),
     slug: varchar("slug", { length: 256 }).notNull(),
+    settings: jsonb("settings").$type<{ keepMonitoring?: boolean | undefined }>(),
```
neoxelox, line 11: shouldnt this use the ProjectSettings type?

packages/platform/db-postgres/src/schema/better-auth.ts
andresgutgon: I'll change later

packages/platform/db-postgres/src/schema/projects.ts
andresgutgon: same

PR #2497, opened 2026-03-24
Reliability Phase 0: Type-safe async foundations (BullMQ + Temporal)

andresgutgon:
## Summary

Extends the shared async substrate (`@domain/queue`, BullMQ, domain-events dispatcher, Temporal workflows) so later reliability phases can define their own topics, events, and workflows by adding registry entries — without reopening core abstractions.

### Type-safe queue system

- **`TopicRegistry`** — single centralized const object in `@domain/queue/src/topic-registry.ts` declaring all 11 topics, their tasks, and typed payloads. Both the `TopicRegistry` type and `TOPIC_NAMES` runtime array are derived from it.
- **`publish(queue, task, payload, options?)`** — generic over topic + task. Wrong payload is a compile error. Supports `dedupeKey` (BullMQ `jobId`) and `debounceMs` (BullMQ `delay` + deduplication).
- **`subscribe<T>(queue, handlers)`** — generic, requires a `TaskHandlers<T>` map covering every task in the topic. Missing handler = compile error.
- **Runtime enforcement** — `start()` validates all `TOPIC_NAMES` have handlers (fails fast with `QueueSubscribeError`). Unknown tasks and missing payloads throw and fail the BullMQ job.
- Removed `topics.ts`, `task-dispatcher.ts`, `MessageHandler`, `QueueMessage`, `createEventHandler`, `createStubDispatcher`.

### Typed domain events

- **`EventPayloads`** in `@domain/events` maps event names to their payload shapes.
- **`KnownDomainEvent`** — discriminated union derived from `EventPayloads`.
- Domain-events dispatcher rewritten as a **`switch` on `event.name`** with full TypeScript narrowing — no routing table, no builders, no casts. Unknown events fail with `UnhandledEventError`.

### Typed Temporal workflows

- **`WorkflowRegistry`** in `@domain/queue/src/workflow-registry.ts` — maps workflow names to typed inputs.
- **`WorkflowStarterShape`** — domain-level typed contract for starting workflows (`start<W>(workflow, input, options)`).
- **`@platform/workflows-temporal`** (new package) — `createTemporalClient`, `createWorkflowStarter`, `runTemporalWorker`, `loadTemporalConfig`.
- **Issue discovery workflow** — dummy but functional end-to-end workflow demonstrating the execution path.

### Workers aligned

[...]

PR #2509, opened 2026-03-25
Phase 8 - Scores Ingestion And Base Reads

neoxelox:


apps/workers/src/workers/scores.test.ts
```diff
@@ -0,0 +1,201 @@
...
+    await Effect.runPromise(handler.handle(message))
+  }
+}
+
+const pg = setupTestPostgres()
+const ch = setupTestClickHouse()
+
+beforeAll(async () => {
+  await ch.client.command({
+    query: `CREATE TABLE IF NOT EXISTS scores
```
andresgutgon: Schema is done in a central place. Do not do in a specific test

apps/workers/src/workers/score-publications.ts
```diff
@@ -0,0 +1,55 @@
...
+export const createScorePublicationsWorker = (
+  consumer: QueueConsumer,
+  deps: ScorePublicationsWorkerDependencies = {},
+) => {
+  const chClient = deps.clickhouseClient ?? getClickhouseClient()
+  const pgClient = deps.postgresClient ?? getPostgresClient()
+  const workerLogger = deps.logger ?? logger
+
+  consumer.subscribe(
+    "score-publications",
```
andresgutgon: Wait for this please https://github.com/latitude-dev/latitude-llm/pull/2497

neoxelox: yess i wait

apps/workers/src/workers/scores.test.ts
neoxelox: what do you mean?

apps/workers/src/workers/domain-events.ts
```diff
@@ -23,7 +23,7 @@ type EventHandlerFn = (e: KnownDomainEvent) => Effect.Effect<void, unknown>
 export const createDomainEventsWorker = (
   consumer: QueueConsumer,
   pub: QueuePublisherShape,
-  workflows: WorkflowStarterShape,
+  _workflows: WorkflowStarterShape,
```
geclos, line 26: remove it

apps/workers/src/workers/scores.ts
```diff
@@ -0,0 +1,39 @@
...
+  readonly logger?: Pick<typeof logger, "error" | "info">
+}
+
+export const createScoresWorker = (consumer: QueueConsumer, deps: ScoresWorkerDependencies = {}) => {
+  const chClient = deps.clickhouseClient ?? getClickhouseClient()
+  const pgClient = deps.postgresClient ?? getPostgresClient()
+  const workerLogger = deps.logger ?? logger
+
+  consumer.subscribe("scores", {
+    publish: (payload) =>
```
geclos: is publish just saving a score in db? why the weird name

apps/workers/src/workers/scores.ts
```diff
@@ -0,0 +1,39 @@
...
+      publishScoreAnalyticsUseCase({ scoreId: payload.scoreId }).pipe(
```
geclos: storing score in analytics should be a side effect of saving the score in pg rather than an imperative call from boundary

apps/workers/src/workers/domain-events.ts
andresgutgon: I'm changing this interface to be an object.

apps/workers/src/workers/scores.ts
neoxelox: its publishing the score to clickhouse db, that usecase is called publishScoreAnalyticsUseCase, llm decided the name, what name do you recommend?

apps/workers/src/workers/domain-events.ts
neoxelox: oopsie, i missed this slop, wtf why it did this

apps/workers/src/workers/scores.ts
neoxelox: we cannot store the score in clickhouse when we store it in postgres, because:

1. it can be a failed, non errored score that it is still waiting for issue discovery. the score is not immutable and thus cannot be published to clickhouse
2. it can be a "draft" score annotation, either produced by the system annotation queue annotation, thus waiting for human confirmation of the preannotation, or the human when it is annotating and changing things, until 5 minutes after last edit, where it becomes non-draft, thus final thus can be publishable to clickhose

Obviously if the score comes already passed or errored and not draft, then it is written directly in both postgres and clickhose because it is already immutable

apps/workers/src/workers/domain-events.ts
neoxelox: ah lol not slop, its probably the linter or knip told the llm that this was unused

apps/workers/src/workers/scores.ts
geclos: create/write/save all the classic names we use to insert a row in a db

apps/workers/src/workers/scores.ts
geclos: don't get doesn't clickhosue handle deduplication in an append-only table? i thought the idea is that we always publish to clickhouse as a side effect event if it's a score update or delete and let the append only table deduplicate

apps/workers/src/workers/domain-events.ts
geclos: just remove the arg not sure what we are debating here

apps/workers/src/workers/scores.ts
neoxelox: mmm wait, we both decided to not use the replacingmergetree, and use mergretree and create this postgres-clickhouse duality.,...

neoxelox: ah always publish even if it by doing updates? it will surely do updates all the time, once per failed issue, i though we didnt want this

apps/workers/src/workers/scores.ts
neoxelox: changed all "publish" things to "save"

andresgutgon (approved):

PR #2511, opened 2026-03-25
Sessions

csansoon:
Added the `sessions` table.

A Session is a collection of spans and traces.

### Data

A Session is similar to a Trace, containing a collection of span, and aggregation of the following attributes:
- tokens (all types)
- cost (input, output and total)
- user id
- tags
- metadata
- providers
- models

Additionally, it also registers:
- traceCount
- traceIds (this is an array of all its trace ids. This is required for the table to work correctly, especially with selection)
- spanCount
- errorCount

---

### UI

A new "Sessions" view has been added to the main Traces page. This works exactly the same as the regular "Traces" view, but traces are grouped by session, and only those with sessions appear.

Sorting, filters, selection and trace detail are shared between them and work in both.

https://github.com/user-attachments/assets/ce975b44-e24c-4abf-88b3-5ac6a1aa49f6

---

### Other changes

Spans' seed script has been configured to include more sessions. 2/3 traces will not be part of sessions, but those 1/3 that are are usually included in sessions that contain 3-6 different traces.


apps/web/src/domains/sessions/sessions.functions.ts
```diff
@@ -0,0 +1,150 @@
...
+import { SessionRepository } from "@domain/spans"
+import { SessionRepositoryLive, withClickHouse } from "@platform/db-clickhouse"
+import { createServerFn } from "@tanstack/react-start"
+import { Effect } from "effect"
+import { z } from "zod"
+import { requireSession } from "../../server/auth.ts"
+import { getClickhouseClient } from "../../server/clients.ts"
+import { errorHandler } from "../../server/middlewares.ts"
+
+export interface SessionRecord {
```
andresgutgon: Instead of this which it gets desyncronized do this 
```typescript
export type SessionRecord = ReturnType<typeof serializeSession>
```

apps/web/src/layouts/ListingLayout/index.tsx
```diff
@@ -80,7 +80,7 @@ interface ListProps {
 }
 
 function List({ children, className }: ListProps) {
-  return <div className={cn("min-h-0 grow p-6 pt-0 pr-0 flex flex-col", className)}>{children}</div>
+  return <div className={cn("min-h-0 min-w-0 grow p-6 pt-0 pr-0 flex flex-col", className)}>{children}</div>
```
andresgutgon, line 83: Why this? How looks in datasets?

csansoon: Something broke the table's horizontal scroll. This fixes that

apps/web/src/lib/hooks/useSelectableRows.ts
```diff
@@ -264,6 +336,8 @@ export function useSelectableRows<T extends string | number>({
     bulkSelection,
     toggleRow,
     toggleAll,
+    selectMany,
+    deselectMany,
```
andresgutgon, line 340: QA dataset rows list selection 🙏

apps/web/src/layouts/ListingLayout/index.tsx
csansoon: It looks fine on datasets too

apps/web/src/routes/_authenticated/projects/$projectId/-components/sessions-view.tsx
```diff
@@ -0,0 +1,385 @@
...
+                <span>
+                  <ProviderIcon provider={p} size="sm" />
+                </span>
+              }
+            >
+              {p}
+            </Tooltip>
+          ))}
+          <span className="truncate">{models.join(", ")}</span>
+        </div>
```
andresgutgon, line 123: Is this column reused in traces? DRY in that case

apps/web/src/routes/_authenticated/projects/$projectId/-components/sessions-view.tsx
```diff
@@ -0,0 +1,385 @@
...
+    return {
+      data: entry.data.map((trace): SessionTableRow => ({ kind: "trace", trace })),
+      isLoading: entry.isLoading,
+    }
+  }
+
+  return (
+    <div className="flex flex-row flex-1 min-h-0 min-w-0 overflow-hidden">
+      {filtersOpen && (
+        <FiltersSidebar
```
andresgutgon, line 357: Try to use `<Layout.Aside>`

apps/web/src/routes/_authenticated/projects/$projectId/-components/sessions-view.tsx
```diff
@@ -0,0 +1,385 @@
...
+    const entry = traceMap.get(row.session.sessionId)
+    if (!entry) return { data: [], isLoading: true }
+    return {
+      data: entry.data.map((trace): SessionTableRow => ({ kind: "trace", trace })),
+      isLoading: entry.isLoading,
+    }
+  }
+
+  return (
+    <div className="flex flex-row flex-1 min-h-0 min-w-0 overflow-hidden">
```
andresgutgon: Let's avoid custom divs for layouts.

apps/web/src/routes/_authenticated/projects/$projectId/-components/traces-view.tsx
```diff
@@ -0,0 +1,190 @@
...
+            trigger={
+              <span>
+                <ProviderIcon provider={p} size="sm" />
+              </span>
+            }
+          >
+            {p}
+          </Tooltip>
+        ))}
+        <span className="truncate">{t.models.join(", ")}</span>
```
andresgutgon, line 90: Looks similar

apps/web/src/routes/_authenticated/projects/$projectId/-components/traces-view.tsx
```diff
@@ -0,0 +1,190 @@
...
+  })
+
+  const onRowClick = (t: TraceRecord) => {
+    const sel = window.getSelection()
+    if (sel && sel.toString().length > 0) return
+    onActiveTraceChange(t.traceId === activeTraceId ? undefined : t.traceId)
+  }
+
+  return (
+    <div className="flex flex-row flex-1 min-h-0 min-w-0 overflow-hidden">
```
andresgutgon: Same pattern. Should be solved in the layout

apps/web/src/routes/_authenticated/projects/$projectId/index.tsx
```diff
@@ -53,317 +51,233 @@ function getTimeFilterValue(filters: FilterSet, op: "gte" | "lte"): string | und
...
-                  variant={filtersOpen ? "outline" : "ghost"}
-                  size="sm"
-                  flat
-                  onClick={() =>
-                    navigate({
-                      search: (prev: TracesSearch) => ({ ...prev, filtersOpen: !filtersOpen || undefined }),
-                      replace: true,
-                    })
+    <Layout>
+      <Layout.Content>
```
andresgutgon: Layout content is not for wrapping actions

apps/web/src/routes/_authenticated/projects/$projectId/index.tsx
```diff
@@ -53,317 +51,233 @@ function getTimeFilterValue(filters: FilterSet, op: "gte" | "lte"): string | und
...
+                <FilterIcon className="h-4 w-4" />
+                Filters
+                {hasActiveFilters && (
+                  <span className="inline-flex items-center justify-center rounded-full bg-primary px-1.5 text-[10px] leading-4 font-medium text-primary-foreground">
+                    {Object.keys(filters).length}
+                  </span>
+                )}
+              </Button>
+            </Layout.ActionRowItem>
+            <Layout.ActionRowItem className="shrink-0">
```
andresgutgon: maybe shrink-0 belongs to row item and works always. If possible layouts in the app should compose in all the cases without ad-hoc CSS

apps/web/src/routes/_authenticated/projects/$projectId/-components/sessions-view.tsx
csansoon: The columns are typed differently. Not only there are some columns only shown in some of the views, but also they have different types. DRYing SOME of them could be possible, but I think it would make it way mor complex than it already is for very little gain

apps/web/src/routes/_authenticated/projects/$projectId/-components/sessions-view.tsx
csansoon: The `FiltersSidebar` is a left-side collapsible panel within the content area, while `Layout.Aside` is designed for a single right-side panel (the drawer). They serve different layout roles.

apps/web/src/routes/_authenticated/projects/$projectId/-components/sessions-view.tsx
csansoon: Right. I added a `Layout.Body` component for drying this

apps/web/src/routes/_authenticated/projects/$projectId/-components/traces-view.tsx
csansoon: ditto

apps/web/src/routes/_authenticated/projects/$projectId/-components/sessions-view.tsx
andresgutgon: You're right but my point is FilterSidebar should not know about positioning. Is a layout concern. We could have `<Layout.Aside position="start">` or another `<Layout.Sidebar>`

apps/web/src/routes/_authenticated/projects/$projectId/-components/traces-view.tsx
csansoon: I think that's fine for now. I am thinking of using the `ModelBadge` component we already have in @repo/ui, but I need to change something in the db before that anyways. Will probably do it in a future PR

PR #2513, opened 2026-03-25
Replace router navigate with useParamState for URL-backed UI state

csansoon:
## Summary

Introduces a `useParamState` hook that manages ephemeral UI state (active tab, open panels, selected rows, sort columns) via URL search parameters using `useSyncExternalStore`, replacing the heavier `router.navigate` + `validateSearch` pattern that routed every state change through TanStack Router's reconciliation cycle.

`router.navigate` should be reserved for actual page navigation — e.g. navigating to a newly created dataset or opening a project page — not for toggling UI controls that happen to be reflected in the URL. For those cases, `useParamState` provides a lightweight `useState`-like API that reads and writes search params directly without involving the router.

Also fixes a pre-existing bug in `ListingLayout` where toggling the aside panel caused React to unmount and remount the entire content subtree (including tab animations) because the component returned a different root DOM structure depending on whether the aside was present.

### DIFF
--- a/apps/web/src/lib/hooks/useParamState.ts
+++ b/apps/web/src/lib/hooks/useParamState.ts
@@ -1,0 +1,200 @@
+import { type SetStateAction, useSyncExternalStore } from "react"
+
+/**
+ * useParamState — a useState-like hook backed by URL search parameters.
+ *
+ * Use this for ephemeral UI state that should be reflected in the URL (active
+ * tab, open panels, sort columns, selected row IDs, etc.). Do NOT use
+ * router.navigate for this — navigate is for actual page transitions (e.g.
+ * opening a dataset after creation). useParamState bypasses the router
+ * entirely, so it avoids TanStack Router's reconciliation overhead.
+ *
+ * Design decisions:
+ *
+ * - Uses `useSyncExternalStore` to treat the URL as an external store. This
+ *   is the React-blessed way to subscribe to non-React state; it handles
+ *   concurrent mode, avoids tearing, and provides an SSR snapshot.
+ *
+ * - The URL is the source of truth. Reads always parse `window.location.search`
+ *   so there's no stale-closure risk and no local state to sync.
+ *
+ * - When the resolved value equals `defaultValue`, the param is removed from
+ *   the URL entirely (via `Object.is` comparison), keeping URLs clean.
+ *
+ * - Writes are batched: multiple setters called in the same synchronous handler
+ *   (e.g. `setSortBy(x); setSortDirection(y)`) produce only one React render.
+ *   Each write updates `history` immediately so subsequent reads within the
+ *   same tick see the latest URL, but the subscriber notification is deferred
+ *   to a microtask so React processes all changes in a single pass.
+ */
+
+type ParamStateValue = boolean | number | string
+
+/**
+ * Widens literal types to their base type so the setter accepts any value of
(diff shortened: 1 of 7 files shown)


apps/web/src/lib/hooks/useParamState.ts
```diff
@@ -0,0 +1,98 @@
...
+        ? (nextValue as (prev: WidenLiteral<T>) => WidenLiteral<T>)(currentValue)
+        : nextValue
+
+    const serialized = Object.is(resolvedValue, defaultValue) ? undefined : String(resolvedValue)
+
+    writeParam(paramKey, serialized, historyMode)
+  }
+
+  return [value, setParamValue]
+}
```
andresgutgon, line 200: This is using new react `useTransition` and keep using typed routes from tanstack. Didn't tried but i think it would be better to stick with tanstack primitives. 
```typescript
// useParamState.ts — stays within TanStack Router primitives
import { useNavigate, useSearch } from '@tanstack/react-router'
import { useCallback, useTransition } from 'react'
import type { RegisteredRouter, RouteIds, RouteById } from '@tanstack/react-router'

type AnyRoute = RouteIds<RegisteredRouter['routeTree']>

export function useParamState
  TFrom extends AnyRoute,
  TKey extends keyof RouteSearch<TFrom>
>(from: TFrom, key: TKey) {
  type TSearch = RouteSearch<TFrom>
  const search = useSearch({ from })
  const navigate = useNavigate({ from })
  const [isPending, startTransition] = useTransition()

  const setValue = useCallback(
    (next: TSearch[TKey] | ((prev: TSearch[TKey]) => TSearch[TKey])) => {
      startTransition(() => {
        navigate({
          search: (prev: TSearch) => ({
            ...prev,
            [key]: typeof next === 'function' ? (next as Function)(prev[key]) : next,
          }),
          replace: true,      // no history entry
          resetScroll: false, // don't jump to top
        })
      })
    },
    [key, navigate]
  )

  return [search[key] as TSearch[TKey], setValue, isPending] as const
}

// Helper type — pull the validated search type off a route
type RouteSearch<TFrom extends AnyRoute> =
  RouteById<RegisteredRouter['routeTree'], TFrom>['types']['fullSearchSchema']
```

Use example
```typescript
const [tab, setTab] = useParamState('/datasets/$id', 'tab')
```

andresgutgon: I think this way we can keep using `validateSearch` not sure

csansoon: Hmm I don't know, I feel like they serve different purposes.

TanStack Router's `validateSearch` + `useNavigate` is designed for params that are part of the main route's data contract: params that loaders depend on, that define what the page fetches, that affect the route match. Things like projectId, datasetId...

But `filtersOpen`, `tab`, `traceId`, `sortDirection`... these are ephemeral UI toggles that happen to be persisted in the URL for shareability. No loader reads them. No route match depends on them. Routing every sidebar toggle and tab click through the router's reconciliation pipeline (validateSearch → search update → loader check → route context update → render) is unnecessary overhead for what is essentially a setState that also writes to the address bar.

csansoon: `useParamState` works from any component without needing to know its route ID. It makes the components reusable.

And also there's no need for `validateSearch` schemas on every route for params that have nothing to do with routing

PR #2518, opened 2026-03-26
Fix workers and workflows build

andresgutgon:
# What?
Address [comments here](https://github.com/latitude-dev/latitude-llm/pull/2497) and fix build of `apps/workers` and `apps/workflows`

## TODO
- [x] Fix build 
- [x] Comments


apps/workers/tsup.config.ts
```diff
@@ -7,9 +7,11 @@ export default defineConfig({
   outDir: "dist",
   platform: "node",
   noExternal: [/@(platform|domain|repo)\/.*/],
+  external: ["@temporalio/worker", "better-auth/tanstack-start"],
```
andresgutgon: Heavy dependencies made `tsup` build choke. Declaring as external to keep it light

apps/web/src/server/clients.ts
```diff
@@ -121,7 +122,7 @@ export const getBetterAuth = () => {
       baseUrl: webUrl,
       basePath: "/api/auth",
       trustedOrigins,
-      enableTanStackCookies: true,
+      extraPlugins: [tanstackStartCookies()],
```
geclos, line 125: what is this change

apps/workers/tsup.config.ts
```diff
@@ -7,9 +7,11 @@ export default defineConfig({
   outDir: "dist",
   platform: "node",
   noExternal: [/@(platform|domain|repo)\/.*/],
+  external: ["@temporalio/worker"],
   splitting: false,
   sourcemap: true,
   clean: true,
+  shims: true,
```
geclos, line 14: is this needed?

packages/platform/db-postgres/src/create-better-auth.ts
```diff
@@ -164,7 +163,7 @@ export const createBetterAuth = (config: BetterAuthConfig) => {
         expiresIn: 3600,
         allowedAttempts: 5,
       }),
-      tanstackStartCookies(),
```
geclos, line 167: totally unecessary change no?

apps/workers/tsup.config.ts
andresgutgon: For this https://github.com/latitude-dev/latitude-llm/blob/main/apps/workers/src/server.ts#L30

`import.meta` is not `cjs` thing and needs transpilation. Btw why we use CJS could we not use ESM? Maybe some old deps?

packages/platform/db-postgres/src/create-better-auth.ts
andresgutgon: Build for workers was failing for me

apps/web/src/server/clients.ts
andresgutgon: https://github.com/latitude-dev/latitude-llm/pull/2518#discussion_r2994626203

apps/workers/tsup.config.ts
geclos: many dependencies are still using cjs so using esm is not easy

geclos: > For this https://github.com/latitude-dev/latitude-llm/blob/main/apps/workers/src/server.ts#L30

this just causes a warning in the build not an error and it will happen in the other packages that use it as well, have a single approach for all packages plz

packages/platform/db-postgres/src/create-better-auth.ts
geclos: oh i see... hmm i wonder if we should move better auth to web actually, rather than having it in platform postgres

andresgutgon: But it has postgresql schemas and shit no?

packages/platform/db-postgres/package.json
```diff
@@ -27,7 +27,7 @@
     "test": "vitest run --passWithNoTests"
   },
   "dependencies": {
-    "@better-auth/stripe": "^1.5.0",
+    "@better-auth/stripe": "^1.5.6",
```
andresgutgon, line 31:
```

packages/platform/db-postgres
├─┬ @better-auth/core 1.5.0
│ ├── ✕ unmet peer better-call@1.3.2: found 1.1.8
│ └── ✕ unmet peer @better-auth/utils@0.3.1: found 0.3.0
└─┬ @better-auth/stripe 1.5.0
  ├── ✕ unmet peer better-call@1.3.2: found 1.1.8
  └── ✕ unmet peer better-auth@1.5.0: found 1.5.6

```

apps/workers/tsup.config.ts
andresgutgon: > ⏺ shims: true is already minimal — tsup scans the CJS output for import.meta.url
   usage and only injects one line:

>  const importMetaUrl = require("url").pathToFileURL(__filename).href;

> That's it. It's not a broad compatibility layer. So it's already scoped to
  exactly what those lines need — no further targeting is possible or necessary.

Didn't know this but looks like a good solution no?

packages/platform/db-postgres/src/create-better-auth.ts
geclos: hmm ya, ok anyhow move on

apps/workers/tsup.config.ts
geclos: yes but other packages have the same issue, either use it everywhere or not

apps/api/src/server.ts
```diff
@@ -16,13 +14,7 @@ import { destroyTouchBuffer } from "./middleware/touch-buffer.ts"
...
 import { logger } from "./utils/logger.ts"
 
-const nodeEnv = process.env.NODE_ENV || "development"
-// Load .env file for local development; skipped in production containers where the file won't exist
-if (import.meta.url) {
-  const envFilePath = fileURLToPath(new URL(`../../../.env.${nodeEnv}`, import.meta.url))
-  if (existsSync(envFilePath)) loadDotenv({ path: envFilePath, quiet: true })
-}
-
+const { nodeEnv } = loadDevelopmentEnvironments(import.meta.url)
```
geclos, line 17: nice

PR #2530, opened 2026-03-27
Optimize span ingestion with inline payloads and async inserts

geclos:
## Summary
This PR optimizes the span ingestion pipeline by introducing inline payload support for small payloads and improving database insert performance. Small payloads (≤50KB) are now inlined directly in queue messages, eliminating unnecessary disk I/O, while large payloads continue to use disk storage. Additionally, ClickHouse async inserts are enabled to reduce merge pressure.

## Key Changes

- **Inline payload support**: Payloads ≤50KB are now base64-encoded and sent directly in queue messages, avoiding disk writes
- **Conditional disk storage**: Only payloads >50KB are written to disk with the new `tmp-ingest/` path structure
- **Simplified cleanup**: Removed file deletion logic since small payloads don't create files and large payloads are auto-expired via S3 lifecycle policy (1 day TTL)
- **ClickHouse async inserts**: Enabled `async_insert` with `wait_for_async_insert` to buffer concurrent inserts and reduce part creation/merge pressure
- **S3 lifecycle policy**: Added automatic expiration of temporary ingest files after 1 day
- **Queue consumer concurrency**: Increased from 10 to 50 to handle higher throughput
- **Updated queue schema**: Modified `span-ingestion` topic to support both `fileKey` and `inlinePayload` fields (both nullable)

## Implementation Details

- The 50KB threshold balances memory usage against disk I/O overhead
- Temporary ingest files use a simplified path structure: `tmp-ingest/{orgId}/{projectId}/{uuid}.{ext}`
- Worker now handles both inline and file-based payloads transparently
- Tests refactored to cover both small (inline) and large (disk) payload paths separately
- No breaking changes to the overall ingestion flow; the optimization is transparent to downstream consumers

## Bottom line
This takes from around 150 span ingestion jobs per second to around 2500 span ingestion jobs per second with a single worker

### DIFF
--- a/apps/workers/src/workers/span-ingestion.test.ts
+++ b/apps/workers/src/workers/span-ingestion.test.ts
@@ -128,6 +128,35 @@ describe("createSpanIngestionWorker", () => {
     })
   })
 
+  it("returns early when neither inlinePayload nor fileKey is provided", async () => {
+    const consumer = new TestQueueConsumer()
+    const disk = new FakeStorageDisk()
+    const pub = createFakeEventsPublisher()
+
+    createSpanIngestionWorker({ consumer, eventsPublisher: pub, clickhouseClient: ch.client, disk })
+
+    await consumer.dispatchTask("span-ingestion", "ingest", {
+      fileKey: null,
+      inlinePayload: null,
+      contentType: "application/json",
+      organizationId: "org_no_payload_test",
+      projectId: "proj_no_payload_test",
+      apiKeyId: "api_key_no_payload_test",
+      ingestedAt: "2026-03-18T10:00:00.000Z",
+    })
+
+    const [count] = await Effect.runPromise(
+      queryClickhouse<{ total: string }>(
+        ch.client,
+        "SELECT count() AS total FROM spans WHERE organization_id = {organizationId:String}",
+        { organizationId: "org_no_payload_test" },
+      ),
+    )
+
+    expect(Number(count?.total ?? 0)).toBe(0)
+    expect(pub.published).toHaveLength(0)
+  })
+
   it("drops invalid payloads without inserting spans", async () => {
     const consumer = new TestQueueConsumer()
     const disk = new FakeStorageDisk()


apps/workers/src/workers/span-ingestion.ts
```diff
@@ -45,7 +45,17 @@ export const createSpanIngestionWorker = ({
...
       return Effect.gen(function* () {
-        const payload = yield* getFromDisk(disk, wire.fileKey)
+        let payload: Uint8Array
+
+        if (wire.inlinePayload) {
+          payload = Buffer.from(wire.inlinePayload, "base64")
+        } else if (wire.fileKey) {
+          payload = yield* getFromDisk(disk, wire.fileKey)
+        } else {
+          workerLogger.error("Span ingestion: no inline payload or fileKey in message")
```
andresgutgon: Test for this

packages/platform/queue-bullmq/src/adapter.ts
```diff
@@ -134,7 +134,7 @@ export const createBullMqQueueConsumer = (config: BullMqRedisConfig): Effect.Eff
               },
               {
                 connection: new Redis(redisConfig),
-                concurrency: 10,
+                concurrency: 50,
```
andresgutgon: I realized all queue share same config no?. Should we allow different config by queue?

geclos: 💯 good catch

packages/domain/queue/src/index.ts
```diff
@@ -43,10 +43,14 @@ export class QueuePublisher extends ServiceMap.Service<QueuePublisher, QueuePubl
   "@domain/queue/QueuePublisher",
 ) {}
 
+export interface SubscribeOptions {
+  readonly concurrency?: number
```
andresgutgon, line 47: Not sure this is related with queues. Is bullmq specific. I always have these dilemas

PR #2531, opened 2026-03-27
Phase 9 - Scores Analytics And Telemetry Queries

neoxelox:
> Read first: Before implementing this phase, first read the full `specs/reliability.md` reliability specification, the full reliability documentation set under `docs/`, and the local markdown task list for this phase in `specs/reliability.md`. Treat the spec/docs plus the dependencies, subtasks, and exit gate below as the authoritative implementation plan.

**Depends on**: Phase 2, Phase 8

**Parallelization notes**: Phase 10 can continue in parallel while this phase finishes; phases 11 and 15 should wait for it.

- [x] Implement repositories and query primitives over canonical Postgres scores plus the initial ClickHouse score analytics query layer, while keeping the exact later materialized analytics tables deferred.
- [x] Implement score-aware filtering for spans using rollups instead of hot joins against raw scores.
- [x] Implement trace and session drilldowns filtered by score state, value, source, and issue.
- [x] Implement evaluation/custom source trend queries and project-wide score aggregates from the immutable ClickHouse score analytics rows, while keeping score-table and drilldown reads in Postgres.
- [x] Implement issue occurrence/time-series aggregates used by issue lifecycle and UI.
- [x] Ensure score queries can include or exclude simulation-generated scores where the product requires it.
- [x] Benchmark and tune the ClickHouse score analytics table sort keys, partitions, and skip indexes against realistic reliability workloads.
- [x] Add query regression tests and benchmark fixtures for Postgres-backed score drilldowns, ClickHouse-backed project aggregates, and include/exclude simulation behavior.

**Exit gate**: spans, traces, and sessions can be filtered by score-derived properties; evaluation, issue, and simulation analytics have a working query path without prematurely locking the final materialized-table design.


packages/platform/db-clickhouse/src/repositories/score-analytics-repository.test.ts
```diff
@@ -0,0 +1,519 @@
...
+    cost: (overrides.cost as number) ?? 50,
+    created_at: (overrides.created_at as string) ?? "2026-03-15 12:00:00.000",
+  }
+}
+
+async function insertScores(rows: ReturnType<typeof makeScoreRow>[]) {
+  await ch.client.insert({ table: "scores", values: rows, format: "JSONEachRow" })
+}
+
+interface ScoreAnalyticsRepoShape {
```
andresgutgon: Dont put generic entities inside particular tests. We call this `fake-x-repository`

PR #2541, opened 2026-03-30
Phase 7 - Simulations Foundations

neoxelox:


packages/platform/db-postgres/src/seeds/simulations/index.ts
```diff
@@ -0,0 +1,77 @@
...
+            target: simulations.id,
+            set,
+          })
+        }
+
+        console.log(`  -> simulations: ${simulationRows.length} canonical lifecycle samples`)
+      },
+      catch: (error) => new SeedError({ reason: "Failed to seed simulations", cause: error }),
+    }).pipe(Effect.asVoid),
+}
```
csansoon, line 75: Shouldn't the seed script create spans with the simulation_id, too?

Take a look at the spans seed script, that is already handling the generation of thousands of spans, traces and even sessions

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00008_simulation_id.sql
```diff
@@ -0,0 +1,215 @@
+-- +goose NO TRANSACTION
+-- +goose Up
+
+-- ═══════════════════════════════════════════════════════════
+-- Spans: add simulation_id column and bloom filter index
+-- ═══════════════════════════════════════════════════════════
+
+ALTER TABLE spans
+    ADD COLUMN IF NOT EXISTS simulation_id FixedString(24) DEFAULT '' CODEC(ZSTD(1)) AFTER api_key_id,
+    ADD INDEX IF NOT EXISTS idx_simulation_id simulation_id TYPE bloom_filter(0.01) GRANULARITY 2;
```
csansoon, line 10: Add this attribute to the `Span`, `Trace` and `Session` types too, in `domain/spans`. Otherwise we still don't know about simulation Ids in the code

csansoon: Also maybe add the attribute to the trace detail drawer

PR #2544, opened 2026-03-30
feat: use project slug

geclos:
- Updated ingest endpoint to expect project slug in X-Latitude-Project header instead of project ID
- Added findBySlug method to ProjectRepository interface and implementation
- Made project slug in breadcrumb copiable with clipboard icon and toast notification

### DIFF
--- a/apps/web/src/domains/projects/projects.functions.ts
+++ b/apps/web/src/domains/projects/projects.functions.ts
@@ -40,6 +40,23 @@ export const listProjects = createServerFn({ method: "GET" })
     return projects.map(toRecord)
   })
 
+export const getProjectBySlug = createServerFn({ method: "GET" })
+  .middleware([errorHandler])
+  .inputValidator(z.object({ slug: z.string() }))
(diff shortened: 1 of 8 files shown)


apps/web/src/routes/_authenticated.tsx
```diff
@@ -84,10 +85,23 @@ function ProjectBreadcrumb({ projectId }: { projectId: string }) {
           <ChevronsUpDown className="h-4 w-4 text-muted-foreground" />
         </button>
       ) : (
-        <span className="text-sm font-medium text-muted-foreground px-2 py-1">
-          {emoji && `${emoji} `}
-          {title}
-        </span>
+        <Button
```
andresgutgon: We already have `<CopyButton>`

geclos: it lacks a lot of stuff. @claude create a github issue to improve the copy button, it should support multiple sizes and text color variations e.g size xs and text foreground muted, size sm and text destructive, yada yada, use existing tokens

PR #2548, opened 2026-03-30
Traces aggregations

csansoon:
Added:
- General trace aggregations
- Spans histogram
- MIN/AVG/MEDIAN/MAX/SUM aggregations on all numeric columns of the table

All of these aggregations react on the currently selected filters.

Spans histogram has a max target of 50 columns. It will select a human bucket size (1 minute, 1 hour, 2 days, etc, not "1232 seconds") that fits better for the plot to display at most 50 columns.

https://github.com/user-attachments/assets/a08cbed2-77d9-4d97-ab1d-dc1a60dd2b50


packages/utils/src/format.ts
```diff
@@ -1,13 +1,23 @@
...
  * Examples: `128000` -> `"128K"`, `1500000` -> `"1.5M"`, `42` -> `"42"`
+ * Fractional values under 1000 use at most two decimal places (e.g. averages).
  */
 export function formatCount(count: number): string {
   if (count < 0) return `-${formatCount(-count)}`
-  if (count < 1000) return String(count)
+  if (count < 1000) {
+    if (Number.isInteger(count)) return String(count)
+    return countFractionFormatter.format(count)
+  }
```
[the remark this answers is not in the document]
csansoon: that edge case is fine for now

packages/domain/spans/src/helpers.ts
```diff
@@ -0,0 +1,184 @@
...
+  24 * 60 * 60, // 1 day
+  2 * 24 * 60 * 60,
+  3 * 24 * 60 * 60,
+  4 * 24 * 60 * 60,
+  5 * 24 * 60 * 60,
+  7 * 24 * 60 * 60, // 1 week
+  14 * 24 * 60 * 60,
+  30 * 24 * 60 * 60, // 30 days
+  60 * 24 * 60 * 60, // 60 days
+  MAX_HISTOGRAM_BUCKET_SECONDS, // 90 days
```
geclos, line 47: what happens if i query a custom value beyond the 90 day mark? say i query data between 6 months and 3 months old

packages/platform/db-clickhouse/src/repositories/session-repository.ts
```diff
@@ -255,6 +303,55 @@ export const SessionRepositoryLive = Layer.effect(
...
+                        sum(span_count) AS span_sum
+                      FROM (
+                        SELECT session_id, ${LIST_SELECT}
+                        FROM sessions
+                        WHERE organization_id = {organizationId:String}
+                          AND project_id = {projectId:String}
+                          ${extraWhere}
+                        GROUP BY organization_id, project_id, session_id
+                        ${havingClause}
+                      )`,
```
geclos, line 339: can this query be unbounded in time? let's not allow it if so

packages/platform/db-clickhouse/src/repositories/session-repository.ts
```diff
@@ -255,6 +303,55 @@ export const SessionRepositoryLive = Layer.effect(
...
+                        max(cost_total_microcents) AS cost_max,
+                        avg(cost_total_microcents) AS cost_avg,
+                        quantileExact(0.5)(cost_total_microcents) AS cost_median,
+                        sum(cost_total_microcents) AS cost_sum,
+                        min(span_count) AS span_min,
+                        max(span_count) AS span_max,
+                        avg(span_count) AS span_avg,
+                        quantileExact(0.5)(span_count) AS span_median,
+                        sum(span_count) AS span_sum
+                      FROM (
```
geclos, line 331: anyway we can do this without a subquery? it increases ram usage of db a lot since the db has to hold the whole dataset in mem while computing the aggregations.

packages/platform/db-clickhouse/src/repositories/trace-repository.ts
```diff
@@ -317,6 +385,112 @@ export const TraceRepositoryLive = Layer.effect(
...
+                        quantileExactIf(0.5)(time_to_first_token_ns, time_to_first_token_ns > 0) AS ttft_median,
+                        sumIf(time_to_first_token_ns, time_to_first_token_ns > 0) AS ttft_sum
+                      FROM (
+                        SELECT trace_id, ${LIST_SELECT}
+                        FROM traces
+                        WHERE organization_id = {organizationId:String}
+                          AND project_id = {projectId:String}
+                          ${extraWhere}
+                        GROUP BY organization_id, project_id, trace_id
+                        ${havingClause}
```
geclos, line 430: ditto questions from the sessions query

packages/ui/src/components/infinite-table/infinite-table.tsx
```diff
@@ -97,19 +98,25 @@ export function InfiniteTable<T>({
...
+    // Lock once headers are measurable so interactive header content changes do not reflow columns.
+    if (table.offsetWidth === 0 || headerCells.some((th) => th.offsetWidth === 0)) return
+
+    for (const th of headerCells) {
       th.style.width = `${th.offsetWidth}px`
     }
     table.style.width = `${table.offsetWidth}px`
     setLayoutFixed(true)
-  }, [layoutFixed, data.length])
+  }, [layoutFixed, hasExpansion, hasSubheaderRow, isLoading, selection, columns.length, data.length])
```
geclos: this layout effect is becoming complex anyway we can simplify it

packages/ui/src/components/charts/bar-chart.tsx
```diff
@@ -0,0 +1,100 @@
+/// <reference path="../../echarts-subpaths.d.ts" />
+import type { EChartsCoreOption } from "echarts/core"
+import EChartsReact from "echarts-for-react"
```
geclos: this lib has granular imports if you use echarts v5 or v6 https://www.npmjs.com/package/echarts-for-react

packages/ui/src/components/charts/bar-chart.tsx
```diff
@@ -0,0 +1,100 @@
...
+   * (useful for embedded previews with a forced theme).
+   */
+  readonly colorScheme?: "light" | "dark"
+  readonly formatTooltip?: (category: string, value: number) => string
+  /** When false, hides y-axis tick labels and frees left grid margin (tooltip still shows values). */
+  readonly showYAxis?: boolean
+}
+
+type EChartsReactBridgeProps = {
+  readonly echarts: typeof import("./register-echarts.ts").echarts
```
geclos: why dynamic import here

packages/domain/spans/src/helpers.ts
csansoon: The system will try to show a histogram with a maximum of 50 bars.
If you select 6 months, for example, it will select the bucket list that fits the most: Natural-looking buckets (specific ranges like hours, days, etc, instead of "perfectly calculated" ranges, like "32.67 hours"). In this case, it would select 4-day buckets, since 6 months are about 180 days, so it would take 45 3-day buckets, just below the target 50.

The only edge case would be selecting a case where even 90-days buckets would surpass the target of 50 bars, but in that case I think it's fine showing more buckets lol. It would require selecting a range of more than 12 years

packages/platform/db-clickhouse/src/repositories/session-repository.ts
csansoon: Currently we don't have a forced "Start time" in the UI.
If we added it, we can start enforcing it. However, I would not enforce the aggregation metrics if the UI doesn't have a Start time selected, it would look broken

packages/platform/db-clickhouse/src/repositories/session-repository.ts
csansoon: Hmmm not really. Without the subquery, it would be aggregating partial rows, because the `AggregatingMergeTree` might not have deduplicated many of the rows yet.

This can also be solved by adding `FINAL` to the end instead of the subquery, but this is often more expensive both on latency and memory anyways.

Once change we could do is changing the MEDIAN from `quantileExact` to `quantileTDigest `, which is cheaper to calculate but ofter more like an approximate median rather than the exact value

packages/platform/db-clickhouse/src/repositories/session-repository.ts
geclos: but what's the time bound of aggregations if no filters are selected in ui? I assume we are not aggregating across the whole trace dataset?

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/aggregations/general-aggregations.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const medianDurationStr = loading ? ellip : traceMetrics ? formatDuration(traceMetrics.durationNs.median) : dash
+
+  const totalTokensStr = loading ? ellip : traceMetrics ? formatCount(traceMetrics.tokensTotal.sum) : dash
+
+  const totalSpansStr = loading ? ellip : traceMetrics ? formatCount(traceMetrics.spanCount.sum) : dash
+
+  const showTtftAggregations = traceMetrics && traceMetrics.timeToFirstTokenNs.max > 0
+
+  return (
+    <div className="flex flex-row flex-wrap gap-4 p-4">
```
andresgutgon, line 58: How works responsive here?

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/aggregations/histogram.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <div className="px-4 py-3">
+        <ChartSkeleton minHeight={160} className="border-0 bg-transparent p-0" />
+      </div>
+    )
+  }
+
+  if (isError) {
+    return (
+      <div className="flex w-full min-h-[80px] items-center justify-center px-4 py-3">
```
andresgutgon, line 48: Try not to hardcode height and width. Make it fluid if possible

apps/web/src/routes/_authenticated/projects/$projectSlug/index.tsx
```diff
@@ -218,6 +219,10 @@ function ProjectPage() {
         </div>
       )}
 
+      <div className="px-6">
```
andresgutgon, line 273: This is a smell that the layout is not flexible enough. Could be this an existing layout item?

packages/ui/src/components/charts/chart-css-theme.ts
```diff
@@ -0,0 +1,80 @@
...
+  tooltipBorder: "hsl(0 0% 89.8%)",
+}
+
+const FALLBACK_DARK: ChartCssThemeColors = {
+  foreground: "hsl(0 0% 98%)",
+  mutedForeground: "hsl(0 0% 63.9%)",
+  border: "hsl(231 9% 12.9%)",
+  primary: "hsl(211 94% 43%)",
+  tooltipBackground: "hsl(0 0% 3.9%)",
+  tooltipBorder: "hsl(231 9% 12.9%)",
```
andresgutgon, line 26: Are these color our theme colors? Maybe we could use `--var` to use it

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/aggregations/histogram.tsx
csansoon: Mostly, yes. But we want the loading placeholder and error message to take a similar size like the actual Histogram, so the rest of the UI doesn't jump suddenly when there's an error

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/aggregations/general-aggregations.tsx
csansoon: It wraps

apps/web/src/routes/_authenticated/projects/$projectSlug/index.tsx
csansoon: It's just that we would want the parent container to have an horizontal padding, except for the table.
So, to fix that, each child has to have the padding instead

packages/ui/src/components/charts/chart-css-theme.ts
csansoon: These are just fallbacks, if the vars don't work. We're using --vars in the same component

apps/web/src/routes/_authenticated/projects/$projectSlug/index.tsx
andresgutgon: I don't like these addhoc wrappers honestly

packages/platform/db-clickhouse/src/repositories/session-repository.ts
csansoon: General aggregations is not bound, it's the whole dataset by default.

We're displaying an infinite table that is not bound in time. It would be weird that aggregations were but the table wasn't

The histogram, though, is bound to 1 week for visibility.

geclos: hmm we have to review this I don't think it makes sense to make clickhouse work that hard to display the same number constantly (across 3 months the average will be very stable if you have some volume) also I think the default should not be unbounded but rather "last hour" or so, this is common in most observ platforms. As a user, you can always unbound if you want, but it should be opt in rather than the default. It's also weird that the table is unbounded but the histogram is, what other observ platforms do is that when the time window is large (months) they decrease the resolution, instead of being traces per day it becomes traces per week, and then per month, and as you drill in, the resolution of the histogram increases.

PR #2553, opened 2026-03-31
Score eligibility validation

neoxelox:


apps/workflows/src/activities/index.ts
```diff
@@ -1,32 +1,118 @@
...
+const isEligibilityError = (error: unknown): error is CheckEligibilityError => {
+  return eligibilityErrorConstructors.some((Ctor) => error instanceof Ctor)
+}
+
+export const checkEligibility = (input: CheckEligibilityInput): Promise<true> =>
+  Effect.runPromise(
+    checkEligibilityUseCase(input).pipe(
+      Effect.tap(() =>
+        Effect.sync(() => {
+          logger.info("Issue discovery eligibility passed", {
```
andresgutgon: Do we need to log this?

apps/workflows/src/activities/index.ts
```diff
@@ -1,32 +1,118 @@
...
+  readonly organizationId: string
+  readonly projectId: string
+  readonly scoreId: string
+  readonly matchedIssueId: string | null
+}): Promise<AssignmentResult> =>
+  Effect.runPromise(
+    createOrAssignIssueUseCase(input).pipe(
+      Effect.tap((result) =>
+        Effect.sync(() => {
+          logger.info("Issue discovery assignment completed", {
```
andresgutgon: I find that server logs are super crowded with info making it harder to see anything. Let's think what really need to be logged

PR #2562, opened 2026-03-31
chore(ci): upload sourcemaps to Datadog on deploy

geclos:
## Summary

- Enable `sourcemap: true` in `apps/web/vite.config.ts` so Vite generates `.js.map` files alongside minified client assets
- Add `staging-upload-sourcemaps` and `production-upload-sourcemaps` jobs to `deploy.yml` that run in parallel with the ECS rollout after each build
- All five services are covered: `web` (browser sourcemaps) and `api`, `ingest`, `workers`, `workflows` (Node.js server sourcemaps — tsup already had `sourcemap: true`)
- Release version is tagged with the same image SHA (`sha-XXXXXXXX`) used for the Docker release, so Datadog can correlate stack frames to the right build

## Required secrets/vars per GitHub environment

| Name | Type | Example |
|---|---|---|
| `DATADOG_API_KEY` | secret | — |
| `DATADOG_SITE` | variable | `datadoghq.eu` |

## Test plan

- [x] Verify `DATADOG_API_KEY` and `DATADOG_SITE` are set on both `staging` and `production` GitHub environments
- [x] Trigger a staging deploy and confirm the `staging-upload-sourcemaps` job completes successfully
- [x] Check [Datadog RUM Debug Symbols](https://app.datadoghq.com/source-code/setup/rum) that sourcemaps appear under the correct service and release version

### DIFF
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -44,6 +44,7 @@ export default defineConfig({
     },
   },
   build: {
+    sourcemap: true,
     rolldownOptions: {
       // These are Node.js-only packages that must never appear in client chunks.
       // Vite follows dynamic imports in @repo/observability/otel.ts and would
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -210,8 +210,54 @@ jobs:
           done
           exit "${ec}"
 
+  staging-upload-sourcemaps:
+    needs: staging-build
+    runs-on: ubuntu-latest
+    environment:
+      name: staging
+      deployment: false
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+
+      - name: Install pnpm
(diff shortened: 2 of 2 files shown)


apps/web/vite.config.ts
andresgutgon, line 54: Can you add `paparse` here? https://github.com/latitude-dev/latitude-llm/blob/main/packages/domain/datasets/src/export-csv.ts#L1 I think is the culprit here for the extra bundle weight https://github.com/latitude-dev/latitude-llm/blob/019acd5e854a66543fdc92749d75bf7da1845a1b/apps/web/src/routes/_authenticated/projects/%24projectSlug/datasets/%24datasetId.tsx#L1

geclos: but this is used client side we cannot externalize it, you mean to split it to its own chunk I guess

PR #2571, opened 2026-04-01
Lat 468/issue discovery workflow

neoxelox:


apps/workers/src/workers/domain-events.ts
```diff
@@ -50,6 +61,19 @@ export const createDomainEventsWorker = ({
...
+    IssueDiscoveryRequested: (event) =>
+      workflowStarter.start(
+        "issueDiscoveryWorkflow",
+        {
+          organizationId: event.payload.organizationId,
+          projectId: event.payload.projectId,
+          scoreId: event.payload.scoreId,
+        },
+        {
+          workflowId: buildIssueDiscoveryWorkflowId(event.payload),
```
andresgutgon: Do not use `buildIssueDiscoveryWorkflowId` this test is better using hardcoded version this way you're also testing this helper

packages/platform/db-weaviate/src/repositories/issue-projection-repository.test.ts
```diff
@@ -0,0 +1,211 @@
...
+}) {
+  let existingTenant = tenantExists
+  const createdTenants: string[] = []
+  const withTenantCalls: string[] = []
+  const insertCalls: unknown[] = []
+  const replaceCalls: unknown[] = []
+  const hybridCalls: Array<{ query: string; options: unknown }> = []
+  const existingIds = new Set<string>()
+
+  const tenantCollection = {
```
andresgutgon, line 40: Move to a `testing/fake-tenant-collection.ts` or something like that

PR #2577, opened 2026-04-01
LAT-469 | Phase 12 | Add evaluation generation from Issues and alignment workflow

csansoon:
## Summary
Implement the full issue-evaluation generation and alignment system, including canonical evaluation persistence, example curation from annotation evidence, baseline script generation, GEPA-backed optimization, long-lived refresh orchestration, polling-based job status, and internal/public trigger surfaces.

This turns issue monitoring evaluations into durable, inspectable artifacts that can be generated on demand, aligned against human feedback, incrementally refreshed when evidence changes, and fully re-optimized when alignment quality degrades.

https://github.com/user-attachments/assets/574a0438-7e18-45d1-bcdb-18b55c135c0c

### Domain and persistence
- Added canonical evaluation lifecycle support for active, archived, and deleted evaluations, including issue-driven archive behavior and `keepMonitoring` semantics.
- Added confusion-matrix storage plus derived alignment metrics, so evaluation quality is computed from canonical counters instead of persisting separate metric snapshots.
- Added a Redis-backed evaluation-alignment job status contract with stable status values and key helpers for background workflow polling.
- Added deterministic split/config contracts for alignment datasets so generation and optimization runs are reproducible.
- Added repository support for canonical evaluation reads/writes, issue reads, and exact alignment-example selection from annotation-backed evidence.

### Alignment example curation
- Added a dedicated alignment example query path that derives positive and negative examples from canonical annotation scores linked to an issue.
- Positive examples are sourced from failed, non-draft, non-errored annotation evidence tied to the issue being aligned.
- Negative examples follow the defined priority order and remain trace-grouped and deterministic.
- Example hydration now joins curated evidence with canonical trace-backed conversation context before generation or optimization runs.

### Baseline generation
- Added a repository-owned baseline generator that builds an evaluation script directly from issue context plus curated examples.
- Added a narrow evaluation runtime used to execute generated scripts against hydrated conversations and accumulate confusion-matrix results.
- Added a post-alignment details generation pass so resulting evaluations are persisted with generated user-facing metadata after alignment completes.

### Optimization architecture
- Added `@domain/optimizations` as the domain contract layer for optimizer inputs, candidates, objectives, results, and errors.
- Added `@platform/op-gepa` as the concrete GEPA integration layer, including:
  - Python runtime assets
  - stdio JSON-RPC bridge
  - optimizer adapter
  - proposer/details-generator prompts
  - named model constants
- Adapted the legacy optimizer behavior into the repository’s domain/platform boundaries rather than restoring the old architecture directly.
- Added multi-objective candidate evaluation across:
  - alignment quality via MCC
  - cost derived from stored microcents
  - duration derived from stored nanoseconds

### Initial generation flow
When generation is triggered for an issue, the implemented workflow now performs this end-to-end sequence:

1. Collect canonical positive/negative examples for the target issue.
2. Hydrate those examples with trace-backed conversation context.
3. Generate a baseline evaluation script from issue context and curated examples.
4. Evaluate the baseline against curated examples and derive confusion-matrix state.
5. Run GEPA optimization over the evaluation script using host-side evaluation/proposal callbacks.
6. Persist the aligned evaluation and generated metadata.
7. Write Redis-backed job status updates for polling clients.

### Long-lived refresh orchestration
The evaluation-alignment workflow was extended from a one-shot path into a long-lived Temporal orchestration that keeps alignment fresh over time.

It now supports:
- debounced annotation-driven refresh
- incremental recomputation when the evaluation hash is unchanged
- one-hour metric recompute cadence
- eight-hour full re-optimization cadence
- manual re-alignment using the same workflow path
- escalation from incremental refresh into full re-optimization when MCC drop exceeds tolerance

This keeps refresh policy inside Temporal workflow state and durable timers rather than external delayed-job scans.

### Internal application surface
- Added internal `apps/web` server-side entrypoints to:
  - read issue-linked evaluations
  - start generation
  - poll job status
  - trigger manual realignment
- Added a temporary but functional issue-level UI that exposes:
  - linked evaluation status
  - generate action
  - realign action
  - polling feedback while background work is running

### Public and machine-facing surface
- Added approved API endpoints for:
  - listing issue-linked evaluations
  - starting generation
  - polling alignment job status
  - manually requesting realignment
- The API returns immediate job handles for asynchronous work and reuses the same Redis-backed status contract as the internal product flow.

## Refresh and Alignment Semantics
- Incremental refresh evaluates only evidence newer than the current alignment watermark.
- Existing confusion-matrix counters are merged forward when the evaluation script remains unchanged and performance remains within tolerance.
- Full re-optimization is triggered only when derived alignment quality shows meaningful degradation.
- Manual realignment is rate-limited and routed through the same durable workflow path as automated refreshes.

## Reliability and Verification
- Added deterministic tests covering:
  - evaluation lifecycle behavior
  - confusion-matrix derivation
  - alignment example curation
  - baseline generation behavior
  - optimizer bridge/protocol behavior
  - long-lived refresh orchestration
  - annotation-driven workflow triggering
  - internal/public generation and polling flows
- Included focused verification around optimizer integration and refresh behavior so the feature is testable as a durable system, not just as isolated helpers.

## Result
This PR introduces a complete issue-to-evaluation pipeline: evidence is curated from annotations, turned into an initial evaluation, optimized against measurable alignment objectives, persisted as a first-class evaluation artifact, refreshed automatically as new evidence arrives, and exposed through both internal and machine-facing surfaces with background-job visibility.


apps/api/src/routes/evaluations.ts
```diff
@@ -0,0 +1,399 @@
...
+  description: "Returns the current evaluations linked to an issue within a project.",
+  security: PROTECTED_SECURITY,
+  request: {
+    params: OrgProjectIssueParamsSchema,
+  },
+  responses: {
+    200: jsonResponse(EvaluationListResponseSchema, "Linked evaluations"),
+    401: errorResponse("Unauthorized"),
+    404: errorResponse("Issue not found"),
+  },
```
andresgutgon: DRY https://github.com/latitude-dev/latitude-llm/blob/main/apps/api/src/routes/annotations.ts#L40-L44

apps/api/src/routes/evaluations.ts
```diff
@@ -0,0 +1,399 @@
...
+    params: OrgProjectEvaluationParamsSchema,
+  },
+  responses: {
+    202: jsonResponse(EvaluationAlignmentJobStatusResponseSchema, "Accepted manual realignment job"),
+    401: errorResponse("Unauthorized"),
+    404: errorResponse("Evaluation not found"),
+  },
+})
+
+const toFailurePayload = (error: unknown): { readonly message: string; readonly code?: string } => {
```
andresgutgon: This should be handle by error middleware to unify how errors are shown in api endpoints.

apps/workers/src/server.ts
```diff
@@ -78,7 +79,10 @@ const bootstrap = async () => {
       workflowStarter,
     }
 
-    createDomainEventsWorker(ctx)
+    createDomainEventsWorker({
+      ...ctx,
+      pgClient,
```
andresgutgon: Review this. In other workers pgClient is created with the helper not us the one that's on the server. Not sure if this has any implications in relation with open connections in this client or something. 

Maybe the right thing is to pass this client always to all workers that need postgresql. Let's think about it.

apps/workers/tsconfig.json
```diff
@@ -2,8 +2,13 @@
...
-    "rootDir": "src",
+    "rootDir": "../..",
     "jsx": "react-jsx"
   },
-  "include": ["src/**/*.ts"]
+  "include": [
+    "src/**/*.ts",
+    "../../packages/domain/evaluations/src/**/*.ts",
+    "../../packages/platform/db-postgres/src/**/*.ts",
+    "../../packages/platform/testkit/src/**/*.ts"
```
andresgutgon: This should not be needed.

apps/workers/tsconfig.json
```diff
@@ -2,8 +2,13 @@
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "dist",
-    "rootDir": "src",
+    "rootDir": "../..",
```
andresgutgon: Weird

packages/platform/op-gepa/python/op_gepa_engine/core/gepa.py
```diff
@@ -0,0 +1,262 @@
...
+    loop = asyncio.get_running_loop()
+
+    timeout_stopper = TimeoutStopCondition(
+        timeout_seconds=min(params.budget.timeSeconds or ENGINE_MAX_TIME, ENGINE_MAX_TIME)
+    )
+    usage_stopper = UsageStopCondition(max_usage=min(params.budget.tokens or ENGINE_MAX_TOKENS, ENGINE_MAX_TOKENS))
+    stagnation_stopper = NoImprovementStopper(max_iterations_without_improvement=ENGINE_MAX_STAGNATION)
+
+    run_optimize = partial[GEPAResult[Output, str]](
+        optimize,
```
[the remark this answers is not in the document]
csansoon: I don't think this is a real runtime bug in the current implementation. The GEPA Python package is pinned to Python `>=3.13,<3.14`, and in that runtime `functools.partial` is subscriptable, so `partial[GEPAResult[Output, str]](...)` does not raise a `TypeError`.

Also, this line is inside the `gepa_optimize` request handler, not at module import / engine boot, so even if it were invalid under an unsupported interpreter, the failure mode would be an optimize RPC error rather than “the GEPA engine cannot start”.

That said, I agree the `partial[...]` syntax is a bit surprising. I'm happy to simplify it to plain `partial(...)` in a follow-up for readability / portability, but I don't think it's a correctness issue for the supported runtime.

packages/platform/ai-vercel/src/ai.ts
```diff
@@ -49,8 +49,8 @@ export const AIGenerateLive = Layer.effect(
                 system: input.system,
                 prompt: input.prompt,
                 output: Output.object({ schema: input.schema }),
-                maxOutputTokens: input.maxTokens ?? 256,
                 temperature: input.temperature ?? 0.3,
+                ...(input.maxTokens !== undefined ? { maxOutputTokens: input.maxTokens } : {}),
```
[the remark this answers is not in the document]
csansoon: That is fine, 256 is way too low as an output token count, making it always fail

apps/web/src/domains/evaluations/evaluation-alignment.functions.ts
```diff
@@ -0,0 +1,257 @@
...
+
+    await Effect.runPromise(
+      Effect.gen(function* () {
+        const issueRepository = yield* IssueRepository
+        const issue = yield* issueRepository.findById(issueId)
+
+        if (issue.projectId !== projectId) {
+          return yield* new BadRequestError({
+            message: `Issue ${issue.id} does not belong to project ${projectId}`,
+          })
```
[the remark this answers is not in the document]
csansoon: This is actually the idiomatic Effect pattern for failing inside `Effect.gen`. `BadRequestError` extends `Data.TaggedError`, which makes every instance an `Effect` value that fails with itself when yielded. So `yield* new BadRequestError(...)` is equivalent to `yield* Effect.fail(new BadRequestError(...))`.

This pattern is used consistently across the codebase (annotations, evaluations, issues, api-key repository, workflow activities, etc.) and is explicitly tested in `packages/domain/shared/src/errors.test.ts`.

apps/web/src/domains/evaluations/evaluation-alignment.functions.ts
```diff
@@ -0,0 +1,257 @@
...
+        if (!evaluation) {
+          return yield* new EvaluationNotFoundError({
+            evaluationId: data.evaluationId,
+          })
+        }
+
+        if (evaluation.projectId !== projectId || evaluation.issueId !== issueId) {
+          return yield* new BadRequestError({
+            message: `Evaluation ${evaluation.id} does not match the requested issue or project`,
+          })
```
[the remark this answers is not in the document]
csansoon: ditto

apps/api/src/routes/evaluations.ts
```diff
@@ -0,0 +1,399 @@
...
+        if (!loaded) {
+          return yield* new EvaluationNotFoundError({
+            evaluationId,
+          })
+        }
+
+        if (loaded.projectId !== projectId) {
+          return yield* new BadRequestError({
+            message: `Evaluation ${loaded.id} does not belong to project ${projectId}`,
+          })
```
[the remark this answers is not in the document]
csansoon: ditto

apps/api/src/routes/evaluations.ts
csansoon: `toFailurePayload` is **not** formatting the HTTP response, it's serializing the error into the Redis-backed job status so the polling endpoint can surface a meaningful failure reason to the frontend. The actual HTTP error response is already handled by `honoErrorHandler` via the `throw error` on the next line, which goes through the actual unified error middleware.

If we removed this and only relied on the middleware, the Redis job status would show `"failed"` with no error details, breaking the polling contract.

apps/workers/src/server.ts
csansoon: yeah you're right

I unified all Postgres-consuming workers to receive `pgClient` as a required parameter from the server, instead of some calling `getPostgresClient()` internally. This way tests already pass the mocked client, instead of having to mock the singleton, there is no hidden init-order coupling (`getPostgresClient(10)` in the server only sets max connections because it happens to be called first; with explicit injection, this temporal dependency goes away) and we have a visible dependency graph

apps/workers/tsconfig.json
csansoon: Yup. This is proof that Opus 4.6 is way better than GPT 5.4 at not fucking things up

apps/api/src/routes/evaluations.ts
```diff
@@ -0,0 +1,394 @@
...
+        error: toFailurePayload(error),
+      })
+      throw error
+    }
+
+    return c.json(toJobStatusResponse(pendingStatus), 202)
+  })
+
+  return app
+}
```
geclos: I do not think we should expose public api endpoints for these just yet, please remove them and move any that you need to apps/web. We will implement proper rest endpoints in the public api afterwards.

apps/web/src/domains/issues/issues.collection.ts
geclos: why isn't this a tanstack db collection

apps/web/src/domains/issues/issues.collection.ts
```diff
@@ -30,3 +30,15 @@ export function useCreateIssue({ projectId }: { readonly projectId: string }) {
...
+  return useQuery({
+    queryKey: ["issues", projectId],
+    queryFn: () =>
+      listIssuesByProject({
+        data: {
+          projectId,
+        },
+      }),
+  })
+}
```
geclos, line 32: ditto why isn't this a tanstack collection db

apps/web/src/domains/issues/issues.functions.ts
```diff
@@ -1,21 +1,40 @@
...
+  escalatedAt: issue.escalatedAt?.toISOString() ?? null,
+  resolvedAt: issue.resolvedAt?.toISOString() ?? null,
+  ignoredAt: issue.ignoredAt?.toISOString() ?? null,
 })
 
-type IssueRecord = ReturnType<typeof toRecord>
+/** @knipignore Used by declaration emit for useIssuesCollection return type */
+export type IssueRecord = ReturnType<typeof toIssueRecord>
+
+/** @knipignore Used by declaration emit for useIssuesCollection return type */
```
geclos: i do not understand this comment the collection does not seem to have any return type in its declaration it's just inferred

apps/web/src/domains/issues/issues.functions.ts
```diff
@@ -1,21 +1,40 @@
...
-  ignoredAt: issue.ignoredAt ? issue.ignoredAt.toISOString() : null,
+  createdAt: issue.createdAt.toISOString(),
+  updatedAt: issue.updatedAt.toISOString(),
+  escalatedAt: issue.escalatedAt?.toISOString() ?? null,
+  resolvedAt: issue.resolvedAt?.toISOString() ?? null,
+  ignoredAt: issue.ignoredAt?.toISOString() ?? null,
 })
 
-type IssueRecord = ReturnType<typeof toRecord>
+/** @knipignore Used by declaration emit for useIssuesCollection return type */
```
geclos: ditto

apps/web/vite.config.ts
```diff
@@ -43,6 +43,12 @@ export default defineConfig({
       tslib: "tslib/tslib.es6.mjs",
     },
   },
+  ssr: {
+    external: ["@swc/core"],
+  },
+  optimizeDeps: {
+    exclude: ["@swc/core"],
+  },
```
geclos, line 43: what's this dep used for and why this configuration change?

apps/workers/src/workers/api-keys.test.ts
```diff
@@ -46,7 +46,7 @@ describe("createApiKeysWorker", () => {
 
   it("persists an API key for the organization", async () => {
     const consumer = new TestQueueConsumer()
-    createApiKeysWorker({ consumer, postgresClient: pg.appPostgresClient })
+    createApiKeysWorker({ consumer, pgClient: pg.appPostgresClient })
```
geclos: avoid this kind of refactors in PRs that are already huge please, i appreciate the effort but put in another pr next time

apps/workflows/src/activities/index.ts
```diff
@@ -139,3 +564,586 @@ export const syncIssueProjections = (input: SyncIssueProjectionsInput): Promise<
...
+        updatedAt: now,
+      })
+
+      yield* evaluationRepository.save(evaluation)
+
+      return {
+        evaluationId: evaluation.id,
+      }
+    }).pipe(withPostgres(EvaluationRepositoryLive, getPostgresClient(), OrganizationId(input.organizationId))),
+  )
```
geclos: why is this file 1100 lines long, organize it well please. Keep workflow handlers thin, delegate to use cases and repositories, just like the rest of boundaries.

packages/domain/evaluations/package.json
```diff
@@ -5,6 +5,10 @@
   "type": "module",
   "main": "src/index.ts",
   "types": "src/index.ts",
+  "exports": {
+    ".": "./src/index.ts",
+    "./constants": "./src/constants.ts"
```
geclos, line 10: why this separate export

packages/domain/issues/src/use-cases/sync-projections.ts
```diff
@@ -37,4 +33,4 @@ export const syncIssueProjectionsUseCase = (input: SyncIssueProjectionsInput) =>
       description: issue.description,
       vector,
     })
-  }) as Effect.Effect<void, RepositoryError>
+  }).pipe(Effect.catchIf(isNotFoundError, () => Effect.void)) as Effect.Effect<void, RepositoryError>
```
geclos: i don't think this type casting is needed

packages/domain/queue/src/index.ts
```diff
@@ -22,6 +27,15 @@ export interface WorkflowStarterShape {
     input: WorkflowInput<W>,
     options: { readonly workflowId: string },
   ) => Effect.Effect<void>
+  readonly signalWithStart: <W extends WorkflowName>(
```
geclos, line 31: what's the difference between this and start?

packages/domain/queue/package.json
```diff
@@ -6,6 +6,7 @@
   "types": "src/index.ts",
   "exports": {
     ".": "./src/index.ts",
+    "./workflow-registry": "./src/workflow-registry.ts",
```
geclos, line 9: why the separate export

packages/domain/shared/src/effect.ts
```diff
@@ -0,0 +1 @@
+export { Data, Effect, Layer, ServiceMap } from "effect"
```
geclos: ?? just import effect where we need it this indirection is unnecessary

apps/workers/src/server.ts
```diff
@@ -78,16 +78,19 @@ const bootstrap = async () => {
       workflowStarter,
     }
 
-    createDomainEventsWorker(ctx)
-    createMagicLinkEmailWorker(ctx)
+    createDomainEventsWorker({
+      ...ctx,
+      pgClient,
```
geclos: let's just add pgclient to the context

apps/workers/src/workers/domain-events.ts
```diff
@@ -66,10 +121,20 @@ export const createDomainEventsWorker = ({
...
+            debounceMs: ISSUE_REFRESH_DEBOUNCE_MS,
+          }),
+          signalEvaluationAlignmentRefreshes({
+            organizationId: event.payload.organizationId,
+            projectId: event.payload.projectId,
+            issueId: event.payload.issueId,
+          }),
+        ],
+        { concurrency: "unbounded" },
+      ).pipe(Effect.asVoid),
```
geclos: hmm this is a problem now that i see it. @cursoragent create a github issue to review how domain event fan outs handle errors because at the moment it seems like we could have partial errors, i.e domain event tries to fan out to 3 other event handlers but only 2 succeed, 3rd fail and we silently ignore the error.

apps/workers/src/workers/domain-events.ts
```diff
@@ -24,11 +32,58 @@ export const createDomainEventsWorker = ({
...
+                {
+                  reason: "debounced-metric-refresh",
+                  jobId: `auto-refresh:${evaluation.id}`,
+                },
+              ],
+            },
+          ),
+        ),
+        { concurrency: "unbounded" },
+      ).pipe(Effect.asVoid)
```
geclos: again, this is eating potential errors faning out to N workflows, what happens if temporal goes down mid-iteration

apps/workflows/src/workflows/evaluation-alignment-workflow.ts
```diff
@@ -0,0 +1,376 @@
...
+  readonly positiveExampleCount: number
+  readonly negativeExampleCount: number
+}> => {
+  const trackJobStatus = shouldTrackJobStatus(input)
+
+  if (trackJobStatus) {
+    await writeEvaluationAlignmentJobStatus({
+      jobId: input.jobId,
+      status: "pending",
+    })
```
geclos: is this necessary if we are immediately updating the status to running in the next line?

apps/web/src/domains/issues/issues.collection.ts
csansoon: It's a search/filter query.

The `nameFilter` changes on every keystroke, results are a capped subset, and nothing mutates through it. Collections are for the canonical entity list you read from and write to, ephemeral search views are just `useQuery`.

apps/web/src/domains/issues/issues.collection.ts
csansoon: this can actually be changed to a collection

apps/web/src/domains/issues/issues.functions.ts
csansoon: fair, these are no longer needed

apps/web/src/domains/issues/issues.functions.ts
csansoon: ditto

apps/web/vite.config.ts
csansoon: `@swr/core` is a .node binary, which can't be bundled by any JS bundler. This binary is used as a peer dependency of `tsup`, which is used by some other packages in the repo. If this happens to be visible to Vite when bundling the app, this tells it to ignore it

packages/domain/evaluations/package.json
csansoon: temporal workflows need to import some of the constants without importing unnecessary methods. This import is now used by temporal workflows.
Temporal failed without this

geclos: > without importing unnecessary methods

why? unless necessary i would avoid this minor optimizations as they break convention

apps/workers/src/workers/domain-events.ts
geclos: cursor esta frito

geclos: @claude create a github issue to review how domain event fan outs handle errors because at the moment it seems like we could have partial errors, i.e domain event tries to fan out to 3 other event handlers but only 2 succeed, 3rd fail and we silently ignore the error.

apps/web/vite.config.ts
geclos: leave a comment with this info please

apps/web/src/domains/issues/issues.collection.ts
geclos: hmm there's something broken in our approach here if we cannot/want to default to tanstack collections for 99% of client-side state. useQuery decouples mutations from state so the developer is forced to manually invalidate state keys that have 0 referential integrity whenever a mutation happen. I understand this state is readonly so you do not worry but at some point someone will introduce a write op and this store will not be notified a bug will be introduced and someone will retroactively add a key invalidation to keep this store in sync. Besides losing realtime since mutations in usequery aren't optimistic. 

I'd say you should default to tanstack db collections unless there's a strong reason not to do so, even if only to keep convention.

packages/domain/queue/package.json
csansoon: same as other comments: because of Temporal bundling constraints

apps/workers/src/workers/domain-events.ts
csansoon: #2665

apps/workers/src/workers/domain-events.ts
csansoon: #2665

packages/domain/queue/src/index.ts
csansoon: `start` simply starts a new workflow execution. If a workflow with the same `workflowId` is already running, it silently no-ops. It's fire-and-forget

`signalWithStart` is an atomic two-in-one operation: it sends a signal to an existing workflow with the given `workflowId`, or if no such workflow is running yet, it starts one and delivers the signal to it. This is critical when you need to guarantee the signal is never lost due to a race condition between checking if a workflow exists and starting/signaling it.

You can see the difference in practice in `apps/workers/src/workers/domain-events.ts`:

**`IssueDiscoveryRequested`** uses `start`: it just needs to kick off an `issueDiscoveryWorkflow`. If one is already running for that ID, skip it.

**`EvaluationAlignmentRequested`** uses `signalWithStart`: the workflow is a long-lived loop that **blocks waiting** for signals to schedule work. Using start would only ensure the workflow is running, but a running workflow sitting in `condition()` has no way to learn that new data arrived. `signalWithStart` both ensures the workflow exists and delivers the "new data, schedule a refresh" message in one atomic call. With `start`, the second and subsequent events would silently no-op (workflow already running), and the loop would starve.

packages/domain/evaluations/package.json
csansoon: They're not minor optimizations. They're Temporal bundling constraints.
They're necessary, Temporal will fail otherwise

apps/api/src/clients.ts
```diff
@@ -68,3 +76,19 @@ export const getQueuePublisher = (): Promise<QueuePublisherShape> => {
   }
   return queuePublisherPromise
 }
+
+function getTemporalConfig(): TemporalConfig {
```
neoxelox: no need for this, it can be done inside `getWorkflowStarter`

Dockerfile
```diff
@@ -75,6 +75,15 @@ ARG VITE_LAT_WEB_URL
...
 RUN pnpm --filter @app/web build
 
+# ---------------------------------------------------------------------------
+# Build GEPA Python venv — install Python deps for the GEPA optimizer
+# ---------------------------------------------------------------------------
+FROM python:3.13-slim-bookworm AS python-gepa
+
+COPY packages/platform/op-gepa/python /build
+WORKDIR /build
+RUN python -m venv --clear .venv && .venv/bin/pip install --no-cache-dir .
```
neoxelox: i do not understand, we are using uv how is this working?

geclos: why are we installing python / gepa in web package

apps/workflows/src/activities/evaluation-alignment-activities.ts
```diff
@@ -0,0 +1,254 @@
...
+        }),
+    }),
+  )
+
+const evaluationManualRealignmentRateLimitKey = (evaluationId: string): string =>
+  `evaluation-alignment:manual-rate-limit:${evaluationId}`
+
+export const assertManualEvaluationRealignmentAllowed = (input: { readonly evaluationId: string }): Promise<void> =>
+  Effect.runPromise(
+    Effect.tryPromise({
```
geclos: weird runPromise wrapping a tryPromise, just do runPromise with a try catch 🤷🏼 in that case

Dockerfile
csansoon: we're not. This is the python-gepa build stage that gets then included into `workflows`, not `web`. The `web` target is defined way down the file, around line ~220, and has no Python at all.

Anyways, this has been changed to use `uv` instead of setting up a python env manually

packages/domain/evaluations/src/entities/evaluation.ts
```diff
@@ -70,10 +75,61 @@ export function emptyEvaluationAlignment(evaluationHash: string): EvaluationAlig
 }
 
 // ---------------------------------------------------------------------------
-// Evaluation entity
+// Evaluation id
+// ---------------------------------------------------------------------------
+
+export const evaluationIdSchema = cuidSchema.transform(toEvaluationId)
+export type EvaluationId = z.infer<typeof evaluationIdSchema>
```
neoxelox: why duplicate this¿

neoxelox (dismissed):

neoxelox (approved):

PR #2689, opened 2026-04-09
feat(annotation-queues): system queue annotate step after flagger triaging

geclos:
## Summary

Implements the system annotation queue annotation step after flagger triaging. When a system queue flagger successfully matches a trace, the workflow now creates a queue item and a queue-backed draft annotation atomically.

### Backend Implementation
- **Contracts & Defaults**: Added annotator input/output contracts with Zod schemas
- **Persistence Primitives**: 
  - `insertIfNotExists` for idempotent queue item creation
  - `incrementTotalItems` for atomic counter maintenance
  - `findQueueDraftByTraceId` for draft lookup
- **Draft Annotation**: Extracted `persistDraftAnnotation` primitive that doesn't auto-publish
- **Annotator Use Case**: `runSystemQueueAnnotatorUseCase` generates feedback using LLM with full conversation context
- **Orchestration**: `systemQueueAnnotateUseCase` coordinates queue item + draft creation transactionally
- **Workflow Wiring**: Extended `systemQueueFlaggerWorkflow` to call annotate activity on successful matches

### Key Behaviors
- **Idempotent**: Retry-safe for `(organizationId, projectId, queueId, traceId)` tuple
- **Atomic**: Queue item insert + counter increment happen in same transaction
- **Non-publishing**: System-created drafts stay drafts until explicit human publication
- **Refactored**: `writeAnnotationUseCase` now uses outbox for transactional publication scheduling

## Testing

- 75 tests in `@domain/annotation-queues`
- 9 tests in `@app/workflows`
- Full regression coverage for happy path, idempotent retry, annotator failure, and no-auto-publication

## No UI Changes

This is a backend-only implementation. No changes to apps/web routes, components, or server functions.

## Related

Extends the existing `system-annotation-queues:fanOut → systemQueueFlaggerWorkflow` architecture without topology changes.


packages/domain/annotation-queues/src/use-cases/system-queue-annotate.ts
```diff
@@ -0,0 +1,165 @@
...
+      }),
+    )
+
+    return systemQueueAnnotateOutputSchema.parse({
+      queueId,
+      traceId: parsedInput.traceId,
+      draftAnnotationId: result.draftAnnotationId,
+      wasCreated: result.wasCreated,
+    }) as SystemQueueAnnotateOutput
+  })
```
geclos, line 144: we can do this better with workflows. Once the flagger has determined it's a match we execute an activity to draft annotate the trace, and when that is complete we run another activity to persist the annotation queue item and the score transaccionally

geclos:
@cursoragent review

PR #2726, opened 2026-04-13
Phase 13 PR2: implement live evaluation enqueue flow

csansoon:
## Summary
- Add the `live-evaluations:enqueue` pipeline so ended traces can be turned into concrete live-evaluation execution jobs.
- Move the selection logic into domain code: load the trace, scan active evaluations, apply filter/sampling/turn rules, and publish the correct `live-evaluations:execute` jobs.
- Add the supporting repository, queue, logging, and test coverage needed to make that enqueue path reliable and reviewable.

### 1. A real enqueue path for live evaluations
The `live-evaluations` worker no longer treats `enqueue` as a stub. It now calls a dedicated domain use case that:
- reloads the ended trace from ClickHouse
- lists active evaluations for the project from Postgres
- applies trigger gates in the required order: `filter -> sampling -> turn/debounce`
- publishes one `live-evaluations:execute` job per eligible `(evaluationId, traceId)` pair

### 2. Batched trace filter matching
This PR adds a batched `TraceRepository` read for checking one trace against many evaluation filter sets in a single call, instead of issuing one filter query per evaluation.

That keeps filter semantics in the trace repository while making the enqueue scan efficient enough for project-wide evaluation checks.

### 3. Correct turn and debounce publication semantics
The enqueue path now distinguishes between:
- `every` without debounce: publish a trace-scoped execute job
- `every` with debounce: publish a scope-scoped delayed job
- `last`: publish a scope-scoped delayed job so the latest eligible trace in the scope wins
- `first`: check persisted score history before publishing
- `first` with debounce: still publish a delayed trace-scoped job, without collapsing into `last`-style scope coalescing

This PR also tightens the trigger invariant so `turn = "last"` requires `debounce > 0`.

### 4. Paused-evaluation handling
Evaluations with `sampling = 0` are now treated as paused and skipped before publication.

### 5. Structured enqueue logging
The worker now logs structured enqueue outcomes, including:
- active evaluations scanned
- filter matches
- paused skips
- sampling skips
- turn skips
- execute jobs published

### 6. Focused automated coverage
This PR adds coverage at two levels:
- domain-level tests for enqueue selection and publication semantics
- worker-level tests using in-memory Postgres and ClickHouse to exercise the real worker composition

## What is still missing for the full live-monitoring phase

This PR only implements the enqueue/selection half of the flow.

Still missing:
- `live-evaluations:execute` is still a stub
- no actual evaluation execution happens yet
- no score persistence happens yet
- no issue assignment for live evaluation results happens yet
- no execute-path logging/telemetry is implemented yet
- `live-traces:end` is still a stub, so real trace completion does not yet emit `TraceEnded`
- because of that, this flow is not yet activated automatically from live traffic
- rollout and end-to-end coverage for the sibling `TraceEnded` consumers are still pending

## Test plan
- [x] `cd packages/domain/evaluations && pnpm test src/helpers.test.ts`
- [x] `cd packages/domain/evaluations && pnpm test src/use-cases/live/enqueue-live-evaluations.test.ts`
- [x] `cd packages/domain/evaluations && pnpm typecheck`
- [x] `cd apps/workers && pnpm test src/workers/live-evaluations.test.ts`
- [x] `cd apps/workers && pnpm typecheck`

### DIFF
--- a/packages/domain/evaluations/src/use-cases/live/enqueue-live-evaluations.ts
+++ b/packages/domain/evaluations/src/use-cases/live/enqueue-live-evaluations.ts
@@ -3,8 +3,19 @@ import { OrganizationId, ProjectId, type RepositoryError, TraceId } from "@domai
 import { TraceRepository } from "@domain/spans"
 import { Effect } from "effect"
 import type { Evaluation } from "../../entities/evaluation.ts"
-import { getLiveEvaluationEligibility, shouldSampleLiveEvaluation } from "../../helpers.ts"
+import type { LiveEvaluationQueuePublishError } from "../../errors.ts"
+import {
+  buildLiveEvaluationExecuteScopeDedupeKey,
+  buildLiveEvaluationExecuteTraceDedupeKey,
+  getLiveEvaluationEligibility,
+  shouldSampleLiveEvaluation,
+  toLiveEvaluationDebounceMs,
+} from "../../helpers.ts"
 import { EvaluationRepository } from "../../ports/evaluation-repository.ts"
+import {
+  LiveEvaluationQueuePublisher,
+  type PublishLiveEvaluationExecuteInput,
+} from "../../ports/live-evaluation-queue-publisher.ts"
 
 const ACTIVE_EVALUATION_SCAN_PAGE_SIZE = 100
 
@@ -36,7 +47,7 @@ export type EnqueueLiveEvaluationsResult =
       readonly summary: EnqueueLiveEvaluationsSummary
     }
 
-export type EnqueueLiveEvaluationsError = RepositoryError
+export type EnqueueLiveEvaluationsError = RepositoryError | LiveEvaluationQueuePublishError
 
 const listAllActiveEvaluations = ({ projectId }: { readonly projectId: ProjectId }) =>
   Effect.gen(function* () {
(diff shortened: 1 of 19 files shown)


packages/domain/evaluations/src/use-cases/live/enqueue-live-evaluations.ts
```diff
@@ -0,0 +1,281 @@
...
+      evaluation,
+      eligibility: getLiveEvaluationEligibility(evaluation),
+    }))
+    const skippedPausedCount = evaluationEligibility.reduce((count, item) => {
+      if (!item.eligibility.eligible && item.eligibility.reason === "paused") {
+        return count + 1
+      }
+
+      return count
+    }, 0)
```
andresgutgon, line 180: This can be done inside the `flapMap`

packages/platform/db-clickhouse/src/repositories/trace-repository.ts
```diff
@@ -282,6 +296,22 @@ function buildTraceFilterClauses(filters: FilterSet | undefined): {
   }
 }
 
+function buildTraceFilterCondition(
+  filters: FilterSet | undefined,
+  paramPrefix: string,
+): {
+  condition: string
+  params: Record<string, unknown>
```
andresgutgon, line 304: Could this be typed?

csansoon: No

andresgutgon (approved):

PR #2733, opened 2026-04-14
feat(ingest): add trace ingestion burst limits

geclos:
## Summary
- add Redis-backed burst limiting to `POST /v1/traces`, scoped by organization, project, and API key
- enforce both request-count and payload-byte caps so large OTLP batches cannot bypass ingest protection
- document the ingest rate-limit env knobs and cover the limiter with focused tests

## Testing
- pnpm --filter @app/ingest typecheck
- pnpm --filter @app/ingest test -- src/rate-limit/trace-ingestion.test.ts


geclos:
🐮 👦🏼

apps/ingest/src/rate-limit/trace-ingestion.ts
```diff
@@ -0,0 +1,150 @@
...
+import { Effect } from "effect"
+
+const productionTraceRateLimitDefaults = {
+  maxRequests: 600,
+  maxBytes: 64 * 1024 * 1024,
+  windowSeconds: 60,
+} as const
+
+const developmentTraceRateLimitDefaults = {
+  maxRequests: 5_000,
```
andresgutgon, line 11: Why is this higher than production?

geclos: it's just a protection rate limit we will implement a per subscription plan rate limit in the #future

PR #2735, opened 2026-04-14
Implement queued live evaluation execution and persistence

csansoon:
## Summary
- Implements the `live-evaluations:execute` path so a queued live evaluation task now loads the current evaluation and trace context, executes the evaluation, and persists exactly one canonical score.
- Replaces the worker stub with real orchestration that wires Postgres, ClickHouse, and AI dependencies at the worker boundary while keeping business logic in the domain layer.
- Adds structured logging, stable AI telemetry, and focused domain and worker coverage for the new execute path.

## What this PR adds
- A new domain-level `runLiveEvaluationUseCase` that owns the full execute flow for one queued run.
- Execute-time lifecycle and duplicate-result rechecks so archived, deleted, paused, or already-scored evaluations skip before spending hosted AI work.
- Canonical score persistence through `writeScoreUseCase` for passed, failed, and errored live evaluation results.
- Direct `issueId` assignment for failed non-errored issue-linked monitor results so those scores become immutable immediately.
- Persistence of execution failures as canonical errored scores with the correct `error` / `errored` semantics plus duration, token, and cost accounting.
- Structured worker logs for skipped and persisted execute outcomes, including evaluation identity, score identity, result kind, issue-assignment path, accounting fields, and session id when present.
- Stable AI telemetry on the hosted evaluation call via the `evaluation.live.execute` span, tagged with queued execution metadata.
- Domain and worker tests that cover duplicate skips, passed/failed/errored outcomes, direct issue assignment, telemetry, logging, and analytics sync timing.

## How it is implemented
- The worker remains orchestration-only: `apps/workers/src/workers/live-evaluations.ts` now routes `live-evaluations:execute` through `runLiveEvaluationUseCase` and provides the required repository, analytics, and AI layers there.
- The new domain use case loads the evaluation, rechecks eligibility, checks canonical duplicate state for `(evaluationId, traceId)`, loads linked issue context, executes through the shared live evaluation executor, and writes results through the shared score writer instead of introducing a parallel persistence path.
- The existing live execution seam was extended just enough to carry telemetry into the shared AI service without changing execution semantics.
- The score-writing path is reused for all result kinds so immutability rules, outbox emission, and analytics sync behavior stay centralized in one canonical place.
- Worker coverage now includes real default-wrapper execute branches on no-AI paths, while domain coverage asserts the exact persistence and analytics timing semantics.

## What is still missing for the overall phase
- The upstream activation step is still not enabled: the debounced `live-traces:end` worker does not yet publish the real `TraceEnded` event that would wake this pipeline up end to end.
- The rollout behavior for sibling `TraceEnded` consumers, especially `live-annotation-queues:curate`, still needs to be decided so activation does not accidentally ship unrelated behavior.
- End-to-end activation coverage is still missing for the full `TraceEnded -> enqueue -> execute` pipeline, including debounce reset behavior, turn selection, lifecycle skips, direct issue assignment, analytics timing, and persisted accounting.
- Final documentation cleanup still needs to happen once the activation behavior and rollout plan are locked.

## Test plan
- [x] `cd packages/domain/evaluations && pnpm typecheck`
- [x] `cd packages/domain/evaluations && pnpm test src/live-execution.test.ts src/use-cases/live/run-live-evaluation.test.ts`
- [x] `cd apps/workers && pnpm test src/workers/live-evaluations.test.ts`
- [x] `cd apps/workers && pnpm typecheck` currently fails because of an existing unrelated Better Auth version literal mismatch in `packages/platform/db-postgres/src/create-better-auth.ts`


packages/domain/evaluations/src/use-cases/live/score-write-outbox-event-writer.ts
```diff
@@ -0,0 +1,12 @@
...
+// Mirror the shared outbox writer service key locally so this package can compose
+// `writeScoreUseCase` without adding a direct dependency on `@domain/events`.
+export interface ScoreWriteOutboxEventWriterShape {
+  write(event: unknown): Effect.Effect<void, unknown>
+}
+
+export class ScoreWriteOutboxEventWriter extends ServiceMap.Service<
+  ScoreWriteOutboxEventWriter,
+  ScoreWriteOutboxEventWriterShape
+>()("@domain/events/OutboxEventWriter") {}
```
andresgutgon: I think this is not what we do normally. Not sure we need this abstraction

andresgutgon (dismissed):

packages/domain/evaluations/src/use-cases/live/score-write-outbox-event-writer.ts
andresgutgon: Change it please

packages/domain/evaluations/src/use-cases/live/run-live-evaluation.ts
```diff
@@ -0,0 +1,273 @@
...
+    RunLiveEvaluationError,
+    | AI
+    | EvaluationIssueRepository
+    | EvaluationRepository
+    | ScoreWriteOutboxEventWriter
+    | ScoreAnalyticsRepository
+    | ScoreRepository
+    | SqlClient
+    | TraceRepository
+  >
```
geclos, line 318: perfect candidate for a workflow with 2 activities one to run the llm eval and one to persist the score

packages/domain/evaluations/src/use-cases/live/score-write-outbox-event-writer.ts
geclos: 💯 ai sloppitty slop

packages/domain/evaluations/src/use-cases/live/run-live-evaluation.ts
csansoon: I've been looking at it. It could be moved to a temporal workflow in the future, yes, but for now it is a bit more complex than just moving the logic. The temporal workflow does not have the scoped dedupe that BullMQ does, which could result in duplicated runs.

Again, this can be improved, but I'll leave it like this for now during the MVP

andresgutgon (dismissed):

andresgutgon (approved):

PR #2742, opened 2026-04-14
Use AWS Bedrock everywhere

neoxelox:


packages/platform/ai-vercel/src/ai.ts
```diff
@@ -14,16 +12,47 @@ import { runWithAiTelemetry } from "@platform/ai-latitude"
...
 type ProviderOptions = NonNullable<GenerateTextCall["providerOptions"]>
 type ProviderModel = GenerateTextCall["model"]
+type BedrockGeographyPrefix = "eu" | "us" | "apac"
 
 const DEFAULT_MAX_OUTPUT_TOKENS = 8192
-const BEDROCK_PROVIDER = "amazon-bedrock"
+const bedrockScopedModelIdPattern = /^(?:(?:eu|us|apac)\.)?([a-z0-9-]+\..+)$/
 
-const isSupportedBedrockProvider = (provider: string) => provider === BEDROCK_PROVIDER
+const bedrockGeographyPrefixForAwsRegion = (region: string): BedrockGeographyPrefix => {
```
andresgutgon, line 26: Are we provisioning bedrock in different places?

neoxelox: @geclos

geclos: we do not provision bedrock we use their on demand inference and you require to specify which region you want to target (they have different prices depending on region)

andresgutgon: But what regions are que requesting? only eu or dependending on something another?

csansoon (approved):

PR #2748, opened 2026-04-15
Improve issue seeds

neoxelox:


packages/platform/db-postgres/package.json
```diff
@@ -48,7 +48,8 @@
     "drizzle-orm": "catalog:",
     "effect": "catalog:",
     "pg": "catalog:",
-    "stripe": "^22.0.1"
+    "stripe": "^22.0.1",
+    "voyageai": "catalog:"
```
andresgutgon, line 52: Is this platform dependency necessary in postgresql?

andresgutgon (approved):

packages/platform/db-postgres/package.json
andresgutgon: At least put in devDependencies

neoxelox: In order to fix some features not working locally on the seeded project, like searching issues or issue discovery, the seeded issues need to be built correctly, by embedding the feedback scores correctly and embedding them if you have voyage api key

andresgutgon: So voyageai is a runtime dependency of postgresql package?

PR #2752, opened 2026-04-15
Add scalable live ingestion seed tool

csansoon:
## Summary
- Add a dedicated `@tools/live-seeds` package for simulating OTLP span ingestion against the seeded organization and project. The tool generates realistic trace families, resolves `traceId`s so deterministic sampling lands in or out as intended, and sends spans with bounded parallelism while preserving each trace's internal timeline.
- This makes it much easier to test the full live ingestion path at scale. A single run can exercise ingest acceptance, trace-end debounce, live evaluation enqueue and execute flows, live annotation queue curation, and system annotation queue fan-out for workflows such as frustration detection, tool-call errors, empty responses, and output-schema validation.
- The CLI supports reproducible high-volume runs with fixture selection, per-fixture counts, concurrency controls, and fixed seeds, while the generated spans still look like real seeded traffic by reusing the same ambient tag and metadata conventions as the default seed data.
- Rename and reorganize the package from `tools/seeds` to `tools/live-seeds` so the structure matches its actual purpose: a single dedicated live-ingestion seeding tool rather than a generic multi-tool seed workspace. The branch also keeps repeated local validation smoother by relaunching Temporal after `docker/reset-postgres.sh` resets the database.

## Test plan
- [x] `pnpm format`
- [x] `pnpm knip`
- [x] `cd tools/live-seeds && pnpm typecheck`
- [x] `cd tools/live-seeds && pnpm test`
- [ ] Run `pnpm seed:live-seeds` against local `ingest`, `workers`, and `workflows` services for an end-to-end smoke test


docker/reset-postgres.sh
```diff
@@ -46,4 +51,12 @@ pnpm --filter @platform/db-postgres pg:migrate
 echo "Seeding database..."
 pnpm --filter @platform/db-postgres pg:seed
 
+if [ "$TEMPORAL_WAS_RUNNING" -eq 1 ]; then
+	# Temporal stores its own state in Postgres, so restart it after the reset.
+	echo "Restarting temporal..."
+	docker compose restart temporal
```
andresgutgon, line 57: Thanks

andresgutgon (approved):

PR #2768, opened 2026-04-16
Optimize live trace filtering

csansoon:
### Problem

After a trace is completed, reliability work was split across different workers for managing Live Evaluations Curation and Execution, Live Annotation Queues Curation and System Annotations FanOut. All of these three jobs follow a common selection logic where, depending on the item's Sampling settings and FilterSet, it decides which of the elements must be selected. This logic is not only duplicated, but also running the filter-set queries is expensive enough once, and unnecessarily expensive to do similar queries for the same trace id accross different workers.

### Solution

Introduce a **single debounced downstream job**, `trace-end`, driven from the debounced **`SpanIngested`** dispatcher.
One worker run **loads the trace once**, builds a **unified candidate set** (live evaluations, live queues, system queues), runs **shared selection logic** in `@domain/spans` (`selectTraceEndItemsUseCase`)—including **one batched trace filter query** when filters apply—then **applies** subsystem-specific side effects: publish `live-evaluations` / `execute`, insert live queue rows, and start the system flagger workflow where appropriate.

`projects` / `checkFirstTrace` remains **immediate** (not debounced with trace-end) so first-trace UX is unchanged.

### New workflow (end-to-end)

1. **Span ingest** — unchanged boundary: spans land via `span-ingestion` / `ingest` and `processIngestedSpansUseCase` still emits **`SpanIngested`** per trace after durable span writes.
2. **`domain-events` / `dispatch`** — `SpanIngested` handler publishes:
   - **`trace-end` / `run`** with **debounce + dedupe** keyed by org/project/trace (constant `TRACE_END_DEBOUNCE_MS` from `@domain/spans`).
   - **`projects` / `checkFirstTrace`** as today.
3. **`trace-end` / `run` worker** — Orchestration only:
   - Resolve trace from ClickHouse; if missing, skip.
   - Load **active evaluations** (`listAllActiveEvaluations`), **live queues**, **system queues** in parallel.
   - Build **namespaced selection keys** and specs (sampling, optional filters, stable sample keys).
   - Call **`selectTraceEndItemsUseCase`**: sample → normalize/prune filters → dedupe by filter fingerprint → **`listMatchingFilterIdsByTraceId`** once when needed → per-key decisions (`selected` / `sampled-out` / `filter-miss`).
   - **Apply**: enqueue **`live-evaluations` / `execute`** (with existing turn/dedupe helpers), **`materializeLiveQueueItemsUseCase`** for selected live queues, **Temporal** `systemQueueFlaggerWorkflow` start with Redis-backed “once per trace/slug” progress.

Selection completes **before** apply so **new scores from the same run cannot affect** the same selection pass.

### Queue topics & tasks

| Topic / task | Change | Why |
|--------------|--------|-----|
| **`trace-end` / `run`** | **Added** | Single debounced unit of work after trace activity settles. |
| **`live-evaluations` / `execute`** | **Kept** | Still the execution rail for running an evaluation against a trace; only the **enqueue** path was removed from the old design. |
| **`live-evaluations` / `enqueue`** (if it existed) | **Removed** | Enqueueing is folded into trace-end’s apply phase. |
| **`live-annotation-queues`**, **`system-annotation-queues`** (and related tasks) | **Removed** | Curation/fan-out selection is folded into trace-end + domain use cases. |
| **`domain-events` / `dispatch`** | **Updated** (`SpanIngested`) | Dispatcher publishes `trace-end:run` + `projects:checkFirstTrace` instead of multiple reliability fan-outs. |

### Domain / package responsibilities

- **`@domain/spans` — `selectTraceEndItemsUseCase`**  
  Pure selection: deterministic sampling, empty-filter normalization (no CH when filters are empty), **deduped filter fingerprints**, one **`TraceRepository.listMatchingFilterIdsByTraceId`** call, structured decisions/reasons.

- **`@domain/evaluations`**  
  **`listAllActiveEvaluations`** for candidate discovery; **`buildLiveEvaluationExecutePublication`** (and helpers) stay the source of truth for **what** to publish to `live-evaluations` / `execute`.

- **`@domain/annotation-queues`**  
  **`materializeLiveQueueItemsUseCase`** performs **Postgres** inserts + counter updates for selected live queues (replaces “curate” worker responsibilities for this path).

- **`apps/workers`**  
  - **`trace-end.ts`**: wiring + apply + logging summaries.  
  - **`live-evaluations.ts`**: **execute-only** consumer.  
  - Removed workers that only existed for the old fan-out.  
  - **`domain-events.ts`**: `SpanIngested` routing only (dispatcher pattern preserved).

### Docs & specs

`docs/spans.md`, `docs/reliability.md`, `docs/evaluations.md`, `docs/annotation-queues.md`, `docs/annotations.md`, and `specs/reliability.md` are updated so **durable docs** describe the unified `trace-end:run` runtime and **sample-before-filter** semantics.

### How to review

- Confirm **no remaining references** to removed queue names/tasks in workers and publishers.  
- Trace **debounce behavior**: repeated spans for the same trace should **reset** the pending `trace-end:run` job, not stack unbounded work.  
- Verify **empty evaluation filters** still mean “match all” **without** issuing a trace filter query.  
- **`live-evaluations` / `execute`** payloads and idempotency (dedupe keys, first-turn checks) still match product expectations.

### Testing

- Worker tests updated for `domain-events`, `trace-end`, live evaluation **execute**, and live monitoring where queue names changed.  
- **`selectTraceEndItemsUseCase`** covered in `@domain/spans` unit tests.


apps/workers/src/workers/trace-end.ts
```diff
@@ -0,0 +1,493 @@
...
+        })
+        .pipe(Effect.catchTag("NotFoundError", () => Effect.succeed(null)))
+
+      if (traceDetail === null) {
+        return {
+          action: "skipped",
+          reason: "trace-not-found",
+          traceId: payload.traceId,
+        } satisfies TraceEndRunResult
+      }
```
geclos, line 171: cool

apps/workers/src/workers/trace-end.ts
```diff
@@ -0,0 +1,493 @@
...
+        workflowId,
+      },
+    )
+
+    yield* progress.markComplete(workflowId)
+    return true
+  })
+}
+
+export const runTraceEndUseCase =
```
geclos: why not move the use cases to the spans domain, is not like you don't have trace-end related use code in there already. keep boundaries thin.

geclos (dismissed):

apps/workers/src/workers/trace-end.ts
csansoon: sure!

geclos (dismissed):
tell the llm to review the docs they are probably stale now also

apps/workers/src/workers/trace-end.ts
```diff
@@ -0,0 +1,375 @@
...
+    yield* progress.markComplete(workflowId)
+    return true
+  })
+}
+
+export const runTraceEndJob =
+  ({ publisher, workflowStarter, postgresClient, clickhouseClient, redisClient }: RunTraceEndDeps) =>
+  (payload: TraceEndPayload) =>
+    Effect.gen(function* () {
+      const loaded = yield* loadTraceForTraceEndUseCase(payload)
```
andresgutgon, line 163: Move this code to the domain and test there. Code in `apps/**` should be just routing

csansoon: talked offline

geclos (approved):

andresgutgon (approved):

PR #2774, opened 2026-04-16
Poll evaluation alignment workflow in the UI

neoxelox:


apps/workflows/src/workflows/evaluation-alignment-workflow.ts
```diff
@@ -1,9 +1,19 @@
+// Import from the `./alignment/workflow` subpath (not from `@domain/evaluations`)
+// so the Temporal workflow bundle doesn't pull in `browser.ts`, which
+// re-exports `helpers.ts` — which imports from `@domain/shared`'s barrel
+// and in turn loads `cache.ts` / `ServiceMap.Service`. The `Service` ctor
+// writes to `Error.stackTraceLimit`, forbidden in the Temporal workflow
+// sandbox. This subpath file is self-contained (constants + a type only).
```
andresgutgon, line 6: Looks like we have a little mess here cc @geclos

neoxelox: Yes but let's fix this alone in another PR please

andresgutgon (approved):

PR #2779, opened 2026-04-17
Try fix web build with temporal client dependencies

andresgutgon:
## What

Staging web was failing to connect to Temporal Cloud with an opaque gRPC error:
\`\`\`
TemporalConnectionError: Failed to connect to Temporal at ZONE.aws.api.temporal.io:7233
  (namespace latitude-llm-staging.SOMETHING): Error
  -- rawError: Error: undefined undefined: undefined
  -- code: undefined, details: undefined
\`\`\`

## Why

\`@temporalio/client\` pulls in \`@grpc/grpc-js\` and \`protobufjs\`. Nitro was **partially** bundling the tree — \`protobufjs\` got bundled into an SSR chunk (\`_ssr/protobufjs-*.mjs\`) while \`@grpc/grpc-js\` resolved from \`node_modules\` at runtime. protobufjs uses \`new Function()\` to codegen encoders/decoders at load time, and those generated functions close over whichever protobufjs module instance was active at codegen time. When \`@temporalio/proto\` (bundled) and \`@grpc/grpc-js\` (external) disagree on that instance, gRPC status deserialization silently drops every field → \`undefined undefined: undefined\`.

## Fix

Force the entire Temporal dep tree to live on one side of the bundler boundary: all external, all resolved at runtime from the same pnpm store entry.

- **\`apps/web/vite.config.ts\`** — Nitro \`rollupConfig.external\` + \`rolldownConfig.external\` + \`ssr.external\` all include \`@temporalio/*\`, \`@grpc/*\`, \`protobufjs\`, \`@protobufjs/*\`, \`long\`.
- **\`.npmrc\`** — \`public-hoist-pattern\` for those same packages so Node's upward resolution from \`.output/server/index.mjs\` finds them at \`/app/node_modules/\` in the Docker image. Without this, the externalized requires would fail because the packages only exist nested under \`packages/platform/workflows-temporal/node_modules/\`.
- **\`apps/web/package.json\`** — \`@temporalio/client\` added as a direct dep. Not imported by web source, but declared so the runtime dependency is explicit at the web boundary (knip ignored accordingly).
- **\`Dockerfile\`** — \`.npmrc\` copied into the \`deps\` stage so \`pnpm fetch\` sees the same hoist rules; \`CI=true\` on the install step to auto-confirm the modules-dir purge when hoist patterns change.

## Why not move Temporal calls to workers?

Considered. Rejected: Redis and Postgres clients work fine from web, so "web can't talk to infra" isn't the right principle. The actual issue was bundler-specific to gRPC + protobufjs codegen. With this fix, adding more Temporal use cases in web server functions keeps working.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,7 @@
+public-hoist-pattern[]=@temporalio/*
+public-hoist-pattern[]=@grpc/*
+public-hoist-pattern[]=protobufjs
+public-hoist-pattern[]=@protobufjs/*
+public-hoist-pattern[]=long
+
+confirm-modules-purge=false
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 # ---------------------------------------------------------------------------
 FROM base AS deps
 
-COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
+COPY pnpm-lock.yaml pnpm-workspace.yaml package.json .npmrc ./
 
 # Populate pnpm store from lockfile only for better cache reuse
 RUN --mount=type=cache,id=pnpm,target=/root/.local/share/pnpm/store \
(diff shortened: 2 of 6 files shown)


csansoon (approved):
lgtm

PR #2810, opened 2026-04-20
hotfix(workflows): inline evaluation-alignment workflow contract to unbreak deploy

andresgutgon:
## Summary

Staging workflows deploys (and eventually prod) are stuck in a restart loop because `Worker.create` crashes at startup:

> Module not found: Error: Can't resolve \`@domain/evaluations/alignment/workflow\` in \`/app/apps/workflows/src/workflows\`

Worker exits with code 1 → ECS restarts the task → new task dies the same way → deploy never stabilizes.

This PR unblocks the deploy by inlining the three symbols the workflow needs (two string constants + one narrow state type) directly into `apps/workflows/src/workflows/evaluation-alignment-workflow.ts`, so the workflow bundle no longer goes through the failing subpath import.

## Why the import existed in the first place

The original author of #2774 deliberately avoided importing the symbols from the main `@domain/evaluations` barrel. That barrel re-exports `browser.ts` → `helpers.ts` → `@domain/shared` → `cache.ts` → `ServiceMap.Service`, and `Service`'s constructor writes to `Error.stackTraceLimit`, which Temporal's workflow sandbox forbids. Going through the barrel would crash the workflow on first run.

The subpath export at `packages/domain/evaluations/src/alignment/workflow.ts` was created to give the workflow a sandbox-safe import path with no transitive dependencies.

## Why the subpath fails at deploy time

The subpath works in local dev and typecheck because the TypeScript/tsx resolver honors the `exports` map. But `@temporalio/worker`'s internal webpack bundler — invoked by `Worker.create({ workflowsPath })` at worker startup inside the container — fails to resolve the subpath. Most likely candidates (either would explain the symptom):

1. The bundled webpack version in the SDK does not read package `exports` subpaths the same way the top-level Node resolver does.
2. The container image does not expose `packages/domain/evaluations/src/` in a location the bundler walks to (pnpm symlink or Docker layer pruning).

Either way, the `Module not found` happens before the worker even registers with Temporal.

## Why inlining is the right hotfix

The three symbols are:

- two string literals (`\"scheduleRefresh\"`, `\"getEvaluationAlignmentWorkflowState\"`)
- one narrow \`readonly\` type

No runtime code, no imports, nothing the sandbox could object to. Inlining them inside the workflow file sidesteps both the sandbox issue (that motivated the subpath in the first place) and the webpack resolution issue (that broke the deploy) with zero risk.

The same symbols are still consumed by \`apps/web\` and \`apps/workers\` via the \`@domain/evaluations\` barrel — those call sites are **not** inside the workflow sandbox, so the barrel is safe for them and no changes are needed there.

## Follow-up (not this PR)

- Decide whether to delete the now-orphaned \`packages/domain/evaluations/src/alignment/workflow.ts\` file and its \`exports\` map entry, since the only consumer is gone.
- Or, investigate the Temporal SDK / Docker build so subpath exports resolve at bundle time again, then restore the shared-contract pattern.

## Test plan

- [x] \`turbo typecheck --filter=@app/workflows\`
- [x] \`turbo test --filter=@app/workflows\`
- [x] \`biome check\` on the touched file
- [ ] After merge: retry the staging deploy and confirm the workflows service reaches steady state without looping on \`Module not found\`


apps/workflows/src/workflows/evaluation-alignment-workflow.ts
```diff
@@ -1,14 +1,3 @@
...
-// re-exports `helpers.ts` — which imports from `@domain/shared`'s barrel
-// and in turn loads `cache.ts` / `ServiceMap.Service`. The `Service` ctor
-// writes to `Error.stackTraceLimit`, forbidden in the Temporal workflow
-// sandbox. This subpath file is self-contained (constants + a type only).
-import {
-  EVALUATION_ALIGNMENT_REFRESH_SIGNAL,
-  EVALUATION_ALIGNMENT_STATE_QUERY,
-  type EvaluationAlignmentWorkflowState,
-} from "@domain/evaluations/alignment/workflow"
 import {
```
neoxelox, line 7: Remove the constants from the other file "@domain/evaluations/alignment/workflow"? now we have the same constants and type in 2 places

neoxelox (approved):

PR #2812, opened 2026-04-20
Annotations API changes

andresgutgon:
# What?

Kick-off PR for the **Latitude dogfood epic** — see the full plan in [`prd/annotate-annotated-flagger-and-enrichment-dogfood.md`](./prd/annotate-annotated-flagger-and-enrichment-dogfood.md). This PR lands **Phase 1 (Annotations API changes)**; later phases (scoreId identity plumbing, Fern SDK, domain/platform packages, worker, UI) land in follow-up PRs.

### `POST /v1/organizations/:organizationId/projects/:projectId/annotations`

- **Publish by default.** New `draft?: boolean` on the request body, defaulting to `false`. Published writes land with `draftedAt = null` and emit `ScoreCreated` with `status: "published"` in the same transaction — issue discovery picks them up immediately instead of waiting for the 5-minute debounced `annotation-scores:publish` task. Pass `draft: true` to keep the existing editable-draft behavior used by the managed UI.
- **Trace resolution union.** The flat `traceId` field is replaced by a discriminated `trace` union:
  - `{ by: "id", id }` — exact trace id, **verified to belong to the caller's org + project** before use (closes a cross-tenant footgun where any caller could write an annotation against any trace id).
  - `{ by: "filters", filters }` — reuses the shared `FilterSet` from `@domain/shared`. `0` matches → 404, `>1` matches → 400, exactly `1` → proceed. This is the primitive the dogfood flows (SDK-via-Fern in Phase 3+) will use to resolve the right Latitude telemetry span from a `scoreId` attribute.
- **`sourceId` is forced to `\"API\"`** inside the use case. The route no longer passes it.

### Domain

- New `submitApiAnnotationUseCase` (`packages/domain/annotations/src/use-cases/submit-api-annotation.ts`) — the single public-API entry point. Resolves the trace, branches on `draft`, and delegates to the existing `writeDraftAnnotationUseCase` or the new publish-now primitive.
- New `writePublishedAnnotationUseCase` (`packages/domain/annotations/src/use-cases/write-published-annotation.ts`) — thin primitive that calls `writeAnnotation(input, null)` under a dedicated span.
- New `submitApiAnnotationInputSchema` (`packages/domain/annotations/src/helpers/annotation-public-api-schema.ts`) — public-API-only. The lower-level `persistDraftAnnotationInputSchema` stays untouched for internal callers.

### Breaking change

`POST /v1/.../annotations` now rejects the old flat `traceId` shape. Per the PRD, we accept this break because the endpoint has no known external consumers and the new default aligns with the documented \"human-created annotations are the strongest human signal\" rule in `docs/annotations.md`.

### Tests

- **8 unit tests** for `submitApiAnnotationUseCase`: by-id happy path, cross-tenant rejection, filter match / 0-match / 2+match, draft vs published branching, `sourceId = \"API\"` invariant.
- **7 API contract tests** against the PGlite + ClickHouse testkit covering all of the above end-to-end (including the 404/400 error shapes).
- `pnpm --filter @app/api typecheck` and `pnpm --filter @domain/annotations typecheck` green.

## Out of scope for this PR

Phases 2–5 — `scoreId` identity plumbing, OpenAPI emission + Fern SDK setup + `@latitude-data/sdk@6.0.0-alpha.0` publish workflow, `@domain/product-feedback` + `@platform/latitude-api` + BullMQ worker, and the post-decision popover UI — are documented in the PRD and will land as follow-up PRs.

## Test plan

- [x] CI green on this branch.
- [ ] Manual smoke: create an annotation by `{ by: \"id\", id }`, confirm `draftedAt = null` in the response and a published `ScoreCreated` event in the outbox.
- [ ] Manual smoke: create an annotation by `{ by: \"filters\", filters }` with a single match, confirm trace is resolved.
- [ ] Manual smoke: attempt a cross-tenant `{ by: \"id\" }` write against a trace in another project, confirm 404.


geclos:
Remember to split in small PRs 🙏🏼

packages/domain/annotations/src/use-cases/submit-api-annotation.test.ts
```diff
@@ -0,0 +1,363 @@
...
+          feedback: "Should not persist",
+          organizationId: cuid,
+          projectId: ProjectId(projectCuid),
+        }).pipe(Effect.provide(layer)),
+      )
+
+      expect(exit._tag).toBe("Failure")
+      if (exit._tag === "Failure") {
+        const failure = exit.cause
+        expect(JSON.stringify(failure)).toContain("BadRequestError")
```
geclos: make sure the error message is human readable and explains why we failed here

packages/domain/annotations/src/use-cases/submit-api-annotation.test.ts
```diff
@@ -0,0 +1,363 @@
...
+          organizationId: cuid,
+          projectId: ProjectId(projectCuid),
+        }).pipe(Effect.provide(layer)),
+      )
+
+      expect(exit._tag).toBe("Failure")
+      if (exit._tag === "Failure") {
+        const failure = exit.cause
+        // Tagged error surfaces inside the failure cause
+        expect(JSON.stringify(failure)).toContain("NotFoundError")
```
geclos: ditto about the error message

packages/domain/annotations/src/use-cases/submit-api-annotation.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    const page = yield* traceRepository.listByProjectId({
+      organizationId,
+      projectId: ctx.projectId,
+      options: { filters: trace.filters, limit: 2 },
+    })
+
+    if (page.items.length > 1) {
+      return yield* new BadRequestError({
+        message: "Multiple traces match the provided filters. Refine the filter set to identify exactly one trace.",
```
geclos: lovely

geclos (dismissed):

packages/domain/annotations/src/use-cases/submit-api-annotation.ts
```diff
@@ -0,0 +1,135 @@
...
+      })
+    }
+
+    const [match] = page.items
+    if (match === undefined) {
+      return yield* new NotFoundError({ entity: "Trace", id: "by-filter" })
+    }
```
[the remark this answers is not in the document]
andresgutgon: Valid concern. I went with the other direction — updated the PRD to match the actual `{ error: string }` shape produced by `honoErrorHandler` rather than introducing structured codes. Adding machine-readable `code` fields touches the shared error handler and every route's error surface, which I'd rather not do in Phase 1. Filed it as a note in the PRD; we can revisit if a future phase actually needs programmatic error discrimination beyond status codes.

packages/domain/annotations/src/use-cases/submit-api-annotation.ts
```diff
@@ -0,0 +1,135 @@
...
+ * `writeDraftAnnotationUseCase` / `writePublishedAnnotationUseCase` primitives
+ * remain for internal callers that already have a resolved trace id and a
+ * different source.
+ */
+export const submitApiAnnotationUseCase = Effect.fn("annotations.submitApiAnnotation")(function* (
+  input: SubmitApiAnnotationRequest,
+) {
+  yield* Effect.annotateCurrentSpan("annotation.projectId", input.projectId)
+  yield* Effect.annotateCurrentSpan("annotation.trace.by", input.trace.by)
+
```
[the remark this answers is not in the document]
andresgutgon: Good catch — fixed. Moved the `annotation.trace.by` / `annotation.draft` span annotations *after* `parseOrBadRequest` so any malformed body (missing `trace`, non-object, etc.) bubbles up as a 400 `BadRequestError` instead of an uncaught `TypeError` → 500.

packages/domain/annotations/src/helpers/annotation-public-api-schema.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * - `{ by: "id", id }` — exact trace id lookup.
+ * - `{ by: "filters", filters }` — resolve a single trace from a `FilterSet`.
+ *
+ * The filter path exists for callers (including our own SDK) that do not have
+ * the raw OpenTelemetry trace id at hand but can identify the trace via
+ * attributes such as `attributes.scoreId` or `metadata.userId`.
+ */
+const traceRefSchema = z.discriminatedUnion("by", [
+  z.object({ by: z.literal("id"), id: z.string().min(1) }),
```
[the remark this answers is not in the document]
andresgutgon: Fixed. Swapped `z.string().min(1)` for the shared `traceIdSchema` (32-char length check + `TraceId` brand transform), so malformed ids now fail with a 400 validation error before ever hitting the repository — matching the pre-change behavior of the top-level `traceId` field.

packages/domain/annotations/src/use-cases/submit-api-annotation.test.ts
andresgutgon: Done. Tightened the message on the multi-match path to *"Trace filter matched more than one trace in this project. Refine the filter set so it identifies exactly one trace."* and updated the test to assert on both the `"more than one trace"` and `"Refine the filter set"` substrings so future refactors can't silently regress the UX.

packages/domain/annotations/src/use-cases/submit-api-annotation.test.ts
andresgutgon: Same treatment on the 0-match path: the `NotFoundError` now carries *"No trace in this project matches the provided filters"* in the `id` field (which the HTTP handler surfaces as the response body), and the test asserts on that exact substring. For the by-id cross-tenant case the error body carries the offending trace id so the caller can diagnose it from the response alone — also asserted.

andresgutgon:
All review comments addressed and pushed as an amend to the Phase 1 commit. Summary:

- **Copilot × 3** (error shape / parse ordering / trace id validation) — all valid, all fixed. Details in each thread.
- **@geclos × 2** (error message readability) — valid, error bodies now explain *why* and *how to fix*, and the tests assert on those substrings so the UX is regression-guarded.
- **@geclos "lovely"** 🫶 — no action (thank you!).
- **@geclos "remember to split in small PRs"** — 100% agreed. This PR intentionally lands only **Phase 1 (Annotations API changes)** of the dogfood epic per the PRD. Phases 2–5 (scoreId identity plumbing, Fern SDK + publish, `@domain/product-feedback` + worker, UI) land as follow-up PRs.

Tests: 8 unit + 8 contract green, both typechecks green, Biome clean.

apps/web/vite.config.ts
```diff
@@ -50,7 +50,7 @@ export default defineConfig({
     external: ["@temporalio/client", "@temporalio/proto", "@grpc/grpc-js", "protobufjs", "long"],
   },
   resolve: {
-    conditions: ["source"],
+    conditions: ["source", "browser"],
```
csansoon, line 53: Nice!

csansoon (approved):

PR #2828, opened 2026-04-22
feat(web): GTM conversion tracking for signup flows

geclos:
## Summary

- Add Google Tag Manager container (head + noscript) to `__root.tsx`, gated by new `VITE_LAT_GTM_CONTAINER_ID` env var.
- New `/api/auth/{google|github}/start` endpoint that captures whitelisted tracking params (`gclid`, `fbclid`, `ttclid`, `li_fat_id`, `msclkid`, `_gl`, `utm_*`, `baker_anon_id`, `baker_session_id`), encodes them into Better Auth's `callbackURL` / `newUserCallbackURL`, and 302s to the OAuth provider with the state cookies returned by `auth.api.signInSocial`.
- Login page now redirects social sign-in through that endpoint and tags the magic-link `newUserCallbackURL` with `signup=email` plus tracking params.
- Global `<SignupCompleteWatcher />` listens for `?signup=<method>` on any route, fetches the active session, and pushes a `signup_complete` event to `dataLayer` once per session before stripping the params from the URL.
- Welcome loader's auto-redirect to `/` now forwards `signup` + tracking params so invited users (single-org case) still trigger the event.

## Notes

- LinkedIn was listed in the spec but isn't currently configured as a Better Auth social provider, so the start endpoint only accepts `google` and `github`. Adding LinkedIn is a config-only follow-up.
- `signup_complete` payload uses `user.email` plus a best-effort first/last split of `user.name`. For email signups landing on `/welcome`, name may be empty until the user fills the create-org form — per spec, partial data is acceptable; GTM dedupes per session.

## Test plan

- [ ] Set `VITE_LAT_GTM_CONTAINER_ID=GTM-5NWGV24H` in production env and verify the GTM container loads on every page (head script + noscript iframe).
- [x] Visit `/login?gclid=test_id&utm_source=test`, click Continue with Google. Confirm Google's consent screen, return to dashboard, and verify `gclid=test_id&utm_source=test` are present on the final URL.
- [x] Repeat the above for GitHub.
- [x] Verify `signup_complete` fires in GTM Preview mode with `signup_method: 'google'|'github'` and the user's email when registering via Google/GitHub.
- [x] Verify `signup_complete` fires with `signup_method: 'email'` when registering via the magic-link form.
- [x] Verify the event does NOT re-fire on subsequent navigations within the same session.


apps/web/src/routes/__root.tsx
```diff
@@ -73,7 +82,19 @@ function RootDocument({ children }: Readonly<{ children: ReactNode }>) {
         <HeadContent />
       </head>
       <body>
+        {GTM_CONTAINER_ID ? (
```
andresgutgon: Do we need GTM for the app? This script is a performance monster. Can we use it only in the signup flow?

geclos: according to los buenos de landing bakery: 

> The following snippets must be present on every page of console.latitude.so.

andresgutgon: Buff

andresgutgon: I don't get it. Why private URLs needs this extra performance penalty load. We convert outside no?

andresgutgon (approved):

PR #2830, opened 2026-04-22
feat(evaluations): gate incremental alignment refresh on script hash match

csansoon:
## Problem

Phase 12 specified that incremental alignment refresh should only run **"when the evaluation hash is unchanged"**. Until now that invariant was held _implicitly_: the refresh workflow trusted that `evaluation.script`, `evaluation.alignment.evaluationHash`, and `evaluation.alignment.confusionMatrix` always land together via `persistAlignmentResultUseCase`, so the persisted matrix was always produced by the script that is live right now.

Nothing actually re-hashed the live script at refresh time to confirm that. The `decideAlignmentRefreshStrategy` helper only compared MCC drop — it never looked at the hash.

If any future code path mutates `evaluation.script` without going through the atomic alignment write — an admin tool, a data migration, a direct repo call — the refresh workflow would silently merge judgments from two different scripts into the same confusion matrix. The resulting metrics would be garbage. Defense against that is a one-line hash comparison; worth doing now rather than once it bites.

## Change

Make the invariant explicit, and fall back cleanly when it's violated.

### Domain layer

`loadAlignmentStateOrInactiveUseCase` now computes `sha1(evaluation.script)` alongside the usual state load, and returns two extra fields:

```ts
| { readonly status: "active"
    readonly state: LoadedEvaluationAlignmentState
    readonly incrementalEligible: boolean   // sha1(script) === evaluationHash
    readonly currentScriptHash: string      // the freshly computed hash
  }
```

Reuses the existing `hashOptimizationCandidateText` helper from `@domain/optimizations` (the same one `generate-baseline-draft.ts` already calls at alignment time), so match logic is guaranteed symmetric.

### Workflow

`refreshEvaluationAlignmentWorkflow` now branches on `incrementalEligible`:

| Branch | Path |
| --- | --- |
| `true` (common case) | Existing incremental flow — collect examples `createdAfter: state.alignedAt`, judge only those, merge into the persisted matrix, decide `metric-only` / escalate / `no-op`. |
| `false` (hash drift) | Rebuild from scratch — collect **all** curated examples (no `createdAfter`), run `evaluateBaselineEvaluationDraft` against the live script, persist the new matrix with `currentScriptHash` as the fresh `evaluationHash`. Return status `full-metric-rebuild`. |

Rebuild never triggers GEPA. MCC drop is re-evaluated cleanly on the next incremental pass, now that hash and script agree again.

New result variant:

```ts
| { readonly status: "full-metric-rebuild"; readonly totalExampleCount: number }
```

### Spec / docs

`specs/reliability.md` Phase 12 checkbox reworded to describe the enforced hash check and the rebuild fallback. `docs/evaluations.md` alignment-rules bullet updated to match. The existing `- [x]` is preserved — the checkbox promised this behavior; PR 1 shipped the perf part, and this PR closes the correctness part.

### Housekeeping

Pre-commit hook `pnpm knip` was already failing on `main` on two pre-existing gaps (unrelated SDK landings) that blocked my local commit. Fixed in the same PR by adding `examples/sdks/typescript` and extending `packages/sdk/typescript` in `knip.json`. Trivial config-only repair, flagged in the commit body.

## Test plan

- [ ] `pnpm turbo typecheck --filter=@app/workflows --filter=@domain/evaluations` — clean.
- [ ] `pnpm turbo test --filter=@app/workflows --filter=@domain/evaluations --force` — workflows 53/53, domain/evaluations 55/55.
- [ ] New tests cover both branches of the domain check and the workflow's rebuild path (asserts the incremental evaluator is skipped, example collection drops `createdAfter`, persist is called with the freshly computed hash, no GEPA escalation).
- [ ] `pnpm knip` — clean (was previously broken on `main` due to the SDK workspaces).
- [ ] End-to-end: persist an evaluation with a deliberately stale `evaluationHash` (e.g. in a dev script that updates `evaluation.script` without touching the alignment row). Fire the refresh workflow. Temporal UI should show `evaluateBaselineEvaluationDraft` running (not the incremental one), and the persisted row should end up with `evaluationHash === sha1(script)` again.

## Out of scope

- Cleaning up `evaluateIncrementalDraftUseCase`'s own `strategy: "no-op"` fast path — still useful when the hashes match but there are no new annotations.
- Auditing for other places that could mutate `evaluation.script` without going through `persistAlignmentResultUseCase`. There are none today; the hash check is a safety net against future changes more than a fix for a current bug.


csansoon:
Re [`apps/workflows/src/workflows/refresh-evaluation-alignment-workflow.ts` L96](https://github.com/latitude-dev/latitude-llm/pull/2830#discussion_r3123452948) (Copilot's inline comment about `requirePositiveExamples`):

Good catch — confirmed by reading [`collect-alignment-examples.ts:55`](https://github.com/latitude-dev/latitude-llm/blob/main/packages/domain/evaluations/src/use-cases/alignment/collect-alignment-examples.ts#L55): without `requirePositiveExamples: false` the use case throws `BadRequestError` on zero positives, which in the rebuild path would fail the activity and push the workflow through retries instead of rebuilding cleanly.

Fixed in 31487fe2f by passing the flag on the rebuild collect call and updating the workflow test's assertion to pin that contract. Went with the "just pass the flag" option (not the "skip persist on total=0" one) because persisting the empty matrix with the fresh hash is actually what we want: it resets the row to match the live script and puts the next refresh back on the incremental-eligible path ready to merge in new annotations.

knip.json
```diff
@@ -45,7 +45,14 @@
...
       "entry": [],
-      "project": []
+      "project": [],
+      "ignoreBinaries": ["tsup"]
+    },
+    "examples/sdks/typescript": {
+      "entry": ["src/**/*.ts"],
+      "project": ["src/**/*.ts"],
+      "ignoreBinaries": ["tsx"],
+      "ignoreDependencies": ["tsx"]
```
andresgutgon, line 55: Thanks

andresgutgon (approved):

PR #2832, opened 2026-04-22
fix(observability): parent Effect spans to active OTel traces

geclos:
## Summary
- bridge the active OpenTelemetry span into `withTracing` so nested Effect spans stay under the worker or request trace instead of starting a separate trace
- add a regression test for the observability package that exercises the bridge with a real OTel context manager and in-memory span exporter
- add the test dependency and lockfile entry needed to run that regression coverage locally and in CI

## Verification
- pnpm check --filter @repo/observability
- pnpm test --filter @repo/observability
- pnpm typecheck --filter @repo/observability


geclos:
🐮 👦🏼

PR #2837, opened 2026-04-22
fix(usage): Usage was incorrectly calculated in UI breakdown

csansoon:
Fixes this huge discrepancy:
image

Basically, `cacheCreate` was being treated as part of the output, instead of the input, which made the total calculation wrong.

I am also using this PR to update the color and position of some of the components, to make it easier to read and understand semantically.


geclos (dismissed):
add a test and 👌🏼

andresgutgon (approved):

PR #2840, opened 2026-04-23
Fix/ts7 lazy instrumentations

geclos:


apps/api/tsdown.config.ts
```diff
@@ -8,6 +8,7 @@ export default defineConfig({
   platform: "node",
   deps: {
     alwaysBundle: [/@(platform|domain|repo)\/.*/],
+    neverBundle: [/^@traceloop\//, /^@langchain\//, /^langchain($|\/)/],
```
andresgutgon, line 11: But is not just this. I think we are putting all node_modules dependecies inside the build. Has to be a way of dealing with this on monorepos

geclos: we are bundling all latitude-data/telemetry dependencies that's to be expected because we pull the package from source rather than installing from npm registry. This is the correct fix.

The long term fix is to have latitude data telemetry not bundle all its instrumentations by default but this is a product question that I'm unsure how to address because, if we do it, then users would need to manually install the necessary traceloop instrumentation for their use case, rather than just installing the latitude telemetry package.

PR #2854, opened 2026-04-23
feat(backoffice): cross-org search for users, orgs, and projects

csansoon:
## Summary

Introduces the v2 backoffice — a staff-only admin area at `/backoffice` — with a single feature: a cross-organisation search over users, organisations, and projects. This is the foundation; impersonation and other v1 features will come in follow-up PRs.

## Implementation

**Three independent guards** (all must pass; all return `notFound()` so non-admins can't fingerprint the surface):

1. **Route loader guard** — `apps/web/src/routes/backoffice/route.tsx` runs `getSession()` in `beforeLoad`/`loader` and throws `notFound()` unless `users.role === "admin"`. TanStack Start code-splitting means non-admins never fetch the backoffice chunk.
2. **Server-function guard** — every admin `createServerFn` calls `requireAdminSession()` (in `apps/web/src/server/admin-auth.ts`) as its first line, before any IO. Throws `NotFoundError` — not 401/403 — so RPC probes are indistinguishable from calling a non-existent function.
3. **Database guard** — admin handlers use `getAdminPostgresClient()` (a separate pool on `LAT_ADMIN_DATABASE_URL`, superuser) piped through `withPostgres(…, client, OrganizationId("system"))`. The `"system"` scope skips the RLS `set_config` step entirely. Misconfigured deploys fail closed.

**Admin domain** (`@domain/admin`) is a new, separately-installable domain package with one port (`AdminSearchRepository`) and one use-case (`unifiedSearchUseCase`). It depends only on `@domain/shared`, `effect`, `zod`. The Drizzle adapter (`AdminSearchRepositoryLive` in `@platform/db-postgres`) is the only place queries run without an `organization_id` filter — guarded by a file-header warning and safe only via the admin client wiring above.

**Routes & UI** (`apps/web/src/routes/backoffice/`):
- `route.tsx` — role-gated layout with a header + collapsible sidebar (reuses `AppSidebar`/`NavItem` from `/settings`).
- `index.tsx` — redirects `/backoffice` → `/backoffice/search`.
- `search.tsx` — debounced input + entity-type tabs (all / user / organization / project) calling `adminSearch` via TanStack Query.
- `-components/search-results.tsx` — result cards with `Avatar`s (photos or hash-colour initials), project emoji tiles, and human cross-entity labels (memberships on users; owning organisation on projects).

**Better Auth extension** — `user.additionalFields.role` (in `create-better-auth.ts`) surfaces the app-extended `users.role` column on the session with `input: false` so the field is DBA-only.

**App-header entry point** — the authenticated avatar dropdown shows a "Backoffice" option only when `user.role === "admin"`.

## Tests

- `unified-search.test.ts` — use-case (short-circuit, trim, relevance sort `exact > prefix > contains`).
- `admin-search-repository.test.ts` — PGlite-backed adapter (cross-org scanning, case-insensitive match, soft-delete exclusion, memberships join, project→organisation join).
- `admin-auth.test.ts` — guard covers admin/user/null/undefined/missing-role, plus the assertion that the error shape never fingerprints the admin surface.
- `admin.functions.test.ts` — input schema (default type, valid/unknown type, max length).

## Test plan

- [ ] `pnpm --filter @domain/admin test` passes
- [ ] `pnpm --filter @platform/db-postgres exec vitest run --passWithNoTests` passes
- [ ] `pnpm --filter @app/web exec vitest run --passWithNoTests src/server/admin-auth.test.ts src/domains/admin/admin.functions.test.ts` passes
- [ ] Promote a user to admin via SQL (`UPDATE latitude.users SET role = 'admin' WHERE email = '…'`), log out / back in, confirm the avatar dropdown shows "Backoffice" and `/backoffice/search` is reachable
- [ ] Confirm a non-admin hitting `/backoffice` gets the default 404 page (no redirect, no 403)
- [ ] Confirm a non-admin calling `adminSearch` directly via `fetch` receives a `NotFoundError` shape, not a 401/403
- [ ] Confirm search returns entities from multiple organisations in a single response


apps/web/src/routes/backoffice/route.tsx
```diff
@@ -0,0 +1,76 @@
...
+const assertAdminSession = async () => {
+  const session = await getSession()
+  const role = (session?.user as { role?: string } | undefined)?.role
+  if (!session || role !== "admin") {
+    // 404, not redirect: the existence of the backoffice surface must not
+    // leak through error types, redirects, or Location headers.
+    throw notFound()
+  }
+  return session
+}
```
andresgutgon, line 19: This will clearly will used in sub rutes. Also dig into tanstack router. It has to be a way of asserting this for `/backoffice/**` without doing it in all routes

apps/web/package.json
```diff
@@ -15,6 +15,7 @@
     "test": "vitest run --passWithNoTests --dir src"
   },
   "dependencies": {
+    "@domain/admin": "workspace:*",
```
andresgutgon, line 18: Not sure if doing all under `@domain/admin` is the best. I would do `@domain/annotations` and under it a `backoffice.functions.ts` and the same with other domains. Otherwise we end with a huge `admin` domain

csansoon: Hmm I don't know. I think I like it all being in a single package, since all functions here will always have the same requirements, which are different from any other domain

andresgutgon: Sorry yes, I meant  to comment on `apps/web`. Is where my concern is. This is fine. This is all backoffice logic use case concentrated. But in terns of functions. I don't know. Maybe at least split functions by domain inside backoffice domain

andresgutgon (approved):

PR #2862, opened 2026-04-24
feat(backoffice): user impersonation with audit and banner

csansoon:
## Summary

- Staff can open a user detail page from search and impersonate that user for support. Better Auth's `admin` plugin (installed here) tracks the original admin in `sessions.impersonated_by`, so **"Stop impersonating" swaps back without a re-login** — a direct upgrade over v1's session-swap-with-no-return-path.
- Every start/stop is recorded through the outbox (`AdminImpersonationStarted` / `AdminImpersonationStopped`), and while impersonating every authenticated page renders a red top-bar banner plus a small overlay badge on the user avatar — two mitigations for v1's biggest footguns (no audit, no visual cue).
- Three-guard discipline is unchanged: route `notFound()` at the UI, `requireAdminSession()` at every server function, `getAdminPostgresClient()` + `OrganizationId("system")` at the DB. `stopImpersonating` is the one exception — it guards on `session.impersonatedBy` being set instead of role, because during impersonation the current session is the target's.

Builds on #2854. New routes: `/backoffice/users/$userId`. Schema adds four plugin-required columns (`users.banned/banReason/banExpires`, `sessions.impersonatedBy`) — no ban UI exposed, the columns exist solely so the admin plugin can function.

## Test plan

- [x] `pnpm --filter @app/web typecheck`, production build, and bundle-size check all pass locally
- [x] `@domain/admin`, `@platform/db-postgres`, `@app/web` test suites: 301+ tests green (34 new)
- [ ] Apply the migration locally (`pnpm --filter @platform/db-postgres pg:migrate`) and walk through the happy path as `owner@acme.com`: search → user detail → Impersonate → banner appears → navigate as target → Stop impersonating → back to admin
- [ ] `SELECT event_name, payload FROM latitude.events_outbox ORDER BY occurred_at DESC LIMIT 2;` shows the Started/Stopped pair with the right admin/target ids
- [ ] Defence-in-depth probe: as a non-admin, `curl -X POST` the `impersonateUser` / `adminGetUser` RPC URLs with a valid non-admin cookie → same "not found" error shape as hitting a nonsense URL (no 401/403 fingerprint)

### DIFF
--- a/apps/web/src/domains/admin/impersonation.functions.ts
+++ b/apps/web/src/domains/admin/impersonation.functions.ts
@@ -1,0 +1,131 @@
+import { getUserDetailsUseCase } from "@domain/admin"
+import { generateId, NotFoundError, UserId } from "@domain/shared"
+import { AdminUserRepositoryLive, SqlClientLive, withPostgres } from "@platform/db-postgres"
+import { withTracing } from "@repo/observability"
+import { createServerFn } from "@tanstack/react-start"
+import { getRequestHeaders } from "@tanstack/react-start/server"
+import { Effect } from "effect"
+import { z } from "zod"
+import { requireAdminSession } from "../../server/admin-auth.ts"
+import { getAdminPostgresClient, getBetterAuth, getOutboxWriter } from "../../server/clients.ts"
+import { ensureSession } from "../sessions/session.functions.ts"
+
+/**
+ * Exported for input-schema tests.
+ */
+export const impersonateUserInputSchema = z.object({
+  userId: z.string().min(1).max(256),
+})
+
+/**
+ * Start impersonating another user.
+ *
+ * The flow:
+ * 1. Guard: `requireAdminSession()` — rejects non-admins with
+ *    `NotFoundError`, matching the three-guard discipline.
+ * 2. Resolve the target's first organization membership via
+ *    {@link AdminUserRepository}. This also verifies the target
+ *    exists (Better Auth would otherwise create an impersonation
+ *    session for a nonsense id). The first-org id is captured in the
+ *    audit event as a best-effort hint of which tenant the admin was
+ *    about to see.
+ * 3. Write `AdminImpersonationStarted` to the outbox. Writing
+ *    **before** the Better Auth swap means if the swap fails we still
+ *    have a record that an attempt was made. The event's envelope
+ *    `organizationId` is `"system"` — impersonation is platform-wide
+ *    audit, not tenant-owned.
+ * 4. Call Better Auth's `auth.api.impersonateUser`, which mints a new
+ *    session cookie for the target user and stores the admin's id in
+ *    `sessions.impersonatedBy`. The client redirects to `/` after
+ *    this returns.
+ */
+export const impersonateUser = createServerFn({ method: "POST" })
+  .inputValidator(impersonateUserInputSchema)
+  .handler(async ({ data }): Promise<{ ok: true }> => {
+    const { userId: adminUserId } = await requireAdminSession()
+
+    const target = await Effect.runPromise(
+      getUserDetailsUseCase({ userId: UserId(data.userId) }).pipe(
+        withPostgres(AdminUserRepositoryLive, getAdminPostgresClient()),
+        withTracing,
+      ),
+    )
+
+    const targetOrganizationId = target.memberships[0]?.organizationId ?? null
+
+    const outboxWriter = getOutboxWriter()
+    await Effect.runPromise(
+      outboxWriter
+        .write({
+          id: generateId(),
+          eventName: "AdminImpersonationStarted",
+          aggregateType: "user",
+          aggregateId: target.id,
+          organizationId: "system",
+          payload: {
+            adminUserId,
+            targetUserId: target.id,
+            targetOrganizationId,
+          },
+          occurredAt: new Date(),
+        })
+        .pipe(Effect.provide(SqlClientLive(getAdminPostgresClient())), withTracing),
+    )
+
+    const headers = getRequestHeaders()
+    const auth = getBetterAuth()
+    await auth.api.impersonateUser({ body: { userId: target.id }, headers })
+
+    return { ok: true }
+  })
+
+/**
+ * Stop an active impersonation and restore the admin's original session.
+ *
+ * NOTE: this handler deliberately does NOT call `requireAdminSession()`.
+ * During impersonation the current session user is the *target* (role
+ * = `"user"`), so an admin-role check would reject the very call that
+ * needs to succeed. The guard is instead "is `session.impersonatedBy`
+ * set?" — which is only true inside an active impersonation.
+ *
+ * The admin's id is recovered from `session.impersonatedBy` *before*
+ * calling Better Auth's `stopImpersonating` (which swaps the cookie
+ * back and clears the field). The audit event is written before the
+ * swap for the same "record the attempt even if the swap fails"
+ * reason as on start.
+ */
+export const stopImpersonating = createServerFn({ method: "POST" }).handler(async (): Promise<{ ok: true }> => {
+  const session = await ensureSession()
+  const adminUserId = (session.session as { impersonatedBy?: string | null }).impersonatedBy ?? null
+  if (!adminUserId) {
+    // Match the backoffice 404 discipline — never fingerprint admin
+    // surfaces with 401/403 shapes.
+    throw new NotFoundError({ entity: "Route", id: "backoffice" })
+  }
+
+  const targetUserId = session.user.id
+
+  const outboxWriter = getOutboxWriter()
+  await Effect.runPromise(
+    outboxWriter
+      .write({
+        id: generateId(),
+        eventName: "AdminImpersonationStopped",
+        aggregateType: "user",
+        aggregateId: targetUserId,
+        organizationId: "system",
+        payload: {
+          adminUserId,
+          targetUserId,
+        },
+        occurredAt: new Date(),
+      })
+      .pipe(Effect.provide(SqlClientLive(getAdminPostgresClient())), withTracing),
+  )
+
+  const headers = getRequestHeaders()
+  const auth = getBetterAuth()
+  await auth.api.stopImpersonating({ headers })
+
+  return { ok: true }
+})
--- a/apps/web/src/domains/admin/impersonation.functions.ts
+++ b/apps/web/src/domains/admin/impersonation.functions.ts
@@ -1,14 +1,13 @@
 import { getUserDetailsUseCase } from "@domain/admin"
-import { generateId, NotFoundError, UserId } from "@domain/shared"
+import { generateId, UserId } from "@domain/shared"
 import { AdminUserRepositoryLive, SqlClientLive, withPostgres } from "@platform/db-postgres"
 import { withTracing } from "@repo/observability"
 import { createServerFn } from "@tanstack/react-start"
 import { getRequestHeaders } from "@tanstack/react-start/server"
 import { Effect } from "effect"
 import { z } from "zod"
-import { requireAdminSession } from "../../server/admin-auth.ts"
+import { adminMiddleware, impersonatingMiddleware } from "../../server/admin-middleware.ts"
 import { getAdminPostgresClient, getBetterAuth, getOutboxWriter } from "../../server/clients.ts"
-import { ensureSession } from "../sessions/session.functions.ts"
 
 /**
  * Exported for input-schema tests.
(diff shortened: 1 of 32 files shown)


apps/web/src/domains/admin/impersonation.functions.ts
```diff
@@ -0,0 +1,131 @@
...
+ *    audit, not tenant-owned.
+ * 4. Call Better Auth's `auth.api.impersonateUser`, which mints a new
+ *    session cookie for the target user and stores the admin's id in
+ *    `sessions.impersonatedBy`. The client redirects to `/` after
+ *    this returns.
+ */
+export const impersonateUser = createServerFn({ method: "POST" })
+  .inputValidator(impersonateUserInputSchema)
+  .handler(async ({ data }): Promise<{ ok: true }> => {
+    const { userId: adminUserId } = await requireAdminSession()
```
andresgutgon: Try to do this to avoid mistakes
```typescript
const adminMiddleware = createMiddleware({ type: 'function' })
  .server(async ({ next }) => {
    const request = getRequest()
    const session = await auth.getSession({ headers: request.headers })

    if (!session?.user) {
      throw new Error('Unauthorized')
    }
    if (session.user.role !== 'admin') {
      throw new Error('Forbidden: admin only')
    }

    return next({
      context: {
        user: session.user,
      },
    })
  })

// Export a pre-wired factory
export const createBackofficeServerFn: typeof createServerFn = (options) =>
  createServerFn(options).middleware([adminMiddleware])
```

andresgutgon (dismissed):

apps/web/src/routes/backoffice/users/$userId.tsx
```diff
@@ -0,0 +1,102 @@
...
+export const Route = createFileRoute("/backoffice/users/$userId")({
+  loader: async ({ params }) => {
+    try {
+      const user = await adminGetUser({ data: { userId: params.userId } })
+      return { user }
+    } catch {
+      // Both "user not found" and "non-admin" throw NotFoundError from
+      // our server function. From the UI's point of view they are the
+      // same: render the standard 404.
+      throw notFound()
```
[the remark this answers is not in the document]
csansoon: Addressed in 63729404a — good catch.

Slight tweak vs. the suggested patch: I match on the Effect-native `_tag === "NotFoundError"` rather than `error.name === "NotFoundError"`. Loaders can also run client-side on navigations, where the thrown error has been rehydrated from JSON and no longer carries the class prototype; `_tag` is an own enumerable property on `Data.TaggedError` so it survives the RPC boundary. Everything else (DB outage, connectivity, serialization bug) now bubbles to the router error boundary as intended.

packages/domain/events/src/event-payloads.ts
```diff
@@ -115,4 +115,28 @@ export interface EventPayloads {
...
+   * tenant ownership. `targetOrganizationId` records the org whose data
+   * the admin is about to see (the target user's active org at the
+   * moment of impersonation), so queries like "who looked at tenant X?"
+   * have a join key.
+   */
+  AdminImpersonationStarted: {
+    readonly adminUserId: string
+    readonly targetUserId: string
+    readonly targetOrganizationId: string | null
+  }
```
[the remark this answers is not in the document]
csansoon: Addressed in 44025dcaf — updated the JSDoc to match the code rather than the other way around.

The target user’s *actual* active org isn’t reliably available at impersonation time: we’d have to fetch their most recent session (if any) from the DB, and even then Better Auth may assign a different `activeOrganizationId` on the impersonation session, and the admin can switch orgs immediately from the banner. Treating the first-membership hint as authoritative would be misleading regardless. The new doc calls it out as a best-effort hint and points audit consumers to the admin’s subsequent request trail for definitive answers.

packages/platform/db-postgres/src/schema/better-auth.ts
```diff
@@ -49,6 +62,14 @@ export const sessions = latitudeSchema.table(
...
       .references(() => users.id, { onDelete: "cascade" }),
     activeOrganizationId: text("active_organization_id"),
+    /**
+     * Set by the Better Auth `admin` plugin when an admin impersonates
+     * another user. Holds the admin's user id for the lifetime of the
+     * impersonation session, so `stopImpersonating` can restore the
+     * original session without requiring a re-login. Nullable in every
+     * other case.
+     */
+    impersonatedBy: text("impersonated_by"),
```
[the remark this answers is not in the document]
csansoon: Not applying this one. Two reasons:

1. **No-FK rule.** The project explicitly bans foreign keys on new columns — see `.agents/skills/database-postgres/SKILL.md` rule 6: *"new Postgres tables must not add foreign key constraints. Do not use `.references()` or manually create `FOREIGN KEY` constraints. Referential integrity is enforced at the application/domain layer."* Adopting the `.references(() => users.id, { onDelete: "set null" })` suggestion would violate this.

2. **`text` is consistent with what’s already here.** `sessions.activeOrganizationId` in the same table (line 64) also holds an id as `text("active_organization_id")` — no FK, no `cuid`. `impersonatedBy` mirrors that precedent. It’s also what Better Auth’s `admin` plugin declares in its own schema (`{ type: "string" }`, see `better-auth/dist/plugins/admin/schema.mjs`), so switching to `cuid` would diverge from the plugin without upside.

Referential integrity for this specific column is already covered at the app layer: `stopImpersonating` revalidates `session.impersonatedBy` against the current session on every call, and the value is only ever written by Better Auth itself during a supervised impersonation flow.

apps/web/src/domains/admin/impersonation.functions.ts
csansoon: Thanks — thinking about this one out loud because I want to make sure we land in the right place.

Two things I want to flag about *this exact shape*:

**(a) The factory wrapping `createServerFn` is what I had to revert on this PR series.** The first cut of the backoffice (PR #2854) shipped a `createAdminServerFn: typeof createServerFn = (options) => createServerFn(options).middleware(...)` — same shape as your snippet. CI failed with `[MISSING_EXPORT] Error: "withTracing" is not exported by "../../packages/observability/src/browser.ts"` from the web-bundle-size job. Why: TanStack Start’s Vite compiler detects server functions by pattern-matching the literal `createServerFn(...).handler(inline-fn)` chain at the call site (see `start-compiler-plugin/compiler.js`, `KindDetectionPatterns.ServerFn = /\bcreateServerFn\b|\.\s*handler\s*\(/`). When the call site is `createAdminServerFn({...})`, the literal `createServerFn` token is gone, so the compiler doesn’t transform the file, and Node-only module-level imports (`withTracing`, `getAdminPostgresClient`, …) leak into the browser bundle. Fix history on the `backoffice-setup` branch and write-up in `.agents/skills/backoffice/SKILL.md` ("Do not wrap `createServerFn` in a factory").

**(b) The 401/403 split in the error shapes fingerprints the admin surface.** `throw new Error("Unauthorized")` vs `throw new Error("Forbidden: admin only")` lets a non-admin probe tell the two cases apart — exactly the thing we want to hide. Our three-guard discipline standardises on `NotFoundError` with entity `"Route"` / id `"backoffice"` for both so the error shape is indistinguishable from any non-existent server function. Same argument applies to the `Unauthorized` vs any-other-error distinction.

**A variant that WOULD work with the compiler**: attach middleware at each call site, keeping `createServerFn` literal:

```ts
export const impersonateUser = createServerFn({ method: "POST" })
  .middleware([adminMiddleware])
  .inputValidator(impersonateUserInputSchema)
  .handler(async ({ data }) => { ... })
```

The compiler recognises this (it pattern-matches `.handler(...)`) and `createMiddleware(...).server(...)` is separately on the allow-list. The middleware itself is server-only and can freely import Node-only code. Trade-off: the admin author still has to remember to add `.middleware([adminMiddleware])` at each call site, so it stays convention-based — same as remembering `await requireAdminSession()` as the first handler line, which is what we do today. The marginal win is that middleware runs before input validation.

Happy to migrate to that form on a follow-up if you prefer — mostly a consistency/style call. Left as-is for this PR to avoid scope creep, and to keep the already-reverted factory shape documented in the skill as a warning for the next backoffice feature.

csansoon: Migrated to the middleware pattern in 1a151e3fe. Good call — it reads cleaner at the call site and the "guard runs before validation" point is a genuine win.

Two things I changed from the snippet, both flagged in my earlier reply:

1. **Kept `createServerFn` literal at each call site** rather than exposing a `createBackofficeServerFn` factory. The factory shape is what I had to revert on PR #2854 — TanStack Start's Vite plugin detects server functions by pattern-matching the literal `createServerFn` token at the call site, and wrapping it in a named factory hides that token and leaks Node-only imports into the browser bundle. The `.middleware([...])` call is part of the chain the compiler already recognises, so attaching at each call site works fine. The skill doc documents the factory trap.
2. **Errors still collapse to `NotFoundError`** — same 404-fingerprint discipline the rest of the backoffice follows, so non-admin probes can't tell "unauthorized" and "forbidden" apart.

Two middlewares in `apps/web/src/server/admin-middleware.ts`:

- `adminMiddleware` — fetches the session with Better Auth's cookie cache bypassed (PR #2862 also picked up that fix), rejects non-admins, injects `{ adminUserId, user }` into context.
- `impersonatingMiddleware` — used only by `stopImpersonating`. Gates on `session.impersonatedBy` rather than `role`, because during impersonation the current session user is the target (role: "user"). Injects `{ adminUserId, targetUserId }` for the audit event.

The route loader in `backoffice/route.tsx` still calls `requireAdminSession()` directly — route loaders aren't server functions so they can't use `createServerFn` middleware. Both paths share the same `getFreshSession` + `assertAdminUser` internals.

apps/web/src/domains/admin/users.functions.ts
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * Guard: {@link adminMiddleware} runs before the input validator and
+ * rejects non-admins with `NotFoundError` (identical to hitting a
+ * non-existent server function). Queries use
+ * {@link getAdminPostgresClient} + `withPostgres` at the default
+ * `OrganizationId("system")` scope — the only sanctioned RLS-bypass
+ * signal.
+ */
+export const adminGetUser = createServerFn({ method: "GET" })
+  .middleware([adminMiddleware])
```
andresgutgon, line 65: So is not possible to combine `createServerFn` + `adminMiddleware` so devs can use without thinking about it?

apps/web/src/routes/backoffice/route.tsx
```diff
@@ -1,21 +1,23 @@
...
+// surfaces share a single source of truth for the admin gate — including
+// the DB-fresh fetch that sidesteps Better Auth's 5-minute session cookie
+// cache (see the comment on `requireAdminSession`). The route-local
+// wrapper just converts the domain `NotFoundError` into TanStack Router's
+// `notFound()` so the response is indistinguishable from any unknown
+// URL — no redirect, no 403, no Location header leak.
+const guardBackofficeRoute = async () => {
+  try {
+    await requireAdminSession()
+  } catch {
```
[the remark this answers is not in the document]
csansoon: Fixed in b12c4504e. You are right — I applied the `_tag` discrimination to the `$userId.tsx` loader earlier in the PR but missed carrying the same fix across to `guardBackofficeRoute`. Both guards now follow the same pattern: `NotFoundError` collapses to `notFound()`, everything else bubbles to the router error boundary so DB-outage / connectivity / serialization failures surface correctly instead of looking like a 404.

apps/web/src/domains/admin/users.functions.ts
csansoon: Short answer: not today, for a framework reason rather than a design choice.

The only shape that would give us "attach once, never forget" — i.e. make a missing guard a *type* error instead of a review error — is hiding `createServerFn` behind a factory such as

```ts
export const createBackofficeServerFn: typeof createServerFn =
  (options) => createServerFn(options).middleware([adminMiddleware])
```

…and having every admin handler call `createBackofficeServerFn(...)` instead of `createServerFn(...)`.

That is exactly the shape I had to revert on this PR series. TanStack Start's Vite plugin detects server functions by grepping the literal `createServerFn` token at the call site (see `start-compiler-plugin/compiler.js`, `KindDetectionPatterns.ServerFn`); when the call site uses `createBackofficeServerFn`, the literal is gone, the compiler skips the file, and the handler body — along with its Node-only imports (`withTracing`, `getAdminPostgresClient`, `@platform/db-postgres`, …) — gets emitted into the client bundle. The web-bundle-size CI catches it as `MISSING_EXPORT: "withTracing" is not exported by "../../packages/observability/src/browser.ts"`.

So the two shapes that *do* work with the compiler both keep `createServerFn` literal at each call site:

1. `.handler(async ({ data }) => { await requireAdminSession(); … })` — inline guard as first handler line (what this PR shipped before the middleware refactor).
2. `.middleware([adminMiddleware]).handler(…)` — what we have now.

Both are convention-based: the author still has to type the guard. The middleware version is nicer (guard visible at the top of the chain, runs before Zod, context injection), but neither can be enforced by the type system as long as `createServerFn` has to appear literally at the call site.

If TanStack Start ever adds an AST-level alternative (e.g. a decorator, a directive, or a `.handler` overload that requires specific middleware), we can revisit. Until then the closest to "attach once" is: code review + the discipline documented in `.agents/skills/backoffice/SKILL.md`. (Note `stopImpersonating` intentionally uses a *different* middleware (`impersonatingMiddleware`) because during impersonation the current session user is the target — a mandatory-admin-guard factory would break that case anyway, so even if the compiler allowed it we would still want per-call-site selection.)

andresgutgon: I don't understand this comment. I want to talk with a human please

csansoon: If with "combine `createServerFn` + `adminMiddleware`" you mean creating an `createAdminFn`, for example, that's not possible.

It's not possible because Tanstack "compiles" these into client and server methods, and finds them by grepping "createServerFn". So it would not find this new function and it would not work

csansoon: allegedly

andresgutgon (approved):

PR #2869, opened 2026-04-24
refactor(annotation-queues): split deterministic flagger fan-out into its own worker

geclos:
## Summary

Redesigns the flagger pipeline so every flagger's deterministic phase runs on 100% of trace-end events (high-throughput worker), while the LLM phase runs only for sampled no-match or rate-limited ambiguous cases (fault-tolerant Temporal workflow).

**Before:** trace-end sampled per-strategy up-front, then started one Temporal workflow per sampled queue that did deterministic + LLM + draft + persist. A separate loop wrote scores for 3 hardcoded deterministic-only detectors.

**After:** trace-end enqueues a single `deterministic-flaggers:run` job. That worker loads the trace once, fans out across every registered strategy with per-strategy `Effect.catch` isolation, and routes each result:
- **matched** → write a `SYSTEM`-authored score directly with the strategy-provided `feedback` string (issue flows via `issues:discovery`). No queue item, no LLM.
- **no-match** (LLM-capable strategies) → sampling → enqueue `start-flagger-workflow`.
- **ambiguous** (LLM-capable strategies) → Redis rate-limit per `{org, slug}` → enqueue `start-flagger-workflow`. Prevents a hot trace pattern from stampeding the workflow queue with thousands of LLM calls/min.

The `start-flagger-workflow` worker is a thin `workflowStarter.start(...)` call with bounded BullMQ retries (`attempts: 4`, exponential 2s) so short Temporal outages don't replay the whole deterministic fan-out. Exhausted attempts log + drop today (TODO to DLQ + alert).

## Deterministic feedback (strict, type-enforced)

`DetectionResult.matched` is `{ kind: "matched"; feedback: string }` — every strategy that returns `matched` must produce a feedback string. The worker writes that exact string to the score's `feedback` column and `metadata.rawFeedback`, which drives issue clustering. This means matched-deterministic doesn't need a draft annotation step; the deterministic detector is authoritative for both the decision and the human-readable explanation.

| Strategy | Matched feedback shape |
| --- | --- |
| `jailbreaking` | "Jailbreak attempt: matched a high-precision bypass pattern (prompt injection / instruction override)" |
| `nsfw` | "NSFW content: matched a high-precision workplace-inappropriate pattern" |
| `tool-call-errors` | Per-match: malformed tool call, duplicate tool_call id, `Tool "X" returned error: <snippet>`, etc. |
| `output-schema-validation` | Per-match: trailing comma, unclosed JSON string, JSON parse failure |
| `empty-response` | "Assistant response was empty or whitespace only" / "…degenerate: only the character X repeated" |
| `trashing` | "Trashing: identical tool+args invocation repeated N times" |

`frustration`, `refusal`, and `laziness` are deliberately ambiguous-only — their deterministic signals are conservative pre-filters that route to LLM rather than write a score directly.

## Cross-strategy suppression (dependency graph)

`QueueStrategy` gains an optional `suppressedBy?: readonly string[]`. The deterministic worker now runs strategies in two phases:

1. **Phase 1** — strategies with no `suppressedBy` run in parallel; their `matched` decisions form a `phase1MatchedSlugs` set.
2. **Phase 2** — strategies with `suppressedBy` short-circuit before `hasRequiredContext` / `detectDeterministically` if any listed suppressor is in that set, emitting `{ action: "suppressed", suppressedBy }` instead of running detection or enqueueing the LLM workflow.

Only `matched` (not `ambiguous`) suppresses, so suppression stays high-precision. Initial wiring:

| Phase-2 strategy | Suppressed by | Reasoning |
| --- | --- | --- |
| `refusal` | `jailbreaking`, `nsfw` | A justified refusal of a jailbreak / NSFW prompt is correct behavior, not a defect |
| `laziness` | `trashing` | Identical-tool-call loop is a different failure mode than punting/deferring work |

This fixes a real false-positive class where the refusal flagger fired on traces in which the assistant correctly refused a jailbreak attempt.

## Key code changes

- The 3 deterministic-only matchers (`tool-call-errors`, `output-schema-validation`, `empty-response`) fold into the unified `QueueStrategy` registry as strategies without `buildSystemPrompt`/`buildPrompt`. `isLlmCapableStrategy()` distinguishes them.
- Missing `detectDeterministically` defaults to `no-match` (was `ambiguous`).
- `classifyTraceForQueueUseCase` drops the deterministic phase — workflow is LLM-only now.
- Generic `checkRedisRateLimit` in `@platform/cache-redis` (fail-open).
- `PublishOptions` extended with `attempts` + `backoff` for per-job BullMQ retry policy.
- New deterministic pre-checks for the 4 previously LLM-only strategies (`trashing`, `refusal`, `laziness`, `frustration`) — only `trashing` returns `matched` (LoopGuard-style: identical tool+args ≥3×); the other three are ambiguous-only.
- Two-phase `processDeterministicFlaggersUseCase` with `suppressedBy` driving phase-2 short-circuits and a new `suppressed` `StrategyDecision` variant for telemetry.

## Behavior change to flag

For `jailbreaking` and `nsfw`, a deterministic match now writes the score directly (with the deterministic feedback string above) and skips the queue-item + draft-annotation creation. Previously those went through the workflow and produced a human-review draft. This matches the existing deterministic-only matcher behavior and was confirmed as intended in design.

## Test plan

- [x] `pnpm turbo typecheck` passes across all 57 packages
- [x] `@domain/annotation-queues` tests pass (215 tests incl. unit coverage for every deterministic branch / threshold boundary / phrase category and the new suppression matrix)
- [x] `@app/workers` tests pass
- [x] `@app/workflows`, `@platform/cache-redis`, `@platform/queue-bullmq` tests pass
- [ ] Manual: deploy to staging, verify `deterministic-flaggers:run` jobs succeed across all strategies
- [ ] Manual: verify per-`{org, slug}` rate-limit keys appear in Redis under `ratelimit:flagger-ambiguous:*`
- [ ] Manual: verify `start-flagger-workflow` retries visible under a simulated Temporal outage
- [ ] Manual: verify `suppressed` decisions appear in logs when jailbreak/nsfw match alongside refusal candidates


apps/workers/src/workers/deterministic-flaggers.ts
```diff
@@ -0,0 +1,170 @@
+import {
```
andresgutgon, line 1: In the benchmark I removed the deterministic part on jailbreak and it passed from 520 fails to 320 or something like that. I fear the deterministic flagger are too simplistics and are going to rule out real issues

geclos: if the deterministic part says no match we still run the llm part of the jailbreak flag, now, so that won't happen again. Do not edit the flaggers in your code changes please.

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00009_trace_search_documents.sql
```diff
@@ -42,7 +42,7 @@ ENGINE = ReplacingMergeTree(indexed_at)
 PARTITION BY toYYYYMM(start_time)
 PRIMARY KEY (organization_id, project_id, trace_id)
 ORDER BY (organization_id, project_id, trace_id)
-TTL start_time + INTERVAL 90 DAY DELETE;
+TTL toDateTime(start_time) + INTERVAL 90 DAY DELETE;
```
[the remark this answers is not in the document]
geclos: acceptable

apps/workers/src/workers/deterministic-flaggers.ts
andresgutgon: We'll talk offline. Feel free to merge. I'll rebase

PR #2898, opened 2026-04-27
fix(spans): scope findMessagesForTrace by project + start_time, drop FINAL

geclos:
## Summary

The `trace-search` worker's `findMessagesForTrace` query was OOMing ClickHouse on large tenants — `(total) memory limit exceeded: would use 11.29 GiB ... While executing MergeTreeSelect`.

Three compounding causes in `packages/platform/db-clickhouse/src/repositories/span-repository.ts`:

- **No partition pruning.** The `spans` table is partitioned by `toYYYYMM(start_time)` and ordered by `(organization_id, project_id, start_time, …, trace_id, span_id)`. The query filtered only on `organization_id` + `trace_id`, so ClickHouse had to scan every monthly partition for the org.
- **`FINAL` on `ReplacingMergeTree`.** Forced a merge across all matching parts at query time — the dominant memory cost given the heavy `input_messages` / `output_messages` columns.
- **Heavy columns over an unbounded time range.** `input_messages` / `output_messages` are ZSTD(3) on disk but expand massively in memory.

### Changes

- Add `project_id` and a `[startTime − 1m, startTime + 1d]` window to the `WHERE` so partition pruning + the `(org, project, start_time)` primary key kick in.
- Replace `FINAL` with a `LIMIT 1 BY span_id` over `ORDER BY span_id, ingested_at DESC` subquery to dedupe ReplacingMergeTree rows without merge-on-read.
- Thread `projectId` + `startTime` through the port and both call sites:
  - Worker (`apps/workers/src/workers/trace-search.ts`) — already had both in the payload.
  - Web (`apps/web/src/domains/spans/spans.functions.ts`) — sequenced the trace fetch before the spans fetch (was running both in parallel via `Effect.all`) so `traceDetail.startTime` is available.

The 1-day upper bound assumes a single trace doesn't outlive 24h — flag if you have long-lived sessions/traces that exceed that.

## Test plan

- [x] Type-check passes across `@domain/spans`, `@platform/db-clickhouse`, `@apps/workers`, `@apps/web` (verified locally).
- [x] Spot-check `mapConversationToSpans` in the web UI for a recent trace — conversation → span attribution still resolves correctly.
- [ ] Run the trace-search worker against a known-large trace and confirm no OOM.
- [ ] Confirm `LIMIT 1 BY span_id` dedupe still produces correct messages when the same span has been ingested more than once.

### DIFF
--- a/packages/platform/db-clickhouse/src/repositories/span-repository.ts
+++ b/packages/platform/db-clickhouse/src/repositories/span-repository.ts
@@ -420,15 +420,9 @@ export const SpanRepositoryLive = Layer.effect(
             )
         }),
 
-      findMessagesForTrace: ({ organizationId, projectId, traceId, startTime }) =>
+      findMessagesForTrace: ({ organizationId, projectId, traceId, startTimeFrom, startTimeTo }) =>
         Effect.gen(function* () {
           const chSqlClient = (yield* ChSqlClient) as ChSqlClientShape<ClickHouseClient>
-          // Bound start_time to a [-1m, +1d] window around the root span so
-          // partition pruning + the (org, project, start_time) primary key kick
-          // in. Without this the FINAL/dedupe step has to scan every monthly
-          // partition for the org and OOMs the server on large tenants.
-          const startTimeFrom = new Date(startTime.getTime() - 60 * 1000)
-          const startTimeTo = new Date(startTime.getTime() + 24 * 60 * 60 * 1000)
           return yield* chSqlClient
             .query(async (client) => {
               const result = await client.query({
--- a/packages/platform/db-clickhouse/src/repositories/span-repository.ts
+++ b/packages/platform/db-clickhouse/src/repositories/span-repository.ts
@@ -420,19 +420,44 @@ export const SpanRepositoryLive = Layer.effect(
             )
         }),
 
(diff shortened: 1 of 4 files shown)


packages/platform/db-clickhouse/src/repositories/span-repository.ts
```diff
@@ -420,19 +420,44 @@ export const SpanRepositoryLive = Layer.effect(
...
-      findMessagesForTrace: ({ organizationId, traceId }) =>
+      findMessagesForTrace: ({ organizationId, projectId, traceId, startTime }) =>
         Effect.gen(function* () {
           const chSqlClient = (yield* ChSqlClient) as ChSqlClientShape<ClickHouseClient>
+          // Bound start_time to a [-1m, +1d] window around the root span so
+          // partition pruning + the (org, project, start_time) primary key kick
+          // in. Without this the FINAL/dedupe step has to scan every monthly
+          // partition for the org and OOMs the server on large tenants.
+          const startTimeFrom = new Date(startTime.getTime() - 60 * 1000)
+          const startTimeTo = new Date(startTime.getTime() + 24 * 60 * 60 * 1000)
```
geclos: make this an argument rather than baking it into the method implementation

PR #2907, opened 2026-04-28
Serialize discovery and assignment with Redis locks

geclos:
## Summary
- Fixes two issue-discovery concurrency bugs: duplicate issues from racing no-match workflows (Weaviate projection lag), and stale Weaviate projections from interleaved assign-to-issue post-tx syncs.
- Adds layered Redis NX locks — feedback, project, and per-issue — covering the Postgres centroid recompute *and* the Weaviate projection sync as one critical section. `SET NX EX` is non-blocking; contention returns `lock-unavailable` immediately.
- Workflow retries `lock-unavailable` via durable Temporal `sleep` with exponential backoff (1s→30s cap) plus 1-100s jitter, so contended workers don't pin DB connections or all wake at the same instant.
- Eligibility is re-checked inside each lock; if a competing worker has already claimed the score the activity returns `{ status: "skipped", reason }` instead of failing, so it doesn't burn Temporal retry budget.
- Projection sync moved into `assignScoreToIssueUseCase` (inside the per-issue lock) and `createIssueFromScoreUseCase` (after the tx; no lock — fresh issues can't be raced). The workflow no longer calls `syncIssueProjections` itself.
- New shared workflow helper `runWithLockRetry` consolidates the retry loop; consumed by both `issueDiscoveryWorkflow` and `assignScoreToKnownIssueWorkflow`.
- Naming: `finalizeIssueDiscovery` → `serializeIssueDiscovery` (use-case, activity, file, span, status), and the assignment action `"assigned-existing"` → `"assigned"`.
- Spec doc: new "Bounded locked serialization" section in `dev-docs/issues.md`.

### DIFF
--- a/apps/workflows/src/activities/index.ts
+++ b/apps/workflows/src/activities/index.ts
@@ -21,6 +21,7 @@ export {
   hybridSearchIssues,
   rerankIssueCandidates,
   resolveMatchedIssue,
+  serializeIssueDiscovery,
   syncIssueProjections,
   syncScoreAnalytics,
 } from "./issue-discovery-activities.ts"
--- a/apps/workflows/src/activities/issue-discovery-activities.ts
+++ b/apps/workflows/src/activities/issue-discovery-activities.ts
@@ -9,12 +9,15 @@ import {
   embedScoreFeedbackUseCase,
   type HybridSearchIssuesInput,
   hybridSearchIssuesUseCase,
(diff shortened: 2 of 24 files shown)


geclos:
@claude review

andresgutgon (dismissed):

neoxelox (dismissed):

neoxelox (approved):

PR #2919, opened 2026-04-29
feat(optimizer): refusal flagger + proposer patches, baseline auto-apply, OOM fix

andresgutgon:
## Summary

Optimizer overhaul that lands the **refusal flagger**, makes the proposer ~5× cheaper via find/replace patches, integrates the benchmark runner into the optimization pipeline, and unblocks long runs that were OOMing around iteration ~129.

### Refusal flagger
- New `refusal` strategy + curated dataset (XSTest mapper) with an initial baseline benchmarked under Nova 2.
- Anthropic added as an internal provider (`LAT_ANTHROPIC_API_KEY`) so optimizer scripts can hit Opus 4.7 directly.

### Proposer: find/replace patches
- `flaggerProposerOutputSchema` now emits an ordered list of literal `{ find, replace }` edits instead of regenerating the full strategy file. For typical 22–27K-char files with 1–3 small edits, output tokens drop from ~6,700 → ~500–1,500 and proposer wall time from ~150s → ~30s.
- New `flagger/apply-edits.ts` enforces sequential application, exactly-one-match, CRLF normalization, and all-or-nothing rejection (15 unit-test cases).
- System prompt rewritten with explicit failure-mode warnings and a worked example. Patch failures route to `rejection.stage = "patch-apply"`, distinct from `proposer-error`.

### GEPA iteration tracking + silent-skip diagnosis
- Added `IterationCounter` stopper that always returns `False` but counts every main-loop entry. `totalIterations` and `proposeCalls` flow through JSON-RPC → `@platform/op-gepa` → `@domain/optimizations` → audit JSON + final banner. Surfaces gepa's `skip_perfect_score` behavior — the diff between "loop iters" and "reached propose" makes premature stagnation diagnosable.
- `cost-meter` adds a coarse "silent skips (est)" line under Judge: `silentSkips × reflectionSize × avgEvalCost`. Big enough estimate → bump `--reflection-size`.
- Throw on AI proposer error instead of returning the parent. The old behavior triggered ~5 subsample + ~187 full-val evals on a parent-equivalent candidate every time the model errored; throwing propagates through `reflective_mutation` so the engine `continue`s without any post-propose evals.

### Benchmark integration into the optimizer
- `benchmark-run` becomes a library used by `benchmark-optimize` after a successful run: when an optimization improves baseline, the optimizer auto-applies + reruns the benchmark, but the optimizer's final report is preserved.
- `benchmark-report` extracted as its own script.

### OOM fix on long runs
The candidate loader compiled each proposed strategy with esbuild's `bundle: true` and no `external`, then imported the bundle via a `data:text/javascript;base64,...` URL. Bare specifiers can't resolve through `data:` URLs, so `@domain/*`, `@repo/utils`, `effect`, `zod`, etc. all had to be inlined — 1–5 MB per candidate. Node's ESM loader pins every imported module forever, so ~129 iterations of a 1-hour stagnation-200 run blew past `--max-old-space-size`.

Two fixes drop per-candidate cost ~50×:
- Externalize the same allowlist the static safety scan permits (the strategy package's `dependencies`). esbuild now emits bare imports; relative imports stay bundled.
- Replace the `data:` URL with a temp `.candidate-<hash>.mjs` written inside the strategy directory and imported via `file:`. Node resolves bare specifiers from the strategy's `node_modules`. Unlinked in `finally` immediately after import — Node holds the module by URL, not by file existence.

The leak isn't gone (Node still pins every URL), but headroom now buys orders of magnitude more iterations. The real fix is the planned worker-isolation v2 this branch is named for.

### debug.log
- 1500-row eval batches collapse into a single dot stream per candidate (`.` deterministic, `o` LLM-reached, `!` error, `R` rejected) instead of 1500 timestamped lines.
- Proposer output dump replaced with per-edit summaries (`[1] '<anchor preview>' -Xc/Yl -> +Zc/Wl`) plus a closing `result: N chars (+delta) · hash <8>` line. Audit JSON keeps full text + edits for forensic review.
- Per-candidate byte counts and a unified diff (input → bundled output) flow into the verbose debug log via a new `onBundle` callback on `loadFlaggerCandidate` instead of fighting ink's TUI on stdout.

### Refusal optimizer runs

Baseline updated by:

\`\`\`
pnpm --filter @tools/ai-benchmarks benchmark:optimize \\
  --target flaggers:refusal --verbose \\
  --model claude-opus-4-7 --provider anthropic \\
  --budget-time 2700 --budget-stagnation 50 --reflection-size 10
\`\`\`

## Test plan

- [x] \`pnpm --filter @platform/op-gepa typecheck && pnpm --filter @platform/op-gepa test\`
- [x] \`pnpm --filter @tools/ai-benchmarks test\` (incl. \`apply-edits.test.ts\`)
- [x] Smoke a short refusal optimization end-to-end and confirm the banner reports \`loop iters / reached propose / silently skipped\`.
- [x] Trigger a deliberate proposer schema error and confirm no post-propose evals are charged (audit shows \`proposer-error\`, no eval rows for that candidate).
- [x] Trigger a non-matching \`find\` anchor and confirm \`rejection.stage = "patch-apply"\`.
- [x] Long run (≥1h, stagnation ≥100) no longer OOMs.


andresgutgon:
@claude review please

andresgutgon:
Going through the review point by point:

**Will fix:**

1. **Duplicate debug block** — confirmed, `benchmark-optimize.ts:501-506` is byte-identical to `:494-499`. Will delete the second copy.
2. **Hardcoded "Opus 4.7" label in `printFinalBanner`** — confirmed. Default proposer is `claude-opus-4-6` / `anthropic.claude-opus-4-6-v1`. Will switch the banner to use `proposerModel.model` at runtime.
3. **`loadFixture` duplicated three times** — agreed. Will export `loadFixture` from `runner/benchmark.ts` and have `benchmark-optimize.ts` and `benchmark-report.ts` consume it. Good catch on the latent divergence — the optimizer's copy skips the freshness check, which auto-fetch makes moot but is worth unifying.
4. **`perTactic` bucketing is jailbreaking-specific** — verified against `baselines/flaggers/refusal.json` (`perTactic` has only `other`: 1221/1221). Will move the bucket resolver onto `BenchmarkTarget` (e.g. `tacticTag?: (tags) => string`) so each target gets a meaningful breakdown.

**Probably:**

5. **`void formatPercent` in `OptimizeView.tsx`** — will remove it. The "may surface train F1 in a follow-up" justification is speculative; easier to re-add when actually needed.

**Disagree:**

6. **`ADAPTIVE_THINKING_MODELS` "unreachable dead code"** — this is not dead code. It's the dispatch table consulted in `buildAnthropicProviderOptions` when a user passes `--model claude-opus-4-7`, which is reachable today. The docstring directly above the declaration already explains its purpose ("Add new model ids here as Anthropic ships them"). No change needed.

**Defer:**

7. **`benchmark-report.ts` argv parsing** — real inconsistency, but the script takes one positional arg with no flags. Worth aligning if it grows options; not now.
8. **XSTest CSV runtime safety** — pinned `XSTEST_REPO_REVISION` SHA plus the empty-prompt guard in `buildRow` covers the practical risk. Worth a Zod row parse if we ever depend on a moving HEAD.

csansoon (dismissed):

csansoon (approved):

PR #2931, opened 2026-04-29
chore(db): backfill flaggers for existing projects

neoxelox:
## Summary
- Adds a custom Drizzle migration that provisions one `flaggers` row per registered strategy slug for every existing project.
- Mirrors what `provisionFlaggersUseCase` does on `ProjectCreated`, but for projects that pre-date the flaggers table — those never went through the worker provisioning step.
- Idempotent: `ON CONFLICT DO NOTHING` on the `(organization_id, project_id, slug)` unique index, so re-runs are no-ops and the migration is safe to ship alongside the worker-driven path.

## Details
- Slug list is hardcoded to match `FLAGGER_STRATEGY_SLUGS` in `@domain/flaggers`. Defaults (`enabled = true`, `sampling = 10`) match `FLAGGER_DEFAULT_ENABLED` / `FLAGGER_DEFAULT_SAMPLING`.
- IDs are generated in SQL via `substr(md5(gen_random_uuid()::text), 1, 24)` — 24-char lowercase hex, a valid subset of the CUID2 charset used by the `cuid()` helper.
- Soft-deleted projects are included so a restored project already has its flagger rows.
- Future strategy slugs added to the registry will get provisioned for new projects via `ProjectCreated`; old projects would need a fresh backfill migration.

### DIFF
--- a/packages/platform/db-postgres/drizzle/20260429152120_backfill-flaggers-for-existing-projects/migration.sql
+++ b/packages/platform/db-postgres/drizzle/20260429152120_backfill-flaggers-for-existing-projects/migration.sql
@@ -1,0 +1,27 @@
+-- Custom SQL migration file, put your code below! --
+-- Backfill: provision one flagger row per registered strategy slug for every existing project.
+-- New projects continue to be provisioned by the worker on ProjectCreated via `provisionFlaggersUseCase`.
+-- Idempotent: ON CONFLICT DO NOTHING on the (organization_id, project_id, slug) unique index.
+-- Strategy slugs mirror `FLAGGER_STRATEGY_SLUGS` in @domain/flaggers; defaults match `FLAGGER_DEFAULT_ENABLED` / `FLAGGER_DEFAULT_SAMPLING`.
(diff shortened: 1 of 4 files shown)


packages/platform/db-postgres/drizzle/20260429152120_backfill-flaggers-for-existing-projects/migration.sql
```diff
@@ -0,0 +1,27 @@
...
+FROM "latitude"."projects" p
+CROSS JOIN (VALUES
+  ('frustration'),
+  ('nsfw'),
+  ('refusal'),
+  ('laziness'),
+  ('jailbreaking'),
+  ('forgetting'),
+  ('trashing'),
+  ('tool-call-errors'),
```
andresgutgon, line 23: Are we also persisting with deterministic flaggers?

neoxelox: it works the same as `packages/domain/flaggers/src/use-cases/provision-flaggers.ts`
it provisions the same flaggers (all)

geclos (approved):

andresgutgon (approved):

PR #2956, opened 2026-05-04
feat(datasets): include user trace metadata in dataset rows

andresgutgon:
## Summary
- Carry `Trace.metadata` (the user-defined `Record<string, string>` set at instrumentation time) into the row metadata produced by `addTracesToDataset` / `createDatasetFromTraces`, exposed under a new `traceMetadata` key.
- Unblocks the LAT-544 use case: users with templated input prompts can now use trace-derived golden datasets for regression testing — the prompt variables travel with the row instead of having to be reverse-engineered from rendered messages.
- Backwards compatible: existing `metadata.traceId` extraction (`findExistingTraceIds`) keeps working since the field is added alongside, not nested under, the existing keys.

## Test plan
- [x] `pnpm typecheck` (whole workspace, 61/61 successful)
- [x] `pnpm --filter @domain/datasets test` (6/6 passing)
- [x] New `addTracesToDataset copies user-defined trace metadata into the row metadata` integration test in `@platform/db-clickhouse` — passes against real PG + CH testkits with a fake `TraceRepository` injecting a known `metadata` map; asserts `row.metadata.traceMetadata` round-trips through ClickHouse JSON serialization.
- [x] Two pre-existing failures in `packages/platform/db-clickhouse/src/use-cases/add-traces-to-dataset.test.ts` (`createDatasetFromTraces creates a new dataset and inserts rows`, `createDatasetFromTraces soft-deletes the dataset when ClickHouse insert fails`) are unrelated — reproducible on `main` without this change. Out of scope here.

Closes LAT-544.


packages/domain/datasets/src/use-cases/add-traces-to-dataset.ts
```diff
@@ -31,6 +31,10 @@ function mapTraceToRow(t: TraceDetail) {
       userId: t.userId,
       systemInstructions: t.systemInstructions,
       allMessages: t.allMessages,
+      // User-defined metadata captured at instrumentation time.
+      // Carries prompt template variables so golden datasets can be re-run
+      // through templated prompts without re-deriving inputs from messages.
+      traceMetadata: t.metadata,
```
andresgutgon, line 37: This is the real change. The reset is fixing broken CI. Test was not really running due to miss configured vitest config

apps/workflows/src/activities/evaluation-alignment-activities.test.ts
```diff
@@ -1,49 +1,81 @@
...
-    generate: vi.fn(),
-    embed: vi.fn(),
-    rerank: vi.fn(),
-  },
-  mockOptimizer: {
-    optimize: vi.fn(),
-  },
+import { describe, expect, it, vi } from "vitest"
+
+// FIXME(workflow-runtime): the activity exports in this file are individual
```
andresgutgon, line 46: I dont want to do this refactor here but activities are not using effect properly

csansoon (approved):

PR #2960, opened 2026-05-04
Fix PH sessions

learningbizz:
- **Problem**: PostHog session replays appeared fragmented (short segments, buffering). PostHog support flagged erroneous posthog.reset() calls as a common cause.
- **Root cause**: reset() was called whenever the authenticated layout's PostHogIdentity mounted (before identify()). Any remount starts a new PostHog session.
- **Fix — logout-only reset**: Removed the reset() from the mount path. On sign-in / app shell load, only identify() and the existing org group logic run — no reset.
- **Fix — identity-swap guard**: Added a sessionStorage-backed guard in identifyUser that detects when the user id changes (e.g. admin impersonation swaps the session cookie without logout). When a different user id is identified, posthog.reset() is called automatically before re-identifying, preventing PostHog from aliasing/merging distinct users.
- **Why sessionStorage**: Impersonation triggers a full page reload (window.location.href = "/"), which destroys module-level state. sessionStorage survives reloads within the same tab but clears on tab close — the right lifetime for this guard.
- **No-op for same user**: Normal page reloads, React Strict Mode double-mounts, and SPA navigations all see the same user id and skip the reset — recordings stay continuous.

### DIFF
--- a/apps/web/src/lib/posthog/posthog-provider.tsx
+++ b/apps/web/src/lib/posthog/posthog-provider.tsx
@@ -1,6 +1,6 @@
 import { useMountEffect } from "@repo/ui"
 import { useEffect } from "react"
-import { identifyOrganization, identifyUser, initPostHog, resetPostHog } from "./posthog-client.ts"
+import { identifyOrganization, identifyUser, initPostHog } from "./posthog-client.ts"
 
 export function PostHogProvider() {
   useMountEffect(() => {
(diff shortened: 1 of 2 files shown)


apps/web/src/lib/posthog/posthog-provider.tsx
```diff
@@ -25,14 +25,11 @@ export function PostHogIdentity({
   organizationName,
 }: PostHogIdentityProps) {
   useMountEffect(() => {
-    void (async () => {
-      await resetPostHog()
```
andresgutgon, line 29: Maybe do this also when changing organizations.

learningbizz: No need to reset when changing orgs, we already to a full page reload, which remounts PH identity. So each org will have its own PH Session

csansoon (approved):

PR #2973, opened 2026-05-05
Several fixes to the existing API/SDK

neoxelox:
# Summary

Tightens the public API surface for annotations and scores ahead of `@latitude-data/sdk@6.0.0-alpha.1`. Both endpoints now share a single trace-resolution model (`TraceRef` discriminated union — by id or by filter set), drop fields that are auto-resolved server-side, and remove power-user knobs that the public API doesn't need to expose yet (drafts, updates, uninstrumented scores).

## Motivation

The API was carrying historical fields that aren't useful for external callers and made the SDK confusing:

- Annotations accepted both flat `messageIndex`/`partIndex`/`startOffset`/`endOffset` AND a nested `anchor` object — duplicate.
- Annotations exposed `sessionId` / `spanId` overrides that are trivially derivable from the trace.
- Annotations exposed an `id`-based upsert path and a `draft: true` flag that are managed-UI concerns, not public API concerns.
- Scores took flat `traceId`/`sessionId`/`spanId` and didn't support filter-based trace resolution like annotations did.
- Scores allowed an "uninstrumented" path (no trace) that's only meaningful once the simulations SDK lands.

### Public API surface

**Annotations** (`POST /v1/projects/{projectSlug}/annotations`):
- Required: `value`, `passed`, `feedback`, `trace` (`TraceRef` — `{ by: "id", id }` or `{ by: "filters", filters }`).
- Optional: `anchor`, `simulationId`, `issueId`, `annotatorId`.
- **Removed**: flat `messageIndex`/`partIndex`/`startOffset`/`endOffset`; `id`; `draft`; `sessionId`; `spanId`; `traceId` (replaced by `trace`).
- Always published (`draftedAt = null`); `sessionId` lifted from the trace, `spanId` pinned to the trace's last LLM completion.

**Scores** (`POST /v1/projects/{projectSlug}/scores`):
- Same `trace` field as annotations (now required — no uninstrumented scores).
- **Removed**: flat `traceId`/`sessionId`/`spanId`.
- `_evaluation` discriminator preserved (still marked internal in OpenAPI).

### Domain

- **New**: `@domain/spans` exports shared `traceRefSchema`, `TraceRef`, `resolveTraceIdFromRef`, and `resolveScoreTraceContext`. Annotations and scores use cases share these.
- **New**: `submitApiScoreUseCase` in `@domain/scores`, plus `baseSubmitApiScoreSchema` exported so the route reuses the same field list instead of re-deriving from `baseWriteScoreInputSchema`.
- `submitApiAnnotationInputSchema` now `omit`s `id`/`projectId`/`sourceId`/`sessionId`/`traceId`/`spanId` — the public-API schema is honest about what it accepts. Doc block consolidated. Schema and use case live in `submit-api-annotation.ts` (deleted the helper file `annotation-public-api-schema.ts` since there was nothing left to host).
- `@domain/scores` now depends on `@domain/spans` (acyclic — spans doesn't depend on scores).
- `writeDraftAnnotationUseCase` is **kept** for internal callers (managed UI's editing flow); only the public API path was simplified.

### API route layer

- Single shared `TraceRefSchema` (with named `FilterCondition` / `FilterSet` components) lives in `apps/api/src/openapi/schemas.ts`. Annotations and scores both import it. The chain has to be rebuilt with named instances to work around a Fern TS-generator bug for anonymous types nested under arrays inside discriminated unions (verified the bug is unfixed in 3.69.0); the long explanation lives in the `schemas.ts` comment.
- `apps/api/src/routes/scores.ts`: ternary in the route handler deduplicated — only `source`/`sourceId`/`metadata` live inside the discriminator branch; the 9 shared fields are listed once outside.
- Route descriptions for annotations and scores aligned in shape; scores keeps its `Annotations use the separate /annotations endpoint.` pointer.

### TypeScript SDK

- **New**: `packages/sdk/typescript/README.md` — install, auth, configuration, examples (trace by id / filters, custom scores, API keys, passthrough), error handling, regeneration pointer.
- **New**: `packages/sdk/typescript/CHANGELOG.md` — Keep-a-Changelog format; alpha.1 entry consolidates all four breaking changes with diff snippets.
- **CI updated**: `.github/workflows/publish-typescript-sdk.yml` now extracts release notes from CHANGELOG via the same awk pipeline as `publish-typescript-telemetry.yml`. The two workflows are structurally identical except for package-specific names.
- **Version**: `6.0.0-alpha.1` (consolidated all in-progress changes since alpha.0 was published 2026-04-22).
- SDK regenerated; `CreateAnnotationBody`, `CreateCustomScoreBody`, `CreateEvaluationScoreBody`, and the new shared `TraceRef` type all reflect the new shape.

### Dogfooding

- `packages/platform/latitude-api/src/client.ts` (the workers' product-feedback wrapper) — dropped the now-removed `draft: false` from the `sdk.annotations.create` call. Tests adjusted accordingly. 9/9 wrapper tests still pass with no other changes needed.
- `examples/sdks/typescript`: deleted `create-annotation-draft.ts` and the `annotate:draft` script; example README cleaned up.


packages/sdk/typescript/src/api/resources/annotations/client/Client.ts
```diff
@@ -35,17 +35,12 @@ export class AnnotationsClient {
...
-     *         feedback: "Approved - the system annotator correctly flagged this as a refusal.",
+     *         feedback: "feedback",
      *         trace: {
-     *             by: "filters",
-     *             filters: {
-     *                 "metadata.scoreId": [{
-     *                         op: "eq",
-     *                         value: "abc123def456ghi789jkl012"
-     *                     }]
-     *             }
```
andresgutgon, line 48: This example was more helpful

andresgutgon (approved):

packages/sdk/typescript/src/api/resources/annotations/client/Client.ts
neoxelox: was contradictory in my opinion, also there are no examples elsewhere

PR #2986, opened 2026-05-06
Stop recording PH events of latitude team

learningbizz:
The first hypothesis didnt work, this should fix it.

Root cause: PostHog persists opt-in state in a cookie. If a staff user was ever opted in (before the exclusion logic existed), PostHog restores that state on next page load and starts capturing immediately — before syncPostHogSession can call opt_out_capturing().
Fix 1 — loaded callback: Added loaded: (ph) => { ph.opt_out_capturing() } to the PostHog init config. This forces opt-out on every single page load, regardless of what the cookie says. No events can fire until syncPostHogSession explicitly opts in.
Fix 2 — Session recording disabled at init: Changed disable_session_recording: true so no recording starts until startSessionRecording() is explicitly called for non-staff users.
Fix 3 — Explicit recording lifecycle: setPostHogCaptureEnabled(true) now calls startSessionRecording(); setPostHogCaptureEnabled(false) calls stopSessionRecording().

### DIFF
--- a/apps/web/src/lib/posthog/posthog-client.ts
+++ b/apps/web/src/lib/posthog/posthog-client.ts
@@ -61,13 +61,15 @@ const loadInstance = (): Promise<PostHog | null> => {
       const posthog = mod.posthog
       posthog.init(env.apiKey, {
         api_host: env.host,
-        // Per product decision: session recordings + autocapture + pageview.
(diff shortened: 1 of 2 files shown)


andresgutgon:
Session recording is a client thing no? This is backend client

PR #2992, opened 2026-05-07
Wrap markdown tables so wide content scrolls inside the message

geclos:
Wide markdown tables in the trace detail drawer's conversation tab were pushing past the drawer width. Wrap the table in an overflow-x-auto container so the message itself stays within bounds and wide tables get their own horizontal scroll instead.

### DIFF
--- a/packages/ui/src/components/genai-conversation/parts/markdown-content.tsx
+++ b/packages/ui/src/components/genai-conversation/parts/markdown-content.tsx
@@ -59,6 +59,13 @@ const markdownComponents: Components = {
       </CodeBlockShell>
     )
   },
+  // Wide tables would otherwise push their parent past the drawer width;
+  // wrap them so overflow scrolls horizontally inside the message.
+  table: ({ children }) => (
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/genai-conversation/parts/markdown-content.tsx
```diff
@@ -59,6 +59,13 @@ const markdownComponents: Components = {
       </CodeBlockShell>
     )
   },
+  // Wide tables would otherwise push their parent past the drawer width;
+  // wrap them so overflow scrolls horizontally inside the message.
+  table: ({ children }) => (
+    <div className="max-w-full overflow-x-auto">
+      <table>{children}</table>
```
andresgutgon, line 66: Will this produce a double `<table>` element? Is not markdown generating the table tag?

geclos: don't know tbh cc @claude answer this

andresgutgon (approved):

PR #2997, opened 2026-05-07
fix(auth): drop _gl from tracking params to prevent magic-link INVALID_CALLBACK_URL

andresgutgon:
## Summary

- A user reported a production magic-link click landing on `https://console.latitude.so/api/auth/magic-link/verify?...` with `{"message":"Invalid callbackURL","code":"INVALID_CALLBACK_URL"}`.
- Root cause: when the user landed on `/login` from a Google-Ads-tagged URL, the `_gl` linker param (`1*rre81b*_gcl_au*MTgz...`) was forwarded into `newUserCallbackURL` (`apps/web/src/routes/login.tsx:58`). The `*` delimiters fail Better Auth's relative-path regex (`[\w\-.+/=&%@]` in the query, no `*`), so `originCheck` middleware on the verify endpoint rejects the link.
- The link generation request (server-side `auth.api.signInMagicLink({ body, headers })` from `apps/web/src/domains/auth/auth.functions.ts`) has no HTTP request object, so Better Auth's global `originCheckMiddleware` early-returns and the bad URL gets baked into the email link silently — only the verify GET catches it.
- Fix: drop `_gl` from `TRACKING_PARAM_KEYS` in `apps/web/src/lib/analytics/gtm.ts`. `_gl` only carries cross-domain GA session identity, which the same-origin auth flow does not need; `gclid` already covers Google Ads attribution.

## Test plan

- [x] `pnpm --filter @app/web test src/lib/analytics/gtm.test.ts` (9 passing) — covers `pickTrackingParams` exclusion, the social-login start flow, and a regression guard with the literal failing production `newUserCallbackURL`.
- [x] `pnpm --filter @app/web typecheck` clean.
- [x] `pnpm --filter @app/web check` (Biome) clean.
- [ ] Manual: visit `/login?_gl=1*foo*bar*baz` in production-like env, request a magic link, click it, verify redirect to `/welcome?signup=email` instead of `INVALID_CALLBACK_URL`.

### DIFF
--- a/apps/web/src/lib/analytics/gtm.test.ts
+++ b/apps/web/src/lib/analytics/gtm.test.ts
@@ -62,18 +62,6 @@ describe("appendTrackingParams", () => {
     expect(appendTrackingParams("/welcome?foo=1", { signup: "email" })).toBe("/welcome?foo=1&signup=email")
   })
 
-  it("inserts the query before any '#fragment' so user-supplied redirects with a hash stay valid", () => {
-    expect(appendTrackingParams("/welcome#pricing", { signup: "email" })).toBe("/welcome?signup=email#pricing")
-    expect(appendTrackingParams("/welcome?foo=1#pricing", { signup: "email" })).toBe(
-      "/welcome?foo=1&signup=email#pricing",
-    )
-    expect(appendTrackingParams("/welcome#", { signup: "email" })).toBe("/welcome?signup=email#")
-  })
-
-  it("returns the path including its fragment unchanged when no params are provided", () => {
-    expect(appendTrackingParams("/welcome#pricing", {})).toBe("/welcome#pricing")
-  })
-
   it("leaves regex-safe values untouched (no spurious encoding for typical tracking IDs)", () => {
     const url = appendTrackingParams("/welcome", {
       gclid: "Cj0KCQiAxxxx",
--- a/apps/web/src/lib/analytics/gtm.test.ts
+++ b/apps/web/src/lib/analytics/gtm.test.ts
@@ -23,11 +23,13 @@ const simulateBetterAuthValidatedValue = (callbackValue: string): string => {
 }
 
 describe("pickTrackingParams", () => {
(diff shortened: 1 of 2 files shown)


andresgutgon:
Added url-safe encoding as defense-in-depth (commit 7072ddeb4) per the suggestion.

`sanitizeForCallback` now pre-encodes any char outside `[\w\-.+/=&%@]` (the regex's allow set) as `%XX` before it reaches `URLSearchParams.toString()`. The non-obvious bit: Better Auth double-decodes the callback value (URL parse + explicit `decodeURIComponent`) before validating, so naively encoding `*` → `%2A` round-trips back to `*` and still fails the regex. Pre-encoding to `%XX` before `URLSearchParams.toString()` adds one `%`→`%25` layer, Better Auth's own `searchParams.set` adds another, and the two decodes reverse them — leaving `%XX` visible to the regex.

Tests now include a `simulateBetterAuthValidatedValue` helper mirroring the verify endpoint's processing, plus an end-to-end preservation check (`gclid=1*foo*bar` and `utm_source=spring(2026)` survive intact to `/welcome`).

The `_gl` exclusion stays — it's still useless for same-origin redirects, and one less special case to worry about — but the encoding now also makes any future tracking key or unusual `utm_*` value safe.

andresgutgon:
Addressed Copilot review (commit 73bdd3e7a):

- **Hash fragments**: `appendTrackingParams` now splits on the first `#` and inserts the query between path and fragment. A user-supplied `?redirect=/foo#bar` no longer produces `/foo#bar?gclid=…` (where the browser would treat the query as part of the fragment).
- **Allocations**: hoisted `TextEncoder` to a module-level singleton; replaced per-char string concatenation with `array.join`. Minor — this runs once per auth redirect URL — but trivially better.

Two new tests cover hash cases (with/without existing query, empty fragment, no-op when params is empty). All 14 tests pass; typecheck + Biome clean.

neoxelox (approved):

PR #3018, opened 2026-05-08
fix(trace-search): explicit phrase / semantic modes (LAT-562)

geclos:
## What changes

Search now has two explicit modes instead of one mixed score:

- **Quoted text** (`"handOffToHuman"`) is an exact phrase filter. Multiple quoted segments must all be present.
- **Unquoted text** (`pricing complaint`) is a semantic search ranked by embedding similarity.
- **Mix the two** (`"handOffToHuman" customer needle`) and the phrases filter, the rest ranks.

Storage swaps the bloom-filter indexes for a ClickHouse text index on `search_text` with the `splitByNonAlpha` tokenizer and a `lower()` preprocessor. That makes whitespace and case differences between the typed query and stored bytes irrelevant at the token layer. The legacy bloom-filter indexes stay around for a stability window before being dropped in a follow-up migration.

## Why

Two real bugs that LAT-562 surfaced:

- The trace UI prettifies JSON, so users were copy-pasting `"key": true` (with a space). Indexed bytes are compact (`"key":true`), and `ILIKE %key": true%` returned zero hits even though the field was right there.
- Filter-style queries like `"handOffToHuman": true` were being ranked by the embedder, which can't reliably tell `:true` apart from `:false`. Results came back mixed.

Letting the user state intent (quotes = filter, unquoted = meaning) removes the heuristic guesswork.

## Notes

- Production runs ClickHouse 26.2; `hasPhrase` lands in 26.4. Implementation uses `hasTokenCaseInsensitive` AND-ed across the phrase's tokens. Exact for single-token phrases like the LAT-562 case; multi-token phrases match as a token bag (no order enforcement). Swap to `hasPhrase` once prod upgrades — comment in `trace-repository.ts` flags the spot.
- Migration `00012_text_index_search_documents.sql` adds the index and runs `MATERIALIZE INDEX`. Existing rows become searchable as the materialize job completes.
- `HYBRID_SEARCH_CONFIG` (lexical/semantic weights) is gone. `TRACE_SEARCH_MIN_RELEVANCE_SCORE` stays as the semantic-only floor.

## Test plan

- [x] `pnpm --filter @domain/spans test` (parser tests, 10 cases)
- [x] `pnpm --filter @platform/db-clickhouse test trace-repository` (4 query-shape tests + LAT-562 whitespace regression + list/count/metrics/histogram consistency)
- [x] Migration applied locally; backfill embedded ~2900 seeded traces
- [x] Manual: paste `"handOffToHuman": true` from a trace into the search bar and confirm matches
- [x] Manual: unquoted query still ranks by similarity
- [ ] Manual: drop the legacy bloom-filter indexes after one week of production traffic on the new path (separate PR)


specs/trace-search.md
```diff
@@ -0,0 +1,173 @@
...
+- [ ] **P4-2**: Author migration `00012_drop_legacy_search_indexes.sql` dropping `idx_search_text_tokenbf` and `idx_search_text_ngrambf`.
+- [ ] **P4-3**: Update the leading comment block in `00009_trace_search_documents.sql` (or supersede it via a comment in the new migration) to reflect the text-index path as the lexical backend.
+
+**Exit gate**: legacy bloom-filter indexes are gone from production; storage footprint of `trace_search_documents` reflects the swap.
+
+## Open Items
+
+- Whether to gate the cutover behind a per-org feature flag or hard-cut after staging validation. Default: hard-cut.
+- Whether to expose a `-"phrase"` exclusion operator in v1. Defer unless explicitly requested.
+- Whether snake_case identifier searchability (e.g., querying `search_properties` as one token) is enough of a regression to justify a custom `splitByString` separator list. Decided after observing real query patterns post-launch; default is to accept the loss.
```
geclos: @claude remove this spec

packages/platform/db-clickhouse/src/repositories/trace-repository.ts
```diff
@@ -68,24 +62,80 @@ const HYBRID_SEARCH_CONFIG = {
...
+ * exact for the motivating problem. The proper "contiguous tokens, in order"
+ * primitive is `hasPhrase`, which lands in CH 26.4 (PR
+ * ClickHouse/ClickHouse#101997); prod runs 26.2 today. When prod upgrades,
+ * swap the per-token loop for one `hasPhrase(search_text, phrase)` call per
+ * phrase to recover phrase-order semantics.
+ *
+ * If a phrase tokenizes to nothing (the user typed only punctuation between
+ * the quotes), the entire query intentionally matches no rows: the user
+ * filtered on a phrase the index can't represent — surfacing zero results is
+ * more honest than silently dropping the filter.
```
geclos: simplify this comment, remove references to linear issues

andresgutgon (commented):
Really nice

PR #3023, opened 2026-05-08
fix(telemetry): unblock npm publish action and default to production ingest

andresgutgon:
## Summary

Two fixes for the TypeScript telemetry release path, both surfaced while debugging why every `publish-typescript-telemetry` job was failing on the runner.

### 1. `publish-npm-package` composite action no longer parses as invalid

The `npm_token` input description contained the literal string `${{ secrets.NPM_TOKEN }}`. GitHub's composite-action loader tried to evaluate that as an expression, but `secrets` is not a valid context inside a composite action manifest, so every job failed before any step ran with:

```
Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.NPM_TOKEN
```

Reworded the description to plain text. The token is still passed via `inputs.npm_token` from the caller workflow, unchanged.

### 2. `@latitude-data/telemetry` defaults to production ingest

Old behavior in `packages/telemetry/typescript/src/env/env.ts`:

| `NODE_ENV` | URL used |
| --- | --- |
| `production` | `https://ingest.latitude.so` |
| `development` | `http://localhost:3002` |
| `test` | `http://localhost:3002` |
| unset / anything else (e.g. `staging`) | `http://localhost:3002` |

So a consumer who installed the published SDK without setting `NODE_ENV=production` (or set it to `staging`) silently shipped traces to localhost and never reached our ingest. Note: `tsdown` does **not** inline `process.env.NODE_ENV` at build time (verified in `dist/index.js` and `dist/index.cjs` — the literal lookup is preserved), so this is read at the consumer's runtime, not on our CI runner. Setting `NODE_ENV` in the publish action would not have helped.

New behavior: default to `https://ingest.latitude.so`. Localhost is only used when `LATITUDE_TELEMETRY_URL` is unset **and** `NODE_ENV` is explicitly `development` or `test`. `LATITUDE_TELEMETRY_URL` continues to take precedence over everything.

## Test plan

- [ ] Re-run the `publish-typescript-telemetry` workflow on this branch and confirm the action manifest validates.
- [ ] Smoke-test the published SDK in a consumer app with `NODE_ENV` unset and confirm traces reach `ingest.latitude.so`.
- [ ] Confirm `NODE_ENV=development` and `NODE_ENV=test` still hit `http://localhost:3002`.
- [ ] Confirm `LATITUDE_TELEMETRY_URL` overrides both paths.

### DIFF
--- a/packages/telemetry/typescript/src/env/env.ts
+++ b/packages/telemetry/typescript/src/env/env.ts
@@ -6,10 +6,7 @@ function getExporterUrl() {
     return process.env.LATITUDE_TELEMETRY_URL
   }
 
-  if (
-    process.env.NODE_ENV === "development" ||
-    process.env.NODE_ENV === "test"
-  ) {
+  if (process.env.NODE_ENV === "development" || process.env.NODE_ENV === "test") {
     return LOCAL_EXPORTER_URL
   }
 
--- a/packages/telemetry/typescript/src/env/env.ts
+++ b/packages/telemetry/typescript/src/env/env.ts
@@ -1,16 +1,19 @@
-const DEFAULT_EXPORTER_URL =
-  {
-    production: "https://ingest.latitude.so",
-    development: "http://localhost:3002",
-    test: "http://localhost:3002",
-  }[process.env.NODE_ENV ?? "development"] ?? "http://localhost:3002"
+const LOCAL_EXPORTER_URL = "http://localhost:3002"
+const PRODUCTION_EXPORTER_URL = "https://ingest.latitude.so"
 
 function getExporterUrl() {
   if (process.env.LATITUDE_TELEMETRY_URL) {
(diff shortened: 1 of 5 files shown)


packages/telemetry/typescript/src/env/env.ts
```diff
@@ -1,16 +1,16 @@
...
 
-function getExporterUrl() {
+export function getExporterUrl() {
   if (process.env.LATITUDE_TELEMETRY_URL) {
     return process.env.LATITUDE_TELEMETRY_URL
   }
 
-  return DEFAULT_EXPORTER_URL
+  if (process.env.NODE_ENV === "development" || process.env.NODE_ENV === "test") {
+    return LOCAL_EXPORTER_URL
```
geclos: aqui cc @andresgutgon

geclos (approved):

PR #3026, opened 2026-05-08
fix(marketing): resolve Loops contact update 400 and 409 swallowing

andresgutgon:
## Summary

- Pass `email` alongside `userId` on Loops `updateContact` from `updateContactOnboarding` and `markContactTelemetryEnabled` so contacts whose Loops record isn't yet linked to the v2 `userId` (e.g. v1-era records) can be resolved by email instead of failing with HTTP 400.
- Detect duplicate-on-create by `APIError.statusCode === 409` rather than regexing the response message — Loops reworded the 409 body and `registerContact` started failing for users already on the list.

## Why

Datadog Error Tracking issue [`f7687552-…`](https://app.datadoghq.eu/error-tracking?query=service%3A%2A&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22f7687552-4ac2-11f1-ae72-da7ad0900005%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D) was firing `MarketingContactsError: updateContact failed for …`. Spans showed `PUT https://app.loops.so/api/v1/contacts/update → 400` (Loops can't find the contact when only `userId` is provided and that id was never linked) plus several `POST /contacts/create → 409` errors that should have been swallowed but weren't because the message regex no longer matched.

## Test plan

- [x] `pnpm --filter @domain/marketing --filter @platform/loops test --run`
- [x] `pnpm --filter @domain/marketing --filter @platform/loops typecheck`
- [ ] After deploy, monitor the error-tracking issue and Loops `register-user` / `update-onboarding` queue tasks for recurrence

### DIFF
--- a/packages/domain/marketing/src/use-cases/mark-contact-telemetry-enabled.test.ts
+++ b/packages/domain/marketing/src/use-cases/mark-contact-telemetry-enabled.test.ts
@@ -76,6 +76,9 @@ describe("markContactTelemetryEnabled", () => {
 
     expect(sender.updates).toHaveLength(3)
     expect(new Set(sender.updates.map((u) => u.userId))).toEqual(new Set([USER_OLDEST, USER_OWNER, USER_NEWEST]))
+    expect(new Set(sender.updates.map((u) => u.email))).toEqual(
+      new Set(["oldest@example.com", "owner@example.com", "newest@example.com"]),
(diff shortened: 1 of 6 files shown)


csansoon (approved):
LGTM

PR #3033, opened 2026-05-08
feat(web): OAuth consent page + root-level AS-metadata redirect

neoxelox:
## Summary

Closes the OAuth/MCP flow end-to-end. After this PR, MCP clients can register, authorize, consent (with org binding), and exchange the auth code for an access token that the API actually accepts.

### `/auth/consent` — TanStack route + server fns

The BA `mcp` plugin's authorize endpoint redirects signed-in users to `/auth/consent?consent_code=…&client_id=…&scope=…`. The page lists the user's organizations, asks them to pick one, and:

- **On Approve** — `decideOAuthConsent({ accept: true, … })` validates membership in the chosen org via `MembershipRepository.isMember`, writes `oauth_applications.organization_id` (admin Postgres connection — the row is RLS-scoped on the very `organization_id` we're about to set), then calls BA's `oAuthConsent({ accept: true })` which returns the redirect URL back to the MCP client's `redirect_uri` with `?code=…`.
- **On Deny** — `decideOAuthConsent({ accept: false, … })` skips the binding step and just calls BA with `accept: false`. BA returns the same redirect URL with `?error=access_denied`.

Either way the server fn returns `{ redirectUrl }`; the page navigates via `window.location.href` (full navigation, not router.navigate — the target is the MCP client's origin, not ours).

**Auth gate** — the loader requires a session. If missing, it bounces through `/login` with the consent URL preserved in `redirect` so the flow resumes after sign-in. BA's `mcp/authorize` already sets the `oidc_consent_prompt` signed cookie, which survives the login round-trip and is what `oAuthConsent` reads server-side.

### `/.well-known/oauth-authorization-server` — root-level redirect

BA serves the RFC 8414 metadata document at `/api/auth/.well-known/oauth-authorization-server`. RFC 8414 §3 lets clients derive the metadata path from the issuer, so most MCP clients honor the path-prefixed form. Some don't — they assume hostname-rooted `/.well-known/...`. The redirect catches root-level probes and 307s them to the BA-served document; HEAD mirrors GET for clients that pre-flight.

The route file lives under `apps/web/src/routes/[.well-known]/` — TanStack file-based routing treats `.` as a segment separator, so the brackets escape the literal dot. Confirmed in the regenerated `routeTree.gen.ts`: `id: '/.well-known/oauth-authorization-server'`.

### Bonus fix: drop `$defaultFn` on FK columns BA leaves undefined

While smoke-testing `POST /api/auth/mcp/register` (the first call any MCP client makes), I hit a 500:

```
insert or update on table "oauth_applications" violates foreign key
constraint "oauth_applications_user_id_users_id_fkey"
Key (user_id)=(lyh100fkhbf9uujurq370yx5) is not present in table "users".
```

Our `cuid()` schema helper attaches a `$defaultFn(() => createId())` so unset id columns get a synthesized CUID2. Correct for primary keys; wrong for nullable FK columns where "the caller didn't set it" should mean NULL, not "synthesize an id that can't possibly satisfy the FK". `oauth_applications.user_id` and `organization_id` are filled by BA from the session and from the consent flow respectively — both are undefined for unauthenticated dynamic-registration calls. Same shape for `oauth_access_tokens.user_id` and `oauth_consents.user_id`.

The fix adds an optional `{ default: false }` knob to `cuid()` and routes the four BA-managed nullable FK columns through it. Same physical column shape (`varchar(24)`), so **no migration needed** (`pnpm pg:generate` reports "No schema changes"). Implemented as overloads, not a single-impl ternary, because a ternary collapses the return into a union and erases Drizzle's `hasDefault` flag.

Folded in here from #3034 (closed) — the consent page is the first surface that exercises the register endpoint, so the FK fix lives where it gets tested.

## Plan reference

`plans/mcp-oauth-api-expansion.md` M2 web changes.

With this PR M2 is **fully shipped**:
- BA `mcp()` plugin install (#3032)
- API auth fall-through (API-key → OAuth → 401, #3032)
- API `/.well-known/oauth-protected-resource` (#3024)
- **Consent page + AS-root redirect + FK-default fix** (this PR)

## Test plan

- [x] `pnpm --filter @app/web test` — 260 passing
- [x] `pnpm --filter @platform/db-postgres test` — 168 passing
- [x] `pnpm --filter @app/web check` (biome)
- [x] `pnpm --filter @platform/db-postgres check` (biome)
- [x] `pnpm knip`
- [x] `pnpm pg:generate` — "No schema changes, nothing to migrate"
- [ ] **Manual smoke** (after merge in dev): full OAuth flow per the recipe — `curl POST /api/auth/mcp/register` returns `{ client_id, ... }`, browser-driven authorize → consent page → org pick → approve → MCP client receives access token → `curl /v1/api-keys -H "Authorization: Bearer <token>"` returns 200.

## Out of scope

- `/v1/mcp` mount handler in `apps/api` (next PR — last M2 item).
- `apps/web/src/routes/_authenticated/settings/keys.tsx` rename + OAuth Keys section (M-Settings, parallel track).


andresgutgon (dismissed):

apps/web/src/routes/api/auth/$.ts
```diff
@@ -1,11 +1,14 @@
...
+import { forceOAuthConsent } from "../../../server/force-oauth-consent.ts"
 
 export const Route = createFileRoute("/api/auth/$")({
   server: {
     handlers: {
       GET: async ({ request }: { request: Request }) => {
-        return getBetterAuth().handler(request)
+        // No-op for everything except GET /api/auth/mcp/authorize — see
+        // `force-oauth-consent.ts` for why we always force `prompt=consent`.
+        return getBetterAuth().handler(forceOAuthConsent(request))
```
andresgutgon: Why you dont do this file `/api/auth/mcp/authorize` instead of a catch all?

apps/web/src/domains/oauth/oauth-consent.functions.ts
```diff
@@ -0,0 +1,201 @@
...
+    // endpoints from `auth.api` — the plugin is at runtime, but TS can't see
+    // it. Cast through a narrow shape rather than `any` so the body / return
+    // contract stays explicit.
+    //
+    // Return shape comes from `better-auth@1.6.9/dist/plugins/oidc-provider/
+    // index.mjs` — both the accept branch (line 358) and the deny branch
+    // (line 331) call `ctx.json({ redirectURI })`, so the field is
+    // `redirectURI`, not the more familiar `{ redirect, url }` pair BA uses
+    // elsewhere for browser-fetch responses.
+    interface OAuthConsentApi {
```
andresgutgon: Uff inline typescript interface. Move to module level please

apps/web/src/routes/api/auth/$.ts
neoxelox: less files, its just an if check dont worry

apps/web/src/domains/oauth/oauth-consent.functions.ts
neoxelox: ok

andresgutgon (approved):

PR #3034, opened 2026-05-08
fix(db-postgres): drop CUID auto-default on FK columns BA leaves undefined

neoxelox:
## What broke

Hitting `POST /api/auth/mcp/register` with `allowDynamicClientRegistration: true` (no session, as MCP clients do) returned a 500. Server log:

```
insert or update on table "oauth_applications" violates foreign key
constraint "oauth_applications_user_id_users_id_fkey"
Key (user_id)=(lyh100fkhbf9uujurq370yx5) is not present in table "users".
```

That `lyh100fkhbf9uujurq370yx5` is a freshly-generated CUID2 that doesn't match any row.

## Why

Our `cuid()` schema helper attaches a `$defaultFn(() => createId())` so unset id columns get a synthesized CUID2. That's correct for **primary keys** but wrong for **nullable FK columns** where "the caller didn't set it" should mean NULL, not "synthesize an id that can't possibly satisfy the FK".

`oauth_applications.user_id` and `organization_id` are filled by BA from the session (which is undefined for unauthenticated dynamic-registration calls) and from our consent flow (which only writes after the org-pick step). When BA passes `userId: undefined`, the `$defaultFn` fires, Drizzle inserts a fresh CUID2, Postgres rejects.

Same shape for `oauth_access_tokens.user_id` and `oauth_consents.user_id` — BA does set them in the standard flow, but the schema shouldn't silently substitute a phantom id if it ever passes undefined.

## Fix

Adds a `cuidRef(name)` helper that's `cuid()` minus the `$defaultFn`, and routes the four BA-managed nullable FK columns through it. Same physical column shape (`varchar(24)`) at the DB level, so no migration is needed (`pnpm pg:generate` reports "no schema changes").

```ts
// schemaHelpers.ts
export function cuidRef(name: string) {
  return varchar(name, { length: 24 })
}

// schema/better-auth.ts
oauthApplications:  userId: cuidRef("user_id").references(...)
                    organizationId: cuidRef("organization_id").references(...)
oauthAccessTokens:  userId: cuidRef("user_id").references(...)
oauthConsents:      userId: cuidRef("user_id").references(...)
```

## Test plan

- [x] `pnpm --filter @platform/db-postgres test` — 168 passing
- [x] `pnpm --filter @platform/db-postgres check` (biome)
- [x] `pnpm pg:generate` — "No schema changes, nothing to migrate"
- [ ] **Manual repro after merge**: `curl -X POST http://localhost:3000/api/auth/mcp/register -H "Content-Type: application/json" -d '{"client_name":"x","redirect_uris":["http://localhost:9999/callback"],"grant_types":["authorization_code"],"response_types":["code"],"token_endpoint_auth_method":"client_secret_basic"}'` should return `{ client_id, client_secret, ... }` instead of 500.

### DIFF
--- a/apps/web/src/domains/oauth/oauth-consent.functions.ts
+++ b/apps/web/src/domains/oauth/oauth-consent.functions.ts
@@ -1,0 +1,159 @@
+/**
+ * Server-fns backing the OAuth consent page (`apps/web/src/routes/auth/consent.tsx`).
+ *
+ * The flow:
+ * 1. The MCP plugin's authorize endpoint redirects the signed-in user to
+ *    `/auth/consent?consent_code=…&client_id=…&scope=…` (and sets a signed
+ *    `oidc_consent_prompt` cookie that BA's `oAuthConsent` endpoint reads).
+ * 2. The page calls {@link getOAuthConsentRequest} to fetch context the UI
+ *    needs (the requesting client's name + icon + the requested scopes
+ *    plus the user's organizations to pick from).
+ * 3. The user picks an org and presses Approve or Deny.
+ * 4. The page calls {@link decideOAuthConsent}. On approve, the server fn
+ *    binds the picked org to the OAuth application (writes
+ *    `oauth_applications.organization_id`) and then asks BA to finish the
+ *    authorize flow with `accept: true`. On deny, it just calls BA with
+ *    `accept: false` — no binding.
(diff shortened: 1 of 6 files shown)


neoxelox:
Folded into #3033 — the consent page is the first thing that exercises the BA register endpoint, so the FK fix lives where it gets tested.

PR #3022, opened 2026-05-08
feat(api): MCP scaffolding and tool manifest generation

neoxelox:
## Summary

Foundation for the MCP server that will mount at `/v1/mcp` in M2. Adds the helpers, the manifest emission pipeline, and migrates the api-keys routes as a smoke test.

- **`defineApiEndpoint` / `mountWithMcp`** — each API route is authored once with a `name` (camelCase) and `description`, then mounted under a prefix via `mountWithMcp`, which both wires the HTTP handler and registers the endpoint as an MCP tool.
- **`pnpm generate:mcp`** — emits `apps/api/mcp.json` alongside `openapi.json`. Mirrors `pnpm openapi:emit`; CI drift-guards both files.
- **api-keys migrated** as proof of concept. OpenAPI diff is just operationId renames (`apiKeys.create` → `createApiKey`, etc.); Fern vendor extensions and request/response shapes are untouched.
- **`/v1/mcp` mount handler deferred to M2** — it needs the auth-middleware rewrite from that milestone, so it's cleaner to ship together.
- **MCP SDK dep deferred to M2** — nothing in this PR consumes `@modelcontextprotocol/sdk` (zod 4's native `z.toJSONSchema` produces the manifest schemas), so we don't add the dep yet.

Plan reference: `plans/mcp-oauth-api-expansion.md` (D11, D12, M1 — MCP scaffolding + emit script + first migrated route).

## Test plan

- [x] `pnpm --filter @app/api typecheck`
- [x] `pnpm typecheck` (whole workspace)
- [x] `pnpm --filter @app/api test`
- [x] `pnpm --filter @app/api check` (biome)
- [x] `pnpm knip`
- [x] `pnpm generate:mcp` produces `mcp.json` with 3 tools (createApiKey, listApiKeys, revokeApiKey) with correct flat input schemas
- [x] `pnpm openapi:emit` diff is just operationId renames (no Fern SDK shape changes expected)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3037, opened 2026-05-08
feat(api): mount /v1/mcp transport — last M2 piece

neoxelox:
## Summary

Adds the MCP HTTP transport endpoint at `/v1/mcp` and wires every auto-generated tool through it. After this PR, MCP clients (Claude Code, Cursor, …) can complete the OAuth + consent flow shipped in #3033 + #3032, present `Authorization: Bearer …`, run `tools/list` to see the routes exposed via `defineApiEndpoint`, and `tools/call` to invoke them. **M2 closes with this**.

## Wiring

- `pnpm-workspace.yaml` — uncomment `@modelcontextprotocol/sdk@1.29.0` catalog entry.
- `apps/api/package.json` — add the SDK as a dependency.
- `apps/api/src/mcp/server.ts` — new `registerMcpRoute(...)`. Mounted on the protected sub-app (auth, rate-limit, org-context already applied upstream). Per-request `McpServer` + stateless `WebStandardStreamableHTTPServerTransport` — small enough to spin up per call, dwarfed by the inner fetch round-trip.
- `apps/api/src/routes/index.ts` — call `registerMcpRoute(...)` after `mountWithMcp(...)` so the tool registry is populated first.

## Tool dispatch

Each `tools/call` re-enters the root Hono app via `rootApp.fetch(internalRequest)` so the full middleware chain runs again on the inner call:

- Outer `Authorization` and `X-Forwarded-For` headers are forwarded — auth and rate-limit semantics carry across.
- `splitFlatInput` partitions the flat tool input back into `{ params, query, body }` buckets using the source map captured at registry time.
- `buildInternalUrl` substitutes path params (URL-encoded) and serialises the query string. Trailing-slash quirk handled — Hono treats `/api-keys` and `/api-keys/` as distinct routes, so templates of `"/"` collapse to the bare prefix.
- Response: 2xx → `isError: false`, otherwise `isError: true`. Body text becomes the tool's `content`.

`splitFlatInput` and `buildInternalUrl` were stubs in the M1 scaffolding PR (#3022) — implementing both here completes that scaffold.

## Tests (77 total, up from 59)

- `flatten-input.test.ts` — 11 new unit tests for `splitFlatInput` + `buildInternalUrl`: source-based partitioning, wrapped-body passthrough, unknown-source drop, path param encoding, query array repetition, null/undefined value drop, leading/trailing slash normalisation.
- `server.test.ts` (new) — 6 integration tests against the in-memory test app:
  1. 401 without bearer.
  2. `initialize` handshake returns server info + capabilities.
  3. `tools/list` returns the api-key tools.
  4. `tools/call listApiKeys` round-trips through the inner middleware (asserts tenant isolation: every returned row's `organizationId` matches the caller's org).
  5. `tools/call createApiKey` forwards JSON body and returns the new key.
  6. `tools/call revokeApiKey` against a cross-tenant id surfaces as `isError: true` from the inner 404.

Tests parse the SSE-framed JSON-RPC response directly so we don't pull in the MCP-client SDK as a test dep.

## Out of scope (M3+)

Migrating `projects.ts`, `scores.ts`, `annotations.ts` to `defineApiEndpoint` to expose them as MCP tools too. M3 starts there per the plan.

## Test plan

- [x] `pnpm --filter @app/api test` — 77 passing
- [x] `pnpm --filter @app/api check` (biome)
- [x] `pnpm knip`
- [x] `pnpm openapi:emit` — no diff (the MCP transport isn't a REST endpoint and shouldn't appear in the OpenAPI spec).
- [x] `pnpm mcp:emit` — no diff (tool list unchanged; this PR adds the *transport*, not new tools).
- [ ] **Manual smoke** (after merge): `npx @modelcontextprotocol/inspector http://localhost:3001/v1/mcp` → discovery → consent → `tools/list` returns 3 tools → `tools/call listApiKeys` returns the org's keys.

### DIFF
--- a/apps/api/scripts/emit-mcp.ts
+++ b/apps/api/scripts/emit-mcp.ts
@@ -39,7 +39,7 @@ const tools = collectToolDescriptors().map((tool) => ({
   // `outputSchema` is optional in the MCP spec; we omit the property entirely
   // (rather than emitting `null`) for 204 / no-JSON-body routes so clients can
   // rely on `"outputSchema" in tool` to mean "structured output is available".
-  ...(tool.output ? { outputSchema: z.toJSONSchema(tool.output, { target: "draft-2020-12" }) } : {}),
+  ...(tool.output ? { outputSchema: z.toJSONSchema(tool.output.schema, { target: "draft-2020-12" }) } : {}),
 }))
 
 const manifest = {
--- a/apps/api/src/mcp/server.ts
+++ b/apps/api/src/mcp/server.ts
@@ -1,0 +1,164 @@
+import type { OpenAPIHono, z } from "@hono/zod-openapi"
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js"
+import { WebStandardStreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/webStandardStreamableHttp.js"
+import type { Context } from "hono"
+import { API_VERSION, MCP_INFO } from "../constants.ts"
+import type { AppEnv, ProtectedEnv } from "../types.ts"
+import { splitFlatInput } from "./flatten-input.ts"
+import { collectToolDescriptors } from "./registry.ts"
+
+/**
+ * Mounts the MCP transport endpoint on `protectedApp` (which already lives
+ * behind the unified auth + org-context middleware). Tool dispatch re-enters
+ * the root Hono app via `rootApp.fetch(internalRequest)` so every tool call
+ * runs through the same middleware chain (validation, rate-limit, auth,
+ * organization context). The outer `Authorization` and `X-Forwarded-For`
+ * headers are forwarded so the inner request authenticates as the same
+ * caller and counts against the same rate-limit bucket.
+ *
+ * `mountPath` is the path on `protectedApp` (e.g. `"/mcp"` when
+ * `protectedApp` is mounted at `/${API_VERSION}` on the root, giving the
+ * public URL `/v1/mcp`). The dispatcher prepends `/${API_VERSION}` to the
+ * descriptor's router prefix when re-entering through `rootApp.fetch()` —
+ * so a tool whose registry entry has `routerPrefix: "/api-keys"` resolves
+ * to `/v1/api-keys`.
+ *
+ * Each request gets its own `McpServer` + `WebStandardStreamableHTTPServerTransport`
+ * + `connect()` + tool registration loop. This is an SDK invariant in stateless
+ * mode, not a stylistic choice:
+ *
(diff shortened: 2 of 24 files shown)


apps/api/src/mcp/server.ts
```diff
@@ -0,0 +1,124 @@
...
+  protectedApp: OpenAPIHono<ProtectedEnv>
+  rootApp: OpenAPIHono<AppEnv>
+  mountPath: string
+  internalPathPrefix: string
+}): void => {
+  const { protectedApp, rootApp, mountPath, internalPathPrefix } = params
+  const toolDescriptors = collectToolDescriptors()
+
+  protectedApp.all(mountPath, async (c) => {
+    const mcpServer = new McpServer({ name: "Latitude API", version: "1.0.0" }, { capabilities: { tools: {} } })
```
andresgutgon: This is doing the instance for each route

neoxelox: for each request

apps/api/scripts/emit-mcp.ts
```diff
@@ -39,7 +39,7 @@ const tools = collectToolDescriptors().map((tool) => ({
   // `outputSchema` is optional in the MCP spec; we omit the property entirely
   // (rather than emitting `null`) for 204 / no-JSON-body routes so clients can
   // rely on `"outputSchema" in tool` to mean "structured output is available".
-  ...(tool.output ? { outputSchema: z.toJSONSchema(tool.output, { target: "draft-2020-12" }) } : {}),
+  ...(tool.output ? { outputSchema: z.toJSONSchema(tool.output.schema, { target: "draft-2020-12" }) } : {}),
```
andresgutgon, line 42: 🔥

apps/api/src/routes/index.ts
```diff
@@ -41,16 +41,16 @@ export const registerRoutes = (app: OpenAPIHono<AppEnv>, options: ApiOptions) =>
...
-  // The org entity is loaded on every protected request from the API key's
-  // resolved org id. No path param feeds into it — see the middleware.
   routes.use("*", createOrganizationContextMiddleware())
 
   routes.route("/projects", createProjectsRoutes())
   routes.route("/projects/:projectSlug/scores", createScoresRoutes())
   routes.route("/projects/:projectSlug/annotations", createAnnotationsRoutes())
   mountWithMcp(routes, "/api-keys", apiKeysEndpoints)
 
+  registerMcpRoute(app, routes, "/mcp")
```
andresgutgon, line 51: Why we have `mountWithMcp` and `registerMcpRoute`?

andresgutgon (approved):

PR #3050, opened 2026-05-11
fix(api,ingest): suppress 4xx HttpError telemetry at the OTel boundary

andresgutgon:
## Summary

Fixes a Datadog Error Tracking issue (`fcf74994-…`) where expected client 4xx errors (a typo'd project slug → `NotFoundError`, an unauthenticated probe → `UnauthorizedError`, etc.) were being recorded as application exceptions on the HTTP server span and grouped into Error Tracking issues.

Diagnosis: `@hono/otel` reads `c.error` after `await next()` and unconditionally calls `span.recordException(c.error)` + `span.setStatus({ code: ERROR })`. Hono's `compose` sets `c.error` whenever a handler throws — even when the central `onError` already mapped it to a clean 4xx response. So *any* thrown domain error reaches telemetry as a span exception, regardless of the final HTTP status.

## Approach

Keep the codebase's convention of "throw domain errors, handle them centrally" — don't push nullable lookup plumbing into every route handler, and don't ship an entity-specific fix that misses the same bug for other entities (`OrganizationNotFoundError`, `UnauthorizedError`, `ValidationError`, …).

Instead, add a tiny Hono middleware that runs *inside* `@hono/otel` (registered immediately after `app.use(otel())`) and clears `c.error` once the global `onError` has produced a 4xx response. OTel then reads `c.error === undefined`, skips `recordException`, and leaves the span clean.

- 4xx `HttpError` (`httpStatus < 500`): suppressed from telemetry.
- 5xx `HttpError` (`RepositoryError`, `CacheError`, `StorageError`, etc.): pass through untouched — those are real infra issues worth seeing in Error Tracking.
- Non-`HttpError` throws: pass through untouched — those are bugs.

Applied to both `apps/api` and `apps/ingest` (both use `@hono/otel` and the same `isHttpError`-based error handler).

## Files

- `apps/api/src/middleware/suppress-http-error-telemetry.ts` — new middleware (+ unit test)
- `apps/ingest/src/middleware/suppress-http-error-telemetry.ts` — same middleware
- `apps/api/src/server.ts`, `apps/ingest/src/server.ts` — register after `otel()`

## Test plan

- [x] `pnpm typecheck` (full workspace)
- [x] `pnpm vitest run apps/api/src/middleware/suppress-http-error-telemetry.test.ts` — 3 cases (4xx HttpError cleared, 5xx HttpError preserved, plain `Error` preserved)
- [x] `pnpm --filter @app/api vitest run src/routes/projects.test.ts src/routes/annotations.test.ts src/routes/scores.test.ts` — all 16 tests still pass (no route changes)
- [ ] After merge: confirm in Datadog that no new `NotFoundError`/`UnauthorizedError` issues open for `service:api` or `service:ingest`, while 5xx issues continue to surface normally

### DIFF
--- a/apps/api/src/middleware/suppress-http-error-telemetry.ts
+++ b/apps/api/src/middleware/suppress-http-error-telemetry.ts
@@ -1,0 +1,11 @@
+import { isHttpError } from "@repo/utils"
+import { createMiddleware } from "hono/factory"
+
+// Must be registered after `app.use(otel())` — clears `c.error` before
+// `@hono/otel` reads it to call `recordException`.
+export const suppressHttpErrorTelemetry = createMiddleware(async (c, next) => {
+  await next()
+  if (c.error && isHttpError(c.error) && c.error.httpStatus < 500) {
+    c.error = undefined
+  }
+})
--- a/apps/ingest/src/middleware/suppress-http-error-telemetry.ts
+++ b/apps/ingest/src/middleware/suppress-http-error-telemetry.ts
@@ -1,0 +1,11 @@
+import { isHttpError } from "@repo/utils"
+import { createMiddleware } from "hono/factory"
+
+// Must be registered after `app.use(otel())` — clears `c.error` before
+// `@hono/otel` reads it to call `recordException`.
+export const suppressHttpErrorTelemetry = createMiddleware(async (c, next) => {
+  await next()
+  if (c.error && isHttpError(c.error) && c.error.httpStatus < 500) {
+    c.error = undefined
+  }
+})
--- a/apps/api/src/middleware/suppress-http-error-telemetry.test.ts
+++ b/apps/api/src/middleware/suppress-http-error-telemetry.test.ts
@@ -1,0 +1,46 @@
+import { NotFoundError, RepositoryError } from "@domain/shared"
+import { Hono } from "hono"
+import { describe, expect, it } from "vitest"
+import { suppressHttpErrorTelemetry } from "./suppress-http-error-telemetry.ts"
+
+const buildApp = (thrown: unknown) => {
+  let observedError: Error | undefined
+  const app = new Hono()
+  app.use(async (c, next) => {
+    await next()
+    observedError = c.error
+  })
+  app.use(suppressHttpErrorTelemetry)
+  app.get("/throws", () => {
+    throw thrown as Error
+  })
+  app.onError((err, c) => c.json({ error: err.message }, 500))
+  return { app, getObservedError: () => observedError }
+}
+
+describe("suppressHttpErrorTelemetry", () => {
+  it("clears c.error for 4xx HttpErrors so @hono/otel skips recordException", async () => {
+    const { app, getObservedError } = buildApp(new NotFoundError({ entity: "Project", id: "missing" }))
+
+    await app.request("/throws")
+
+    expect(getObservedError()).toBeUndefined()
+  })
+
+  it("leaves c.error in place for 5xx HttpErrors", async () => {
+    const { app, getObservedError } = buildApp(new RepositoryError({ operation: "findById", cause: "boom" }))
+
+    await app.request("/throws")
+
+    expect(getObservedError()).toBeInstanceOf(RepositoryError)
+  })
+
+  it("leaves c.error in place for non-HttpErrors", async () => {
+    const plain = new Error("kaboom")
+    const { app, getObservedError } = buildApp(plain)
+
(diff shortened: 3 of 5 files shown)


packages/domain/projects/src/ports/project-repository.ts
```diff
@@ -6,7 +6,7 @@ export class ProjectRepository extends Context.Service<
   ProjectRepository,
   {
     findById(id: string): Effect.Effect<Project, NotFoundError | RepositoryError, SqlClient>
-    findBySlug(slug: string): Effect.Effect<Project, NotFoundError | RepositoryError, SqlClient>
+    findBySlug(slug: string): Effect.Effect<Project | null, RepositoryError, SqlClient>
```
geclos: this is breaking the pattern of repositories methods findBy* always return value or error. You have to use findOptionalBy* (check the relevant skill)

geclos (commented):
change the method to findOptionalBy*, use it only on the boundaries that strictly need it (e.g web functions don't need it)

packages/domain/projects/src/ports/project-repository.ts
```diff
@@ -7,6 +7,15 @@ export class ProjectRepository extends Context.Service<
...
     findById(id: string): Effect.Effect<Project, NotFoundError | RepositoryError, SqlClient>
     findBySlug(slug: string): Effect.Effect<Project, NotFoundError | RepositoryError, SqlClient>
+    /**
+     * Like {@link findBySlug} but resolves to `null` when the slug is unknown
+     * instead of failing with `NotFoundError`. Use this at HTTP boundaries
+     * where a missing slug is an expected client error: returning a 404
+     * directly avoids surfacing the exception on the OTel span, which would
+     * otherwise show up in Datadog Error Tracking as a real application
+     * error.
+     */
```
geclos: unnecessary comment

packages/platform/db-postgres/src/repositories/project-repository.ts
```diff
@@ -113,6 +113,27 @@ export const ProjectRepositoryLive = Layer.effect(
             )
         }),
 
+      findOptionalBySlug: (slug: string) =>
+        Effect.gen(function* () {
```
geclos: use effect.fn instead of effect.gen

geclos (changes requested):
now i that i think about this, this PR breaks the conventional way we handle errors in api boundaries. By convention we allow errors to throw and handle them centrally in the error handler middleware. This change now forces us to handle nullable case in project find bys of every api route handler, and it does not protect us from other entities not being found (e.g organization not found error will surface the same not found error in telemetry). I think the proper solution is to not report NotFoundError to telemetry or ignore them in the telemetry dashboard (datadog).

apps/api/src/middleware/suppress-http-error-telemetry.ts
```diff
@@ -0,0 +1,21 @@
...
+import { createMiddleware } from "hono/factory"
+
+/**
+ * Clears `c.error` for client-side `HttpError`s (4xx) after the error handler
+ * has produced its response. `@hono/otel` reads `c.error` to decide whether to
+ * call `span.recordException` and mark the span as failed, which surfaces in
+ * Datadog Error Tracking. Expected client errors (missing slug, missing auth,
+ * etc.) are part of normal API behavior — they're mapped to a clean 4xx by
+ * `honoErrorHandler` and should not pollute issue tracking. 5xx HttpErrors
+ * pass through untouched.
```
geclos: remove the yapping

apps/ingest/src/middleware/suppress-http-error-telemetry.ts
```diff
@@ -0,0 +1,21 @@
...
+import { createMiddleware } from "hono/factory"
+
+/**
+ * Clears `c.error` for client-side `HttpError`s (4xx) after the error handler
+ * has produced its response. `@hono/otel` reads `c.error` to decide whether to
+ * call `span.recordException` and mark the span as failed, which surfaces in
+ * Datadog Error Tracking. Expected client errors (missing slug, missing auth,
+ * etc.) are part of normal API behavior — they're mapped to a clean 4xx by
+ * the global `onError` handler and should not pollute issue tracking. 5xx
+ * HttpErrors pass through untouched.
```
geclos: remove the yapping

geclos (dismissed):
remove the yapping comments and ok

neoxelox (approved):

PR #3054, opened 2026-05-11
feat(issues): seasonal anomaly detection for issue escalation

csansoon:
## what we had

The escalation detector was a single rule keyed on a flat 7-day baseline:

**Entry** — a non-escalating issue starts escalating when, in the trailing 24 hours, its occurrence count clears `max(20, floor(baselineAvg × 1.33) + 1)`, where `baselineAvg` is the average daily occurrences over days 1-8 ago. Issues younger than 7 days are excluded outright (their baseline window hasn't filled in yet).

**Exit** — an escalating issue stops escalating when its trailing-24h count drops below `floor(entryThreshold × 0.7)`. The 0.7 factor is hysteresis: exit must clear meaningfully below entry, otherwise an issue right at the boundary would flap between states.

Both transitions are emitted as outbox events (`IssueEscalated` / `IssueEscalationEnded`) consumed asynchronously by the alert-incidents worker, which is what actually opens or closes the `alert_incidents` row.

## what's wrong with it

The detector has three concrete failure modes:

- **Slow.** The signal is a 24h count. A real spike at 09:00 needs hours of bad traffic to clear an effective threshold; an incident starting at midnight may not trip until late morning. We wanted "react within minutes," and the window is the limit.
- **No notion of seasonality.** A flat 7-day average ignores time-of-day and day-of-week. Any issue with a natural cycle — deploy cadence, weekday traffic, evaluation runs — produces false positives at its weekly peaks and false negatives at its troughs. The `1.33` multiplier is the same whether the bin's natural variance is 5% or 200%.
- **Blind to low-volume spikes.** The hard floor at 20 occurrences means an issue at ~3 occurrences/day can 10× without ever clearing the threshold. Below the floor, no relative spike triggers anything.

## how others do this

These are the patterns we surveyed before designing the replacement:

- **Datadog / CloudWatch / Grafana anomaly detection** all compute a *per-metric expected band* from historical data, usually with separate weekday/weekend or hour-of-week seasonality. The alert trips when the live value falls outside `expected ± k × sigma`. Hours where the metric is naturally noisy get wider bands automatically.
- **Sentry's issue escalation** keeps a per-issue weekly forecast (Sentry calls it "predicted events per hour") retrained in the background, and trips on observed-vs-predicted ratios. Closer to a learned-baseline approach than a band one.
- **Multi-window-multi-burn-rate alerting** (Google SRE workbook) — two windows of different lengths must both cross their thresholds for the alert to fire. The short window proves "this is happening *now*", the long window proves "this is *sustained*", and the AND kills single-noise-bin false positives that either window alone would trigger on.
- **Temporal dwell on recovery** (Prometheus `keep_firing_for`, Datadog `recovery_window`) — alerts don't close the instant the condition stops holding. They wait for the condition to *stay* false for a configured duration before closing. Avoids the inverse-flapping case where a single-bin dip closes an active incident.

## the new detector

The replacement is a seasonal anomaly detector with multi-window confirmation and a three-layer exit. Same outbox event surface; only the trip rule changes.

**Entry** — both windows must cross their bands simultaneously:

- **1h sliding window**: `recent_1h > expected_1h + k_short × sigma_1h`
- **6h sliding window** (per-hour rate): `recent_6h / 6 > expected_per_hour + k_long × sigma_per_hour`

`k_long = k_short - 1` is derived internally so the contained short window can't dominate the long one (this is the SRE multi-window pattern). `expected_*` and `sigma_*` come from the same `(day_of_week, hour)` bin across the prior 4 weeks, pooled with the two neighbouring hours so each estimate uses 12 samples instead of 4. Effective sigma is floored at `max(observed, sqrt(expected), 1.0)` — `sqrt(expected)` is the Poisson lower bound for count data; `1.0` prevents the first non-zero sample after a quiet stretch from tripping the detector.

`k_short` is the only user-facing knob: a 1-6 slider on project settings (the "escalation sensitivity"), default `3`, where lower is noisier and higher is quieter. `3` approximates 99% confidence under a normal assumption.

image

_Issue trend with a dynamic threshold, showing the maximum of either the 1h or 6h sliding windows, which are based on the past week(s) of occurrences_

**Exit** — evaluated in priority order:

1. **Timeout** (72h) — backstop against ghost incidents. Matches what Datadog / CloudWatch / GCP Monitoring all do.
2. **Absolute-rate drop** — `recent_24h < entry_count_24h × 0.5`. Catches the case where the seasonal baseline climbs to meet a declining-but-still-elevated incident; the band-shape exit would silently miss that.
3. **Band-shape + dwell** — both windows must hold below `expected + 0.7 × k × sigma` *continuously for 30 minutes* (~2 evaluation bins) before the incident closes. The dwell kills single-bin dip false-closes.

Cold-start: `isNew` (issue younger than 7 days) blocks entry outright. When fewer than 2 of the last 4 weeks contributed any data to the relevant bin, `k` is inflated by +1 — wider bands where we have less evidence. When zero weeks contributed, the detector falls back to the pre-rewrite absolute occurrence floor.

The exit reason (`threshold` / `absolute-rate-drop` / `timeout`) flows through `IssueEscalationEnded` and `IncidentClosed` for observability. It's not surfaced in V1 notification copy.

## how this is built

**ClickHouse materialized view for seasonality.** At ~15 min per active issue, reading raw `scores` for each check at 1M scores/day would scan ~28M rows per check. The new `scores_hourly_buckets` MV (`AggregatingMergeTree` keyed on `(organization_id, project_id, issue_id, ts_hour)`, partitioned monthly) brings each check down to ~672 rows regardless of underlying volume. The MV populates insert-time via `scores_hourly_buckets_mv`; the migration also runs a one-shot backfill so historical hours are queryable from day one.

Sliding 1h/6h/24h counts still read from raw `scores` — that's a single-issue PK lookup, a tiny scan. The MV is only for the 4-weeks-of-history side, where we'd otherwise be reading millions of rows.

**Persisted state on the incident row.** Two new columns on `alert_incidents`:

- `entry_signals jsonb` — snapshot of `expected_*`, `sigma_*`, `k_*`, `entry_threshold_*`, `entry_count_24h` frozen at entry. The absolute-rate backstop compares against this snapshot, not against live signals — that's how we keep a sustained incident from being closed by the rolling baseline catching up.
- `exit_eligible_since timestamptz` — start of the band-shape exit dwell. Cleared whenever the exit shape fails; advanced silently across check ticks until the dwell duration is met.

Incidents created before this migration get `NULL` on both columns. The detector treats them as legacy: skips the absolute-rate backstop (no snapshot to compare to), but still honours band-shape exit and timeout, so they resolve naturally.

**Architecture.** The decision math is a pure function (`evaluateSeasonalEscalation` in `@domain/issues/helpers.ts`). The use case is a thin I/O boundary that reads the issue, the open incident, the seasonal signals, and the project sensitivity in parallel, calls the helper, and either emits the transition event or writes a single `exit_eligible_since` delta. Incident open/close still happens asynchronously in the alert-incidents worker via outbox events — same separation as before.

**Settings UI.** New 1-6 slider in the Issues section of project settings, debounced through `updateProjectMutation`, gated by the existing `notifications` feature flag.

## out of scope

- The legacy `getEscalationOccurrenceThreshold` / `getEscalationExitThreshold` helpers stay in place — they still drive the `escalationOccurrenceThreshold` *display* field on `IssueListItem` (separate from the detector). Whether that field is still meaningful with the new bands is a product call for a follow-up.
- Email / Slack delivery, per-user notification subscriptions, adaptive sensitivity, and a "lower than expected" anomaly signal are all on the notifications V2 roadmap, not this PR.
- No killswitch back to the legacy detector. The outbox event surface is unchanged, so semantic continuity for downstream consumers makes a flag unnecessary.


packages/domain/shared/src/settings.ts
```diff
@@ -17,8 +17,25 @@ export const organizationSettingsSchema = z.object({
...
+ * the named `escalationSensitivity` key live at the same level — no name
+ * collision is possible since the kinds are dotted identifiers.
  */
-export const alertNotificationsSettingSchema = z.partialRecord(z.enum(ALERT_INCIDENT_KINDS), z.boolean())
+export const alertNotificationsSettingSchema = z.intersection(
+  z.partialRecord(z.enum(ALERT_INCIDENT_KINDS), z.boolean()),
+  z.object({
+    escalationSensitivity: z.number().int().min(1).max(6).optional(),
+  }),
+)
```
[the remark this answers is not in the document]
csansoon: Good catch — verified with `safeParse({ "issue.new": true, escalationSensitivity: 3 })` and the record half of the intersection was indeed rejecting every payload that carried the new key. Swapped to a `z.object` built from `ALERT_INCIDENT_KINDS` so the kind keys stay derived from the constant. Fixed in f119f3d0b.

packages/platform/db-clickhouse/src/repositories/score-analytics-repository.ts
```diff
@@ -594,6 +718,87 @@ export const ScoreAnalyticsRepositoryLive = Layer.effect(
...
+      //   2. Same-(dow, hour ± 1) bucket counts from `scores_hourly_buckets`
+      //      for the prior `SEASONAL_HISTORY_WEEKS` weeks. 6h-band samples are
+      //      reconstructed from six consecutive bucket rows ending at each
+      //      anchor; aggregation to mean/stddev happens in TS because the
+      //      sample set is small (12 anchors × N issues).
+      escalationSignalsByIssues: ({ organizationId, projectId, issueIds, now, options }) =>
+        Effect.gen(function* () {
+          const chSqlClient = (yield* ChSqlClient) as ChSqlClientShape<ClickHouseClient>
+          if (issueIds.length === 0) return []
+          const nowMs = (now ?? new Date()).getTime()
```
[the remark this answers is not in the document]
csansoon: Right, this was a real desync risk for any test overriding `now`. Bound the SQL sliding-window edges to a `{now:String}` query parameter parsed with `toDateTime(..., UTC)` so the JS anchors and the SQL `countIf` windows pin to the same instant. Fixed in 05e25f296.

packages/platform/db-clickhouse/src/repositories/score-analytics-repository.ts
```diff
@@ -242,6 +282,90 @@ const toIssueOccurrenceBucket = (row: IssueOccurrenceBucketRow): IssueOccurrence
...
+      let sixH = 0
+      for (let i = 0; i <= 5; i++) {
+        sixH += perIssueBuckets.get(anchorMs - i * HOUR_MS) ?? 0
+      }
+      oneHourSamples.push(oneH)
+      sixHourPerHourSamples.push(sixH / 6)
+      if (oneH > 0 || sixH > 0) {
+        // bucket "week" = how many full weeks ago — i.e. floor((now - anchor) / 7d)
+        // We don't care about offset within the pool when counting samples.
+        const weekIndex = Math.round((input.nowMs - anchorMs) / WEEK_MS)
```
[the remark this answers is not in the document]
csansoon: Comment was the wrong half here — `Math.round` is the right operator. With `SEASONAL_BUCKET_POOLING_HOURS = 1`, the `+1h` offset anchor of week N lands at `(now − anchor) / 7d ≈ N − 0.006`, which `floor` would push into bucket N-1 and undercount `samplesCount`. Rewrote the comment to explain why round groups the whole pool. Fixed in ac3a94c4a.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
```diff
@@ -305,3 +345,46 @@ function ProjectSettingsPage() {
...
+  readonly value: number
+  readonly isSaving: boolean
+  readonly onChange: (value: number) => void
+}
+
+// Local presentational state so dragging the slider feels responsive while
+// the actual save is debounced through the parent's `onChange`.
+function EscalationSensitivityControl({ value, isSaving, onChange }: EscalationSensitivityControlProps) {
+  const [draft, setDraft] = useState(value)
+
```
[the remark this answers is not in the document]
csansoon: Agreed — `useState(value)` only seeds once. Added a `useEffect([value])` that re-syncs `draft` whenever the prop changes (project data finishing its load, saved value coming back from the collection, navigation). Fixed in 9156c427e.

apps/web/src/routes/_authenticated/projects/$projectSlug/issues/-components/issue-trend-bar.tsx
```diff
@@ -428,16 +521,37 @@ export function IssueTrendBar({
...
+                viewBox={`0 0 ${chartBuckets.length} 100`}
+                preserveAspectRatio="none"
+                role="img"
+                aria-label="Escalation threshold"
+              >
+                <title>Escalation threshold</title>
+                {thresholdSegments.map((segment, segmentIndex) => (
+                  <path
+                    key={`th-seg-${segmentIndex}`}
+                    d={buildSmoothThresholdPath(segment)}
```
andresgutgon, line 566: Maybe `useMemo` of all the segments?

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
```diff
@@ -169,6 +173,32 @@ function ProjectSettingsPage() {
...
   }
 
+  const handleSensitivityChange = (value: number) => {
+    // Slider fires on every step; debounce so we don't flood the API with one
+    // update per tick. Pattern mirrors `handleProjectRename` above.
+    if (sensitivityDebounceRef.current) clearTimeout(sensitivityDebounceRef.current)
+
+    sensitivityDebounceRef.current = setTimeout(() => {
+      setIsSavingSensitivity(true)
+      const transaction = updateProjectMutation(currentProject.id, {
```
andresgutgon: This is a weird pattern. We have https://github.com/latitude-dev/latitude-llm/blob/development/apps/web/src/lib/form-server-action.ts#L29

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
```diff
@@ -305,3 +345,54 @@ function ProjectSettingsPage() {
...
+function EscalationSensitivityControl({ value, isSaving, onChange }: EscalationSensitivityControlProps) {
+  const [draft, setDraft] = useState(value)
+
+  // Sync the local draft when `value` changes from outside (project data finishing
+  // its initial load, the saved value coming back from the collection, navigation
+  // between projects). `useState(value)` only seeds once, so without this the
+  // slider can sit on a stale number after any of those transitions.
+  useEffect(() => {
+    setDraft(value)
+  }, [value])
```
andresgutgon: https://github.com/latitude-dev/latitude-llm/blob/development/packages/ui/src/hooks/use-value-with-default.ts#L44

apps/web/src/routes/_authenticated/projects/$projectSlug/issues/-components/issue-trend-bar.tsx
csansoon: Good call. Wrapped `chartBuckets`, `hasSeasonalThresholds`, `visibleBucketLabelIndices`, `maxCount`, and `thresholdSegments` in `useMemo`, hoisted above the early-return branches. Fixed in 575607bca.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
csansoon: `createFormSubmitHandler` is wired into TanStack Form `onSubmit`, so it doesn't quite fit a debounced slider mutation, but the underlying smell was real: the manual `isSavingSensitivity` state and the chained `.then/.catch/.finally` on top of an awaited promise was extra ceremony. Aligned the handler with `handleProjectRename` above (`useCallback` + minimal transaction promise handling, no per-handler saving flag). Fixed in 2ef69b7a0. Happy to wrap it in a real form if you want that pattern instead.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
csansoon: Nice — swapped the `useState + useEffect` sync for `useValueWithDefault`. Same external-sync semantics, no effect round-trip. Fixed in 5205591a5.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
```diff
@@ -169,6 +173,31 @@ function ProjectSettingsPage() {
     }
   }
 
+  const handleSensitivityChange = useCallback(
+    (value: number) => {
+      // Slider fires on every step; debounce so we don't flood the API with one
+      // update per tick. Pattern mirrors `handleProjectRename` above.
+      if (sensitivityDebounceRef.current) clearTimeout(sensitivityDebounceRef.current)
+
+      sensitivityDebounceRef.current = setTimeout(() => {
```
andresgutgon: Super weird way of doing a debounce. 
```javascript

```

packages/platform/db-clickhouse/src/repositories/score-analytics-repository.ts
```diff
@@ -800,6 +1133,102 @@ export const ScoreAnalyticsRepositoryLive = Layer.effect(
...
+                    bucketEndMs,
+                    kShort,
+                  })
+                : Number.NaN
+              const iso = `${new Date(bucketStartMs).toISOString().slice(0, 19)}.000Z`
+              return { bucket: iso, thresholdCount }
+            })
+            return { issueId, buckets }
+          })
+        }),
```
andresgutgon, line 1231: HUGE file full of changes. 0 tests

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
andresgutgon: Not tied to TanStack form. Is tied to our functions. It returns the correct shape for validation errors

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
csansoon: We don't have `use-debounce` in the project, and all other project settings follow the same pattern

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
csansoon: Got it — pulled in `extractFieldErrors` on the sensitivity catch so server-function Zod failures surface their real message instead of `toUserMessage` returning the raw JSON-encoded issues. Falls back to `toUserMessage` for non-validation errors. Fixed in 92f76ddf8.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
csansoon: Pulled `use-debounce` into `apps/web` and switched both manual `useRef + setTimeout` handlers in this file (`handleProjectRename` and `handleSensitivityChange`) to `useDebouncedCallback`. The two account/organization settings pages still carry the same manual pattern — left them for a follow-up so this PR's scope stays on the seasonal detector. Fixed in e45a99e47.

packages/platform/db-clickhouse/src/repositories/score-analytics-repository.ts
csansoon: Fair. Added 10 integration tests for the new repo methods (`escalationSignalsByIssues` x5, `escalationThresholdHistogramByIssues` x5) via the existing chdb-backed setup, covering: zero-history fallback, sliding 1h/6h/24h windows respecting the overridable `now`, samplesCount counting distinct prior weeks, the pooled (dow, hour ± 1) × 4-week mean, multi-issue batching, sub-hour bucket short-circuit, NaN thresholds when prior pool is empty, finite thresholds with history + aligned bucket keys, and k_short widening the band. Also backported the `scores_hourly_buckets` table + MV into the testkit schema so the insert-time trigger works end-to-end. 128 tests passing. Fixed in 590000c58.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
```diff
@@ -179,7 +179,16 @@ function ProjectSettingsPage() {
...
+        // Server-function Zod failures arrive as JSON-encoded issues; running
+        // them through `toUserMessage` would surface the raw `[{path,message}]`
+        // payload in the toast. Pull the first validation message when present
+        // and fall back to the generic message otherwise — same shape
+        // `createFormSubmitHandler` uses to drive field-level error UI.
+        const fieldErrors = extractFieldErrors(error)
+        const message = fieldErrors
+          ? (Object.values(fieldErrors).flat()[0] ?? toUserMessage(error))
+          : toUserMessage(error)
+        toast({ variant: "destructive", description: message })
```
andresgutgon: https://github.com/latitude-dev/latitude-llm/blob/development/apps/web/src/lib/form-server-action.ts#L29 

Use like this 

```
createFormSubmitHandler(
      async (value) => {
        const { projectId, transaction } = createProjectMutation(value.name)
        await transaction.isPersisted.promise
        return { projectId }
      },
    ),

```

csansoon: Got it — wrapped the slider save in a TanStack Form so it goes through `createFormSubmitHandler` like the rest. One-field shell: `defaultValues` seeds from the current setting, each slider tick calls `form.setFieldValue` + a debounced `form.handleSubmit`, so the helper-wrapped action runs with the freshest value on every save. The slider's displayed value stays driven by `useValueWithDefault(value)` on the prop, so the helper's reset-on-success doesn't affect what the user sees. Fixed in 3bef9aaf7.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings.tsx
```diff
@@ -131,26 +136,19 @@ function ProjectSettingsPage() {
...
-  )
+    const transaction = updateProjectMutation(currentProject.id, { name: trimmedName })
+    void transaction.isPersisted.promise
+      .then(() => {
+        toast({ description: "Project name updated" })
+      })
+      .catch((error) => {
+        toast({ variant: "destructive", description: toUserMessage(error) })
+      })
+  }, 600)
```
andresgutgon, line 151: If you use `useForm` + `createFormSubmitHandler` here you can have validation error under the name input field.

andresgutgon (approved):

PR #3063, opened 2026-05-11
feat: add Latitude telemetry bootstrap class

geclos:
## Summary

Adds the class-based TypeScript telemetry bootstrap API:

```ts
const latitude = new Latitude({
  apiKey,
  projectSlug,
  instrumentations: ["openai"],
})
```

`initLatitude()` is kept as a `@deprecated` compatibility wrapper around `new Latitude(...)`, and the public option type is now `LatitudeOptions`.

## Existing tracing SDK integration

`Latitude` now checks for an existing OpenTelemetry provider before creating its own provider. When an existing provider is present, it attaches the Latitude span processor to that provider instead of replacing the application's tracing setup. This keeps the existing provider, context manager, propagator, sampler, and vendor span processors in place.

The integration path covers the common Node tracing SDK shapes used by Sentry, Datadog, New Relic, Honeycomb, and custom OTel SDK setups:

- public `addSpanProcessor()` providers
- OTel JS v2 `MultiSpanProcessor` internals used by `NodeSDK`/`BasicTracerProvider` setups
- Datadog's `_activeProcessor._processors` bridge shape
- explicit `tracerProvider` injection for custom setups

Provider detection uses the public `ProxyTracerProvider.getDelegateTracer()` API (replacing earlier `constructor.name` / private `_TRACER_PROVIDER` reads), so it survives OTel internal refactors and minified bundles. When a provider is discovered but none of the known attach hooks match, Latitude logs a clear warning and falls back to its own non-globally-registered provider so instrumentations still reach Latitude.

`flush()` and `shutdown()` only affect Latitude's processor when Latitude shares a provider owned by another SDK.

## service.name handling

`service.name` is treated as a resource attribute per OTel semantic conventions, not a span attribute:

- When Latitude owns the provider, `serviceName` is set on the provider's `Resource`.
- When piggy-backing on an existing provider, `serviceName` is ignored — the host provider's resource is the source of truth, so Latitude does not silently relabel spans the host SDK also processes.
- Advanced use of `LatitudeSpanProcessor({ serviceName })` directly (without the bootstrap) applies the value via a span-exporter wrapper that overrides only spans flowing through Latitude's exporter, leaving the host provider's other processors untouched.

## Docs and examples

Updates TypeScript telemetry docs, provider/framework guides, package README, examples, root README, and web onboarding snippets to use `new Latitude(...)`.

## Validation

Local:

- `pnpm knip`
- `pnpm --filter @latitude-data/telemetry test`
- `pnpm --filter @latitude-data/telemetry typecheck`
- `pnpm --filter @latitude-data/telemetry check`
- `cd packages/telemetry/python && uv run pytest tests/`
- `cd packages/telemetry/python && uv run pyright src`
- `cd packages/telemetry/python && uv run ruff check src tests`

CI:

- check: pass
- typecheck: pass
- test: pass
- knip: pass
- web-bundle-size: pass
- check-migration-versions: pass
- CodeQL: pass

### DIFF
--- a/docs/telemetry/otel-exporter.md
+++ b/docs/telemetry/otel-exporter.md
@@ -222,7 +222,7 @@ Latitude expects the standard **parts-based** GenAI message format. Each message
 
 ## Existing Observability Stack
 
-Latitude works alongside your existing observability tools. Add `LatitudeSpanProcessor` as an additional span processor so traces go to both Latitude and your current backend.
+Latitude works alongside your existing observability tools. In TypeScript, `new Latitude()` detects common OpenTelemetry-compatible providers (Sentry, Datadog, New Relic, Honeycomb, and custom OTel SDKs) and attaches Latitude when possible; custom setups can also add `LatitudeSpanProcessor` as an additional span processor so traces go to both Latitude and your current backend.
 
 ### With Datadog (TypeScript)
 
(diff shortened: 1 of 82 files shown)


packages/telemetry/typescript/README.md
geclos, line 333: the code example for datadog is not using the new Latitude approach, why?

geclos: Addressed in a013e87b4: the Datadog examples now use the class-based `new Latitude(...)` bootstrap instead of manual `LatitudeSpanProcessor` wiring.

andresgutgon:
Update skills repo also

docs/telemetry/otel-exporter.md
```diff
@@ -222,63 +222,66 @@ Latitude expects the standard **parts-based** GenAI message format. Each message
...
+  apiKey: process.env.LATITUDE_API_KEY!,
+  projectSlug: process.env.LATITUDE_PROJECT_SLUG!,
+  instrumentations: ["openai"],
 })
 
-await registerLatitudeInstrumentations({
+await latitude.ready
+```
+
+### With New Relic or Honeycomb (TypeScript)
```
geclos: let's add concrete sections for new relic and honeycomb with real code examples rather than a vague "// Initialize new relic or honeycomb first" comment

docs/telemetry/otel-exporter.md
```diff
@@ -222,63 +222,66 @@ Latitude expects the standard **parts-based** GenAI message format. Each message
...
-  ),
-)
+const latitude = new Latitude({
+  apiKey: process.env.LATITUDE_API_KEY!,
+  projectSlug: process.env.LATITUDE_PROJECT_SLUG!,
+  instrumentations: ["openai"],
+})
 
-provider.register()
+await latitude.ready
```
geclos: is this strictly necessary?

geclos:
Addressed in ed23381c8: updated `AGENTS.md` so the skill glossary covers all 21 `.agents/skills/*/SKILL.md` entries, including the missing skills.

docs/telemetry/otel-exporter.md
geclos: Addressed in ed23381c8: split New Relic and Honeycomb into separate sections with concrete TypeScript setup examples.

docs/telemetry/otel-exporter.md
geclos: Addressed in ed23381c8: it is not strictly necessary, so I removed it from the snippets and added a note that `await latitude.ready` is optional when startup needs to wait for instrumentation registration.

andresgutgon (changes requested):
Not bumping package versions

andresgutgon (approved):
No `pnpm-lock.yaml` change. I guess is fine?

PR #3064, opened 2026-05-11
Add screen redirect for mobile users

learningbizz:
We're having a good % of users entering with mobile to Latitude, and its not built for it

This PR adds a screen detecting if its a mobile phone to gatekeep users to try it on their PC and not have a bad first experience


andresgutgon (changes requested):
Simple css media query to hide right panel

andresgutgon (approved):

PR #3068, opened 2026-05-12
fix(flaggers): silence false-positive Datadog error for tolerated schema mismatch

andresgutgon:
## Summary

- Datadog Error Tracking issue [bd420c54-3f4a-11f1-8634-da7ad0900005](https://app.datadoghq.eu/error-tracking?query=service%3A%2A&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22bd420c54-3f4a-11f1-8634-da7ad0900005%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D) was firing on `AI_NoObjectGeneratedError` from `flagger.classify`. The flagger already treats this as `matched: false` in business logic, but the Latitude `capture` span recorded the exception *before* `Effect.catchIf` recovered it — so Datadog kept reporting a tolerated outcome as an error.
- Recover the SDK error inside the ai-vercel adapter via an opt-in `tolerateSchemaMismatch` flag on `GenerateInput`. When set, `generateText` rejections that match `NoObjectGeneratedError` are caught inside `execute` (before the capture span sees them) and resolved as `{ object: undefined, tokens: 0, ... }`.
- Flagger opts in and branches on `result.object === undefined`; the cause-introspecting `Effect.catchIf` workaround is gone. Schema-mismatch unknowns from any *other* `ai.generate` call site remain visible as errors.

## Test plan

- [x] `pnpm --filter @platform/ai-vercel --filter @domain/ai --filter @domain/flaggers --filter @platform/ai --filter @domain/annotations --filter @domain/evaluations --filter @domain/issues --filter workflows typecheck`
- [x] `pnpm --filter @domain/flaggers test` — 164/164 passing
- [x] `pnpm --filter @platform/ai --filter @platform/ai-vercel test` — 13/13 passing
- [ ] After deploy: confirm in Datadog that the `flagger.classify` resource no longer surfaces `NoObjectGeneratedError`, while unrelated `ai.generate` callers still report it on failure

### DIFF
--- a/packages/domain/ai/src/index.ts
+++ b/packages/domain/ai/src/index.ts
@@ -59,6 +59,14 @@ export interface GenerateInput<T> {
    * Does not affect generation semantics; excluded from AI cache keys (see `withAICache`).
    */
   readonly telemetry?: GenerateTelemetryCapture
+  /**
+   * When `true`, an `AI_NoObjectGeneratedError` from the underlying SDK (model output that does
+   * not parse against the schema) is recovered inside the adapter: the call resolves with
+   * `object: undefined` and zeroed token usage instead of rejecting. Callers that opt in must
+   * handle the `undefined` case. Used by flows where a schema mismatch is a valid "no result"
+   * signal and should not surface as an error on the telemetry span.
+   */
+  readonly tolerateSchemaMismatch?: boolean
 }
 
 export interface GenerateResult<T> {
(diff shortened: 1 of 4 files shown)


geclos (changes requested):
even though we gracefully handle this error we need to investigate why it happens rather than silence it.

PR #3097, opened 2026-05-12
feat(api): projects routes through defineApiEndpoint with paginated list

neoxelox:
Migrates the projects routes to `defineApiEndpoint` so the five operations (create, list, get, update, delete) surface as MCP tools alongside the HTTP endpoints. List response switches to the standard `Paginated(ProjectSchema, "PaginatedProjects")` shape so SDK consumers see a consistent envelope across resources. Adds field descriptions on every schema property, exposes the previously-dropped `settings`, `firstTraceAt`, and `lastEditedAt` fields on reads, and lets PATCH update settings (the use-case already supported it). Applies the `medium` rate-limit tier and wires the Fern SDK group naming so the generated SDK lands at `client.projects.*`.

### DIFF
--- a/apps/api/mcp.json
+++ b/apps/api/mcp.json
@@ -3,6 +3,553 @@
   "version": "1.0.0",
   "tools": [
     {
+      "name": "createProject",
+      "title": "Create project",
+      "description": "Creates a new project within the organization. The name must be unique within the org.",
+      "inputSchema": {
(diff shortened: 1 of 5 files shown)


apps/api/mcp.json
andresgutgon, line 1: 🔥 🙏

andresgutgon (approved):

PR #3098, opened 2026-05-12
Add segmented trace search editor

geclos:
## What changed

Adds explicit trace search syntax for literal and ordered token-phrase matching, then updates the search box to make that syntax visible while typing. Double quotes now represent case-sensitive literal substring matches, while backticks represent ordered contiguous token phrase matches. Unquoted text remains semantic search text.

The trace search input now renders quote/backtick sections as inline editable pills instead of a plain text-only input. Users can type `"` or `` ` `` to enter a pill, press Enter to leave it, and paste mixed syntax such as `"example" `followed by a phrase` followed by semantic text` into the editor. Literal pills use the primary color; token phrase pills use purple and include tooltips explaining their matching behavior.

## Search behavior

Quoted text is serialized as `"..."` and matched with indexed case-sensitive `LIKE` against normalized search text. Backtick text is serialized as `` `...` `` and matched as adjacent ordered tokens using `hasAllTokens(...)` plus `hasSubstr(tokens(...), ...)`, which approximates `hasPhrase` on ClickHouse 26.2 while retaining an indexed prefilter.

This keeps copy-pasted exact text and punctuation-flexible phrase search as separate operators instead of overloading quotes with token-bag behavior.

## Validation

- `pnpm exec biome check apps/web/src/routes/_authenticated/projects/\$projectSlug/search/index.tsx`
- `pnpm exec vitest run packages/domain/spans/src/use-cases/parse-search-query.test.ts`
- Manual browser testing on `localhost:3000/projects/default-project/search`, including login via Mailpit and pasted mixed query syntax.

Attach screenshots or a short screen recording of the segmented search editor before requesting review.


apps/web/src/routes/_authenticated/projects/$projectSlug/search/index.tsx
```diff
@@ -409,35 +409,244 @@ function SearchPage() {
...
+        focusSegment(next.id)
+        return [next]
+      }
+
+      const next = current.filter((item) => item.id !== segment.id)
+      const focusTarget = next[Math.max(0, index - 1)] ?? next[0]
+      if (focusTarget) focusSegment(focusTarget.id)
+      return next
+    })
+  }
```
andresgutgon: Beautiful big nice logic. Maybe could be a hook with tests

csansoon (approved):

andresgutgon (approved):

PR #3107, opened 2026-05-13
feat(api): annotations + scores routes through defineApiEndpoint

neoxelox:
Migrates the two remaining old-shaped routes to `defineApiEndpoint` so they surface as MCP tools (`createAnnotation`, `createScore`) alongside the HTTP endpoints. The manifest grew from 16 → 18 tools and both routes gain the `medium` rate-limit tier (projects drops to `low`).

Behavior changes:

- Annotations created via OAuth callers now carry `annotatorId = auth.userId`; API-key callers stay unauthored (the use-case takes the value from the trusted route context, never the body).
- Mount-path constants (`annotationsPath`, `scoresPath`) declared once in Hono's `:param` form and reused for both the parent `routes.route(...)` and `defineApiEndpoint`. The factory normalizes to OpenAPI `{param}` form internally so the MCP dispatcher only has to substitute one placeholder shape.
- Plain (non-transformed) `traceIdSchema`, `spanIdSchema`, `sessionIdSchema` exposed from `apps/api/src/openapi/schemas.ts` so request / response bodies keep the domain's length validation without dragging the branded-ID transforms into JSON Schema — the MCP SDK can't represent transforms and was previously throwing.
- Request and response schemas for both routes are now declared explicitly field-by-field with `.describe()` on every property (instead of spreading the un-described domain shapes). Shared anchor field bag lets the metadata schema reuse the same five fields without duplication. Verified field-by-field that no top-level property was added or dropped relative to the previous OpenAPI spec.

### DIFF
--- a/apps/api/mcp.json
+++ b/apps/api/mcp.json
@@ -572,6 +572,929 @@
       }
     },
     {
+      "name": "createScore",
+      "title": "Create project score",
+      "description": "Creates a score against a target trace. The trace is resolved by explicit id (`trace.by = \"id\"`) or by a filter set (`trace.by = \"filters\"`, exactly one match required). Annotations use the separate `/annotations` endpoint.",
+      "inputSchema": {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "type": "object",
+        "properties": {
+          "projectSlug": {
+            "type": "string",
+            "description": "Project slug (human-readable identifier)"
+          },
+          "body": {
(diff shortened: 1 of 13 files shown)


csansoon (approved):

apps/api/mcp.json
andresgutgon, line 1: 🔥 🙏

andresgutgon (approved):

PR #3108, opened 2026-05-13
fix(issues): guard discovery creation with Postgres centroids

geclos:
## Summary

Adds a final Postgres centroid duplicate guard in the project-locked issue discovery no-match path. When Weaviate retrieval still returns no candidate, discovery now compares the incoming normalized score embedding against recent project issue centroids from Postgres and assigns to the best above-threshold match before creating a new issue.

## Why

The Redis serialization fix prevents concurrent brand-new issue writes, but it still depended on Weaviate read-after-write visibility. Under high-volume bursts of similar traces, the next serialized workflow can acquire the project lock before the previous issue projection is searchable, see another no-match, and create a duplicate issue. Postgres already has the canonical issue centroid immediately after creation, so this PR uses it as the final source-of-truth duplicate guard.

## Validation

- `pnpm --filter @domain/issues test -- serialize-issue-discovery.test.ts`
- `pnpm --filter @domain/issues typecheck`
- `pnpm --filter @platform/db-postgres typecheck`
- `pnpm --filter @domain/issues check`
- `pnpm --filter @platform/db-postgres check`
- commit hook: `pnpm format`, `pnpm knip`

### DIFF
--- a/packages/domain/issues/src/constants.ts
+++ b/packages/domain/issues/src/constants.ts
@@ -138,6 +138,16 @@ export const ISSUE_REFRESH_THROTTLE_MS = 8 * 60 * 60 * 1000
 export const MIN_OCCURRENCES_FOR_VISIBILITY = 3
 
 // ---------------------------------------------------------------------------
+// Postgres centroid duplicate guard
+// ---------------------------------------------------------------------------
+
+/** Maximum recent project issues compared by the Postgres centroid duplicate guard. */
+export const ISSUE_DISCOVERY_POSTGRES_CENTROID_GUARD_CANDIDATES = 250
+
+/** Minimum cosine similarity for assigning to an existing issue during locked no-match serialization. */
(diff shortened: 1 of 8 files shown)


packages/domain/issues/src/constants.ts
```diff
@@ -137,6 +137,16 @@ export const ISSUE_REFRESH_THROTTLE_MS = 8 * 60 * 60 * 1000
  */
 export const MIN_OCCURRENCES_FOR_VISIBILITY = 3
 
+// ---------------------------------------------------------------------------
+// Postgres centroid duplicate guard
+// ---------------------------------------------------------------------------
+
+/** Maximum recent project issues compared by the Postgres centroid duplicate guard. */
+export const ISSUE_DISCOVERY_POSTGRES_CENTROID_GUARD_CANDIDATES = 250
```
neoxelox, line 145: i mean, at this point i dont know why we using weaviate anymore xddd

neoxelox (approved):

PR #3110, opened 2026-05-13
feat(web): sessions section + OAuth keys settings page

neoxelox:
Ships M-Settings from the MCP/OAuth/API expansion plan, plus a small batch of related settings polish.

New surfaces:

- /settings/account gains a Sessions section above Delete Account. Lists the user's BA sessions with UA-parsed browser/OS/device, geoip city/region/country, last-active and expiration timestamps. Per-row revoke + a "Sign out everywhere else" action, both gated behind destructive confirmation modals modeled on the existing API-key delete modal. Current session is pinned to the top and its revoke control is hidden (handled by the regular sign-out flow).
- /settings/keys (renamed from /settings/api-keys) is now a combined page with two sections: an OAuth Keys table (one row per (client, user) for the active org, with a revoke action) on top, and the existing API Keys table underneath. /settings/api-keys is preserved as a redirect-only route so old bookmarks survive.

Backing pieces:

- apps/web/src/domains/sessions/user-sessions.functions.ts — listUserSessions / revokeUserSession / revokeAllOtherUserSessions built on BA's session APIs, with UA-parser + geoip resolution server- side so the client bundle stays lean. Rejects revoking the current session token to prevent accidental self-sign-out.
- apps/web/src/domains/oauth/oauth-keys.functions.ts + oauth-keys.collection.ts — listOAuthKeys (admin connection plus explicit organization_id filter, GROUP BY (client, user) with last activity from oauth_access_tokens.updated_at) and revokeOAuthKey (verify the app belongs to the caller's org, delete every access token for the pair, mark oauth_applications.disabled = true when no tokens remain for that client).
- packages/platform/db-postgres exports desc and max from drizzle-orm to keep the OAuth-keys list query off direct drizzle imports.

Related polish (touched while we were in the area):

- Move the light/dark theme switcher from the top navbar into the user-menu dropdown.
- Settings nav: "API Keys" -> "Keys".
- Light tightening on the members and billing settings pages.


apps/web/src/routes/_authenticated/settings.tsx
```diff
@@ -15,7 +15,7 @@ const navItems = [
   { label: "Organization", to: "/settings/organization", icon: Building2 },
   { label: "Members", to: "/settings/members", icon: Users },
   { label: "Billing", to: "/settings/billing", icon: CreditCard },
-  { label: "API Keys", to: "/settings/api-keys", icon: Key },
```
andresgutgon, line 18: Why? was more clear this way

neoxelox: it has oauth keys and api keys underneath

andresgutgon (dismissed):

andresgutgon (approved):

PR #3115, opened 2026-05-14
chore(infra): finalize encrypted production Aurora

geclos:
## Summary

Finalizes the production Aurora encryption migration in Pulumi after the live cutover completed.

The production RDS model now treats `latitude-production-aurora` as the normal managed Aurora Serverless cluster instead of a temporary encrypted migration cluster. It removes migration-specific outputs and resource fields, keeps the production cluster protected, and keeps aliases from the temporary migration resource names so Pulumi retains ownership of the existing encrypted cluster under the stable logical names.

## Context

Production was cut over from the old unencrypted Aurora cluster to the encrypted Aurora cluster during the maintenance window. The old cluster has been deleted, production secrets point at the encrypted cluster, and the app is running against it.

This PR only cleans up the infrastructure code so future changes do not have to reason about migration-era `encryptedMigration*` resources.

## Validation

- `cd infra && pnpm typecheck`
- `cd infra && pulumi preview --stack production` showed no changes after the cleanup apply
- Verified production health after apply:
  - `https://api.latitude.so/health`
  - `https://ingest.latitude.so/health`
  - `https://console.latitude.so/api/health`

### DIFF
--- a/infra/Pulumi.production.yaml
+++ b/infra/Pulumi.production.yaml
@@ -4,6 +4,8 @@ config:
   latitude-infra:domainName: latitude.so
   latitude-infra:githubOwner: latitude-dev
   latitude-infra:githubRepo: latitude-llm
+  # Pinned to avoid automatic bastion replacement when AWS publishes newer AL2023 AMIs.
+  latitude-infra:bastionAmiId: ami-00b105a12aa2a60eb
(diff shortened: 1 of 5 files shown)


geclos:
🐮 👦🏼 already released

PR #3120, opened 2026-05-14
Remove static AWS credential dependency

geclos:
## Summary
- remove app-level AWS access/secret key configuration for S3, CloudWatch metrics, static asset upload, Bedrock, and AWS KB integration
- rely on the AWS SDK/CLI default credential chain so ECS task roles can provide AWS permissions
- update environment examples, Helm values, docs, and Bedrock provider UI/tests

## Verification
- cd packages/core && pnpm tc
- cd apps/gateway && pnpm tc
- cd apps/web && pnpm tc

## Notes
- Scoped core test run failed during DB migration because the local test Postgres connection was refused (ECONNREFUSED).


apps/web/src/app/(private)/settings/_components/ProviderApiKeys/New/_components/Configuration/AmazonBedrock/index.tsx
```diff
@@ -4,31 +4,11 @@ import { buildConfigFieldName } from '../../../buildProviderPayload'
...
-          fieldNamespace: 'secretAccessKey',
-        })}
-        placeholder='sk-0dfdsn23bm4m23n4MfB'
-      />
       <Input
         required
         type='text'
         label='Region'
-        info='The Amazon Bedrock region'
+        info='Credentials are resolved through the AWS SDK default credential chain. In ECS, attach an IAM task role with Amazon Bedrock access.'
```
geclos: This change is not ok this is a feature that actually allows user to set their aws credentials for running inference via latitude, bring it back.

geclos: Good catch. Restored the Amazon Bedrock credential fields and provider validation/passing in 2a01cb7d2. The PR now only removes deployment/app-level AWS credentials, while preserving user-supplied Bedrock inference credentials.

PR #3125, opened 2026-05-14
feat(notifications): hover-mark-seen, header, link fallback, timestamps, histogram

csansoon:
Four improvements to the bell popover. The big one is hover-to-mark-seen; the other three are smaller fixes that landed together since they share the same surface.

## per-row mark-seen on hover

Opening the popover used to fire `markAllNotificationsSeen` and zero the unread badge for every row, including ones the user never looked at. Now:

- Hovering an unread row for **400ms** marks just that row seen (debounced, deduped per mount). Quick mouse-passes don't fire.
- Focusing a row via keyboard does the same.
- Clicking through to a notification's URL marks it seen synchronously (also covers touch devices, which don't fire `pointerenter`).
- The popover gained a header — title on the left, **"Mark all as read"** on the right, disabled when `unread === 0`. That's the explicit bulk action; opening alone no longer marks anything.

The optimistic update patches the cached list (`seenAt`) and decrements the unread count; `onSettled` invalidates both query keys to reconcile. The existing `openedAt` context survives: once a row is optimistically marked during the current open, `computeIsUnseen` keeps it visually highlighted until the popover closes — no mid-open flicker.

New backend pieces, all in `@domain/notifications` / `@platform/db-postgres`:

- `markNotificationSeenUseCase({ organizationId, userId, notificationId })`. Idempotent. The repo's WHERE clause requires `seen_at IS NULL` and matches `(id, organizationId, userId)` together, so cross-tenant / cross-user / re-mark calls are silent no-ops — no error, no over-write.
- `NotificationRepository.markSeen` port method, live SQL impl, fake impl.
- New `markNotificationSeen` POST server function.

## timestamps on every row

Three of the five renderer types (`issue.new`, `issue.regressed`, `issue.escalating`) didn't show a timestamp. The other two (`custom_message`, `wrapped_report`) had hand-rolled inline copies. Centralized `createdAt: Date` as a required prop on `BaseNotification`, rendered inline at the right edge of the title row, and dropped the inline duplicates.

## incident notifications missing their issue link

`buildIssueUrl` requires both `payload.projectSlug` and `payload.issueId`, but both are optional in `incidentNotificationPayloadSchema` — comment there explicitly says *"existing rows from before this field landed don't have them"*. Legacy rows + rows where the issue/project lookup failed at fan-out time render as non-interactive cards.

Fix: derive the same fields server-side from the alert_incident itself. The notification's `sourceId` is the `alert_incident.id`, and `alert_incident.sourceId` is the issue id (for `sourceType: "issue"`, which is all current incident kinds). New `resolveIncidentNotificationTarget` GET server fn does the walk; `useIncidentLinkFallback` hook fires only when the payload is incomplete (healthy rows never trigger the network call). `buildIssueUrl` and `useLiveIssueSummary` accept the resolved fallback so the issue name + lifecycle states also hydrate live.

## escalation histogram had two buckets

`getIncidentTrend` clamped the time window to `[incident.startedAt, now]`. For `event: "opened"` the center is `startedAt`, so `from = max(startedAt - 1d, startedAt) = startedAt` and the window collapsed to one day = exactly two 12h buckets.

The chart shows score occurrences (issue activity), not incident state. Pre-incident history is precisely what makes an "escalation" chart legible. Dropped the lower-bound clamp. Window is now ±1 day around the center, capped at `now` on the upper end → **4 buckets at 12h**.

## verified

- `pnpm --filter @domain/notifications test` — 11/11 pass. 4 new `markNotificationSeenUseCase` cases cover happy-path, idempotency on already-seen rows, and silent no-ops for cross-user + cross-org calls.
- `pnpm --filter @domain/notifications typecheck`, `--filter @platform/db-postgres typecheck`, `--filter @app/web typecheck` — clean (only the pre-existing `@latitude-data/telemetry` baseline noise from `packages/ai`).
- Manual browser verification not yet done — runbook covers each change against seeded `owner@acme.com`.

## known limitations / follow-ups

- No frontend integration test for the popover; hover/focus/click wiring is currently only covered by manual verification.
- No-URL notification cards on touch devices have no mark-seen affordance (no hover, no click target). Out of scope here.

## Test plan

- [x] Open the bell — badge does **not** drop on open.
- [x] Header shows "Notifications" + "Mark all as read"; button disabled when `unread === 0`.
- [x] Hover an unread row for >400ms → `markNotificationSeen` POST fires, badge decrements, blue dot stays visible.
- [x] Quick hover (<400ms) → no POST.
- [x] Close popover mid-debounce → no POST (timer cleared on unmount).
- [x] Click a row with a URL → navigation + `markNotificationSeen` POST.
- [x] Tab-focus a row → POST fires after debounce.
- [x] "Mark all as read" → bulk POST, badge → 0, button disabled.
- [x] Close + reopen → previously-hovered rows render seen.
- [x] Every notification (all 5 renderer types + parse-fail fallback) shows a relative timestamp.
- [x] Legacy-style incident notification (payload missing `issueId`/`projectSlug`, real `source_id`) → `resolveIncidentNotificationTarget` resolves, card becomes interactive, click navigates to `/projects/<slug>/issues?issueId=<id>`.
- [x] Modern incident notifications do **not** trigger `resolveIncidentNotificationTarget`.
- [x] `issue.escalating` (opened) notification's histogram shows 4 bars instead of 2; ClickHouse query returns `buckets` of length 4.


apps/web/src/domains/notifications/notifications.functions.ts
```diff
@@ -168,3 +195,59 @@ export const getIncidentTrend = createServerFn({ method: "GET" })
...
+ * landed) and rows whose fan-out lookup failed don't carry `issueId` /
+ * `projectSlug`, so the renderer can't build a link from the payload alone.
+ *
+ * The alert_incident row is authoritative: `incident.sourceId` is the issueId
+ * (for `sourceType: "issue"`, which is all current incident kinds) and
+ * `incident.projectId` is the project's id. Missing issues/projects (deleted
+ * since incident creation) surface as `null` fields — the renderer stays
+ * non-interactive in that case.
+ */
+export const resolveIncidentNotificationTarget = createServerFn({ method: "GET" })
```
andresgutgon: This method looks a mutation. Is confusing

apps/web/src/routes/_authenticated/-components/notifications/base-notification.tsx
```diff
@@ -58,6 +151,12 @@ function BaseNotificationContent({ seenAt, icon, title, description, children }:
           )
         ) : null}
         {children}
+        {/* Timestamp anchored at the bottom of the body column — separates
```
andresgutgon: Less japping comments

andresgutgon (dismissed):

apps/web/src/domains/notifications/notifications.functions.ts
csansoon: Good call — renamed to `getIncidentNotificationTarget` to match the file's existing `getIncidentTrend` / `getUnreadNotificationCount` / `listNotifications` pattern. Done in `dad85e7`.

apps/web/src/routes/_authenticated/-components/notifications/base-notification.tsx
csansoon: Fair — trimmed the block comments across this PR's new code. Kept one-liners only where the WHY is non-obvious. Done in `c53da49a`.

andresgutgon (approved):

PR #3128, opened 2026-05-14
Fix static asset S3 upload credentials

geclos:
## Summary
- restore AWS BuildKit secret mounts for the web Docker static asset upload step
- export the mounted credentials only for the upload command so static asset publishing keeps working without baking credentials into the image

## Validation
- git diff --cached --check before commit


geclos:
🐮 👦🏼 fixing latitude-v1 deployments

PR #3136, opened 2026-05-14
MCP docs

neoxelox:


docs/llms.txt
andresgutgon, line 1: So how is generated this now?

andresgutgon (approved):

docs/llms.txt
neoxelox: it was always generated (i think), because check /llms-full.txt, it does render and it is not harcoded, lets see

andresgutgon: But will `/llms.txt` be automatically generated?

PR #3138, opened 2026-05-14
fix(telemetry): LAT-581 pass LLM SDK modules via instrumentations

andresgutgon:
## Summary

- `instrumentations` on `new Latitude(...)` / `registerLatitudeInstrumentations(...)` now takes a plain object mapping integration name → LLM SDK module the consumer imports — e.g. `{ openai: OpenAI, anthropic: AnthropicSDK }`. Replaces the magic-string-array form, which silently failed in modern apps because the SDK's internal `require` resolved the CJS build of dual-bundled LLM SDKs while the consumer app loaded the ESM build (the patch landed on a class the app never invoked).
- `InstrumentationsInput` is typed as `Partial<Record<InstrumentationName, object | undefined>>` so unknown keys and primitive values (`true`, `42`, `"openai"`, …) are rejected at compile time. Non-object input throws at register time. Compile-time `@ts-expect-error` assertions in the test file lock the type behavior in place.
- Removed `tryRequire`, the legacy `LEGACY_INSTRUMENTATION_MAP`, and the `modules` option on `registerLatitudeInstrumentations`. All 18 example files migrated. Version bump to `3.0.0-alpha.11`.

Linear: [LAT-581](https://linear.app/latitude-dev/issue/LAT-581)

## Test plan

- [x] `pnpm --filter @latitude-data/telemetry typecheck` — passes
- [x] `pnpm --filter @latitude-data/telemetry test` — 115 tests passing
- [x] `pnpm --filter @latitude-data/telemetry check` — biome clean
- [x] `pnpm --filter @latitude-data/telemetry build` — build artifacts produced
- [x] `pnpm knip` — no unused exports / unlisted deps
- [x] Manual smoke test from \`examples/test_openai.ts\` against a local Latitude instance — \`@traceloop/instrumentation-openai\` spans land in the trace UI under the \`capture()\` envelope

### DIFF
--- a/packages/telemetry/typescript/examples/README.md
+++ b/packages/telemetry/typescript/examples/README.md
@@ -4,33 +4,33 @@ This directory contains examples demonstrating how to use the Latitude Telemetry
 
 ## Quick Start (Happy Path)
 
-The simplest way to use Latitude is with `LatitudeTelemetry` — no existing OpenTelemetry setup required.
+The simplest way to use Latitude is with `new Latitude({...})` — no existing OpenTelemetry setup required.
 
 ```typescript
-import { LatitudeTelemetry } from "@latitude-data/telemetry"
 import OpenAI from "openai"
+import { Latitude, capture } from "@latitude-data/telemetry"
 
-const telemetry = new LatitudeTelemetry(
-  process.env.LATITUDE_API_KEY!,
-  process.env.LATITUDE_PROJECT_SLUG!,
-  {
-    instrumentations: ["openai"], // Auto-instrument OpenAI
-  }
-)
+const latitude = new Latitude({
+  apiKey: process.env.LATITUDE_API_KEY!,
+  projectSlug: process.env.LATITUDE_PROJECT_SLUG!,
+  instrumentations: { openai: OpenAI }, // Pass the LLM SDK module you use in app code.
+})
 
+await latitude.ready
 const client = new OpenAI()
 
-await telemetry.capture(
-  { tags: ["production"], userId: "user-123" },
+await capture(
+  "chat-call",
   async () => {
     return await client.chat.completions.create({
       model: "gpt-4o-mini",
       messages: [{ role: "user", content: "Hello!" }],
     })
-  }
+  },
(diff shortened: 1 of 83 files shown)


packages/telemetry/typescript/examples/README.md
```diff
@@ -4,33 +4,33 @@ This directory contains examples demonstrating how to use the Latitude Telemetry
...
-const telemetry = new LatitudeTelemetry(
-  process.env.LATITUDE_API_KEY!,
-  process.env.LATITUDE_PROJECT_SLUG!,
-  {
-    instrumentations: ["openai"], // Auto-instrument OpenAI
-  }
-)
+const latitude = new Latitude({
+  apiKey: process.env.LATITUDE_API_KEY!,
+  projectSlug: process.env.LATITUDE_PROJECT_SLUG!,
```
geclos, line 15: wait you didn't change it to project al final?

geclos (changes requested):
missing updating the docs

packages/telemetry/typescript/examples/README.md
csansoon: xd

packages/telemetry/typescript/examples/README.md
```diff
@@ -143,7 +144,7 @@ const provider = new NodeTracerProvider({
 })
 
 await registerLatitudeInstrumentations({
-  instrumentations: ["openai"],
+  instrumentations: { openai: OpenAI },
   tracerProvider: provider,
 })
```
csansoon, line 149: I get you're updating the examples package, but we do have documentation in docs/ about it, don't we? If so, that needs to be updated too.

packages/telemetry/typescript/examples/README.md
andresgutgon: 3 reasons. 
1. Pereza
2. I think is better this way, more clear
3. Also it was already this way before my changes

geclos (dismissed):
missing updating the docs

csansoon (approved):

PR #3152, opened 2026-05-15
feat(notifications): multi-channel delivery pipeline (in-app + email)

csansoon:
## Summary

Reshapes the in-app notification system into a kind + group + channel pipeline. In-app and email delivery now share one producer → creator → channel-worker flow; adding a future channel (Slack, SMS) is a new queue + worker + renderer registry with no source-side changes. The rich Wrapped email is migrated through the same pipeline, so users get one rich email per report (respecting per-user opt-out) instead of the previous "rich-email + teaser" double-send that would have shipped when the notifications flag flipped on.

The system was behind the `"notifications"` feature flag with no real users on it, so the data model was reshaped freely. No data migration is needed beyond the schema rewrite.

## New shape

**Kinds** are now a flat enum (`incident.opened`, `incident.closed`, `wrapped.report`, `custom.message`). Each kind has its payload schema and a group via `NOTIFICATION_KIND_META` in `@domain/notifications`.

**Groups** (`incidents`, `wrapped_reports`, `custom_messages`) live in `@domain/shared` so the `User` entity can carry typed preferences without circular deps. The user-prefs settings page renders one switch per group from `NOTIFICATION_GROUPS` — new groups appear there automatically.

**Per-user prefs** are stored on `users.notification_preferences` (jsonb). Missing entries default to opt-in. Today only email exists as a channel; the shape supports more without a migration.

**Project-level gates** moved from `projectSettings.alertNotifications` (incidents-only, with `escalationSensitivity` bundled in) to `projectSettings.notifications.incidents.<alertKind>` + a separate `projectSettings.escalation.sensitivity`. The umbrella mirrors the user-prefs structure so new groups can slot in. The leaf granularity at the project level differs by group — for incidents, that's per-`AlertIncidentKind` because users want to mute noisy alert types per project.

**Project anchor** (`notifications.project_id`, nullable) ties a notification to a project. Three jobs:
- **Cascade**: `ProjectDeleted` → `notifications:delete-by-project` removes every anchored row.
- **Bell footer**: the in-app row shows the project's live name (via `useProjectsCollection`) next to the relative time.
- **Email rendering**: the email worker looks up the project once and threads `{ id, name, slug }` into the renderer's context.

No FK constraint (per `database-postgres` skill); a partial index on `(organization_id, project_id) WHERE project_id IS NOT NULL` keeps the cascade query cheap.

**Idempotency** is owned by a single `idempotency_key` column on `notifications`. Producer computes `\`\${kind}:\${entityId}\`` for kinds with a natural source (incident/wrapped) or `\`\${kind}:\${generateId()}\`` for one-shot kinds. Unique index on `(organization_id, user_id, idempotency_key)`. The producer-side queue dedupe keys, the unique-index conflict path, and the email worker's `emailed_at IS NULL` claim absorb at-least-once redelivery across the three layers.

## Pipeline

```
source event (IncidentCreated/Closed, wrapped runForProject, ...)
  → notifications:request-{incident,wrapped-report}-notifications
       gate (incidents only) + recipient resolution + payload snapshot
  → notifications:create-notification (one per recipient)
       insertIfAbsent + (if prefs allow) publish notification-email:send
  → notification-email:send
       markEmailed claim → render via per-kind renderer Effect → send

ProjectDeleted
  → notifications:delete-by-project
       DELETE WHERE organization_id = $1 AND project_id = $2
```

Per-kind email renderers are `Effect`s — they declare service deps via Effect's `R` channel and yield whatever they need server-side. The `wrapped.report` renderer yields `WrappedReportRepository.findById` and renders the rich React Email layout (the existing versioned `TEMPLATE_BY_TYPE_VERSION` dispatcher is preserved). Payload-only kinds (incidents, custom) stay trivial: `Effect.tryPromise(() => buildHtml(...))`. The email worker provides the union of all renderers' service deps as one `rendererLayer`; adding a kind that needs new repos = write the renderer + one line in `notification-emailer.ts`.

## Wrapped email migration

Previously, `runWrappedUseCase` resolved recipients and sent rich emails inline, then the wrapped worker also published a teaser via the notification pipeline. After this PR:

- `runWrappedUseCase` shrinks to "compute report → persist". Returns `{ status, reportId, projectName }` (no more `recipientCount`; that's the pipeline's job).
- `runForProject` worker just runs the use case and publishes the wrapped notification request.
- The rich email layout moved under `templates/notifications/wrapped-report/`. The renderer fetches the report row via `WrappedReportRepository` and produces the same HTML as before.

Side effects of the move:
- Users with `wrapped_reports.email = false` no longer receive the rich email. Correct outcome — that's what prefs are for. No real users affected since notifications is feature-flagged off.
- `runForProject` retries no longer duplicate emails (the `emailed_at IS NULL` claim absorbs them).
- SMTP failure inside the emailer is silently dropped (the `markEmailed` claim already happened) — design's stated trade-off of zero duplicates over zero misses.

## Notable subsystems

**Domain events**: `IncidentCreated`/`Closed` route to the producer task; the previously-no-op `ProjectDeleted` handler routes to the cascade task. No source-side changes when adding new channels.

**Frontend renderers**: dispatch on the flat `kind`; `incident.opened` + `incident.closed` share the `IncidentNotification` component which passes `event` down to sub-renderers (`issue-new`, `issue-regressed`, `issue-escalating`). `BaseNotification` accepts `projectId` and renders the live project name (`noWrap` + `ellipsis`) next to the relative time.

**Settings UIs**:
- `/settings/account` gains an "Email notifications" section — one toggle per group, label/description from `NOTIFICATION_GROUP_META`.
- `/projects/<slug>/settings` reads/writes the new `notifications.incidents` + `escalation.sensitivity` paths; the visible UI is unchanged.

**API surface**: `apps/api/src/routes/projects.ts` splits the old `AlertNotificationsSettingSchema` into `IncidentNotificationsSettingSchema` + `NotificationsSettingSchema` + `EscalationSettingSchema`. `openapi.json` + `mcp.json` regenerated.

## Documentation

- `dev-docs/notifications.md` — durable reference (concepts, pipeline, file index, naming conventions, "Adding a new kind / group / channel" walkthroughs, anti-patterns).
- `specs/notifications-multi-channel.md` — design spec (decisions, trade-offs).
- `.agents/skills/notifications/SKILL.md` — agent-facing how-to.
- The skill is registered in the glossary in `AGENTS.md`.

## Verification

- Full workspace `pnpm typecheck` green.
- Domain test suites: `@domain/notifications` (17), `@domain/spans` (582), `@app/workers` (68) — all pass.
- `pnpm db:reset && pnpm seed` runs cleanly; the seed produces 945 `incident.opened` + 14 `incident.closed` rows across the seeded org members with the new shape. Re-running the seed leaves row counts unchanged (unique-index dedupe verified).

End-to-end Mailpit verification (trigger a wrapped run via the backoffice button at `/backoffice/projects/<id>`, confirm exactly one rich email per opted-in org member at `localhost:8025`, confirm the bell row shows up with the project name in the footer) is the remaining manual check before flipping the feature flag.

## Out of scope (tracked follow-ups)

- React Email polish for the incident + custom-message templates (still skeleton HTML).
- Per-(user, project, kind) subscription filtering — `resolveRecipients` keeps the seam.
- Slack / other channels — architecture supports them; not implemented.
- Email digesting / batching — one email per notification, real-time.


packages/domain/email/src/templates/notifications/custom-message/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const recipient = ctx.recipient.name ?? "there"
+  const body = payload.content ?? ""
+  const link = payload.link
+    ? payload.link.startsWith("http")
+      ? payload.link
+      : `${ctx.webAppUrl}${payload.link}`
+    : undefined
+  const subject = `[Latitude] ${payload.title}`
+  const text = `Hi ${recipient},\n\n${body}${link ? `\n\n${link}` : ""}\n\n— Latitude`
+  const html = `<p>Hi ${escapeHtml(recipient)},</p>${body ? `<p>${escapeHtml(body)}</p>` : ""}${link ? `<p><a href="${link}">Open</a></p>` : ""}<p>— Latitude</p>`
```
andresgutgon: Use real react HTML

andresgutgon (approved):

PR #3180, opened 2026-05-18
fix(alerts): collapse eventful incidents to a single point in time

csansoon:
Today every `alert_incidents` row is created with `ended_at = NULL`, even kinds that never close. That conflates two states:

- `issue.escalating` while open → genuinely "still running, end unknown"
- `issue.new` / `issue.regressed` → a single event that just happened

The rendering layer worked around it by keeping a hard-coded `RANGED_KINDS = { "issue.escalating" }` set and branching everywhere on `kind`. Adding a new eventful or lifecycle kind meant touching both the data and the renderer in lock-step.

## what changes

Eventful kinds (`issue.new`, `issue.regressed`) are now created with `ended_at = started_at`. After this PR, `ended_at` alone tells you the shape:

- `ended_at IS NULL` → open lifecycle incident (only `issue.escalating` while active)
- `ended_at > started_at` → closed lifecycle incident
- `ended_at = started_at` → point-in-time event

The decision lives in the creation use-case (`endedAt = kind === "issue.escalating" ? null : occurredAt`); no per-kind set or registry. The `closeAlertIncidentFromIssueEventUseCase` still only runs for `issue.escalating`, so eventful rows are never accidentally "closed" twice — they were never close-able in the first place.

Bonus: the partial index `alert_incidents_open_by_kind_idx ON kind WHERE ended_at IS NULL` (used by the hourly escalation sweep) now contains only genuinely open `issue.escalating` rows instead of every historical eventful incident.

## frontend rendering

`incident-markers.ts` drops `RANGED_KINDS` in favor of an `isRangedIncident(incident)` helper that looks only at timestamps. The tooltip "ongoing" / "→ end" / single-time branches use the same predicate. Adding a new eventful or lifecycle kind no longer requires editing the renderer.

## related bug: stale `isRegressed`

While in the area: `isRegressedExpr` previously checked only that an `issue.regressed` incident exists, so an issue that went `resolved → regressed → resolved again` kept showing as regressed forever. It now also gates on `issues.resolved_at IS NULL`. The historical incident row stays in the table for audit; only the derived lifecycle flag clears.

## migration

A custom Drizzle migration (`20260518120913_backfill-ended-at-on-eventful-incidents`) backfills existing rows:

```sql
UPDATE "latitude"."alert_incidents"
SET "ended_at" = "started_at"
WHERE "kind" IN ('issue.new', 'issue.regressed')
  AND "ended_at" IS NULL;
```

Idempotent — the `IS NULL` predicate makes reruns a no-op. The column stays nullable because open `issue.escalating` still legitimately needs `NULL`.

## verification

- `packages/domain/alerts` — added coverage that `issue.escalating` keeps `endedAt: null` and that eventful kinds mirror it onto `startedAt`
- `packages/platform/db-postgres` — added a test that `isRegressed` clears when `resolved_at` is set; existing escalating-lifecycle tests still pass
- `@app/web` typecheck clean

### DIFF
--- a/apps/web/src/domains/alerts/incident-markers.ts
+++ b/apps/web/src/domains/alerts/incident-markers.ts
@@ -20,8 +20,8 @@ type TopSymbol = NonNullable<BarChartOverlayLine["topSymbol"]>
 const KIND_TOP_SYMBOL: Record<AlertIncidentKind, TopSymbol> = {
   "issue.new": { shape: "circle", size: 7 },
   "issue.regressed": { shape: "diamond", size: 8 },
-  // Escalating renders as an area, but we still render a tiny tick at the start so a 1-bucket
-  // escalation that snaps to a single cell stays visible.
+  // Escalating typically renders as an area, but we still render a tiny tick at the start so a
+  // 1-bucket escalation that snaps to a single cell stays visible.
   "issue.escalating": { shape: "rect", size: 6 },
 }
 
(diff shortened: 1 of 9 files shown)


andresgutgon (approved):

geclos:
> After this PR, ended_at alone tells you the shape:
> ended_at IS NULL → open lifecycle incident (only issue.escalating while active)
> ended_at > started_at → closed lifecycle incident
> ended_at = started_at → point-in-time event

I understand that the started_at time of an incident is just its created_at no?

PR #3193, opened 2026-05-18
fix(deps): upgrade protobufjs to 8.0.2

geclos:
Upgrades the workspace protobufjs resolution to 8.0.2 and adds a pnpm override so transitive consumers resolve to the patched major version. This removes the vulnerable 7.x and 8.0.1 protobufjs lockfile entries reported by Dependabot.

Validated with:
- `pnpm why protobufjs --recursive`
- `pnpm --filter @domain/spans typecheck`
- `pnpm --filter @domain/spans test`


geclos:
checked against changelog and nothing looks offensive, safe upgrade

PR #3196, opened 2026-05-18
fix(ui)+perf(issues): expandable large markdown + skip ClickHouse on empty projects

geclos:
Two independent improvements bundled in one PR.

## 1. `fix(ui)`: render large markdown content as expandable head + tail

- The previous fallback for content above `LARGE_MARKDOWN_CONTENT_THRESHOLD` (20k chars) rendered the entire message as plain text inside a `<pre>`, which looked out of place in the trace conversation drawer.
- Now the head (~6k chars) and tail (~2k chars) render as proper Markdown via the same `prose` wrapper used by normal-sized content, with a clickable pill separator in between that expands the middle on demand.
- Slice boundaries snap to the nearest newline within ±400 chars so we don't cut Markdown constructs (lists, fences, headings) mid-syntax.
- Threshold itself (`LARGE_MARKDOWN_CONTENT_THRESHOLD`) is unchanged — only the fallback rendering differs.

## 2. `perf(issues)`: short-circuit empty projects + loading empty state

- `listIssuesUseCase` was hitting `ScoreAnalyticsRepository` and `TraceRepository` (ClickHouse) on every call, even when the project had zero issues, only to assemble a zeroed-out analytics payload.
- Now we probe with a cheap `issueRepository.list({ limit: 1 })` first and, when empty, return a scaffolded empty result without touching ClickHouse. New unit test asserts ClickHouse paths and `traceRepository.countByProjectId` are never called in that case.
- On the web side, the Issues page used to render `null` during its initial fetch with no cached data, producing a blank flash before the empty state appeared. We now render `IssuesEmptyState` with a new `isLoading` flag so the layout stays stable.

## Test plan

- [x] `pnpm --filter @repo/ui typecheck` + full vitest suite (77/77)
- [x] `pnpm --filter @domain/issues typecheck` + `list-issues.test.ts` (12/12, including new test)
- [x] `pnpm --filter @app/web typecheck`
- [x] `biome check` on all 6 changed files
- [x] Manual QA (markdown): shipped a 27,583-char Markdown assistant message via the claude-code-telemetry CLI to the local stack; verified head/tail render as Markdown, pill expands to reveal the middle, code fences/tables/blockquotes render correctly.
- [ ] Manual QA (issues): visit a fresh project's Issues page and confirm the loading empty state appears, then transitions cleanly to the no-issues empty state without a blank flash.

### DIFF
--- a/apps/web/src/routes/_authenticated/projects/$projectSlug/issues/-components/issues-empty-state.tsx
+++ b/apps/web/src/routes/_authenticated/projects/$projectSlug/issues/-components/issues-empty-state.tsx
@@ -1,7 +1,7 @@
 import { Button, Icon, Text } from "@repo/ui"
 import { ExternalLinkIcon, SearchAlert } from "lucide-react"
 
-export function IssuesEmptyState() {
+export function IssuesEmptyState({ isLoading = false }: { readonly isLoading?: boolean }) {
   return (
     <div className="h-full w-full flex items-center justify-center p-8">
       <div className="max-w-lg flex flex-col items-center gap-6 text-center">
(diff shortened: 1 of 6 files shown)


geclos:
🐮 👦🏼

PR #3222, opened 2026-05-20
feat(sessions): materialization parity with traces (LAT-604)

andresgutgon:
## Summary

Brings `sessions_mv` to column-level parity with `traces_mv` so every session row carries the same kind of information a trace row does. This is the data-layer foundation that the rest of the [session-problems epic](../blob/development/specs/session-problems/0-problems.md) builds on (filter parity, session-level search, session detail panel, freshness-sorted results).

Scope of this PR is **only the ClickHouse schema/MV layer** — domain entity, repository, filter-registry and tests are tracked as PR 2 in the spec's task checklist.

### Two semantic shifts ship together

- **`duration_ns` redefinition.** Was a wall-clock `Int64 ALIAS` (`max_end_time - min_start_time`). Now a real `SimpleAggregateFunction(sum, Int64)` storing sum of root-span durations across the session's traces. Distinguishes a 2-day session that worked 5 min/day from a 2-day session that worked 12 h/day. Wall-clock stays trivially recoverable on the fly. See spec §"On `duration_ns` semantics".
- **Orphan-trace-as-session.** Drops the `WHERE session_id != ''` filter and groups on `coalesce(nullIf(session_id, ''), toString(trace_id))`. Every trace surfaces as a session (orphans as 1-trace sessions). This is the precondition for session-paginated listings to cover every trace per the [core constraint](../blob/development/specs/session-problems/0-problems.md#core-constraint--pagination-is-by-session-not-by-trace).

### Other columns added

`time_of_first_token` + sentinel-aware `time_to_first_token_ns` ALIAS, `root_span_id`, `root_span_name`, `input_messages` / `last_input_messages` / `output_messages` / `system_instructions`, `retention_days`. Session TTL aligned with traces (`retention_days + 30` grace).

Backfill is **forward-only**: pre-migration session rows read new columns as 0/empty/sentinel until they ingest a new span. Matches the trace-search posture.

## Test plan

- [x] `pnpm --filter @platform/db-clickhouse ch:up` applies 00016 in <500ms
- [x] `pnpm --filter @platform/db-clickhouse test` passes (133/133)
- [x] `pnpm --filter @domain/spans --filter @platform/db-clickhouse typecheck` clean
- [x] Smoke test against live CH (`org_check` project) verified:
  - [x] Orphan span (empty `session_id`) → session row keyed on `toString(trace_id)` with `trace_count = 1`
  - [x] Real session keyed on supplied `gen_ai.conversation.id`
  - [x] `sum(duration_ns)` ≈ 10s vs wall-clock ≈ 1h05s for a 2-trace session with idle gap (~360× divergence)
  - [x] TTFT sentinel: 0 when no first-token instrumented, positive when it is
  - [x] `root_span_name` captured via `argMinIfMerge`
- [x] Reviewer to double-check `schema.sql` patch is acceptable in lieu of full regen — `ch:schema:dump` reintroduces the CH 26.2 `text` index from migration 00012 that chdb can't parse and that was hand-stripped from the prior dump. PR 2 (or a separate cleanup) should harden `dump-schema.sh` to sanitize chdb-incompatible features

## Follow-up

PR 2 — read path consumers — is tracked in the spec checklist at `specs/session-problems/1-parity-traces-sessions.md`. It covers the domain entity widening, `LIST_SELECT` / `DETAIL_SELECT` / `findBySessionId`, filter registry (`ttft`, `name`), and the new `session-repository.test.ts`.

### DIFF
--- a/packages/platform/db-clickhouse/clickhouse/migrations/clustered/00016_session_parity.sql
+++ b/packages/platform/db-clickhouse/clickhouse/migrations/clustered/00016_session_parity.sql
@@ -1,0 +1,160 @@
+-- +goose NO TRANSACTION
+-- +goose Up
+
+-- ═══════════════════════════════════════════════════════════
+-- Sessions parity with traces (LAT-604) — clustered variant.
+-- See unclustered/00016_session_parity.sql for full notes.
+-- ═══════════════════════════════════════════════════════════
+
+ALTER TABLE sessions ON CLUSTER default
+    DROP COLUMN IF EXISTS duration_ns;
+
+ALTER TABLE sessions ON CLUSTER default
+    ADD COLUMN IF NOT EXISTS duration_ns
+        SimpleAggregateFunction(sum, Int64) DEFAULT 0 CODEC(T64, ZSTD(1)) AFTER max_end_time,
+    ADD COLUMN IF NOT EXISTS time_of_first_token
+        SimpleAggregateFunction(min, DateTime64(9, 'UTC')) CODEC(Delta(8), ZSTD(1)) AFTER duration_ns,
+    ADD COLUMN IF NOT EXISTS time_to_first_token_ns Int64 ALIAS if(
+        time_of_first_token < toDateTime64('2261-01-01', 9, 'UTC'),
+        reinterpretAsInt64(time_of_first_token) - reinterpretAsInt64(min_start_time),
+        0
+    ) AFTER time_of_first_token,
+    ADD COLUMN IF NOT EXISTS root_span_id
+        AggregateFunction(argMinIf, FixedString(16), DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(1)) AFTER simulation_id,
+    ADD COLUMN IF NOT EXISTS root_span_name
+        AggregateFunction(argMinIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(1)) AFTER root_span_id,
+    ADD COLUMN IF NOT EXISTS input_messages
+        AggregateFunction(argMinIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(3)) AFTER root_span_name,
+    ADD COLUMN IF NOT EXISTS last_input_messages
+        AggregateFunction(argMaxIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(3)) AFTER input_messages,
+    ADD COLUMN IF NOT EXISTS output_messages
+        AggregateFunction(argMaxIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(3)) AFTER last_input_messages,
+    ADD COLUMN IF NOT EXISTS system_instructions
+        AggregateFunction(argMinIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(3)) AFTER output_messages,
+    ADD COLUMN IF NOT EXISTS retention_days
+        SimpleAggregateFunction(max, UInt16) DEFAULT 90 CODEC(T64, ZSTD(1));
+
+ALTER TABLE sessions ON CLUSTER default
+    MODIFY TTL toDateTime(min_start_time) + toIntervalDay(retention_days + 30) DELETE;
+
+DROP VIEW IF EXISTS sessions_mv ON CLUSTER default;
+
+CREATE MATERIALIZED VIEW IF NOT EXISTS sessions_mv ON CLUSTER default TO sessions
+AS SELECT
+    s.organization_id,
+    s.project_id,
+    coalesce(nullIf(s.session_id, ''), toString(s.trace_id))         AS session_id,
+
+    uniqExactState(s.trace_id)                                       AS trace_count,
+    groupUniqArrayState(s.trace_id)                                  AS trace_ids,
+    count()                                                          AS span_count,
+    countIf(s.status_code = 2)                                       AS error_count,
+
+    min(s.start_time)                                                AS min_start_time,
+    max(s.end_time)                                                  AS max_end_time,
+
+    sum(if(
+        (s.parent_span_id = '' OR s.parent_span_id = '0000000000000000')
+            AND s.end_time > s.start_time,
+        reinterpretAsInt64(s.end_time) - reinterpretAsInt64(s.start_time),
+        toInt64(0)
+    ))                                                               AS duration_ns,
+
+    min(if(
+        s.time_to_first_token_ns > 0,
+        addNanoseconds(s.start_time, toInt64(s.time_to_first_token_ns)),
+        toDateTime64('2261-01-01 00:00:00.000000000', 9, 'UTC')
+    ))                                                               AS time_of_first_token,
+
+    sum(s.tokens_input)                                              AS tokens_input,
+    sum(s.tokens_output)                                             AS tokens_output,
+    sum(s.tokens_cache_read)                                         AS tokens_cache_read,
+    sum(s.tokens_cache_create)                                       AS tokens_cache_create,
+    sum(s.tokens_reasoning)                                          AS tokens_reasoning,
+    sum(s.tokens_total)                                              AS tokens_total,
+
+    sum(s.cost_input_microcents)                                     AS cost_input_microcents,
+    sum(s.cost_output_microcents)                                    AS cost_output_microcents,
+    sum(s.cost_total_microcents)                                     AS cost_total_microcents,
+
+    argMaxIfState(s.user_id, s.start_time, s.user_id != '')          AS user_id,
+    groupUniqArrayArray(s.tags)                                      AS tags,
+    maxMap(s.metadata)                                               AS metadata,
+    groupUniqArrayIfState(s.model, s.model != '')                    AS models,
+    groupUniqArrayIfState(s.provider, s.provider != '')              AS providers,
+    groupUniqArrayIfState(s.service_name, s.service_name != '')      AS service_names,
+    argMaxIfState(s.simulation_id, s.start_time, s.simulation_id != '') AS simulation_id,
+
+    argMinIfState(s.span_id, s.start_time,
+        s.parent_span_id = '' OR s.parent_span_id = '0000000000000000') AS root_span_id,
+    argMinIfState(s.name, s.start_time,
+        s.parent_span_id = '' OR s.parent_span_id = '0000000000000000') AS root_span_name,
+    argMinIfState(s.input_messages, s.start_time, s.input_messages != '')  AS input_messages,
+    argMaxIfState(s.input_messages, s.end_time, s.output_messages != '')   AS last_input_messages,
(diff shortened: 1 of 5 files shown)


geclos (dismissed):

andresgutgon:
@claude review, review also the doc to get full idea

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00016_session_parity.sql
```diff
@@ -0,0 +1,193 @@
...
+        s.parent_span_id = '' OR s.parent_span_id = '0000000000000000') AS root_span_name,
+    argMinIfState(s.input_messages, s.start_time, s.input_messages != '')  AS input_messages,
+    argMaxIfState(s.input_messages, s.end_time, s.output_messages != '')   AS last_input_messages,
+    argMaxIfState(s.output_messages, s.end_time, s.output_messages != '')  AS output_messages,
+    argMinIfState(s.system_instructions, s.start_time, s.system_instructions != '') AS system_instructions,
+
+    max(s.retention_days)                                            AS retention_days
+
+FROM spans AS s
+GROUP BY s.organization_id, s.project_id, session_id;
```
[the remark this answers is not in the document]
andresgutgon: Good catch — fixed in f8b0b6fc5. GROUP BY now repeats the explicit `coalesce(nullIf(s.session_id, ''), toString(s.trace_id))` expression instead of referencing the `session_id` SELECT alias, so the materialization is deterministic regardless of `prefer_column_name_to_alias`. Verified locally with two distinct orphan spans (different `trace_id`s, both `session_id = ''`) — they split into two session rows as expected, and `SHOW CREATE TABLE sessions_mv` confirms the persisted MV body now carries the explicit expression in GROUP BY.

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00016_session_parity.sql
```diff
@@ -0,0 +1,160 @@
...
+        s.parent_span_id = '' OR s.parent_span_id = '0000000000000000') AS root_span_name,
+    argMinIfState(s.input_messages, s.start_time, s.input_messages != '')  AS input_messages,
+    argMaxIfState(s.input_messages, s.end_time, s.output_messages != '')   AS last_input_messages,
+    argMaxIfState(s.output_messages, s.end_time, s.output_messages != '')  AS output_messages,
+    argMinIfState(s.system_instructions, s.start_time, s.system_instructions != '') AS system_instructions,
+
+    max(s.retention_days)                                            AS retention_days
+
+FROM spans AS s
+GROUP BY s.organization_id, s.project_id, session_id;
```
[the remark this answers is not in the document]
andresgutgon: Good catch — fixed in f8b0b6fc5. GROUP BY now repeats the explicit `coalesce(nullIf(s.session_id, ''), toString(s.trace_id))` expression instead of referencing the `session_id` SELECT alias, so the materialization is deterministic regardless of `prefer_column_name_to_alias`. Verified locally with two distinct orphan spans (different `trace_id`s, both `session_id = ''`) — they split into two session rows as expected, and `SHOW CREATE TABLE sessions_mv` confirms the persisted MV body now carries the explicit expression in GROUP BY.

packages/platform/testkit/src/clickhouse/schema.sql
```diff
@@ -144,13 +164,19 @@ AS SELECT
...
+    argMaxIfState(s.output_messages, s.end_time, s.output_messages != '') AS output_messages,
+    argMinIfState(s.system_instructions, s.start_time, s.system_instructions != '') AS system_instructions,
+    max(s.retention_days) AS retention_days
 FROM spans AS s
-WHERE s.session_id != ''
 GROUP BY
     s.organization_id,
     s.project_id,
-    s.session_id;
+    session_id;
```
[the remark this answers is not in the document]
andresgutgon: Good catch — fixed in f8b0b6fc5. GROUP BY now repeats the explicit `coalesce(nullIf(s.session_id, ''), toString(s.trace_id))` expression instead of referencing the `session_id` SELECT alias, so the materialization is deterministic regardless of `prefer_column_name_to_alias`. Verified locally with two distinct orphan spans (different `trace_id`s, both `session_id = ''`) — they split into two session rows as expected, and `SHOW CREATE TABLE sessions_mv` confirms the persisted MV body now carries the explicit expression in GROUP BY.

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00016_session_parity.sql
```diff
@@ -0,0 +1,160 @@
...
+ALTER TABLE sessions ON CLUSTER default
+    MODIFY TTL toDateTime(min_start_time) + toIntervalDay(retention_days + 30) DELETE;
+
+DROP VIEW IF EXISTS sessions_mv ON CLUSTER default;
+
+CREATE MATERIALIZED VIEW IF NOT EXISTS sessions_mv ON CLUSTER default TO sessions
+AS SELECT
+    s.organization_id,
+    s.project_id,
+    coalesce(nullIf(s.session_id, ''), toString(s.trace_id))         AS session_id,
```
csansoon, line 46: This could conflict.
There's no OTEL standard for session_id, so there might be some cases where a span has a `session_id` that is the same as another span's `trace_id`. Not likely, sure, but possible.

Maybe it's so unlikely that we can ignore it, but worth a look?

packages/platform/testkit/src/clickhouse/schema.sql
```diff
@@ -83,6 +85,13 @@ CREATE TABLE sessions
     `providers` AggregateFunction(groupUniqArrayIf, String, UInt8) CODEC(ZSTD(1)),
     `service_names` AggregateFunction(groupUniqArrayIf, String, UInt8) CODEC(ZSTD(1)),
     `simulation_id` AggregateFunction(argMaxIf, FixedString(24), DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(1)),
+    `root_span_id` AggregateFunction(argMinIf, FixedString(16), DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(1)),
+    `root_span_name` AggregateFunction(argMinIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(1)),
+    `input_messages` AggregateFunction(argMinIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(3)),
+    `last_input_messages` AggregateFunction(argMaxIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(3)),
+    `output_messages` AggregateFunction(argMaxIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(3)),
+    `system_instructions` AggregateFunction(argMinIf, String, DateTime64(9, 'UTC'), UInt8) CODEC(ZSTD(3)),
```
csansoon, line 93: I get that we're storing now the "messages in a session", but have in mind that this is only "the messages FROM THE LAST INTERACTION of the session". Some sessions might have context reduced, so messages are lost in the way, and these payloads will NOT contain them.

Still worth to show this as "session's messages", if we're doing Sessions UI, but Search and other features must still keep searching over TRACE messages, not SESSION messages.

csansoon (approved):

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00016_session_parity.sql
andresgutgon: Good flag — this is explicitly addressed in the spec at `1-parity-traces-sessions.md` §"On orphan-trace sessions":

> **No risk of collision with real session ids.** Real `session_id`s are arbitrary tenant-supplied strings (commonly UUIDs, short slugs, or database ids). A 32-char lowercase hex trace_id matches the `FixedString(32)` format only by coincidence; in practice tenants don't pick OTel trace ids as their conversation ids. If a collision ever happens, the two rows merge — same `(org, project, session_id)` key — which is a no-op for every aggregation in the MV.

Bounded blast radius:
- A collision requires a tenant-supplied `session_id` that is *exactly* 32 lowercase hex chars (matching OTLP trace_id format). Common patterns — UUIDs with dashes, ObjectIds (24 chars), DB ints, kebab-case slugs — can't collide.
- When it does happen, the `AggregatingMergeTree` engine merges the two rows on `(org, project, session_id)`. All aggregations (`sum`, `min`, `max`, `argMax`, `groupUniqArrayIf`) are well-defined under merge; nothing corrupts.
- The visible artifact is one orphan trace ending up nested inside an unrelated real session in the UI. That's not great, but it's contained to *that one tenant's* tenant-controlled namespace.

If we wanted to make it impossible, the cheapest mitigation would be a prefix in the coalesce — `concat('trace:', toString(s.trace_id))` — but the spec explicitly argued against it because:
1. Every downstream surface that derives "this trace's session_id" (score writes in `./3-session-issue-dedup.md`, search rollup in `./2-session-level-search.md`) would have to know about the prefix too.
2. The UI navigation pattern (`trace_ids[0]` to open the underlying trace from a session row) becomes a tiny bit more roundabout.
3. `'trace:'` is itself a magic prefix — a paranoid tenant could still pick it.

My read: accept and document. Happy to revisit in a follow-up if you'd rather we mitigate — happy to be wrong here.

packages/platform/testkit/src/clickhouse/schema.sql
andresgutgon: Yep, this is a real and known limitation — thanks for putting it in writing on the PR. The aggregate semantics are exactly what you describe:

- `input_messages`        = `argMinIfState(input_messages, start_time, …)`         → the **first** span with input (session opener).
- `last_input_messages`   = `argMaxIfState(input_messages, end_time, output_messages != '')` → input of the **last responsive** span.
- `output_messages`       = `argMaxIfState(output_messages, end_time, …)`             → the **last** assistant response.
- `system_instructions`   = `argMinIfState(system_instructions, start_time, …)`       → the **first** non-empty system prompt.

So we get *four anchor points* across the session — not a continuous transcript. Context-reduction in the middle, tool calls between turns, intermediate assistant turns, etc. are all absent from these columns.

That's intentional per spec `1-parity-traces-sessions.md` (the table calls them out as "detail-only") and `0-problems.md` confirms the rendering strategy: the session detail panel's Conversation tab renders **one trace at a time** using the existing per-trace `<Conversation>` primitive — not a merged session-wide stream. So the panel doesn't try to reconstruct a continuous timeline from these aggregates; it uses them for header metadata ("session opened with X", "current state is Y").

For search — agreed, and the spec is aligned with what you're saying:
- `./2-session-level-search.md` (referenced from `0-problems.md`) keeps the **rollup** at session grain but the **per-row match data** comes from per-trace search documents.
- `./5-search-highlights.md` notes highlights operate on session-level matches, but the underlying text comes from each constituent trace's payload, not from these session aggregate columns.

These four columns are explicitly **not** intended as a search corpus or a conversation reconstruction. If anyone tries to use `sessions.output_messages` for substring search later, that PR should fail review and point back at this thread.

PR #3225, opened 2026-05-20
Imp/remove title boxes

learningbizz:
Adding a textbox instead of premade categories, as after research we noticed we get richer titles if we let the user write them

image


geclos:
i hate this interaction couldn't we have some prefilled buttons and an other category with the input box?

geclos (approved):

PR #3229, opened 2026-05-20
fix(clickhouse): unblock 00016 against replicated cluster

andresgutgon:
## Summary

- Add `SYSTEM SYNC REPLICA … LIGHTWEIGHT` between dependent `ALTER`s in the clustered variant of `00016_session_parity.sql` to break the metadata-version race that was failing staging deploy with code 517.
- Add `max_start_time` column to `sessions` for "most recently active" sorting — populated as `max(spans.start_time)`, mirroring traces' use of `start_time` as the activity-recency signal.

## What was happening

Staging deploy of #3222 looped 20× retrying `00016_session_parity.sql` and exited with:

```
code: 517, message: Looks like this replica doesn't catchup with latest
ALTER query updates: metadata version on replica is N, while common
metadata is N+1. Please retry this query
```

The clustered variant of the migration fires three dependent `ALTER TABLE sessions ON CLUSTER default` statements back-to-back (`DROP COLUMN` → `ADD COLUMN …` → `MODIFY TTL`). Each `ALTER` increments Keeper's "common" metadata version for `sessions`. ClickHouse checks `replica_version == common_version` at **submission time** for the next `ALTER` — *before* `alter_sync=2`'s wait would kick in. So `ALTER #2` 517s because the local replica hasn't yet applied `#1`'s bump.

The existing retry loop in `up.sh` re-runs the whole file on 517, but `ADD COLUMN IF NOT EXISTS` still bumps the metadata version even when the column already exists. Each retry advanced `common` by ≥1 while the replica caught up by ~1 per retry window. The gap never closed — death spiral across all 20 attempts.

## Fix

**1. `SYSTEM SYNC REPLICA sessions ON CLUSTER default LIGHTWEIGHT;` between every dependent `ALTER`** (clustered variant Up + Down). Forces each replica's local metadata to match Keeper before the next `ALTER` is submitted, so the 517 check has no room to fire. The opening `SYNC` also drains any pending bumps left over from prior failed attempts in staging, making the file safe to re-run from its partial state.

`LIGHTWEIGHT` mode waits only for metadata + part-movement queue entries and skips merges, so `SYNC` stays fast even on a large/busy `sessions` table. It does not block readers or writers on the table — it only stalls this migration's own connection.

**2. New column: `max_start_time SimpleAggregateFunction(max, DateTime64(9, 'UTC'))`**, populated as `max(spans.start_time)` in `sessions_mv`. Sort by this `DESC` to surface "most recently active" sessions — mirrors how traces use `start_time` as the activity-recency signal (consistent with the existing `min_start_time` / `max_end_time` columns, all client-clock). Mirrored in `testkit/schema.sql`.

## Verified locally

Round-tripped on local unclustered ClickHouse (`pnpm --filter @platform/db-clickhouse ch:down && ch:up`). Live `DESCRIBE TABLE sessions` after re-apply shows `max_start_time` with `SimpleAggregateFunction(max, DateTime64(9, 'UTC'))` + `Delta(8), ZSTD(1)` codec, and all the rest of the parity columns intact. `SHOW CREATE TABLE sessions_mv` confirms `max(s.start_time) AS max_start_time` is in the projection.

Clustered path was not exercised locally (single-node dev CH), but is the variant containing the `SYSTEM SYNC REPLICA` changes — those will only take effect against the cluster on the next staging deploy.

## Test plan

- [ ] Re-trigger the staging migration task — `00016_session_parity.sql` completes without code 517 (the opening `SYNC` drains the lag from the prior failed attempts).
- [ ] Confirm production migration (first attempt against clean state) applies cleanly.
- [ ] Local dev: `pnpm --filter @platform/db-clickhouse ch:up` applies cleanly. ✅
- [ ] Testkit consumers that snapshot `schema.sql` continue to pass.


packages/platform/db-clickhouse/clickhouse/migrations/clustered/00016_session_parity.sql
```diff
@@ -106,11 +136,19 @@ GROUP BY
...
+SYSTEM SYNC REPLICA sessions ON CLUSTER default LIGHTWEIGHT;
+
 DROP VIEW IF EXISTS sessions_mv ON CLUSTER default;
 
+SYSTEM SYNC REPLICA sessions ON CLUSTER default LIGHTWEIGHT;
+
 ALTER TABLE sessions ON CLUSTER default
     REMOVE TTL;
 
+SYSTEM SYNC REPLICA sessions ON CLUSTER default LIGHTWEIGHT;
```
geclos, line 150: is this a long-lasting configuration change or just for the scope of this migration?

csansoon (approved):

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00016_session_parity.sql
andresgutgon: https://clickhouse.com/docs/sql-reference/statements/system#sync-replica

> table will be synced with other replicas in a cluster, but no more than `receive_timeout` seconds

PR #3230, opened 2026-05-20
fix(web): trace overlay reopening + bundled settings/navbar/annotation/org UX improvements

geclos:
Bundles several small, independent web/UX fixes and improvements that accumulated on this branch.

## Fixes

- **Trace overlay stuck selection** (`fix(issues)`): the issue-detail trace overlay tore itself down only from a CSS `transitionend` listener. When that event never fired (reduced motion, interrupted transition, backgrounded tab), `traceOverlayTraceId` stayed set, the row stayed active, and re-clicking the same trace was a no-op — with a single trace this fully blocked reopening. Replaced with a timer that fires after the slide-out, cleared on (re)open and unmount; also removes a latent risk where a child's bubbling transform transition could close the overlay early.
- **Settings form submit button loading state** (`fix(web)`): the submit button read `form.state.isSubmitting` non-reactively, so it never re-rendered when the flag flipped back to false after a save — it only cleared on the next toast-driven re-render. Wrapped the button in `form.Subscribe` across the organization, project, and account name forms. A follow-up commit restores this on the org name button after a later change reintroduced the non-reactive read.
- **Errored span name color** (`fix(web)`): in the trace detail drawer's span tree, the span name kept its default color when the span errored, unlike the icon, waterfall bar, and duration. Now colored destructive when `statusCode` is `"error"`.

## Improvements

- **Highlight active org/project in navbar dropdowns** (`feat(web)`): added a `selected` option to the shared `DropdownMenu` item (primary-muted background, primary label) and wired it into the organization and project switchers. Clicking the already-selected project no longer triggers a redundant redirect.
- **Identify the flagger behind automatic annotations** (`feat(annotations)`): flagger-created annotations were written as `sourceId: "SYSTEM"` and rendered with a generic "Latitude / Agent" badge, indistinguishable from each other. Now the originating flagger slug is persisted on the annotation score metadata (`annotationScoreMetadataSchema.flaggerSlug`) and threaded through the single `upsertFlaggerAnnotationScore` choke point used by both the deterministic (`process-flaggers`) and LLM (`save-flagger-annotation`) write paths. In the UI, such annotations show the flagger's name as a clickable badge deep-linking to its row in project settings (`settings/flaggers?flagger=<slug>`), where the target is scrolled into view and highlighted. Older rows without a slug keep the "Agent" badge.
- **Allow owners to delete empty organizations** (`feat(organizations)`): added a danger zone to organization settings letting an owner permanently delete an org, gated server-side on the caller being the owner and sole member (active org only). Deleting tears down its projects (soft-delete + `ProjectDeleted` cascade) before removing the org row, whose FK cascade drops members, invitations, and OAuth apps; the client then bounces to `/welcome`. Project teardown is extracted into a shared `purgeOrganizationProjectsUseCase` and also wired into account deletion, so `cleanupUserMembershipsUseCase` returns deleted sole-member org ids for the user-deletion worker to purge with a per-org-scoped `SqlClient`.

### DIFF
--- a/apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
+++ b/apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
@@ -72,9 +72,13 @@ function ProjectGeneralSettingsPage() {
           )}
         </form.Field>
         <div className="self-start">
-          <Button type="submit" isLoading={form.state.isSubmitting}>
-            Save
-          </Button>
+          <form.Subscribe selector={(state) => state.isSubmitting}>
+            {(isSubmitting) => (
+              <Button type="submit" isLoading={isSubmitting}>
+                Save
+              </Button>
+            )}
+          </form.Subscribe>
         </div>
       </form>
       <DeleteProjectSection projectId={currentProject.id} projectName={currentProject.name} />
--- a/apps/web/src/domains/organizations/organizations.functions.ts
+++ b/apps/web/src/domains/organizations/organizations.functions.ts
@@ -1,10 +1,12 @@
 import {
   generateUniqueOrganizationSlugUseCase,
+  MembershipRepository,
   OrganizationRepository,
   provisionOrganizationWorkspaceUseCase,
   updateOrganizationUseCase,
 } from "@domain/organizations"
-import { OrganizationId, UserId } from "@domain/shared"
+import { purgeOrganizationProjectsUseCase } from "@domain/projects"
+import { BadRequestError, ForbiddenError, OrganizationId, UserId } from "@domain/shared"
 import {
   ApiKeyRepositoryLive,
   MembershipRepositoryLive,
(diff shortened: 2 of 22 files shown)


apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
```diff
@@ -72,9 +72,13 @@ function ProjectGeneralSettingsPage() {
           )}
         </form.Field>
         <div className="self-start">
-          <Button type="submit" isLoading={form.state.isSubmitting}>
-            Save
-          </Button>
+          <form.Subscribe selector={(state) => state.isSubmitting}>
+            {(isSubmitting) => (
```
andresgutgon, line 76: Wasn't working the other way?

geclos: terribly slow we should avoid shipping this kind of xp, updates should be optimistic 99% of times

andresgutgon: Oh ok so state checking is slow in tanstack form.

andresgutgon (dismissed):

PR #3241, opened 2026-05-21
Add sandboxed evaluation script runtime

geclos:
## summary

Adds the hosted evaluation script runtime used by live evaluations. Evaluation scripts now run in an isolated worker-backed sandbox with Latitude-provided helpers such as `Passed`, `Failed`, `llm`, and `parse`, while the domain layer depends on the `EvaluationScriptRuntime` port instead of a concrete runtime.

The runtime keeps host/sandbox data transfer to plain JSON and the supported JSON Schema subset. `llm()` calls route through the Latitude-managed model configuration, and Zod stays outside the sandbox boundary.

## live evaluation execution

Live evaluation execution now uses the script runtime for hosted execution and persists successful, failed, and errored results through the existing canonical score write path. Runtime failures are wrapped as live evaluation execution errors so a bad script records an errored score instead of crashing the worker.

## resource limits

The first hosted runtime is bulkheaded for the current shared workers allocation:

- `live-evaluations:execute` uses BullMQ concurrency `4` per workers task
- `@platform/evaluation-runtime` allows at most `2` active sandbox workers per process
- extra work waits behind the runtime semaphore or in BullMQ instead of spawning unbounded worker threads

The docs call out the short-term ECS resource assumptions and the follow-up to split live evaluations into a dedicated worker service or remote runtime adapter.

## validation

Targeted tests cover:

- simple `Passed` / `Failed` script execution
- `llm()` bridge calls through the AI service with JSON Schema output
- sandbox-side `parse(value, schema)` validation
- blocked Node global access such as `process`
- infinite-loop termination
- live evaluation persistence and runtime-failure handling


packages/domain/evaluations/src/runtime/evaluation-execution.ts
```diff
@@ -143,90 +158,36 @@ export const toEvaluationExecutionResult = (result: EvaluationScriptExecution):
...
+      metadata: {
+        duration: 0,
+        usage: {
+          input: 0,
+          output: 0,
+          reasoning: 0,
+          cacheRead: 0,
+          cacheWrite: 0,
+        },
+        cost: 0,
```
geclos: why is this metadata hardcoded to 0

packages/domain/evaluations/src/runtime/evaluation-execution.ts
```diff
@@ -143,90 +158,36 @@ export const toEvaluationExecutionResult = (result: EvaluationScriptExecution):
...
+  }) {
+    yield* Effect.annotateCurrentSpan("evaluation.conversationMessageCount", input.conversation.length)
+
+    const runtime = yield* EvaluationScriptRuntime
+    const conversationText = formatEvaluationConversationForPrompt(input.conversation)
+
+    return yield* runtime.execute({
+      script: normalizeLegacyEvaluationScript(input.script),
+      conversation: input.conversation,
+      conversationText,
```
geclos: why are we introducing this conversationText if we are already passing the conversation

packages/domain/evaluations/src/runtime/json-schema.ts
```diff
@@ -0,0 +1,280 @@
...
+
+export const evaluationVerdictJsonSchema = {
+  type: "object",
+  additionalProperties: false,
+  required: ["passed", "feedback"],
+  properties: {
+    passed: { type: "boolean" },
+    feedback: { type: "string", minLength: 1 },
+  },
+} as const satisfies EvaluationJsonSchema
```
geclos, line 222: what is this absolutely crazy handcrafter parsing mechanism why can't we do simple JSON.parse ? JSON.stringify steps between comms boundaries between host and sandbox?

packages/platform/evaluation-runtime/src/worker-source.ts
```diff
@@ -0,0 +1,147 @@
...
+    pending.set(id, { resolve, reject })
+    parentPort.postMessage({ type: "request", id, method, payload })
+  })
+}
+
+function isRecord(value) {
+  return typeof value === "object" && value !== null && !Array.isArray(value)
+}
+
+function validateJsonSchema(value, schema, path = "$") {
```
geclos: again what is this crazyness it's not ok to manually parse / validate json

packages/platform/evaluation-runtime/src/worker-source.ts
```diff
@@ -0,0 +1,147 @@
+export const workerSource = String.raw`
+const { parentPort, workerData } = require("node:worker_threads")
```
geclos: move the script into a js file that is at least linted and typechecked and we can import as a string here, stringified code escapes all build time safety guards

packages/domain/evaluations/src/runtime/evaluation-execution.ts
geclos: Addressed in d254303ee. Live evaluation execution now passes trace-derived metadata (duration, usage, cost, turn count, trace/session/span/simulation ids) into the runtime; the zeroed metadata is only the fallback for alignment/optimization paths that currently hydrate conversations without trace accounting fields.

packages/domain/evaluations/src/runtime/evaluation-execution.ts
geclos: Addressed in d254303ee. `conversationText` no longer crosses the host/sandbox boundary; the runtime input only carries structured `conversation`, and the sandbox derives the `conversationText` global from it for scripts that use prompt-oriented judging.

packages/domain/evaluations/src/runtime/json-schema.ts
geclos: Addressed in d254303ee. Host/sandbox payloads now go through JSON stringify/parse at the message boundaries, and the domain-side schema parser is a Zod recursive schema instead of a handcrafted parser.

packages/platform/evaluation-runtime/src/worker-source.ts
geclos: Addressed in d254303ee. The sandbox no longer has handwritten JSON Schema validation; `parse()` and score result validation use `@cfworker/json-schema`, with values/schemas JSON-cloned before validation.

packages/platform/evaluation-runtime/src/worker-source.ts
geclos: Addressed in d254303ee. The worker source now lives in `sandbox-worker.ts` as typechecked/linted code that is imported as a source string, instead of a raw string literal that bypasses build-time checks.

geclos:
@claude review

geclos:
@cursoragent review

geclos:
Addressed the critical review points in `125213a47`: documented Node `vm` sandbox isolation limits, fixed boolean JSON Schema enum conversion, added compile-time syntax validation, added compile/LLM-call-limit tests, and clarified the runtime layer wiring/comments.

PR #3256, opened 2026-05-22
chore: add changelog convention to AGENTS.md and create CHANGELOG.md

guillemwilly:
Instructs coding agents to append a structured [area] [type] line to CHANGELOG.md after every user-facing change. Creates the file with an Unreleased section as the initial target.


AGENTS.md
```diff
@@ -14,6 +14,14 @@ At a glance: **`apps/*`** own HTTP boundaries (validation, authz, routing to use
...
+- **After completing any user-facing change, append one line to `CHANGELOG.md`** under the `## Unreleased` section (create the file and section if absent). Format:
+  ```
+  - [area] [type] Short description of the change
+  ```
+  **Area tags** (pick the closest): `evaluations` · `traces` · `sdk` · `annotations` · `platform` · `api` · `internal`
+  **Type tags**: `feat` · `fix` · `improvement` · `breaking`
+  Example: `- [evaluations] [feat] Add support for LLM-as-judge scoring templates`
+  Skip only for changes invisible to users: tests, CI config, tooling, docs-only.
```
andresgutgon: Tell it to add the date of the feature to be able to organize later. Also add the commit of the change or the PR

guillemwilly: done

andresgutgon (dismissed):

geclos (dismissed):
Instead of this we will create a skill for prod releases that updates the changelog on release before actually deploying to prod, this keeps it more honest with what is really implemented in production

geclos:
Addressed in 8fc7c509e: replaced the per-change changelog requirement with a production-release skill. The changelog is now updated at release time as a human-readable diff of what is being pushed to production since the previous production deploy, focused on the major aspects.

PR #3255, opened 2026-05-22
feat(web): trace-search-highlights endpoint (LAT-601 part 1/5)

andresgutgon:
## Summary

- First slice of **LAT-601** (search highlights). Adds the design spec ([`specs/session-problems/5-search-highlights.md`](../tree/LAT-601/highlights-backend-endpoint/specs/session-problems/5-search-highlights.md)) and the backend endpoint `getTraceSearchHighlights` that emits `search-literal` + `search-token` ranges for a given trace + query.
- New pure use-case `computeTraceSearchHighlights` in `@domain/spans` (walks `allMessages`, skips `role === "system"`, matches phrases against text parts only). Relocates `tokenizePhrase` to `@domain/spans/helpers` so the indexer and the new highlight use-case share one canonical tokenizer — no behavioral change to the lexical indexer.
- **Zero user-visible change.** No caller yet — PR 3 (frontend wiring) lights up the endpoint via a `useTraceSearchHighlights` React Query hook. `@public` JSDoc on the export keeps `pnpm knip` happy until then.

## Rollout context

Part 1 of 5 PRs under LAT-601. The spec's `## Rollout (PR sequence)` section documents the full sequence and dependency graph. After this lands, PR 2 (backend semantic-region) and PR 3 (frontend wiring) can proceed in parallel.

## Test plan

- [x] `pnpm --filter @domain/spans test` → 596/596 pass (9 new vitest cases covering literal/token/hybrid/system-filter/multi-part/non-text-parts/empty-query/semantic-only/empty-tokenization)
- [x] `pnpm --filter @domain/spans typecheck` → clean
- [x] `pnpm --filter @platform/db-clickhouse typecheck` → clean (`tokenizePhrase` relocation behaviorally identical)
- [x] `pnpm --filter web typecheck` → clean
- [x] `pnpm knip` → clean (exit 0)
- [x] Reviewer: confirm the endpoint contract in `apps/web/src/domains/traces/traces.functions.ts` matches the spec's "Single unified `getTraceSearchHighlights` endpoint" section.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3257, opened 2026-05-22
feat(search): semantic-region highlights pipeline (LAT-601 part 2/5)

andresgutgon:
## Summary

Second slice of **LAT-601** (search highlights). Builds the backend half that unlocks `search-semantic-region` highlights in the trace drawer — schema, ingest, read path, per-trace lookup, and a reusable use-case all land together. Existing search-listing callers don't opt into the new shape, so listing SQL is byte-identical.

**Depends on [#3255](https://github.com/latitude-dev/latitude-llm/pull/3255)** (PR 1). Branched off it, so until PR 1 merges this diff includes its commits — review can focus on the `860402b3b`/`97c59a2c3` two-commit delta on top.

### Schema — migration `00017`, unclustered + clustered

- `first_message_index Nullable(UInt32)`, `last_message_index Nullable(UInt32)` on `trace_search_embeddings`, `CODEC(T64, ZSTD(1))`.
- Nullable so pre-migration rows are distinguishable from rows that legitimately match at message 0 — no backfill needed; pre-migration coverage stays at "lexical only" until rows re-embed organically.

### Ingest — `@domain/spans/use-cases/build-trace-search-document.ts`

- `extractTurns` → `{ text, firstMessageIndex, lastMessageIndex }[]` indexed against the original `allMessages` (system rows included), so indices line up with the renderer.
- `selectHeadTailTurns` preserves the metadata when picking turns.
- `packChunks` emits union ranges across packed turns; sub-chunks of one oversized turn inherit that turn's full range — turns are atomic for the mapping.
- `TraceSearchChunk` + `TraceSearchEmbeddingRow` + live `upsertEmbedding` + the `trace-search` worker all carry the new fields.

### Read path — `@platform/db-clickhouse/repositories/search-plan.ts`

- `SearchPlan.semanticMetadata: boolean` (default `false`). When `true`, `buildSemanticSearchSubquery` adds `argMax(chunk_index/first_message_index/last_message_index, semantic_score)` rolled up alongside `max(semantic_score)`.
- All three plan branches (phrase-only / semantic-only / hybrid) project matched-* columns through with NULL-typed casts for shape parity. Hybrid `LEFT JOIN` uses `max(sem.matched_*)` — sem is already 1-per-trace so MAX is a no-op pass-through; lexical-only matches naturally return NULL.
- `planSearch(parsed, opts?)` accepts `opts.semanticMetadata?`. Default keeps the listing query plan byte-identical.

### Per-trace lookup — `TraceSearchRepository.findSemanticHighlightForTrace`

- Focused per-trace `argMax` query (no listing scan). Returns `{ chunkIndex, firstMessageIndex, lastMessageIndex, relevanceScore } | null` with nullable range columns on pre-migration chunks.

### Use-case — `@domain/spans/use-cases/get-trace-search-highlights.ts`

- `getTraceSearchHighlightsUseCase` owns the full orchestration: parse query → parallel `findByTraceId` + (embed via AI service + `findSemanticHighlightForTrace`) → `computeTraceSearchHighlights`.
- Pure Effect over domain ports, no HTTP/TanStack concerns. **Reusable by a future REST/MCP endpoint** with the same three layers — apps/web is a thin layer-providing wrapper.
- AI unavailable in dev/CI → semantic match drops to `null`, literal/token still flow.

### `computeTraceSearchHighlights` extended

- Optional `semanticMatch` input. Emits one `search-semantic-region` highlight per non-system message in `[firstMessageIndex, lastMessageIndex]`. NULL range columns → skip semantic-region. Match-nothing phrases suppress semantic-region too (the trace wouldn't be in search results).

### Handler — `apps/web/src/domains/traces/traces.functions.ts`

- Slimmed down to `requireSession()` + delegate to the use-case + provide layers. `Effect.orElseSucceed` to empty result for repository errors (decorative layer; shouldn't break the drawer).

## User-visible change

None. No caller of the search-listing path sets `semanticMetadata: true`, so listing SQL is byte-identical. The new endpoint sits unused until PR 3 / PR 5 wire it from the frontend.

## Test plan

- [x] `pnpm --filter @domain/spans test` → 613 / 613 (12 new cases: 5 chunk index threading + 7 semantic-region emission incl. block-per-message, system-row skipping, NULL-range fallback, literal composition, match-nothing suppression, semantic-only-no-match)
- [x] `pnpm --filter @platform/db-clickhouse test` → 154 / 154
- [x] `pnpm --filter @domain/spans typecheck` → clean
- [x] `pnpm --filter @platform/db-clickhouse typecheck` → clean
- [x] `pnpm --filter web typecheck` → clean
- [x] `pnpm --filter @app/workers typecheck` → clean
- [x] `pnpm knip` → exit 0
- [x] Reviewer: confirm the migration is safe — Nullable additive ALTER on `trace_search_embeddings`; no backfill needed.
- [ ] Manual on staging once merged: diff existing search-listing result IDs and `relevance_score` against a pre-PR baseline — must be identical (gated by the default-false flag).

### DIFF
--- a/apps/web/src/domains/traces/traces.functions.ts
+++ b/apps/web/src/domains/traces/traces.functions.ts
@@ -18,19 +18,18 @@ import type {
   TraceTimeHistogramBucket,
 } from "@domain/spans"
 import {
-  computeTraceSearchHighlights,
   getTraceCohortSummaryByTagsUseCase,
+  getTraceSearchHighlightsUseCase,
   mergeTraceHistogramTimeFilters,
-  parseSearchQuery,
   TraceRepository,
 } from "@domain/spans"
 import { withAi } from "@platform/ai"
 import { AIEmbedLive } from "@platform/ai-voyage"
 import { RedisCacheStoreLive } from "@platform/cache-redis"
-import { TraceRepositoryLive, withClickHouse } from "@platform/db-clickhouse"
+import { TraceRepositoryLive, TraceSearchRepositoryLive, withClickHouse } from "@platform/db-clickhouse"
 import { withTracing } from "@repo/observability"
 import { createServerFn } from "@tanstack/react-start"
-import { Effect } from "effect"
+import { Effect, Layer } from "effect"
 import type { GenAIMessage, GenAISystem } from "rosetta-ai"
 import { z } from "zod"
 import { enforceExportRequestRateLimit } from "../../domains/exports/export-rate-limit.ts"
@@ -355,8 +354,6 @@ export const getTraceTimeHistogramByProject = createServerFn({ method: "GET" })
     )
   })
 
-const EMPTY_HIGHLIGHTS_RESULT: TraceSearchHighlightsResult = { highlights: [], firstMatchIndex: -1 }
-
 /**
  * @knipignore
  */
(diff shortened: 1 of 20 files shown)


andresgutgon:
@claude do a review please

csansoon (approved):

PR #3283, opened 2026-05-26
feat(sessions): freshness-weighted ordering for session search

neoxelox:
## Summary

When you search sessions, results are now sorted by **relevance bucket → recency → session_id** instead of pure relevance. Within a 0.1-wide score bucket, recently-active sessions float above equally-relevant stale ones; relevance still dominates across bucket boundaries.

This fixes the "ageing project" pathology described in [`specs/session-problems/7-freshness-weighted-sort.md`](specs/session-problems/7-freshness-weighted-sort.md): a long-lived project accumulates high-scoring historical matches that crowd out fresh, live conversations on every query.

Concrete example: a search for `"refund"` against a project with
- a 0.87-relevance support thread from 18 months ago, and
- a 0.85-relevance customer conversation from this morning

now puts the live conversation first (both share bucket 0.8; freshness wins). Before, the 18-month-old artifact won.

## What changes

- New outer `SELECT` columns in the search query:
  - `relevance_bucket = floor(max(relevance_score) / bucketWidth) * bucketWidth`
  - `last_activity_at = least(coalesce(any(sf.sess_max_end_time), max(end_time)), now() + clockSkewMs)`
- New `session_freshness` CTE pre-aggregates `sessions.max_end_time` per matched session and joins via `LEFT JOIN ... USING (session_id)`. Orphan traces fall back to `max(trace_rollup.end_time)`.
- `ORDER BY` becomes `(relevance_bucket, last_activity_at, session_id) DESC`.
- `HAVING` cursor predicate matches the new tuple.
- New cursor shape `SessionSearchCursor { relevanceBucket, lastActivityAt, sessionId }`; `SessionListPage.nextCursor` widened to `SessionListCursor | SessionSearchCursor`.
- Wire validator (`sessions.functions.ts`) accepts either shape via a Zod union; React Query `initialPageParam` widened.
- Stale cursors of the wrong shape are silently dropped — pagination restarts from the top instead of producing a broken keyset comparison.

## Why session-level `max_end_time`, not matching-trace `end_time`

A live conversation whose only matching turn is from six months ago should still rank as fresh — the session is alive, the query just happened to hit an old quote. Using `sessions.max_end_time` captures that; using `max(trace_rollup.end_time)` would penalize it. Test #4 in the new suite exercises this exact scenario.

## Knobs

Two new constants in `@domain/spans/constants.ts`:
- `SESSION_SEARCH_RELEVANCE_BUCKET_WIDTH = 0.1` — wider favors freshness, narrower favors relevance
- `SESSION_SEARCH_MAX_CLOCK_SKEW_MS = 60 * 60 * 1000` — clamps future-dated spans (bad client clocks) to `now() + 1h`

### DIFF
--- a/packages/domain/spans/src/ports/session-repository.ts
+++ b/packages/domain/spans/src/ports/session-repository.ts
@@ -60,17 +60,33 @@ export interface SessionListCursor {
   readonly sessionId: string
 }
 
+/**
+ * Cursor shape used by the search-active list path, which sorts by
+ * `(relevance_bucket, last_activity_at, session_id) DESC`. The bucket
+ * snaps `best_score` into fixed-width tiers (see
+ * `SESSION_SEARCH_RELEVANCE_BUCKET_WIDTH`) so recently-active sessions
+ * float above stale ones of equivalent relevance. `lastActivityAt` is the
+ * session's `max_end_time` (clamped against client clock skew) serialized
+ * as an ISO-8601 string.
+ */
+export interface SessionSearchCursor {
+  readonly relevanceBucket: number
+  readonly lastActivityAt: string
+  readonly sessionId: string
+}
+
 export interface SessionListOptions {
   readonly limit?: number
-  readonly cursor?: SessionListCursor
+  readonly cursor?: SessionListCursor | SessionSearchCursor
   readonly sortBy?: string
   readonly sortDirection?: "asc" | "desc"
   readonly filters?: FilterSet
   /**
    * When present, the repository switches into the search code path: results
-   * are session-rollups of matching traces, ranking is forced to
-   * `bestScore DESC, sessionId DESC` regardless of `sortBy`, and the returned
-   * `SessionListPage` carries a parallel `searchMatches` map.
+   * are session-rollups of matching traces, ranking is forced to the
+   * freshness-weighted `(relevance_bucket, last_activity_at, sessionId) DESC`
+   * tuple regardless of `sortBy`, and the returned `SessionListPage` carries
+   * a parallel `searchMatches` map.
    */
   readonly searchQuery?: string
 }
@@ -78,7 +94,7 @@ export interface SessionListOptions {
 export interface SessionListPage {
   readonly items: readonly Session[]
   readonly hasMore: boolean
-  readonly nextCursor?: SessionListCursor
+  readonly nextCursor?: SessionListCursor | SessionSearchCursor
   /**
    * Per-result search match metadata, keyed by `sessionId`. Present only when
    * `SessionListOptions.searchQuery` was active for the request. Surfaced as
--- a/packages/domain/spans/src/ports/session-repository.ts
+++ b/packages/domain/spans/src/ports/session-repository.ts
@@ -60,15 +60,6 @@ export interface SessionListCursor {
   readonly sessionId: string
 }
 
-/**
- * Cursor shape used by the search-active list path, which sorts by
- * `(relevance_bucket, last_activity_at, session_id) DESC`. The bucket
- * snaps `best_score` into fixed-width tiers (see
- * `SESSION_SEARCH_RELEVANCE_BUCKET_WIDTH`) so recently-active sessions
- * float above stale ones of equivalent relevance. `lastActivityAt` is the
- * session's `max_end_time` (clamped against client clock skew) serialized
- * as an ISO-8601 string.
- */
 export interface SessionSearchCursor {
   readonly relevanceBucket: number
   readonly lastActivityAt: string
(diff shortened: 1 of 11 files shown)


packages/domain/spans/src/ports/session-repository.ts
```diff
@@ -60,25 +60,41 @@ export interface SessionListCursor {
...
 
+/**
+ * Cursor shape used by the search-active list path, which sorts by
+ * `(relevance_bucket, last_activity_at, session_id) DESC`. The bucket
+ * snaps `best_score` into fixed-width tiers (see
+ * `SESSION_SEARCH_RELEVANCE_BUCKET_WIDTH`) so recently-active sessions
+ * float above stale ones of equivalent relevance. `lastActivityAt` is the
+ * session's `max_end_time` (clamped against client clock skew) serialized
+ * as an ISO-8601 string.
+ */
```
andresgutgon: Remove slop comments

packages/platform/db-clickhouse/src/repositories/search-by-project.ts
```diff
@@ -274,13 +322,33 @@ export const listSessionsBySearchQuery = ({
...
+                    SELECT
+                      session_id,
+                      max(max_end_time) AS sess_max_end_time
+                    FROM sessions
+                    WHERE organization_id = {organizationId:String}
+                      AND project_id = {projectId:String}
+                      AND session_id IN (
+                        SELECT session_id FROM trace_rollup WHERE session_id != ''
+                      )
+                    GROUP BY session_id
```
andresgutgon, line 320: Now that you're here. Can you review the whole query and see if there is room for improvement?

andresgutgon (dismissed):

packages/domain/spans/src/ports/session-repository.ts
neoxelox: Trimmed in 7040af5d20d307578ea47102e2e6a7e111031dff — kept only the non-obvious bit on `searchQuery` (the `sortBy` override), dropped the rest.

packages/platform/db-clickhouse/src/repositories/search-by-project.ts
neoxelox: Did a pass — three concrete candidates, none applied yet, would like your call on each before pushing:

**1. Drop the dead `WHERE session_id != ''` in the `session_freshness` IN-clause.** `trace_rollup` already coalesces `session_id` to `toString(t.trace_id)` for orphans, so the column is never empty. Pure cleanup, zero behavior change.

**2. Reference the SELECT aliases (`relevance_bucket`, `last_activity_at`) in `HAVING` instead of repeating the expressions.** ClickHouse lets `HAVING` see SELECT aliases, so the bucket/freshness math can live in exactly one place. Removes the two template-string consts (`bucketExpr`, `lastActivityExpr`) — fewer moving parts, easier to read.

**3. Dedupe the `arrayMap(... arrayReverseSort(... groupArray(...)))` chain by materializing the sorted-and-sliced pairs once.** Today the same `arraySlice(arrayReverseSort(groupArray(...)))` runs in two `arrayMap`s back-to-back (one for ids, one for scores). Extract a `top_matching_pairs` alias and project both arrays from it. Performance-wise CH likely CSEs this already; the win is purely readability. Cleanest shape requires wrapping the aggregate in a `session_rollup` CTE so the helper alias doesn't leak into the JSONEachRow output — slightly more structure for less duplication.

Happy to apply any subset. (1) and (2) are basically free; (3) is the bigger restructure.

neoxelox: Applied (1) and (2) in 14f0ea41e72393d6760005fcb6751305093a60ce. (3) skipped — the structural change felt out of scope for this PR. Tests still green (166/166).

packages/platform/db-clickhouse/src/seeds/spans/freshness-sort-demo.ts
```diff
@@ -0,0 +1,238 @@
...
+      yield* insertJsonEachRow(ctx.client, "spans", spans)
+      yield* insertJsonEachRow(ctx.client, "trace_search_documents", docs)
+
+      if (!ctx.quiet) {
+        console.log(
+          `  -> spans/freshness-sort-demo: ${spans.length} sessions seeded (search "freshmark" to verify the freshness sort)`,
+        )
+      }
+    }),
+}
```
andresgutgon, line 238: Update https://github.com/latitude-dev/latitude-llm/blob/development/dev-docs/qa/session-search.md with this to tell user what to search for check freshness is working

neoxelox: Okay

andresgutgon (approved):

PR #3294, opened 2026-05-27
fix(db-postgres): isolate SqlClient transactions by fiber

geclos:
SqlClientLive tracked the active transaction in a closure shared by every fiber using the same SqlClient instance. Concurrent sibling fibers could therefore mistake another fiber's transaction for a nested transaction and reuse its connection, which can corrupt transaction-local RLS state when the first transaction completes.

This changes active transaction tracking to use Effect context. Nested transaction calls in the same fiber still reuse the current Postgres transaction, while concurrent sibling fibers open independent transactions. The SqlClient transaction type no longer advertises ConcurrentSqlTransactionError because concurrent sibling transactions are supported instead of rejected.

The db-postgres transaction tests now cover concurrent sibling transactions and verify each fiber keeps queries on its own transaction. db-postgres package tests run with file parallelism disabled to avoid PGlite migration hook contention, and the shared Vitest hook timeout has more headroom for PGlite/chdb startup under Turbo package concurrency.

Validation:
- pnpm check
- pnpm typecheck
- pnpm --filter @platform/db-postgres test
- pnpm --filter @app/api test

Note: I did not rerun the full test suite after the final targeted checks because it overloads local machines.

### DIFF
--- a/packages/platform/db-postgres/package.json
+++ b/packages/platform/db-postgres/package.json
@@ -26,7 +26,7 @@
     "check": "biome check src",
     "format": "biome format --write src && biome check --fix src",
     "typecheck": "tsgo -p tsconfig.json --noEmit",
-    "test": "vitest run --passWithNoTests --dir src"
+    "test": "vitest run --passWithNoTests --dir src --no-file-parallelism"
   },
   "dependencies": {
     "@better-auth/core": "catalog:",
--- a/packages/platform/db-postgres/package.json
+++ b/packages/platform/db-postgres/package.json
@@ -26,7 +26,7 @@
     "check": "biome check src",
     "format": "biome format --write src && biome check --fix src",
     "typecheck": "tsgo -p tsconfig.json --noEmit",
-    "test": "vitest run --passWithNoTests --dir src --no-file-parallelism"
+    "test": "vitest run --passWithNoTests --dir src"
   },
   "dependencies": {
     "@better-auth/core": "catalog:",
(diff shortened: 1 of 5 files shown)


packages/platform/db-postgres/package.json
```diff
@@ -26,7 +26,7 @@
     "check": "biome check src",
     "format": "biome format --write src && biome check --fix src",
     "typecheck": "tsgo -p tsconfig.json --noEmit",
-    "test": "vitest run --passWithNoTests --dir src"
+    "test": "vitest run --passWithNoTests --dir src --no-file-parallelism"
```
andresgutgon: Will not this make tests slower?

andresgutgon (dismissed):

packages/platform/db-postgres/src/sql-client.ts
```diff
@@ -46,97 +37,81 @@ export const SqlClientLive = (client: PostgresClient, organizationId: Organizati
...
+          Effect.withFiber((fiber) => {
+            if (fiber.context.mapUnsafe.has(ActiveSqlTransaction.key)) return effect
 
-            const txReady = new Promise<Operator>((resolve) => {
-              resolveTxReady = resolve
-            })
-            const effectDone = new Promise<{ ok: true; value: A } | { ok: false; error: unknown }>((resolve) => {
-              resolveEffectDone = resolve
-            })
+            return Effect.gen(function* () {
```
geclos, line 47: use effect.fn instead of effect.gen

PR #3297, opened 2026-05-27
Add flagger selection to onboarding

geclos:
## Summary

Adds a flagger selection step to project onboarding before telemetry setup. New users can choose which automatic monitors are enabled for the default project instead of starting with every flagger enabled by default.

The onboarding save path defensively provisions missing flagger rows before applying the selected enabled state. That keeps the flow safe when the background `ProjectCreated` provisioning worker has not run yet, and the existing `(organization_id, project_id, slug)` conflict handling prevents duplicate rows if both paths race. Provisioning remains insert-only on conflict, so later worker retries do not overwrite the user's choices.

https://www.loom.com/share/73635f10a1e14efa996a984de7287305

## Validation

- `pnpm --filter @app/web test -- src/domains/flaggers/flaggers.functions.ts`

Please attach a screenshot or short screen recording of the new onboarding step for reviewer context.

### DIFF
--- a/apps/web/src/domains/flaggers/flaggers.functions.ts
+++ b/apps/web/src/domains/flaggers/flaggers.functions.ts
@@ -1,4 +1,5 @@
 import {
+  configureProjectFlaggersForOnboardingUseCase,
   FLAGGER_STRATEGY_SLUGS,
   FlaggerRepository,
   type FlaggerSlug,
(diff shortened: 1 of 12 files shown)


geclos:
@claude review

geclos:
Addressed the N+1 onboarding flagger update path in `9345d3003`: added `FlaggerRepository.updateEnabledForProject`, implemented as a single SQL `UPDATE ... CASE ... RETURNING` over the project flaggers, and changed `configureProjectFlaggersForOnboarding` to provision + bulk-update inside one transaction with a single cache eviction after commit. The server function now returns `void` as the caller only invalidates/refetches.

geclos:
Addressed the remaining actionable items from the review in `591120895`:

- Fixed telemetry Back buttons to return to the flaggers step instead of stack selection.
- Typed preset slug arrays against `FLAGGER_STRATEGY_SLUGS` so stale/mistyped slugs fail typecheck.
- Made `FlaggerToggled.payload.sampling` required now that all emitters include the persisted sampling value.

I left the loading flash note as-is because the existing loading guard already prevents interaction/rendering of the cards until both flagger queries complete.

apps/web/src/domains/flaggers/flaggers.functions.ts
```diff
@@ -71,12 +95,73 @@ export const listFlaggersByProject = createServerFn({ method: "GET" })
...
+  )
+  .handler(async ({ data }): Promise<void> => {
+    const { organizationId, userId } = await requireSession()
+    const orgId = OrganizationId(organizationId)
+    const projectId = ProjectId(data.projectId)
+    const enabledSlugs = new Set<FlaggerSlug>(data.enabledSlugs)
+    const client = getPostgresClient()
+
+    await Effect.runPromise(
+      Effect.gen(function* () {
```
geclos: wrap in a domain use case, keep app boundaries thin wrappers

geclos: Addressed in `b70c03bb0`: moved the onboarding flagger provisioning + bulk enabled-state update + outbox writes + cache eviction into `configureProjectFlaggersForOnboardingUseCase` under `@domain/flaggers`. The web server function is now just the boundary wrapper: session/input coercion, layer provisioning, and tracing.

PR #3299, opened 2026-05-27
feat(wrapped): V2 report — token count, week-over-week deltas, leaderboard, and backoffice analytics redesign

csansoon:
## What this adds

This is the V2 Wrapped report schema and a full pass on the web/email renderers, the backoffice analytics page, and the OG image. V1 blobs and their frozen renderers are untouched — old share links keep working.

## V2 report schema

`reportVersion: 2` adds two fields to every new Wrapped run:

**`totals.tokensTotal`** — sum of `tokens_total` across all LLM spans for the window. Sourced from the existing ClickHouse `spans.tokens_total` materialized column; one extra `sum()` in `getTotalsForProject`.

**`lastReport`** — snapshot of the previous week's headline stats (sessions, toolCalls, durationMs, filesTouched, locWritten, tokensTotal), all independently nullable. Null means "not available" — first week, or the previous report was V1 and didn't carry that field. Populated at generation time in `run-wrapped.ts`: fetch the most recent prior report with `created_at ≤ now − 6 days` (the 6-day gap prevents same-day re-triggers from polluting the comparison), load the full record, pass it through `buildReportUseCase` → `assembleReport`.

## Web renderer (WrappedReportV2)

New renderer at `-components/claude-code/v2/WrappedReportV2.tsx`, dispatched when `reportVersion === 2`. Visually identical to V1 except:

- **Token headline** sits between the breadth strip and lines written, wrapped in a `#E8E4D8` rounded card so it reads as the featured stat. Shows: big token count → WoW delta ("30% more than last week") → leaderboard position.
- **Week-over-week deltas** on all stat cards (Sessions, Claude time, Files, Commands) as small `▲ 32%` / `▼ 12%` indicators. LOC written also gets a delta line. Shown in both public and member views — all delta data is public-safe.
- **Leaderboard position** (`#3 this week` for rank ≤ 10, `top 8% this week` for rank > 10) computed fresh on each page load via `getWrappedPageData`. Compares all reports created on the same UTC calendar day as this report, deduplicated to one per project, sorted by `tokensTotal ?? toolCalls`. Hidden when cohort < 5. The cohort is stable forever — it's always "same day as the report" not "today."

## Email renderer (ClaudeCodeWrappedEmailV2)

New email template with a standalone Token section (big number + WoW delta caption) placed before the Lines written section. Headline stats grid stays 4 cards (Session, Claude time, Files, Commands), each with a delta caption. No leaderboard in email.

## OG image

V2 reports replace the Lines written + Sessions two-stat row with a single `"29,841,572 tokens"` line in the same cream/semibold style as the archetype title. V1 reports keep the original layout.

## Backoffice analytics page

The right-side stats panel is replaced with a narrative four-section report:

1. **Summary** — 4 stat cards; window label shows just the date when all reports are from the same UTC day.
2. **This week's archetypes** — horizontal CSS bars sorted by count, with rank, conditional/fallback chip, count, %, and score strength (Strong/Moderate/Weak).
3. **How people work** — per-bucket comparison sorted by |drift|: "Expected 1% → Saw 21%, +20pp, Retune." One-sentence description per bucket.
4. **Personality confidence** — conditional gates as progress bars with "won/passed" cross-reference and an amber "never won" callout when a gate fires but always loses. Always-fires fallbacks as mini-cards with count and score strength.

The list column is now sorted by `tokensTotal ?? toolCalls` descending with a `#N` rank prefix per row. The "Tool calls" column is replaced with "Tokens" (null-safe for V1 rows).

Raw data (excess histograms, p10/p90 percentiles, score percentile tables, median signal values) stays in the JSON payload accessible via the Copy button — removed only from the UI.

## Seeds

`pnpm seed` now creates 100 V2 wrapped reports for the Acme org, all with `createdAt = today at 09:00 UTC`. This gives a valid 100-report leaderboard cohort for local testing: any seeded wrapped URL will show `#N this week`.

## Test plan

- [ ] Visit a seeded wrapped URL as logged-out → public layout, token card visible, leaderboard shows rank among 100 seed reports
- [ ] Visit as org member → same plus private section (Detailed report separator, workspaces, top command)
- [ ] Visit `/backoffice/wrapped` → 100 rows sorted by token count, narrative analytics panel renders all four sections
- [ ] Verify `▲ / ▼` deltas on stat cards, full-sentence delta on token headline and LOC section
- [ ] Check the OG image at `/wrapped/{id}/og/png` → shows token count, not lines/sessions
- [ ] Old V1 share URL → V1 renderer unchanged


csansoon:
Addressed all Copilot review comments in `77e175aacf1c4c245c14732c0ed809b1b415b0e7`:

- **lte → lt boundary** — range is now half-open `[dayStart, dayEnd)`; midnight-UTC reports fall into the next day's cohort only.
- **Redundant `as ReportV2` cast** — removed from seed.
- **Dead `?? "2026-05-22"` fallback** — removed.
- **Missing `findLeaderboardRankForReport` tests** — 4 integration tests added: multi-project ranking, rank-1 winner, per-project dedup (older report for same project is excluded and returns null), cross-day isolation. Also upgraded `sampleReport` to V2 shape to match `CURRENT_REPORT_VERSION = 2`.

csansoon:
Re: the `findLatestForProject` lower-bound concern (thread on `run-wrapped.ts:98`): the absence of a lower bound is intentional. The design goal is 'compare against the most recent prior run regardless of age' — if a project ran a Wrapped in January and then again today, we want the January numbers as the baseline so the delta is meaningful rather than silently absent. A lower bound (e.g. 'must be from within the last 8 days') was considered and rejected: it would produce null deltas after a gap, which is more confusing than a stale comparison. The 6-day upper bound already handles the same-day race condition.

andresgutgon (approved):

PR #3302, opened 2026-05-27
fix(flaggers): bind classifier feedback to annotations

geclos:
## What changed

LLM-backed flagger matches now carry the annotation feedback that should be saved for the matched trace. The classifier contract accepts `matched`, optional `feedback`, and optional `messageIndex`; `matched=true` is only accepted when feedback is present.

Before the workflow saves a matched annotation, a second adversarial AI reviewer checks the proposed annotation against the flagger guidance and the same evidence shown to the classifier. If the reviewer says the annotation does not make sense for the flagger/evidence, the match is dropped and no annotation is saved.

The flagger workflow passes approved classifier feedback into the existing draft annotation activity. When feedback is present, the draft step reuses it and only loads trace context for metadata instead of asking a second LLM to write a new annotation. This keeps the saved system annotation tied to the decision that triggered it.

## Why

The previous flow made one LLM call to classify a trace and a second call to draft the annotation. The second call could contradict the match decision, producing saved annotations like “No jailbreaking behavior detected...” for a matched jailbreaking score. That failure mode was cross-cutting for all LLM flaggers, not specific to jailbreak detection.

## Notes

The contradiction check is model-judged rather than phrase-matched in code, so it can catch category drift, unsupported annotations, and explicit contradictions without maintaining brittle string patterns.

The workflow still calls the draft activity after a match so billing, score id generation, and trace context loading stay in the existing activity boundary.

## Validation

- `pnpm --filter @domain/flaggers test -- run-flagger.test.ts draft-flagger-annotation.test.ts`
- `pnpm --filter @app/workflows test -- flagger-workflow.test.ts`
- `pnpm --filter @app/workflows typecheck`

### DIFF
--- a/packages/domain/flaggers/src/use-cases/run-flagger.ts
+++ b/packages/domain/flaggers/src/use-cases/run-flagger.ts
@@ -51,11 +51,6 @@ export interface ClassifyTraceForFlaggerInput {
   readonly strategyOverride?: FlaggerStrategy
 }
 
-const selfNegatingFeedbackPattern =
-  /\b(?:no\s+(?:[\w-]+\s+){0,6}(?:detected|found|present)|not\s+(?:a|an)?\s*(?:[\w-]+\s+){0,6}(?:issue|problem|violation|jailbreak|refusal|error)|does\s+not\s+(?:show|indicate|contain|demonstrate)|normal\s+(?:software\s+development\s+)?workflow|legitimate\s+(?:request|workflow|behavior)|correctly\s+(?:refused|handled))\b/i
-
-const isSelfNegatingFeedback = (feedback: string): boolean => selfNegatingFeedbackPattern.test(feedback)
-
 const baseFlaggerOutputSchema = z.object({
   matched: z.boolean().optional().default(false),
   feedback: z.string().min(1).nullable().optional(),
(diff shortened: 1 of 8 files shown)


packages/domain/flaggers/src/use-cases/run-flagger.ts
```diff
@@ -98,9 +157,9 @@ export const classifyTraceForFlaggerUseCase = Effect.fn("flaggers.classifyTraceF
     .generate({
       ...FLAGGER_MODEL,
       maxTokens: FLAGGER_MAX_TOKENS,
-      system: strategy.buildSystemPrompt!(input.trace),
+      system: `${strategy.buildSystemPrompt!(input.trace)}\n\n${FLAGGER_OUTPUT_CONTRACT}`,
```
geclos: centralise in buildSystemPropmt rather than concatenating here

packages/domain/flaggers/src/use-cases/run-flagger.ts
```diff
@@ -49,10 +51,67 @@ export interface ClassifyTraceForFlaggerInput {
   readonly strategyOverride?: FlaggerStrategy
 }
 
-const flaggerOutputSchema = z.object({
+const selfNegatingFeedbackPattern =
```
geclos: very brittle defense, what if instead we match the result with an adversarial AI reviewer that ensures the annotation makes sense?

packages/domain/flaggers/src/use-cases/run-flagger.ts
geclos: Addressed in `cdfe44565`: moved the classifier system prompt composition behind `buildClassificationSystemPrompt(...)` instead of concatenating inline at the generate call.

packages/domain/flaggers/src/use-cases/run-flagger.ts
geclos: Addressed in `cdfe44565`: removed the self-negating regex gate and replaced it with an adversarial AI annotation reviewer. The reviewer sees the flagger guidance, the classifier evidence, and the proposed annotation; if it rejects the annotation, the match is dropped and nothing is saved.

andresgutgon (approved):

PR #3304, opened 2026-05-27
fix(flaggers): break flagger-on-flagger recursion with no-reflag marker

geclos:
## Problem

The **\`latitude\`** project collects the telemetry emitted by Latitude's own production flaggers — every \`flagger.classify\` / \`flagger.draft\` LLM call lands there as a trace. We want to run flaggers *on that project* to catch quality issues in our production flaggers. But the meta-flaggers we'd run emit flagger traces too, which would be flagged again, and so on — an unbounded recursion (only dampened today by sampling + rate limits).

## Fix

Bound flagger-on-flagger to **exactly one level** using trace provenance. No queue/workflow payload changes — the decision is local to the flagger run, which already loads the trace and its tags.

- **New \`flagger:no-reflag\` telemetry tag** (\`packages/domain/ai/src/ai-generate-telemetry.ts\`).
- **Stamp on the way out** — when \`classifyTraceForFlaggerUseCase\` / \`annotateTraceForFlaggerUseCase\` run on a trace that is *itself* flagger-generated (carries \`flagger:classify\` / \`flagger:draft\`), the spawned LLM call's telemetry is tagged \`flagger:no-reflag\`.
- **Skip on the way in** — \`processFlaggersUseCase\` returns no decisions (and annotates \`flaggers.skipped = "reflag-suppressed"\`) for any trace carrying the marker.
- Provenance helpers live in \`packages/domain/flaggers/src/reflag.ts\`.

### How it breaks the loop

| Level | Tags | Flagged? |
| --- | --- | --- |
| **1** — production flagger trace | \`flagger:classify\` | ✅ yes (we keep tracking production flagger performance) |
| **2** — output of flagging a Level-1 trace | \`flagger:classify\` + \`flagger:no-reflag\` | ❌ skipped → loop closed |

## Tests

- \`reflag.test.ts\` — pure policy helpers, incl. an explicit "Level-2 is skipped" case.
- \`run-flagger.test.ts\` — asserts the no-reflag tag is appended when the flagged trace is flagger-generated.
- \`process-flaggers.test.ts\` — no-reflag trace skipped even on a deterministic match; first-level flagger trace still flagged.

\`@domain/flaggers\` typecheck + full suite (185 tests) green.

## Notes for review

- **Scope** of "is this a flagger trace" is the flagger tags only (\`flagger:classify\` / \`flagger:draft\`), matching the production scenario. If the \`latitude\` project should also collect eval/judge traces and treat them the same, it's a one-line change in \`FLAGGER_PROVENANCE_TAGS\`.
- **Where the skip lives**: the domain use-case (\`processFlaggers\`), single-sourced. A Level-2 trace still enqueues the \`deterministic-flaggers\` job, which loads the trace and no-ops (≤ Level-1 volume). If preferred, the same \`isReflagSuppressed\` guard can be added at the enqueue point in \`apps/workers/.../trace-end.ts\` to skip enqueuing entirely.

### DIFF
--- a/packages/domain/ai/src/ai-generate-telemetry.ts
+++ b/packages/domain/ai/src/ai-generate-telemetry.ts
@@ -9,6 +9,10 @@ export const AI_GENERATE_TELEMETRY_TAGS = {
   annotationEnrichPublication: ["annotation:enrichment"],
   flaggerClassify: ["flagger:classify"],
   flaggerDraft: ["flagger:draft"],
+  // Appended to a flagger's own LLM calls when the trace being flagged is itself
+  // flagger-generated. Marks the output so it is never re-flagged, bounding the
+  // flagger-on-flagger recursion to a single level. See @domain/flaggers reflag helpers.
+  flaggerNoReflag: ["flagger:no-reflag"],
   evaluationJudgeLive: ["eval:execute", "live"],
   evaluationJudgeAlignment: ["eval:execute", "alignment"],
   evaluationJudgeOptimization: ["eval:execute", "optimization"],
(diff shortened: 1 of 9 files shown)


geclos:
@claude review

PR #3315, opened 2026-05-28
Improve changelog popover interactions

geclos:
## Summary

Updates the in-app changelog popover so recent entries behave more like article links and the sidebar entry can signal newly published updates.

The popover now keeps expansion state at the list level, so moving from one changelog row to another collapses the previous row immediately instead of leaving the initially opened item expanded. Each recent entry links directly to its changelog article using the entry slug.

The sidebar “What’s new” trigger now shows a pulsing dot when any fetched changelog entry was published in the last 48 hours. This gives users a lightweight cue that there is fresh product news without changing the existing popover flow.

## Validation

- `pnpm exec biome check apps/web/src/routes/_authenticated/-components/changelog/whats-new-button.tsx`

### DIFF
--- a/apps/web/src/routes/_authenticated/-components/changelog/whats-new-button.tsx
+++ b/apps/web/src/routes/_authenticated/-components/changelog/whats-new-button.tsx
@@ -6,18 +6,37 @@ import { useChangelogEntries } from "../../../../domains/changelog/changelog.col
 import type { ChangelogEntryRecord } from "../../../../domains/changelog/changelog.functions.ts"
 
 const dateFormatter = new Intl.DateTimeFormat(undefined, { month: "short", day: "numeric", year: "numeric" })
+const RECENT_ENTRY_WINDOW_MS = 48 * 60 * 60 * 1000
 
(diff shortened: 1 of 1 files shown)


apps/web/src/routes/_authenticated/-components/changelog/whats-new-button.tsx
```diff
@@ -104,6 +132,15 @@ export function WhatsNewButton({ collapsed = false }: { collapsed?: boolean }) {
               What's new
             </Text.H5M>
           ) : null}
+          {hasRecentEntry ? (
+            <span
+              aria-hidden="true"
+              className={cn("h-2 w-2 rounded-full bg-primary", {
+                "absolute right-2 top-2": collapsed,
+                "shrink-0 animate-pulse": true,
```
geclos: i think we have a component for a pulsing dot?

geclos: Good catch — switched this to the shared `DotIndicator` with `ping` in `fd656414a`.

andresgutgon (approved):

PR #3318, opened 2026-05-28
feat(flaggers): rework configuration UX in onboarding and settings

csansoon:
Reworks how teams configure flaggers in both the onboarding flow and the project settings page. The two surfaces stay shape-distinct (onboarding = card grid, settings = grouped list) but now share use-case presets and the same semantic grouping of strategies.

## settings page

https://github.com/user-attachments/assets/e8fe0414-7192-49b7-beed-efc69d017969

Per-row autosave + edit modal is replaced with a buffered-edit model. Switches and inline sliders write to a local pending buffer keyed by flagger id. The page header — normally indistinguishable from any other settings page — becomes sticky and surfaces `N unsaved changes · Discard · Apply` on the right as soon as the buffer is non-empty. Apply fires one `updateFlaggerMutation` per dirty row in parallel and clears the buffer on success.

Other settings page changes:

- Flaggers are grouped into three sections: **Response validity** (deterministic, always-on, free), **User-side signals**, **Agent behavior**. Each section has a one-line subtitle that carries the free-vs-LLM context, replacing the per-row badge.
- Use-case preset chips (Support agent / Coding agent / etc.) appear above the list and bulk-edit the buffer while preserving sampling on rows that stay enabled. The chip whose `enabledSlugs` matches the current selection highlights as active.
- Each LLM row shows a compact cost line next to its slider — `30 credits per scan · runs on N% of eligible traces` — so users tuning sampling see the cost shape. Deterministic rows show `Free · Runs on 100% of eligible traces` in the same slot, no slider.
- `Cmd/Ctrl+S` applies, `Esc` discards. `useBlocker` plus `beforeunload` guard against losing unsaved changes on navigation or tab close.
- Deep-link `?flagger=<slug>` continues to scroll the targeted row into view; the highlight is a left-edge accent instead of a ring so it doesn't clip against the group card's \`overflow-hidden\`.

## onboarding

Same card grid, but cards now sort user-side → agent-side → programmatic so the easier-to-grasp behavioral categories lead. The use-case preset chip whose selection matches the current state highlights as active, fixing the previous "did clicking the chip do anything?" feedback gap. A small line under the page title tells users they can tune sampling rates later in settings, so they don't feel obliged to think about sampling here.

## new SettingsPage prop

\`SettingsPage\` gains an optional \`headerSticky\` boolean. When false (every other settings page), the header looks identical to before. When true, the header gets \`position: sticky\`, a background, and a bottom border. Flaggers is currently the only consumer.

## what's deferred

- **Absolute cost projection.** The cost line shows "30 credits per scan · runs on N%" but not "≈ N scans / month" — \`countTracesByProject\` only returns a total count, not a daily rate. A \`countTracesByProjectSinceDays\` endpoint + projection math is a clean follow-up.
- **Bulk update endpoint.** Apply uses N parallel \`updateFlaggerMutation\` calls. A \`updateProjectFlaggers\` mirror of \`configureProjectFlaggersForOnboarding\` would be cleaner but isn't load-bearing at 10 rows.

## verification

Walked through onboarding and settings manually — preset highlight, dirty buffer, Apply/Discard, keyboard shortcuts, navigation/tab-close blocker, deep link. Typecheck and biome clean.

### DIFF
--- a/apps/web/src/domains/flaggers/presets.ts
+++ b/apps/web/src/domains/flaggers/presets.ts
@@ -70,3 +70,34 @@ export const FLAGGER_USE_CASE_PRESETS = [
     enabledSlugs: ["nsfw", "jailbreaking", "refusal", "frustration", "empty-response"],
   },
 ] as const satisfies ReadonlyArray<FlaggerUseCasePreset>
+
+interface FlaggerGroup {
+  readonly id: string
+  readonly label: string
+  readonly description: string
+  readonly slugs: ReadonlyArray<FlaggerPresetSlug>
+}
+
+// Groups must collectively cover every FLAGGER_STRATEGY_SLUG. Add new strategies here when they ship.
+export const FLAGGER_GROUPS = [
+  {
+    id: "response-validity",
+    label: "Response validity",
+    description: "Free deterministic checks; always run on every trace.",
+    slugs: ["empty-response", "tool-call-errors", "output-schema-validation"],
+  },
+  {
+    id: "user-signals",
+    label: "User-side signals",
+    description: "LLM-based detection of risky or unhappy user behavior.",
+    slugs: ["frustration", "jailbreaking", "nsfw"],
+  },
+  {
+    id: "agent-behavior",
+    label: "Agent behavior",
+    description: "LLM-based detection of failure modes in the agent's own output.",
+    slugs: ["refusal", "laziness", "forgetting", "trashing"],
+  },
+] as const satisfies ReadonlyArray<FlaggerGroup>
+
+export const FLAGGER_DISPLAY_ORDER: ReadonlyArray<FlaggerPresetSlug> = FLAGGER_GROUPS.flatMap((group) => group.slugs)
--- a/apps/web/src/domains/flaggers/presets.ts
+++ b/apps/web/src/domains/flaggers/presets.ts
@@ -78,7 +78,6 @@ interface FlaggerGroup {
   readonly slugs: ReadonlyArray<FlaggerPresetSlug>
 }
 
-// Groups must collectively cover every FLAGGER_STRATEGY_SLUG. Add new strategies here when they ship.
 export const FLAGGER_GROUPS = [
   {
     id: "response-validity",
@@ -99,6 +98,15 @@ export const FLAGGER_GROUPS = [
     slugs: ["refusal", "laziness", "forgetting", "trashing"],
   },
 ] as const satisfies ReadonlyArray<FlaggerGroup>
+
+// Compile-time assertion: FLAGGER_GROUPS must cover every FLAGGER_STRATEGY_SLUG. If a new slug
+// is added but missing from any group, the type below resolves to the missing slug name(s)
+// instead of `true`, and the assignment fails typecheck with the missing slug surfaced in the
+// diagnostic. Keeps settings from silently dropping rows when a strategy ships.
+type _MissingFromFlaggerGroups = Exclude<FlaggerPresetSlug, (typeof FLAGGER_GROUPS)[number]["slugs"][number]>
+const _assertFlaggerGroupsExhaustive: [_MissingFromFlaggerGroups] extends [never] ? true : _MissingFromFlaggerGroups =
+  true
+void _assertFlaggerGroupsExhaustive
 
 // Onboarding sorts the flat card grid by user-side first, then agent-side, then deterministic
 // programmatic checks — easier-to-grasp categories lead so the user can scan and pick fast.
--- a/apps/web/src/domains/flaggers/presets.ts
+++ b/apps/web/src/domains/flaggers/presets.ts
@@ -100,4 +100,14 @@ export const FLAGGER_GROUPS = [
   },
 ] as const satisfies ReadonlyArray<FlaggerGroup>
 
-export const FLAGGER_DISPLAY_ORDER: ReadonlyArray<FlaggerPresetSlug> = FLAGGER_GROUPS.flatMap((group) => group.slugs)
+// Onboarding sorts the flat card grid by user-side first, then agent-side, then deterministic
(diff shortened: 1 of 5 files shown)


andresgutgon:
How looks in onboarding?

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
```diff
@@ -811,29 +736,41 @@ export function OnboardingFlow({
...
+                      <button
+                        key={preset.id}
+                        type="button"
+                        aria-pressed={isActive}
+                        onClick={() => applyFlaggerPreset(preset.enabledSlugs)}
+                        className={cn(
+                          "inline-flex cursor-pointer rounded-full border px-3 py-1.5 text-sm font-medium transition-colors",
+                          isActive
+                            ? "border-primary bg-primary-muted/40 text-primary"
+                            : "border-border bg-background text-foreground hover:border-primary/40 hover:bg-accent/10",
```
andresgutgon: Can this be a design system button?

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/-components/settings-page.tsx
```diff
@@ -26,14 +32,20 @@ export function SettingsPage({ title, description, actions, children }: Settings
...
-          {header}
-          <div className="shrink-0">{actions}</div>
-        </div>
-      ) : (
-        header
-      )}
+      <div
+        className={cn(
+          "flex flex-row items-center justify-between gap-4",
+          headerSticky && [
```
andresgutgon: Use `{}` syntax on `cn`

csansoon:
> How looks in onboarding?

Just two minor changes:
1. Flaggers are sorted so that similar ones are more close to each other
2. The selected Template is highlighted, instead of click-and-forget

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/flaggers.tsx
```diff
@@ -46,181 +39,272 @@ function ProjectFlaggersSettingsPage() {
...
+                            ) : (
+                              <div className="flex flex-row flex-wrap items-center gap-x-4 gap-y-2">
+                                <div className="flex flex-row items-center gap-3">
+                                  <div className="w-48">
+                                    <Slider
+                                      min={0}
+                                      max={100}
+                                      step={1}
+                                      value={[row.viewSampling]}
+                                      onValueChange={(values) => setRowChange(row.id, { sampling: values[0] ?? 0 })}
```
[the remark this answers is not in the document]
csansoon: Addressed in 331fe20ad. Added `aria-label={`Sampling rate for ${row.name}`}` on the Slider. The same commit also makes the disabled state visually distinct (50% opacity, cursor-not-allowed, muted range fill) by switching to `data-[disabled]:*` modifiers — Radix uses `data-disabled` on its span-based primitives, so the previous `disabled:*` modifiers were dead code.

apps/web/src/domains/flaggers/presets.ts
```diff
@@ -0,0 +1,113 @@
...
+
+interface FlaggerGroup {
+  readonly id: string
+  readonly label: string
+  readonly description: string
+  readonly slugs: ReadonlyArray<FlaggerPresetSlug>
+}
+
+// Groups must collectively cover every FLAGGER_STRATEGY_SLUG. Add new strategies here when they ship.
+export const FLAGGER_GROUPS = [
```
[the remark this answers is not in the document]
csansoon: Addressed in 0f4e78c9f. Replaced the comment-only contract with a compile-time assertion: a derived `_MissingFromFlaggerGroups` type that resolves to `never` (assignable to `true`) when groups are exhaustive, and to the missing slug name(s) otherwise. Verified by dropping a slug and watching tsc fail with the missing slug name in the diagnostic.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/flaggers.tsx
```diff
@@ -46,181 +39,272 @@ function ProjectFlaggersSettingsPage() {
...
+        event.preventDefault()
+        void applyRef.current()
+      } else if (event.key === "Escape" && !inField) {
+        event.preventDefault()
+        setPending({})
+      }
+    }
+    window.addEventListener("keydown", handler)
+    return () => window.removeEventListener("keydown", handler)
+  }, [hasDirty])
```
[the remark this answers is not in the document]
csansoon: Addressed in 0f4e78c9f. Added an `isApplyingRef` and gated the Escape handler on `!isApplyingRef.current`, so Escape is a no-op while apply is in flight. `Cmd/Ctrl+S` already short-circuits inside `apply()` via `if (!hasDirty || isApplying) return`, so no extra guard there.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
csansoon: Addressed in a577b475e. Replaced the chip with `<Button variant={isActive ? "default-soft" : "outline"} size="sm" aria-pressed={isActive}>`, in both onboarding and the new settings preset row so they stay consistent. The `default-soft` variant already renders `bg-primary-muted` + primary text — exactly the active state I was hand-rolling — and `outline` covers the inactive state. The pill (rounded-full) shape goes away in favor of the DS `rounded-lg`; happy to revisit if a chip-shaped variant is preferred.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/-components/settings-page.tsx
csansoon: Done in 43e5bfd17.

andresgutgon (approved):

PR #3319, opened 2026-05-28
feat(onboarding): optional Slack step with channel routing

csansoon:
## summary

Adds an optional Slack step to the project onboarding wizard, between **flaggers** and **telemetry**. After choosing which Flaggers should monitor their traces, users can connect their Slack workspace and route the resulting incident alerts to a channel without leaving the flow. The step is fully skippable; the rest of onboarding behaves as before.

The new step is hidden when either the `slack` feature flag is off or `LAT_SLACK_CLIENT_ID` / `LAT_SLACK_CLIENT_SECRET` are unset (self-hosted deployments without a Slack app). Channel routing for all three notification groups (`incidents`, `wrapped_reports`, `custom_messages`) still lives at `/settings/integrations` — the onboarding step is a thin gateway, not a replacement.

https://github.com/user-attachments/assets/4f4a0e20-160c-4ed8-9020-c02f8f0bc337

## new step UX

Three visible states, all using the same card primitive as the settings page so users build one mental model:

1. **Disconnected** — \"Connect Slack\" button + ghost \"Skip for now\".
2. **Connected, no incidents routed** — team name + connected timestamp, single channel picker for **Incidents** (the group the step is selling). CTA stays a ghost \"Skip for now\" because the pitch isn't delivered until a route exists.
3. **Connected, incidents routed** — the **Wrapped reports** and **Announcements** rows reveal progressively, and the CTA flips to a primary \"Continue\". A muted footer points users to settings for full control later.

Disconnect is intentionally **not** offered inside onboarding; users can manage that from settings once they're past the wizard.

## OAuth round-trip

Slack OAuth bounces the user out of the app, so onboarding state has to survive a full reload. Two pieces wire this up:

- **Step in URL.** The route now has a `validateSearch` for `step`, `installed`, `error`. A `goToStep(next)` helper centralises React-state + URL sync, so every transition is refresh-safe. Deep-link gate: if the URL points past `stack` but `project.settings.onboardingType` is unset, fall back to `role` — earlier steps have prerequisites that URL hopping shouldn't skip. `onboardingType` is read from the parent route's loader, which is synchronously available on first render.
- **`returnTo` in OAuth state.** `/integrations/slack/install?return_to=/projects/{slug}/onboarding` validates the path (allow-list `/projects/...`, reject protocol-relative, control chars, length > 512) and stores it in the existing Redis-backed CSRF state record. The callback honors `stateEntry.returnTo` on every post-state exit — success, workspace conflict, missing config, **and** the user-denied-on-consent path (Slack returns `state` without `code`, which the old early-return discarded). Only the no-state-at-all branch (replay or hand-typed URL) still falls back to the historic `/?next=integrations`.

Settings keeps its current behaviour: it doesn't pass `return_to`, so its OAuth round-trip continues to land where it always did.

## shared SlackRouteRow

The settings page's per-group channel picker was extracted to `settings/-components/slack-route-row.tsx` and is reused by the onboarding card. Both surfaces invalidate the same `[\"slack-integration\"]` query key on save, so a route change in either place propagates immediately. Settings still renders all three groups in full; the progressive reveal is onboarding-only.

## tests

- `validateReturnTo` accepts `/projects/...` paths, rejects open-redirect-shaped inputs (`//`, backslash, absolute URLs, control chars, paths outside the allow-list, oversize).
- OAuth state round-trip preserves a valid `returnTo`, drops invalid ones at both generate and consume time.
- `buildPostInstallRedirect` correctly composes the `Location` header for paths with/without an existing query string, across all three flash statuses.

## out of scope

- The `UserOnboardingCompleted` event still fires on the `stack → flaggers` transition (i.e. before the slack step), as it did before. Left as-is per discussion — touching analytics timing belongs in its own change.
- No channel auto-routing; the user explicitly picks.

### DIFF
--- a/apps/web/src/domains/integrations/integrations.functions.ts
+++ b/apps/web/src/domains/integrations/integrations.functions.ts
@@ -82,16 +82,7 @@ const toRecord = (row: SlackIntegration): SlackIntegrationRecord => ({
   routes: row.routes ?? {},
 })
 
-/**
- * Probes whether Slack OAuth env vars are present in this deployment.
- * Self-hosters without a Slack app registered get `false` so the UI
- * can hide the Connect button (e.g. on the onboarding Slack step)
- * instead of letting users hit a 503 on `/integrations/slack/install`.
- *
- * Dynamic import keeps `@platform/slack` off the client bundle —
- * `@platform/slack`'s barrel re-exports `client.ts` which transitively
- * pulls in `@slack/web-api` (uses `require("node:path")`).
- */
+// Dynamic import keeps `@platform/slack` (and its `@slack/web-api` transitive dep) off the client bundle.
 export const isSlackConfigured = createServerFn({ method: "GET" }).handler(async (): Promise<boolean> => {
   await requireSession()
   const { loadSlackConfig } = await import("@platform/slack")
--- a/apps/web/src/domains/integrations/integrations.functions.ts
+++ b/apps/web/src/domains/integrations/integrations.functions.ts
@@ -82,6 +82,23 @@ const toRecord = (row: SlackIntegration): SlackIntegrationRecord => ({
   routes: row.routes ?? {},
 })
 
+/**
+ * Probes whether Slack OAuth env vars are present in this deployment.
+ * Self-hosters without a Slack app registered get `false` so the UI
+ * can hide the Connect button (e.g. on the onboarding Slack step)
+ * instead of letting users hit a 503 on `/integrations/slack/install`.
+ *
+ * Dynamic import keeps `@platform/slack` off the client bundle —
+ * `@platform/slack`'s barrel re-exports `client.ts` which transitively
+ * pulls in `@slack/web-api` (uses `require("node:path")`).
+ */
+export const isSlackConfigured = createServerFn({ method: "GET" }).handler(async (): Promise<boolean> => {
+  await requireSession()
+  const { loadSlackConfig } = await import("@platform/slack")
+  const config = await Effect.runPromise(loadSlackConfig)
+  return config !== undefined
+})
+
 export const getActiveSlackIntegration = createServerFn({ method: "GET" }).handler(
   async (): Promise<SlackIntegrationRecord | null> => {
     const { organizationId } = await requireSession()
--- a/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
+++ b/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
@@ -393,15 +393,6 @@ function SdkIntegrationInstructions({
   )
 }
 
-/**
- * Optional onboarding step that prompts the user to connect Slack so
- * they get notified when their Flaggers detect issues. Hidden via the
- * `slack` feature flag + env-config probe in {@link OnboardingFlow}.
- *
- * Connect leaves the page (Slack OAuth redirect); the callback
- * bounces back to `/projects/{slug}/onboarding?step=slack&installed=ok`
- * which `OnboardingFlow` toasts on mount.
- */
 function SlackOnboardingStep({
   projectSlug,
   onBack,
@@ -416,9 +407,7 @@ function SlackOnboardingStep({
     queryFn: () => getActiveSlackIntegration(),
   })
   const connected = integration != null
-  // The step's pitch only materialises once `incidents` is routed
-  // somewhere. Until then, leaving counts as skipping — so the
-  // primary CTA stays a ghost "Skip for now" even after connect.
+  // CTA reflects routing, not just connection — no route = nothing gets delivered.
   const incidentsConfigured = (integration?.routes.incidents?.length ?? 0) > 0
 
   const returnTo = `/projects/${projectSlug}/onboarding?step=slack`
@@ -447,11 +436,7 @@ function SlackOnboardingStep({
             title="Slack"
             subtitle="Send Latitude notifications to your Slack workspace."
             actions={
-              // `/integrations/slack/install` is a server-handler-only
-              // route that 302s the browser to Slack — needs a full-
-              // page GET, not client-side routing. Plain `<a>` (inside
-              // a `Button asChild`) gives cmd/middle-click + copy-link
-              // affordances while keeping the full-page nav.
+              // Server-handler route — needs full-page nav, not `<Link>` client routing.
               <Button asChild>
                 <a href={connectHref}>Connect Slack</a>
               </Button>
@@ -476,21 +461,6 @@ function SlackOnboardingStep({
   )
 }
 
-/**
- * Onboarding-only variant of the settings page's connected card.
- * Differs in two ways:
- *
- * - No "Disconnect" affordance — onboarding is not the place to undo
- *   a connection the user just made.
- * - Progressive group reveal: `incidents` shows immediately (the
- *   group this step is selling), while `wrapped_reports` and
- *   `custom_messages` only appear once the user has set at least one
- *   route. The second condition also keeps already-routed groups
- *   visible for users who reconnect after a previous setup.
- *
- * Full routing for all groups (and disconnect) lives at
- * `/settings/integrations`; the footer note points users there.
- */
 function SlackConnectedOnboardingCard({ integration }: { readonly integration: SlackIntegrationRecord }) {
   const hasAnyRoute =
     (integration.routes.incidents?.length ?? 0) > 0 ||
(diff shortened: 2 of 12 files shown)


apps/web/src/routes/integrations/slack/oauth/callback.ts
```diff
@@ -24,13 +24,32 @@ const logger = createLogger("slack-oauth-callback")
...
+  readonly returnTo: string | null
+  readonly status: FlashStatus
+  readonly webUrl: string
+}): Response => {
+  const path = input.returnTo ?? DEFAULT_RETURN_PATH
+  const separator = path.includes("?") ? "&" : "?"
   const headers = new Headers()
-  headers.set("Location", `${webUrl}/?next=integrations&${status}`)
+  headers.set("Location", `${input.webUrl}${path}${separator}${input.status}`)
   headers.set("Cache-Control", "no-store")
```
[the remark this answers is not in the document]
csansoon: Fixed in 667ca35. `validateReturnTo` now rejects `#` outright (added a test case for it). Tightening the validator instead of restructuring the URL builder keeps the failure mode loud — a fragment in `returnTo` is never a legitimate caller anyway.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/-components/slack-route-row.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  const {
+    data: rawChannels = [],
+    isFetching,
+    refetch,
+  } = useQuery({
+    queryKey: SLACK_CHANNELS_QUERY_KEY,
+    queryFn: () => listSlackChannels(),
+    staleTime: 30_000,
+  })
```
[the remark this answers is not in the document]
csansoon: Updated the docstring in 4a39218. The eager-mount fetch + open-time `refetch()` is the intended behavior — the row was pulled over from the settings page where you want freshly-invited bots' new channels to appear without waiting for `staleTime`. Comment was wrong, behavior is fine.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
```diff
@@ -565,15 +784,15 @@ export function OnboardingFlow({
         },
       })
       await invalidateProjectFlaggers(projectId)
-      setStep("telemetry")
+      goToStep(slackStepEnabled ? "slack" : "telemetry")
     } catch (error) {
```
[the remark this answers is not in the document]
csansoon: Closed in 18fcf4e7. Moved `isSlackConfigured` from a client `useQuery` to the route loader and pass the resolved boolean down as a prop — same pattern `onboardingType` already uses. The race window was vanishingly small in practice (the query starts on mount and the user spends seconds on each prior step) but the loader is the structurally correct home, and it removes the transient-`false` defense the gate previously had to carry.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
```diff
@@ -473,17 +485,228 @@ function SdkIntegrationInstructions({
...
+          <IntegrationCard
+            icon={SlackIcon}
+            title="Slack"
+            subtitle="Send Latitude notifications to your Slack workspace."
+            actions={
+              <Button
+                onClick={() => {
+                  window.location.href = connectHref
+                }}
+              >
```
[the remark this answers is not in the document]
csansoon: Switched both onboarding (18fcf4e7) and the existing settings page (5b568dab) to `<Button asChild><a href="…">`. `<Link>` itself doesn't apply here — `/integrations/slack/install` is a server-handler-only route and TanStack's client-side `<Link>` would try to route into a component that doesn't exist; the OAuth flow needs a full-page GET so the server can 302 us to Slack. Plain `<a>` keeps the full-page navigation while giving the cmd/middle-click + copy-link affordances the skill calls out.

apps/web/src/domains/integrations/integrations.functions.ts
```diff
@@ -82,6 +82,23 @@ const toRecord = (row: SlackIntegration): SlackIntegrationRecord => ({
...
+/**
+ * Probes whether Slack OAuth env vars are present in this deployment.
+ * Self-hosters without a Slack app registered get `false` so the UI
+ * can hide the Connect button (e.g. on the onboarding Slack step)
+ * instead of letting users hit a 503 on `/integrations/slack/install`.
+ *
+ * Dynamic import keeps `@platform/slack` off the client bundle —
+ * `@platform/slack`'s barrel re-exports `client.ts` which transitively
+ * pulls in `@slack/web-api` (uses `require("node:path")`).
+ */
```
andresgutgon: Yapping

andresgutgon (commented):
Probably posthog will need this setup

apps/web/src/domains/integrations/integrations.functions.ts
csansoon: Fair, trimmed in 3e29901. Kept only the dynamic-import bit since that's the only non-obvious part.

apps/web/src/routes/integrations/slack/oauth/callback.ts
```diff
@@ -58,23 +77,45 @@ export const Route = createFileRoute("/integrations/slack/oauth/callback")({
...
+          logger.warn("slack oauth callback missing state")
           return redirectToSettings("error=oauth_failed", webUrl)
         }
 
+        // Consume state *before* the missing-code check. If the user
+        // denied access on Slack's consent screen, Slack redirects here
+        // with `state` set but no `code` — we still want to honor the
+        // caller's `returnTo` so onboarding doesn't dump them on the
+        // settings page when they bail out of the OAuth dance.
         const stateEntry = await consumeSlackOAuthState({ redis: getRedisClient(), state })
```
andresgutgon, line 83: Yapping is strong

apps/web/src/routes/integrations/slack/oauth/callback.ts
```diff
@@ -58,23 +77,45 @@ export const Route = createFileRoute("/integrations/slack/oauth/callback")({
...
-        const code = url.searchParams.get("code")
         const state = url.searchParams.get("state")
-        if (!code || !state) {
-          logger.warn("slack oauth callback missing code or state")
+        const code = url.searchParams.get("code")
+
+        // No state at all → either a replay attempt or a manually-typed
+        // URL. We have no `returnTo` to honor, so fall back to the
+        // default settings redirect.
+        if (!state) {
```
andresgutgon, line 77: Yap

apps/web/src/routes/_authenticated/projects/$projectSlug/onboarding.tsx
```diff
@@ -1,21 +1,43 @@
...
+  error: z.enum(["workspace_taken", "oauth_failed"]).optional(),
+})
+
 export const Route = createFileRoute("/_authenticated/projects/$projectSlug/onboarding")({
+  validateSearch: searchSchema,
+  // Resolve the env probe at load time so the slack step's visibility
+  // gate is synchronous on first render. A client `useQuery` defaults
+  // to `false` while loading, which would race with the user clicking
+  // "Continue" on the flaggers step and silently skip the slack step
+  // even when it should be shown.
```
andresgutgon: 🗣️

apps/web/src/routes/integrations/slack/oauth/callback.ts
csansoon: image

apps/web/src/routes/_authenticated/projects/$projectSlug/onboarding.tsx
csansoon: 📢

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
```diff
@@ -473,17 +482,227 @@ function SdkIntegrationInstructions({
...
+  // deep-links would skip past. `onboardingType` is fed in from the
+  // parent route's loader so it's synchronously available on first
+  // render (the route's loader awaits the project lookup).
+  //
+  // We deliberately do NOT also gate on `slackStepEnabled` here: the
+  // user only reaches `?step=slack` by already having seen the
+  // (env-gated) step, so if the URL says `slack`, trust it. Forward
+  // and back transitions still consult `slackStepEnabled` to route
+  // around the step when truly disabled.
+  const onboardingTypeSet = onboardingType != null
```
andresgutgon, line 523: 🗣️

apps/web/src/routes/integrations/slack/oauth/callback.ts
csansoon: Trimmed (along with the rest) in 07e0224. Kept only one-line WHYs.

apps/web/src/routes/integrations/slack/oauth/callback.ts
csansoon: Trimmed (along with the rest) in 07e0224. Kept only one-line WHYs.

apps/web/src/routes/_authenticated/projects/$projectSlug/onboarding.tsx
csansoon: Trimmed (along with the rest) in 07e0224. Kept only one-line WHYs.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
csansoon: Trimmed (along with the rest) in 07e0224. Kept only one-line WHYs.

andresgutgon (approved):

PR #3320, opened 2026-05-28
Replace search by session search

andresgutgon:
## Summary
- Retire legacy trace-flat `/search` in favor of the session-rollup view that was running behind the `session-search-v2` flag. Saved searches and the Export Traces / Add to Dataset bulk actions are ported over so the new `/search` is a strict superset of the old route.
- Drop Live Taxonomy panel (plus `live-taxonomy-search-recommendations` flag and the web-side `domains/taxonomy/` hooks); clusters are trace-centric and don't map onto sessions cleanly.
- Selection in the session view is already trace-id-based via `useSessionSelectionAdapter`, so existing bulk-export plumbing works unchanged. `useTracesCount` feeds the real trace count for select-all semantics.

## What ships
- `/session-search` is removed; its code becomes the canonical `/search`.
- Saved-search backend (`@domain/saved-searches`, API, Postgres table) is **untouched** — `FilterSet` + `query` are reusable as-is for sessions. The external `GET /searches/:slug/traces` endpoint is intentionally left alone.
- `session-search-v2` + `live-taxonomy-search-recommendations` removed from the feature-flags registry.
- Sidebar collapses to a single Search entry.

## Test plan
- [x] Empty `/search` shows the saved-searches list. With zero saved searches, the empty state renders cleanly (no taxonomy/cluster copy).
- [x] Apply filters + a free-text query → click **Save search** → modal saves; new entry appears in the list on the empty state.
- [x] Open a saved search from the list → URL gets `?savedSearch=<slug>&q=...&filters=...`. Results render. **Update Saved Search** is disabled.
- [x] Modify filters → **Update Saved Search** enables → clicking persists; reload preserves the change.
- [x] **Save as new Search** from the SplitButton creates a fresh saved search.
- [x] Rename + Delete from the row's overflow menu work.
- [x] **Assign to** dropdown changes the assignee on a saved search.
- [x] Every filter from `TRACE_FILTER_FIELDS` is reachable; percentile filters on duration/ttft/cost work; `hasLlmActivity` toggle behaves as before.
- [x] Select multiple sessions → **Export Traces** opens confirmation → enqueues a job (verify a confirmation toast). The selected count reflects the underlying trace IDs, not the session count.
- [x] Stale URLs to `/projects/<slug>/session-search` 404 (route is gone). Sidebar shows only one **Search** entry.
- [x] **Add to Dataset** for the same selection writes the correct trace IDs to the chosen dataset.
- [x] `grep -r session-search-v2` and `grep -r live-taxonomy-search-recommendations` return empty in the repo.
- [x] `pnpm typecheck` clean (74 packages), `pnpm --filter @app/web test --run` 344/344 pass, `pnpm --filter @domain/saved-searches test --run` 24/24 pass, Biome clean.


andresgutgon:
@claude do review please

csansoon (approved):

PR #3324, opened 2026-05-29
Project-level trace sampling at the ingest boundary

csansoon:
## What

A new project setting, **Trace sampling**, that drops a configurable fraction of incoming OTLP batches at the HTTP ingest endpoint. When enabled with a rate below 100%, sampled-out batches never reach object storage, the worker queue, ClickHouse, or any downstream consumer — and they don't count toward billing.

Two new fields on `projects.settings`:

```ts
sampling: {
  enabled?: boolean
  rate?: number  // 0..1, only consulted when enabled
}
```

`enabled` and `rate` are split so the UI can remember the configured rate while the toggle is off.

## Why

High-volume customers ingest more spans than they want to store. Storage and ClickHouse write costs scale linearly with span volume, and there's no existing lever to reduce them. This adds a customer-controlled knob: enable sampling, set a rate, pay for ~that fraction of the original volume.

## How the decision is made

For every incoming OTLP batch, after slug → project resolution (which already runs today), pick the first resolvable project — preferring the `X-Latitude-Project` header default, falling back to the first per-span project. If that project has sampling enabled with `rate < 1`:

1. Walk to the first span in the decoded payload.
2. Extract a sampling key: `session.id` from span attrs (or any of the known session-id aliases: `gen_ai.session.id`, `langfuse.session.id`, `traceloop.association.properties.session_id`, etc. — the same list used by `transformOtlpToSpans`), falling back to resource attrs, then to `trace_id`.
3. Hash the key with SHA-256, take the first 8 bytes as a uniform `BigUInt64`, divide by `2^64` to get a draw in `[0, 1)`.
4. Drop the batch if the draw is `>= rate`; otherwise proceed as today.

The hash gives deterministic sampling: the same session always lands the same way, so all spans of a session land together as long as they ship in the same batch.

## Per-batch vs per-span

The decision is per-OTLP-batch, not per-span. In practice OTLP exporters used with Latitude (Vercel AI SDK, Agents SDK, OpenInference, Langfuse-compatible) send one trace per batch, so per-batch = per-session for the common case.

For multi-session batches (server-side `BatchSpanProcessor` flushing across concurrent requests), the decision is taken from the first span — meaning other sessions in the same batch get bundled with the first session's decision. We accept this imprecision in exchange for keeping the code path tiny (no per-span filtering, no OTLP re-encoding, no acceptance bitmap to ship to the worker). The use-case has a comment documenting the tradeoff. If a customer reports broken sessions later, the worker-side filter is the obvious upgrade path.

## What sampled-out batches skip

- The base64 inlining and disk write (`putInDisk` for >50KB payloads)
- The queue publish, so the worker never decodes or transforms them
- The ClickHouse insert
- The `TracesIngested` event and everything downstream of it: issue discovery, evaluations, scoring, **billing usage events** (one credit per persisted trace — sampled-out traces aren't persisted, so they're not billed)

## OTLP response

Sampled-out spans are reported as `accepted` in the response. The customer's exporter sees a clean 200 and has nothing to retry. This is intentional: sampling is a server-side cost lever, not a partial failure to surface to clients.

## Why HTTP, not the worker

The HTTP endpoint already fetches the full `Project` via `findBySlug` to resolve slug → projectId. `findBySlug` returns the entity with `settings`, but the previous code threw everything but `.id` away. Keeping the rest costs zero DB roundtrips and lets the sampling decision land before any expensive work happens — including storage. Putting the filter in the worker would still write the full payload to disk and run the second decode pass; doing it at HTTP is strictly cheaper.

## UI

A new section in the project's **General settings** page (same page as project name and the delete card): a toggle to enable sampling, and a percentage slider that appears when enabled. Description text explains the session-level grouping and the billing implication.

## Tests

- **`deterministic-sampler.test.ts`** — determinism, distribution at 50% and 10% over 10k random keys, edge rates (0 and 1).
- **`extract-sampling-key.test.ts`** — span attr wins, resource attr fallback, `trace_id` final fallback, only first span inspected, alternative session-id keys, empty payload returns null.
- **`ingest-spans.test.ts` → "trace sampling"** — rate=0 drops, rate=1 keeps, `enabled=false` is a no-op, same `session_id` is deterministic across repeats, ~50% distribution over 200 distinct sessions, large sampled-out batch skips disk + queue, header-default project's settings govern the batch, empty payload short-circuits.

## Out of scope / follow-ups

- Org-level sampling cascade. The cascade plumbing exists for `keepMonitoring` but project-only is enough for now.
- Per-span filtering of multi-session batches (worker-side `acceptedSpanIds` mask). Documented as the upgrade path if multi-session batches become a real concern.
- Redis caching of `ProjectRepository.findBySlug`. Today every batch hits Postgres once per unique slug; if profiling shows it matters we can mirror the `resolveEffectivePlanCached` pattern.


apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
```diff
@@ -1,22 +1,38 @@
...
   component: ProjectGeneralSettingsPage,
 })
 
+interface Draft {
+  readonly name: string
+  readonly samplingEnabled: boolean
+  readonly samplingRate: number
+}
+
+function DirtyDot() {
```
andresgutgon: We have dot component already

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
```diff
@@ -27,65 +43,198 @@ function ProjectGeneralSettingsPage() {
...
+        <Input
+          type="text"
+          label={
+            <span className="flex flex-row items-center gap-2">
+              Name
+              {nameIsDirty ? <DirtyDot /> : null}
+            </span>
+          }
+          value={view.name}
+          onChange={(e) => setField("name", e.target.value)}
```
andresgutgon, line 168: Please use `useForm` a lot of redundant state here

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
```diff
@@ -27,65 +43,198 @@ function ProjectGeneralSettingsPage() {
...
+      setIsApplying(false)
+    }
+  }
+
+  const applyRef = useRef(apply)
+  applyRef.current = apply
+  const isApplyingRef = useRef(isApplying)
+  isApplyingRef.current = isApplying
+
+  useEffect(() => {
```
andresgutgon: This want I feel can be avoided enterely

packages/domain/shared/src/settings.ts
```diff
@@ -60,11 +60,24 @@ export const escalationSettingSchema = z.object({
 })
 export type EscalationSetting = z.infer<typeof escalationSettingSchema>
 
+/**
+ * Trace sampling. Drops a configurable fraction of ingested OTLP batches at the HTTP
+ * boundary. Deterministic on `session_id || trace_id`, so all spans sharing a session
+ * share the keep/drop decision. `enabled` and `rate` are split so the UI can remember
+ * the configured rate while the toggle is off.
+ */
```
andresgutgon: Yapping

packages/domain/spans/src/use-cases/ingest-spans.ts
```diff
@@ -142,13 +143,31 @@ export const ingestSpansUseCase = (
...
+    const sampling = samplingProject?.settings?.sampling
+    if (sampling?.enabled && (sampling.rate ?? 1) < 1) {
+      const samplingKey = extractSamplingKey(decoded)
+      if (samplingKey && !deterministicSample(samplingKey, sampling.rate ?? 1)) {
+        yield* Effect.annotateCurrentSpan("sampling.dropped", true)
+        yield* Effect.annotateCurrentSpan("sampling.rate", sampling.rate ?? 1)
+        // Sampled-out: don't store, don't enqueue. Spans still count as accepted in the OTLP
+        // response — the customer's exporter has nothing to retry. Billing is naturally
+        // skipped because no `TracesIngested` event will fire downstream.
+        return { totalSpans, acceptedSpans, rejectedSpans }
```
andresgutgon, line 160: So inside the same trace some spans are ingested and others no?

csansoon: No! The decision is based on the span's `session_id` (or `trace_id`, if there's no session). So all spans with the same session will return the same decision.

packages/domain/spans/src/sampling/deterministic-sampler.ts
```diff
@@ -0,0 +1,17 @@
...
+ * SHA-256's avalanche property makes any 8 consecutive output bytes uniform over
+ * [0, 2^64), so dividing by 2^64 yields a uniform float in [0, 1).
+ */
+export function deterministicSample(key: string, rate: number): boolean {
+  if (rate >= 1) return true
+  if (rate <= 0) return false
+  const digest = createHash("sha256").update(key).digest()
+  const high = digest.readBigUInt64BE(0)
+  const draw = Number(high) / Number(1n << 64n)
+  return draw < rate
```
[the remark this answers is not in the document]
csansoon: Good catch. Switched to the top 53 bits (`Number(digest.readBigUInt64BE(0) >> 11n) / 2 ** 53`) so the BigInt → Number conversion is lossless and the draw is uniform over [0, 1). Fixed in 01db5e6ae.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
```diff
@@ -81,11 +82,107 @@ function ProjectGeneralSettingsPage() {
...
 }
 
+function TraceSamplingSection({ projectId, settings }: { projectId: string; settings: ProjectRecord["settings"] }) {
+  const { toast } = useToast()
+  const sampling = settings.sampling
+  const enabled = sampling?.enabled ?? false
+  const rate = sampling?.rate ?? 1
+  const [isSavingEnabled, setIsSavingEnabled] = useState(false)
+  const [draftPercent, setDraftPercent] = useValueWithDefault(Math.round(rate * 100))
+
```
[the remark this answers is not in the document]
csansoon: The slider now uses `min={0}`, so a persisted `rate = 0` is a valid in-range value and the control no longer glitches. Addressed in 2d5971be8.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
```diff
@@ -81,11 +82,107 @@ function ProjectGeneralSettingsPage() {
...
+              Percentage of traces to store. 100% stores everything; 10% stores roughly 1 in 10.
+            </Text.H6>
+          </div>
+          <div className="flex w-full flex-row items-center gap-4">
+            <Slider
+              id="trace-sampling-rate"
+              min={1}
+              max={100}
+              step={1}
+              value={[draftPercent]}
```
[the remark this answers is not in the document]
csansoon: Done — the slider now goes down to 0%, and when 0% is selected we show a warning right under it that no traces will be processed or stored. 2d5971be8.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
csansoon: Swapped my custom dot for the existing `DotIndicator` (`variant="primary"`). 2d5971be8.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
csansoon: This is deliberate — the page mirrors the Flaggers settings page (`settings/flaggers.tsx`), which uses the same manual `pending`/overlay state instead of `useForm`. I copied that pattern so the two settings pages stay consistent. (Your review predates this rework.)

Happy to migrate to `useForm`, but I'd suggest doing it for both pages together in a separate change rather than diverging here. Let me know which you'd prefer.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
csansoon: Same story as the state comment: the Cmd/Ctrl+S + Escape handlers are copied verbatim from the Flaggers page for parity. I can drop them here, but it'd be inconsistent with Flaggers — better to remove from both at once if we don't want them. Your call.

packages/domain/shared/src/settings.ts
csansoon: Trimmed to a one-liner. 0825b97d3.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
andresgutgon: Do not AI me bro 😂 I think we should unify if possible and use `useForm` if possible for this pending state. tanstack form probably have already figured out

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
csansoon: Follow-up: regardless of the keep/drop decision, the implementation was also violating our `useEffect` policy (web-frontend skill — raw `useEffect` should be `useMountEffect`). Converted it to a mount-only `useMountEffect` listener that reads latest state via refs, so it no longer re-subscribes on every dirty-state change. 8381f68db.

Still happy to remove the shortcuts entirely if you'd prefer — but I'd do that for the Flaggers page too in the same change.

apps/web/src/routes/_authenticated/projects/$projectSlug/settings/general.tsx
csansoon: xdd
noted for a future PR

andresgutgon (approved):

PR #3331, opened 2026-05-29
feat(onboarding): interactive right-pane previews for the onboarding flow

csansoon:
Reworks the project onboarding flow so the right-hand pane of each step is a live preview of what the user is configuring on the left, instead of the same static screenshot gallery shown on every step.

Onboarding is a five-step split-screen flow — `role → stack → flaggers → slack → telemetry` — with the interaction on the left and examples on the right. Before, the right side was one shared image gallery regardless of step. Now it reacts to the user's input and animates between steps.

https://github.com/user-attachments/assets/0a0e75a8-f243-4832-a40b-fdf7dc89a10f

## structure

`onboarding-flow.tsx` was a single ~1,100-line component. It's now a thin orchestrator that owns cross-step state (the role form, stack choice, flagger selection, trace polling) and renders a per-step `Left`/`Right` pair. Each step lives in its own file under `-components/onboarding/steps/`, with shared pieces (`carousel-track`, `onboarding-gallery`, `motion`, and the `mocks/`) alongside. No behavior changed in that first commit — it's a straight extraction.

## what each right pane does now

- **role + stack** share one "intro" slide: an auto-advancing product-tour gallery (crossfade, dots, pause-on-hover, reduced-motion aware). These are the first impression, so they stay screenshot-based.
- **flaggers** shows a mock Issues feed driven by the selected flaggers — one example issue per enabled flagger, appearing/disappearing as the user toggles. Rows reuse the real `Status` pill, `TagList`, and `formatCount`, plus a mini trend bar, a regressed-row tint, and a live issue count. Titles are concrete scenarios ("Checkout total fails on string amounts") rather than restatements of the flagger name.
- **slack** shows a depth-stacked incident-notification queue — newest on top, older cards receding via opacity + scale, cycling on a timer. Card content mirrors what Latitude actually posts (modeled on `packages/domain/integrations/src/templates/notifications`): a severity-colored accent, a lifecycle kind (new / escalating / regressed / resolved) with emoji, the issue name, and a mini trend sparkline on the events that embed one.
- **telemetry** shows a live "trace tail": shimmering skeleton rows until the first trace arrives, then a confirmed-trace row. This is the only pane backed by real data — it reuses the existing trace-count polling, which already redirected the user on first trace.

The right pane also slides horizontally between steps (a `translateX` carousel), snapping instead of sliding under `prefers-reduced-motion`. role and stack map to the same slide so the gallery doesn't move between them.

## notable decisions

- **List height animation without a library.** There's no framer-motion in the repo, so the feeds animate height with the grid-rows `0fr↔1fr` collapse trick. This fixes two glitches the naive `animate-in/out` approach had: sibling rows snapping into place on add/remove, and the list ballooning then collapsing when a flagger preset toggles many rows at once.
- **The slack queue evicts invisibly.** Cards are removed from the DOM only after they've already faded to opacity 0 at the bottom of the depth stack, so there's no exit animation to time and the stack height stays constant (the entering card grows by the same amount the leaving one shrinks). Its timer is gated on the slack step being active, so it never builds a backlog while mounted off-screen in the carousel.
- All preview data except the telemetry trace count is decorative; there are no backend or domain changes in this PR.

## scope

Onboarding right-pane only. The left-pane interactions, the flagger/onboarding server functions, and the trace-count polling/redirect are unchanged (the polling just moved onto the orchestrator). Each commit passed `typecheck` and `biome`.


apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding/mocks/mock-issues-feed.tsx
```diff
@@ -0,0 +1,275 @@
...
+      changed.forEach((f, i) => {
+        map.set(f.slug, Math.min(i * STAGGER_STEP_MS, STAGGER_MAX_MS))
+      })
+    }
+    return map
+  }, [mockRows, enabledFlaggerSlugs])
+
+  useEffect(() => {
+    prevOpenRef.current = new Set(mockRows.filter((f) => enabledFlaggerSlugs.has(f.slug)).map((f) => f.slug))
+  })
```
andresgutgon: No ⚠️ ⚠️ ⚠️ ⚠️ ⚠️

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding/mocks/mock-slack-queue.tsx
```diff
@@ -0,0 +1,283 @@
...
+  index,
+  children,
+}: {
+  readonly card: RenderedCard
+  readonly index: number
+  readonly children: ReactNode
+}) {
+  // Seeded cards render in place; freshly-added cards start collapsed and expand next frame.
+  const [entered, setEntered] = useState(card.seeded)
+  useEffect(() => {
```
andresgutgon: `useMountEffect`

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding/motion.ts
```diff
@@ -0,0 +1,23 @@
...
+export function usePrefersReducedMotion(): boolean {
+  const [reduced, setReduced] = useState(false)
+
+  useEffect(() => {
+    const query = window.matchMedia("(prefers-reduced-motion: reduce)")
+    setReduced(query.matches)
+    const onChange = (event: MediaQueryListEvent) => setReduced(event.matches)
+    query.addEventListener("change", onChange)
+    return () => query.removeEventListener("change", onChange)
+  }, [])
```
andresgutgon: `useMountEffect`

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding/motion.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Tracks `prefers-reduced-motion: reduce`. Initializes `false` for SSR and reads the real
+ * value after mount, so callers can disable JS-driven motion (e.g. carousel auto-advance)
+ * that CSS `motion-reduce:` utilities can't reach.
+ */
+export function usePrefersReducedMotion(): boolean {
+  const [reduced, setReduced] = useState(false)
+
+  useEffect(() => {
+    const query = window.matchMedia("(prefers-reduced-motion: reduce)")
```
andresgutgon, line 16: This can be done in CSS I think

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding/mocks/mock-issues-feed.tsx
csansoon: Good call — that was a ref read during render plus a no-deps effect rewriting it every render. Dropped the previous-set tracking entirely in e8f93125e and switched to a static per-position delay: a preset still cascades top-to-bottom, a single toggle keeps a barely-perceptible delay, and there's no render-phase ref access anymore.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding/mocks/mock-slack-queue.tsx
csansoon: Switched to `useMountEffect` in d1d15feae.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding/motion.ts
csansoon: Switched to `useMountEffect` in d1d15feae.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding/motion.ts
csansoon: nah this is for JS-driven timers, like the gallery automatically switching the image. now, if user has motion reduced, we don't change it automatically!
(i changed the `useEffect` for a `useMountEffect` too btw)

andresgutgon (approved):

PR #3333, opened 2026-05-29
Fix flagger false positives on nested evaluator traces

geclos:
## what changed

Fixes false positives when LLM flaggers evaluate traces produced by other evaluators or annotation reviewers. The classifier prompt includes the evaluated agent's system prompt or extracted context, then wraps the trace evidence so user/source material and the evaluated assistant response are separate targets.

The prompt tells the flagger to judge the evaluated assistant response itself. If that response classifies, reviews, summarizes, or transforms supplied content, the flagger should judge that response's behavior rather than the nested content it discusses.

Other changes:
- extracts long evaluated-agent system prompts with an LLM and caches the result by prompt hash
- includes short evaluated-agent system prompts verbatim as tagged context
- skips LLM flaggers when no evaluated-agent context can be inferred
- removes the old trace-reading blocks and uses explicit XML-style trace tags instead
- narrows annotation review to the evaluated assistant response(s), not the full nested classifier evidence
- avoids scanning system messages in jailbreak evidence extraction
- wires the flagger workflow activity with `RedisCacheStoreLive` for extraction caching
- switches live flagger classification from MiniMax M2.5 to Bedrock Claude Haiku 4.5 after checking current Bedrock model IDs
- gates the live CSV regression behind `pnpm test:regression` so regular `pnpm test` does not require the downloaded artifact

## validation

- `pnpm --filter @domain/flaggers check`
- `pnpm --filter @domain/flaggers test`

The local downloaded 52-row CSV regression also passes with Claude Haiku 4.5 via the new regression test path; the CSV is intentionally not included in this PR.


packages/domain/flaggers/src/flagger-strategies/jailbreaking.ts
```diff
@@ -315,6 +315,8 @@ function extractJailbreakSuspiciousSnippets(trace: Pick<TraceDetail, "allMessage
 
   for (const message of trace.allMessages) {
     const role = message.role
+    if (role === "system") continue
```
geclos, line 318: evaluate the rest of flagger strategies to make sure we arent' matching against system role messages, since those are codigied instructions that should not include any of the failure models flaggers try to detect.

geclos: Addressed in f6cdaa507. I reviewed the other strategies and tightened the remaining system-role paths: jailbreak/NSFW no longer treat system-only traces as classifier context, tool-call errors now only scans assistant tool calls plus tool/function responses, and the new system-role regression covers content strategies plus deterministic matchers.

andresgutgon (dismissed):

PR #3336, opened 2026-05-30
fix(sdk): eliminate AbortSignal race condition in anySignal()

faisalaffan:
## Problem                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                       
  `anySignal()` has a race condition: if a source signal aborts between the                                                                                                                                                                                                                            
  `signal.aborted` check and the `addEventListener("abort", …)` call, the                                                                                                                                                                                                                              
  abort event is already dispatched and the listener never fires. The combined                                                                                                                                                                                                                         
  signal stays unaborted indefinitely — timeouts and cancellation stop working.                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                       
  ## Root Cause                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                       
  In Node.js, adding an "abort" listener to an already-aborted AbortSignal                                                                                                                                                                                                                             
  does NOT fire the callback synchronously. The gap between the `aborted`                                                                                                                                                                                                                            
  check and listener registration creates a window where aborts are                                                                                                                                                                                                                                    
  silently dropped.                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                       
  ## Fix                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                       
  Re-check `signal.aborted` immediately after `addEventListener`. If the                                                                                                                                                                                                                               
  signal aborted during the registration window, manually propagate the
  abort to the combined controller.                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                       
  Also replaces `break` with `return controller.signal` to avoid useless                                                                                                                                                                                                                               
  iteration after the combined signal is already aborted.                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                       
  ## Test Plan                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                       
  - [x] Happy path: abort forwarded correctly                                                                                                                                                                                                                                                          
  - [x] Already-aborted signal detected on entry                                                                                                                                                                                                                                                     
  - [x] Signal aborting between check and listener (was BROKEN, now fixed)                                                                                                                                                                                                                             
  - [x] Works with `getTimeoutSignal`                                                                                                                                                                                                                                                                  
  - [x] Accepts both spread and array arguments

### DIFF
--- a/packages/sdk/typescript/src/core/fetcher/signals.test.ts
+++ b/packages/sdk/typescript/src/core/fetcher/signals.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest"
-import { anySignal, getTimeoutSignal } from "./signals"
+import { anySignal, getTimeoutSignal } from "./signals.js"
 
 describe("anySignal", () => {
   it("forwards an abort from a source signal to the combined signal", () => {
--- a/packages/sdk/typescript/src/core/fetcher/signals.test.ts
+++ b/packages/sdk/typescript/src/core/fetcher/signals.test.ts
@@ -1,0 +1,93 @@
+import { describe, it, expect } from "vitest"
+import { anySignal, getTimeoutSignal } from "./signals"
+
+describe("anySignal", () => {
+  it("forwards an abort from a source signal to the combined signal", () => {
+    const controller = new AbortController()
+    const combined = anySignal(controller.signal)
+
+    expect(combined.aborted).toBe(false)
+    controller.abort("test-reason")
+    expect(combined.aborted).toBe(true)
+    expect(combined.reason).toBe("test-reason")
+  })
+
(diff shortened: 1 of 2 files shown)


andresgutgon (changes requested):
The problem with this fix is that we use https://github.com/fern-api/fern for auto generating our SDK. 

Our SDK is derived from OpenAPI. So fixing this code will do not fix the problem because next time we add a new endpoint we'll run regenerate and old code will appear again. 

The real fix will bring this case to Fern repo so we can update the generator

geclos:
@claude open an issue in https://github.com/fern-api/fern reporting this issue

PR #3338, opened 2026-06-01
fix(sessions): rebuild sessions from spans to recover pre-migration rows [LAT-637]

andresgutgon:
## Context

[LAT-637](https://linear.app/latitude/issue/LAT-637/backfill-sessions-rows-for-pre-migration-orphan-traces)

ClickHouse materialized views are **insert triggers** — they only aggregate spans ingested *after* the view exists. That leaves two classes of historical rows wrong in `sessions`:

1. **Orphan traces** (spans with no `gen_ai.conversation.id`) ingested before `00016_session_parity` (LAT-604) taught `sessions_mv` to synthesize a session per orphan trace — these have **no session row at all**, so they fall off the session-paginated listings entirely (the original bug).
2. **Named sessions** ingested before 00016 have a row, but it was written by the old MV and is missing every column 00016 added (`duration_ns` as a real `SimpleAggregateFunction`, `max_start_time`, `time_of_first_token`, `root_span_id/name`, `input/last_input/output_messages`, `system_instructions`, `retention_days`).

The table is small in production (**≈331k sessions from ≈4.3M spans**), so the cleanest fix is a **full rebuild from the source-of-truth `spans` table**, which repairs both classes at once.

## Approach

A one-time ClickHouse goose migration `00019_backfill_sessions_from_spans.sql` (both `clustered/` and `unclustered/`).

**Mechanism — keep `sessions_mv` running, rebuild only the contents:**

- We do **not** drop/recreate `sessions_mv` or the `sessions` table. The MV is already the correct post-00016 definition; only the historical *contents* are stale. (Dropping the table while the MV exists would make every `INSERT` into `spans` fail, since the MV can't write its target.)
- `TRUNCATE TABLE sessions` then `INSERT … SELECT` the **post-00016 `sessions_mv` body verbatim** (`coalesce(session_id, trace_id)` so orphans and named sessions are both rebuilt), `FROM spans FINAL` to collapse ReplacingMergeTree duplicate versions. Columns listed explicitly so the `time_to_first_token_ns` ALIAS is skipped and the insert is robust to reordering.
- Starting from an empty table is what makes it safe: `sessions` is an `AggregatingMergeTree`, so re-inserting into existing rows would merge partial states and double-count — there's nothing to merge into after `TRUNCATE`.
- Because the MV is never stopped, spans ingested *during* the rebuild keep flowing into `sessions` via the MV → **no ingestion gap**. The only theoretical overlap is a span ingested in the sub-millisecond window between `TRUNCATE` committing and the `INSERT` snapshot (counted once by the MV and once here); at production ingest rates this is effectively never, and active sessions self-heal as later spans' partial states merge in.

### Why a rebuild migration over the per-org backfill

An earlier revision of this branch implemented a per-org backoffice button → BullMQ worker that backfilled *orphans only* via an anti-join. It's been **reverted** in favor of the rebuild because: (a) given the tiny data size the rebuild's `GROUP BY` runs in seconds, removing the "heavy scan" objection; and (b) the rebuild also repairs the stale columns on pre-00016 *named* sessions, which the orphan-only approach left untouched (and which can't be fixed in place on an AggregatingMergeTree).

### Why `TRUNCATE`+reinsert over shadow-table + `EXCHANGE`

At this size the few-seconds "incomplete list" blip during the deploy migration is negligible, and `TRUNCATE`+`INSERT` is dramatically simpler and more robust than wiring a second MV + atomic swap across both the replicated `clustered/` and `unclustered/` files. (If `sessions` were large/high-traffic, the `EXCHANGE` variant would be worth the complexity.)

## Verification

Validated locally against the real ClickHouse dev schema **without touching the shared `sessions` table** — the exact migration `INSERT` (explicit column list) was run into a throwaway `CREATE TABLE … AS sessions` copy:

- Produced **21,378 rows — identical to the live `sessions` count**, confirming the column mapping and aggregation compile and round-trip correctly.
- Full aggregate diff (rebuilt vs live, `FULL OUTER JOIN` on org/project/session_id) returned **zero mismatches** across `trace_count`, `span_count`, `error_count`, `tokens_total`, `cost_total_microcents`, `min_start_time`, `max_end_time`, `root_span_name`, and **no missing/extra sessions**.
- Throwaway table dropped; dev data untouched.

Goose was intentionally **not** run (so as not to truncate shared dev state); the only parts not exercised that way are `TRUNCATE`, the `-- +goose` markers, and the no-op Down, all structurally identical to existing migrations (`00016`/`00018`).

Suggested QA: apply on a staging copy with known pre-00016 data and confirm (a) orphan traces now appear in session listings and (b) old named sessions show correct duration / root span / messages.

### DIFF
--- a/packages/platform/db-clickhouse/clickhouse/.migration-lock
+++ b/packages/platform/db-clickhouse/clickhouse/.migration-lock
@@ -4,6 +4,6 @@
 # create migrations in parallel, preventing incompatible
 # migrations from being merged without conflict resolution.
 #
-# Generated: 2026-05-24T17:02:07.688Z
-# Hash: 5c6f1c5a-2d38-4df7-bbe4-8e181d2e48be
+# Generated: 2026-06-01T08:39:08.256Z
+# Hash: 67008b2f-cd21-4ca0-9ed7-758d1917a98d
 # Do not manually edit this file.
--- a/packages/platform/db-clickhouse/clickhouse/migrations/clustered/00019_backfill_sessions_from_spans.sql
+++ b/packages/platform/db-clickhouse/clickhouse/migrations/clustered/00019_backfill_sessions_from_spans.sql
@@ -1,0 +1,171 @@
+-- +goose NO TRANSACTION
+-- +goose Up
+
+-- ═══════════════════════════════════════════════════════════
+-- Backfill `sessions` from `spans` (LAT-637) — clustered variant.
+-- See unclustered/00019_backfill_sessions_from_spans.sql for full notes.
+--
+-- Same rebuild as the unclustered variant, with cluster differences:
+--   - DROP VIEW, TRUNCATE, and CREATE MATERIALIZED VIEW all run
+--     ON CLUSTER default so every replica applies the same change.
+--   - The INSERT ... SELECT is plain DML (no ON CLUSTER): on the replicated
+--     `sessions` table the inserted parts replicate to the other replicas on
+--     their own.
+--
+-- The MV is dropped for the duration of the rebuild and recreated at the end.
+-- The CREATE MATERIALIZED VIEW body MUST stay in lockstep with
+-- clustered/00016_session_parity.sql so live aggregation resumes with the same
+-- definition. Span ingestion is unaffected (the MV is independent of the spans
+-- table); only session-row production pauses during the rebuild window.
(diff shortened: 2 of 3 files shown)


andresgutgon:
Not worth it

PR #3339, opened 2026-06-01
feat(monitors): backend foundation

neoxelox:
## feat(monitors): backend foundation — schema, entities, repositories, read-path [LAT-630]

Milestone **M2** of the Monitors feature: the read-path backend plus the schema it sits on. No write/evaluation path yet — this lands the data model, domain package, repositories, and the read hooks the Monitors UI consumes.

> Base: `development`. Part of the LAT-630 Monitors plan (`specs/monitors.md`).

### Schema (`@platform/db-postgres`)
- **New `monitors` + `monitor_alerts` tables**, both org-scoped via RLS and soft-deletable (`deleted_at`). `monitor_alerts` is **soft-delete-only** so historical `alert_incidents.monitor_alert_id` pointers stay resolvable after an alert is removed.
- **`alert_incidents` gains `monitor_alert_id` + `condition`** (condition = open-time snapshot, so mid-incident monitor edits don't change close-side logic or notification copy). The owning monitor is **not denormalised** — recovered by joining through `monitor_alerts`.
- Composite partial index `(monitor_alert_id, started_at DESC, id DESC) WHERE monitor_alert_id IS NOT NULL` serves both the saved-search equality lookups and the incidents-list **keyset pagination**.

### Domain
- **`@domain/shared`**: alert-incident kinds/severities (6 kinds; `low`/`medium`/`high`), `AlertIncidentCondition` discriminated union (threshold modes `absolute`/`multiplier`/`expected`, baselines `average`/`period`, `issue.escalating` sensitivity), and `MonitorId` / `MonitorAlertId` branded ids.
- **New `@domain/monitors` package**: `Monitor` / `MonitorAlert` entities, `MonitorRepository` port, and read-path use-cases — `listMonitors`, `getMonitorBySlug`, `getMonitorIncidents` (keyset paginated + idempotency-key hydration).
- **`AlertIncidentRepository`**: explicit org-scoping on every method (using `sqlClient.organizationId`), new `listByMonitorId` (join through `monitor_alerts`, surfaces incidents from soft-deleted alerts) and `listByMonitorAlertId`; keyset cursor over `(started_at, id)` replaces offset/`totalCount`.
- **`NotificationRepository.findExistingIdempotencyKeys`** for incident hydration.

### Web (`apps/web`)
- `@domain/monitors` server functions + collection hooks (`useMonitors`, `useMonitorIncidents`) on `useInfiniteQuery`; Monitors dashboard + incidents wired to infinite scroll.

### API (`apps/api`)
- Widened incident `sourceType` / `kind` / `severity` enums + descriptions; regenerated `openapi.json` / `mcp.json`.

### Design notes for reviewers
- **Soft-delete + no denormalized `monitor_id`**: an incident's monitor is recovered via `alert_incidents → monitor_alerts(monitorId)`. Keeping `monitor_alerts` soft-deletable means that join survives alert removal, so the monitor's incident history stays intact.
- **Keyset over offset**: a monitor can accumulate thousands of incidents; `(started_at, id)` keyset avoids deep-offset scans. `totalCount` was dropped (not needed for infinite scroll).
- **`condition` snapshot** is frozen at incident open time on purpose.

### Verification
- `pnpm typecheck` (75/75), `pnpm knip`, biome — all clean.
- `@domain/monitors` (20) + repository tests (31) pass.
- `openapi.json` / `mcp.json` regenerated; manifest CI in sync.


packages/domain/monitors/src/entities/monitor.ts
```diff
@@ -0,0 +1,79 @@
...
+  monitorId: monitorIdSchema,
+  kind: alertIncidentKindSchema,
+  source: z.object({
+    type: alertIncidentSourceTypeSchema,
+    id: cuidSchema.nullable(),
+  }),
+  condition: alertIncidentConditionSchema.nullable(),
+  severity: alertSeveritySchema,
+  createdAt: z.date(),
+})
```
csansoon, line 31: why does a monitor have a `kind` like in `alertIncident`?

neoxelox: talked offile

packages/domain/shared/src/alert-incident-condition.ts
csansoon, line 1: Take a look at how "Estalating" issues are identified. They can be used to detect "anomalies" in a monitor. An "anomaly" is a dynamic threshold, so it's not just "2 times more than yesterday", but it takes into account the day of the week and stuff like that. I think "2 times more than yesterday" is not a good alert, because it will be very common that Mondays can have more than twice the sessions than the previous day, a Sunday.

apps/web/src/domains/monitors/monitors.collection.ts
```diff
@@ -0,0 +1,103 @@
...
+  readonly limit?: number
+  readonly offset?: number
+  readonly searchQuery?: string
+  readonly enabled?: boolean
+}) {
+  const limit = input.limit ?? DEFAULT_MONITORS_PAGE_SIZE
+  const offset = input.offset ?? 0
+  const trimmedSearchQuery = input.searchQuery?.trim() || undefined
+
+  return useQuery({
```
andresgutgon: Is this not infinite paginated?

apps/web/src/domains/monitors/monitors.collection.ts
```diff
@@ -0,0 +1,103 @@
...
+export function useMonitorIncidents(input: {
+  readonly monitorId: string
+  readonly limit?: number
+  readonly offset?: number
+  readonly enabled?: boolean
+}) {
+  const limit = input.limit ?? DEFAULT_MONITORS_PAGE_SIZE
+  const offset = input.offset ?? 0
+
+  return useQuery({
```
andresgutgon: Same question

apps/web/src/routes/_authenticated/projects/$projectSlug/monitors/index.tsx
```diff
@@ -38,25 +90,31 @@ function MonitorsPage() {
...
                   size="sm"
                   className="w-64 pl-8 rounded-lg"
-                  disabled
                 />
                 <SearchIcon className="pointer-events-none absolute left-2.5 top-1/2 h-4 w-4 -translate-y-1/2 text-muted-foreground" />
               </div>
             </Layout.ActionRowItem>
             <Layout.ActionRowItem>
+              {/* The create-monitor modal lands in M5 — keep the button visible
+                  but disabled for now so the dashboard layout is final. */}
```
andresgutgon: Yapping

packages/domain/monitors/src/use-cases/get-monitor-incidents.ts
```diff
@@ -0,0 +1,93 @@
...
+  GetMonitorIncidentsResult,
+  RepositoryError,
+  SqlClient | AlertIncidentRepository | NotificationRepository
+> =>
+  Effect.gen(function* () {
+    const alertIncidentRepository = yield* AlertIncidentRepository
+    const notificationRepository = yield* NotificationRepository
+
+    const limit = Math.min(input.limit ?? DEFAULT_PAGE_SIZE, MAX_PAGE_SIZE)
+    const offset = Math.max(input.offset ?? 0, 0)
```
andresgutgon: no cursor pagination?

packages/platform/db-postgres/src/repositories/alert-incident-repository.ts
```diff
@@ -127,6 +137,70 @@ export const AlertIncidentRepositoryLive = Layer.effect(
...
+          // Join through `monitor_alerts` to recover the owning monitor — the id
+          // is not denormalised onto the incident. `monitor_alerts` is
+          // soft-deleted, so this still resolves incidents whose firing alert
+          // was later removed/edited (we intentionally do NOT filter
+          // `monitor_alerts.deleted_at`).
+          const where = and(
+            eq(alertIncidents.organizationId, sqlClient.organizationId),
+            eq(monitorAlerts.monitorId, monitorId),
+          )
+          const [rows, totals] = yield* sqlClient.query(async (db) => {
```
andresgutgon: second argument in query is organizationID. usually is how we get it and force the system to be correct

packages/platform/db-postgres/src/repositories/alert-incident-repository.ts
```diff
@@ -127,6 +137,70 @@ export const AlertIncidentRepositoryLive = Layer.effect(
...
+          }
+        }),
+      listByMonitorAlertId: ({ monitorAlertId, limit, offset }) =>
+        Effect.gen(function* () {
+          const sqlClient = (yield* SqlClient) as SqlClientShape<Operator>
+          const where = and(
+            eq(alertIncidents.organizationId, sqlClient.organizationId),
+            eq(alertIncidents.monitorAlertId, monitorAlertId),
+          )
+          const [rows, totals] = yield* sqlClient.query(async (db) => {
```
andresgutgon: Same

packages/platform/db-postgres/src/repositories/monitor-repository.ts
```diff
@@ -0,0 +1,168 @@
...
+                  isNull(monitorAlerts.deletedAt),
+                ),
+              ),
+          )
+          return toMonitor(row, alertRows.map(toMonitorAlert))
+        }),
+      list: ({ projectId, limit, offset, searchQuery }) =>
+        Effect.gen(function* () {
+          const sqlClient = (yield* SqlClient) as SqlClientShape<Operator>
+          const { organizationId } = sqlClient
```
andresgutgon, line 114: This is not the most used pattern. We do more 
```typescript
yield* sqlClient.query(async (db, organizationId) => {
  // ...
}
```

packages/platform/db-postgres/src/repositories/notification-repository.ts
```diff
@@ -173,6 +173,24 @@ export const NotificationRepositoryLive = Layer.effect(
           )
           return { deleted: deleted.length }
         }),
+
+      findExistingIdempotencyKeys: ({ organizationId, keys }) =>
```
andresgutgon, line 177: No need to pass `organizationId` no?

packages/platform/db-postgres/src/schema/alert-incidents.ts
```diff
@@ -24,6 +25,17 @@ export const alertIncidents = latitudeSchema.table(
...
+    // for legacy / flag-off issue-event incidents. The owning monitor is NOT
+    // denormalised here — it's recovered by joining through `monitor_alerts`
+    // (which is soft-deleted, never hard-deleted, so the join survives alert
+    // removal). Backs both "list incidents for monitor X" (join) and the
+    // saved-search state machines' "open incident for this alert" lookups.
+    monitorAlertId: cuid("monitor_alert_id", { default: false }),
+    // Snapshot of the firing alert's `condition` at the moment the incident
+    // opened. Monitor edits mid-incident don't move the goalposts — the
+    // close-side and the notification templates read the snapshot.
+    condition: jsonb("condition").$type<AlertIncidentCondition>(),
```
andresgutgon, line 34: Yapping to the max

packages/platform/db-postgres/src/schema/monitor-alerts.ts
```diff
@@ -0,0 +1,40 @@
...
+  (t) => [
+    organizationRLSPolicy("monitor_alerts"),
+    // Non-partial: serves both active-alert loads (filtered in the query) and
+    // the `alert_incidents` → monitor join, which must still resolve alerts
+    // that were soft-deleted after firing an incident.
+    index("monitor_alerts_monitor_idx").on(t.monitorId),
+    // Hot path: "find active monitor alerts watching savedSearch X" for the
+    // saved-search firing pipeline and the source-cascade handler. Partial on
+    // live rows — soft-deleted alerts never fire, so they're excluded here.
+    index("monitor_alerts_source_idx").on(t.organizationId, t.sourceType, t.sourceId).where(sql`deleted_at IS NULL`),
```
andresgutgon, line 27: Un yapp this please

packages/platform/db-postgres/src/repositories/notification-repository.ts
neoxelox: follows the same convention as all the other notif repo methods

packages/platform/db-postgres/src/repositories/monitor-repository.ts
neoxelox: we do use `yield* sqlClient.query`, its just the llm wanted to reuse some "where" conditions so it build them outside (because look at it, we do 2 queries to join monitor and monitor alerts. The real question here is if we should make just one query instead

packages/platform/db-postgres/src/repositories/alert-incident-repository.ts
neoxelox: it does that

packages/platform/db-postgres/src/repositories/alert-incident-repository.ts
neoxelox: it does that

packages/platform/db-postgres/src/repositories/monitor-repository.ts
neoxelox: The two-query parent+children pattern is the established idiom here.** `admin-organization-repository.ts` does exactly this — loads the org, then its members and projects as separate `sqlClient.query` calls, grouping with `new Map<string, …>`. Our `monitor-repository` matches that convention. The joins that *do* exist in the repo (`dataset-repository`, `organization-repository`) are 1-to-1 lookups or membership filters, not child-collection aggregation.

**A naïve `JOIN` would actually be worse, especially for `list`:**
- `monitors ⋈ monitor_alerts` fans out to one row per (monitor, alert) pair. You can't `LIMIT 50` that — it limits *pairs*, not monitors — so paginating `list` correctly would need a windowed/subquery dance. The two-query approach paginates monitors cleanly and fetches alerts for just the page's ids.
- For `findById`/`findBySlug` a join works, but you still dedupe the repeated monitor columns in app code — i.e. the same grouping logic we already have.

**The genuinely-clean single-query option isn't really available:**
- Drizzle's relational `with: { alerts: true }` (which *would* be one tidy query) needs the db created as `drizzle(client, { schema })` plus `relations()` definitions — this repo does `drizzle({ client })` with no schema and zero `relations()` anywhere. Wiring that up touches the shared client for everyone.
- A correlated `json_agg` subquery collapses it to one query and paginates fine, but `json_agg(a.*)` returns snake_case keys and ISO-string timestamps, so I'd need a second, parallel mapping path (manual snake_case + date parsing) before `monitorSchema.parse` — more fragile than reusing `toMonitorAlert` on typed Drizzle rows.

**On round-trips** (the real thing your reviewer's instinct is poking at): `findById` already issues the monitor + alerts together in one `Promise.all`. `findBySlug` and `list` do a sequential second query because the alerts query needs the resolved id(s) — but these are cold detail-panel / dashboard reads with tiny child sets (≤50 monitors, a handful of alerts each), on a pooled connection. The extra round-trip is negligible and buys type-safe, convention-matching code.

So my recommendation: **keep it.** It's idiomatic, correct, and a join/json_agg trades clarity for no meaningful win.

If you'd still rather minimize round-trips, the one I'd consider is a `json_agg` correlated subquery (one query per method, pagination intact) — I'd just write a small `parseAlertJson` mapper for the snake_case/date conversion. Say the word and I'll do that variant; otherwise I'll leave the two-query version as-is.

packages/domain/monitors/src/use-cases/get-monitor-incidents.ts
neoxelox: fixed

apps/web/src/domains/monitors/monitors.collection.ts
neoxelox: fixed

apps/web/src/domains/monitors/monitors.collection.ts
neoxelox: fixed

andresgutgon (approved):

PR #3365, opened 2026-06-02
feat(monitors): details panel, consistent kind naming, and system-monitor reset action [LAT-630]

neoxelox:
Part of the Monitors MVP (LAT-630). Two pieces of work on the Monitors page and its backoffice tooling, building on the M4 backend (monitor + alert mutations) already merged.

## Monitor details panel

The Monitors dashboard now opens a detail drawer for the selected monitor, mirroring the issue details panel:

- **Header** — monitor name, muted-state description, and a slug row laid out as `[System tag] slug [Muted badge]`. System monitors carry a Latitude-logo "System" tag.
- **Drawer actions** — next/prev navigation chevrons (`Alt+↑/↓`, `J`/`K`) on the left; a Mute/Unmute button on the right, in the same position as the issue panel's lifecycle actions. Muting/unmuting goes through a shared confirmation modal.
- **Alerts section** (megaphone icon) — read-only cards rendering each alert's human-readable summary.
- **Incidents section** (shield-alert icon) — an infinite table of the monitor's incidents (Started / Status / Source / Notified), keyset-paginated.

The dashboard's three-dots options column exposes Mute/Unmute (functional) plus Rename/Delete (present but disabled — wired up in the next milestone).

## Consistent kind naming

Unified the alert-incident kind labels behind a single source of truth, `ALERT_INCIDENT_KIND_LABEL` in `@domain/shared`, and reworded the system-monitor copy so every surface agrees:

- `issue.new` → **Issue discovered**, `issue.regressed` → **Issue regressed**, `issue.escalating` → **Issue escalating** (plus the saved-search kinds).
- Alert descriptions now start with "Alerts …", say "detected" (not "occurred"), and refer to "traces" — applied across the detail drawer, incident markers, the incident-event email/Slack templates, the system-monitor definitions, and the not-yet-shipped backfill migration.
- API/MCP entity descriptions reworded; `openapi.json` and `mcp.json` regenerated (idempotent — no SDK regen, that stays batched to the end of the plan).

## Backoffice: reset system monitors

A staff-only org action (next to **Create demo project**) that re-provisions the three system monitors to their **current** definitions across every project in a chosen org. Unlike the idempotent provision path (skip-if-exists), this **overwrites** the monitors' titles, descriptions, and default alert-condition values — so edits to the definitions can be pushed onto an org's existing system monitors.

- `buildSystemMonitors({ organizationId, projectId })` was extracted from the provision use-case so the definitions live in one place and both paths share them.
- `MonitorRepository.resetSystemMonitors` upserts each monitor in one transaction: it keeps the existing row (and its `mutedAt`), overwrites name/description, soft-deletes the live alerts and inserts fresh ones from the definition. **Mute state and incident history are preserved**; a slug already held by a user-created monitor is left untouched.
- The repo filters by the entity's `projectId` rather than the SqlClient org, because admin handlers run RLS-off under the "system" scope.
- The web surface is a confirmation modal behind `adminResetSystemMonitors` (admin middleware → input validation → admin Postgres client), reporting `Reset N system monitors across M projects`.

## Validation

typecheck (`@app/web`, `@domain/admin`, `@domain/monitors`, `@platform/db-postgres`), biome, and knip are clean. Tests: `@domain/monitors` 38, `@domain/admin` 39, `@platform/db-postgres` 248 (incl. 5 new `resetSystemMonitors` integration tests), web schema 20.

### DIFF
--- a/apps/web/src/domains/alerts/incident-markers.ts
+++ b/apps/web/src/domains/alerts/incident-markers.ts
@@ -1,4 +1,5 @@
 import type { AlertIncidentKind, AlertSeverity } from "@domain/alerts"
+import { ALERT_INCIDENT_KIND_LABEL } from "@domain/shared"
 import type { BarChartOverlay, BarChartOverlayArea, BarChartOverlayLine } from "@repo/ui"
 import type { AlertIncidentRecord } from "./alerts.functions.ts"
 
@@ -33,14 +34,8 @@ const KIND_TOP_SYMBOL: Record<AlertIncidentKind, TopSymbol> = {
   "savedSearch.escalating": { shape: "rect", size: 7 },
 }
 
-export const KIND_LABELS: Record<AlertIncidentKind, string> = {
-  "issue.new": "New issue",
-  "issue.regressed": "Issue regressed",
-  "issue.escalating": "Issue escalating",
-  "savedSearch.match": "Search match",
-  "savedSearch.threshold": "Search threshold",
-  "savedSearch.escalating": "Search escalating",
-}
+/** Canonical kind labels live in `@domain/shared`; re-exported so chart markers stay in sync with the monitor UI. */
+export const KIND_LABELS = ALERT_INCIDENT_KIND_LABEL
 
 export const SEVERITY_LABELS: Record<AlertSeverity, string> = {
   low: "Low",
--- a/apps/web/src/domains/alerts/incident-markers.ts
+++ b/apps/web/src/domains/alerts/incident-markers.ts
@@ -34,9 +34,6 @@ const KIND_TOP_SYMBOL: Record<AlertIncidentKind, TopSymbol> = {
   "savedSearch.escalating": { shape: "rect", size: 7 },
 }
 
-/** Canonical kind labels live in `@domain/shared`; re-exported so chart markers stay in sync with the monitor UI. */
-export const KIND_LABELS = ALERT_INCIDENT_KIND_LABEL
-
 export const SEVERITY_LABELS: Record<AlertSeverity, string> = {
   low: "Low",
   medium: "Medium",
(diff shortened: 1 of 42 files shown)


apps/web/src/domains/alerts/incident-markers.ts
```diff
@@ -33,14 +34,8 @@ const KIND_TOP_SYMBOL: Record<AlertIncidentKind, TopSymbol> = {
...
-export const KIND_LABELS: Record<AlertIncidentKind, string> = {
-  "issue.new": "New issue",
-  "issue.regressed": "Issue regressed",
-  "issue.escalating": "Issue escalating",
-  "savedSearch.match": "Search match",
-  "savedSearch.threshold": "Search threshold",
-  "savedSearch.escalating": "Search escalating",
-}
+/** Canonical kind labels live in `@domain/shared`; re-exported so chart markers stay in sync with the monitor UI. */
+export const KIND_LABELS = ALERT_INCIDENT_KIND_LABEL
```
csansoon: Why not use `ALERT_INCIDENT_KIND_LABEL` directly?

csansoon (dismissed):

apps/web/src/domains/alerts/incident-markers.ts
neoxelox: Good call — no reason. This file used to have its own inline `KIND_LABELS` map (pre-PR), and when I unified the labels onto `ALERT_INCIDENT_KIND_LABEL` I kept the name as an alias instead of replacing the usages. Removed the alias in `71cb3c3ad`: `formatIncidentKindLabel` now reads `ALERT_INCIDENT_KIND_LABEL` directly, and `incident-marker-popover.tsx` (the only other consumer) uses that helper. Also swept the rest of the PR — every other reference already indexes the canonical constant directly, so this was the only alias.

PR #3371, opened 2026-06-02
refactor(traces): scope outlier baselines project-wide

andresgutgon:
## Summary

- Outlier badges (p90/p95/p99) on traces and sessions now compare against a single **project-wide** percentile baseline instead of a per-tag-set cohort.
- On low-volume or noisy-tagged projects, per-tag cohorts often never reached the sample gate (30 for p90, 100 for p95), so users never saw a badge. Collapsing to a project-wide cohort fixes that and matches our "one project per AI flow" recommendation.
- One Redis key per project (was per tag combination) and one TanStack Query per project (was N per visible tag set), so fewer round-trips from the traces/sessions table.

## What changed

- Drop the `HAVING length(tags) = N AND hasAll(tags, [...])` filter from the cohort SQL in both `trace-repository.ts` and `session-repository.ts`; rename the method to `getCohortBaseline`.
- Rename use-cases, server functions, and hooks accordingly (`...ByTags` → `...`), drop `tags` from their inputs.
- Drop the `tags` prop from `TraceOutlierBadge` / `SessionOutlierBadge` and update every call site (`project-traces-table.tsx`, `sessions-view.tsx`, trace + session detail drawers).
- Tooltip copy now reads "of the traces in this project" / "of the sessions in this project."
- Update fakes, port interfaces, and tests; rewrote the cohort test suites to cover project-wide aggregation, zero-value filtering, sample gating, and `excludeTraceId` / `excludeSessionId`.

## Test plan

- [x] `pnpm typecheck` (75/75 packages clean locally)
- [x] `pnpm biome check` clean locally
- [x] `pnpm --filter @platform/db-clickhouse test -- trace-repository session-repository` — cohort tests pass
- [x] Manual: open a low-volume project's Traces tab → expect to now see `p90/p95` badges next to outlier rows
- [x] Manual: open a high-volume project's Sessions tab → badge tooltip says "of the sessions in this project"

### DIFF
--- a/apps/web/src/domains/sessions/sessions.collection.ts
+++ b/apps/web/src/domains/sessions/sessions.collection.ts
@@ -6,7 +6,7 @@ import { keepPreviousData, useInfiniteQuery, useQuery } from "@tanstack/react-qu
 import { useMemo } from "react"
 import {
   countSessionsByProject,
-  getSessionCohortSummaryByTags,
+  getSessionCohortSummary,
   getSessionDetail,
   getSessionDistinctValues,
   getSessionDistribution,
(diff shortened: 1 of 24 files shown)


andresgutgon:
@claude do review

andresgutgon:
Thanks for the review — addressed in `044995be`.

**Finding 1 (stale docstring)** — fixed. Rewrote the `cohort-baselines.ts` header to describe a cohort as "every trace (or session) in the project."

**Finding 2 (test data accumulating across subtests)** — false alarm. `setupTestClickHouse` in `packages/platform/testkit/src/clickhouse/test-clickhouse.ts:166-169` registers a `beforeEach` that calls `truncateClickHouseTables`, which `TRUNCATE`s every user table between subtests. The local `beforeEach` in `trace-repository.test.ts` then re-inserts `BASELINE_SPANS`/`BASELINE_SCORES` from scratch. So each `it` starts with only the seeded data plus its own inserts — the prior subtest's rows are gone. No change needed.

**Finding 3 (unused `excludeTraceId`/`excludeSessionId`)** — intentional. Documented it in both `getTraceCohortSummaryUseCase` and `getSessionCohortSummaryUseCase`: the badge is meant to compare against a stable project-wide reference, so the row being viewed is included in its own baseline. Left the port-level support in place for a hypothetical future "compare this trace to everything else" view.

csansoon (approved):

PR #3380, opened 2026-06-03
perf(search): kick traces off the 2 GB cliff (LAT-649)

andresgutgon:
## TL;DR

Every session search on an XL project was reading **2.19 GB** from `traces`. Even when the search returned 2 results. Even when it returned 0 results. ClickHouse was just… reading everything, every time.

After this PR: **14 MB**.

That's a **~156× reduction in bytes scanned** and a **~25× speedup** on the heavy part of the search query.

Closes [LAT-649](https://linear.app/latitude/issue/LAT-649).

## The receipts

Real measurements against real project. Varying how many candidate trace_ids the search index returns to the rollup ("fan-out"):

| Fan-out | Old (single JOIN query) | New (literal-IN rollup) | Speedup | Bytes saved |
|--|--|--|--|--|
| 2 | 111 ms / 117 MB | ~25 ms / 13 MB | ~4× | ~9× |
| 412 | 775 ms / 2.15 GB | 82 ms / 19 MB | ~9× | **~113×** |
| **8,464** | **1,381 ms / 2.19 GB** | **54 ms / 14 MB** | **~25×** | **~156×** |

The bigger the search, the bigger the win. The old code's read cost was **flat in candidate count**: always ~2 GB. The new code's read cost is **proportional to actual matches**.

## What was going wrong

The old code wraps the candidate set and the `traces` rollup in one statement, joined inline:

```sql
WITH search_results AS (SELECT trace_id, score FROM <plan.subquery>),
     trace_rollup AS (
       SELECT ...everything you'd ever want about a trace...
       FROM traces t
       INNER JOIN search_results ON t.trace_id = search_results.trace_id
       WHERE org = ? AND project = ?
       GROUP BY ...
     )
SELECT ... FROM trace_rollup GROUP BY session_id
```

This looks elegant. ClickHouse looked at it and said:

> *"Cool. I can narrow `traces` by `(organization_id, project_id)` because those are the primary key. The JOIN's `trace_id` set, though — no idea what that is until I actually scan. I'll just read every AggregateFunction column for every trace in the org+project subset and filter post-join."*

For this project that's 398k trace rows × ~5 KB of AggregateFunction state ≈ **2.19 GB, every search, regardless of how many traces actually matched**. Confirmed via `EXPLAIN indexes=1`: 1,270 / 6,907 granules selected, `trace_id` never reaches the storage filter.

## What we changed

`search-by-project.ts` now does **two queries** instead of one:

1. **Candidate fetch** — run `plan.subquery` on its own. Returns the small `(trace_id, relevance_score)` set the search engine produced. Same lexical/semantic work as before, just on its own roundtrip.

2. **Rollup with parameter-bound IN** — feed the candidate ids and scores back as arrays:
   ```sql
   WITH
     mapFromArrays({traceIds:Array(String)}, {relevanceScores:Array(Float64)}) AS scoreByTrace,
     trace_rollup AS (
       SELECT ...,
         any(scoreByTrace[toString(t.trace_id)]) AS relevance_score
       FROM traces t
       WHERE org = ? AND project = ?
         AND t.trace_id IN ({traceIds:Array(String)})    -- ← this is the fix
       GROUP BY org, project, trace_id
     )
   SELECT ... FROM trace_rollup GROUP BY session_id
   ```

Because the trace_id set arrives as a **literal array at query-plan time**, ClickHouse PREWHEREs the cheap `trace_id` column first and only materializes the heavy AggregateFunction columns for matching rows. The 2 GB read evaporates.

The `fetchFullSessions` step at the end of the listing path is unchanged. `relevance_score` propagation moves from a JOIN-column reference to a Map lookup keyed by `trace_id` — same data, different plumbing.

## Test plan

- [x] `pnpm --filter @platform/db-clickhouse typecheck` clean
- [x] All 46 tests in `session-repository.test.ts` pass — covers literal phrases, token phrases, semantic prompts, hybrid plans, cursors, sort axes, orphan-trace fallback, and the search-active count path
- [ ] Smoke-test on staging: phrase, semantic, and hybrid searches with pagination
- [ ] Watch `query_log.read_bytes` on the rollup query post-deploy — expect a 2-order-of-magnitude drop on XL projects

### DIFF
--- a/packages/platform/db-clickhouse/src/repositories/search-by-project.ts
+++ b/packages/platform/db-clickhouse/src/repositories/search-by-project.ts
@@ -26,14 +26,18 @@ import { Effect } from "effect"
 import { buildClickHouseWhere } from "../filter-builder.ts"
 import { SESSION_FIELD_REGISTRY } from "../registries/session-fields.ts"
 import { buildScoreRollupSubquery, splitScoreFilters } from "../score-filter-subquery.ts"
-import { planSearch } from "./search-plan.ts"
+import { MAX_SEARCH_CANDIDATES, planSearch, type SearchPlan } from "./search-plan.ts"
 
 /**
  * Session-level search read path. The trace-level `search-plan.ts` provides
- * `(trace_id, relevance_score)` candidates; here we wrap that subquery in a
- * `search_results → trace_rollup → session_rollup` CTE chain (spec §4.3) so
- * results collapse to one row per session with the matching-trace metadata
- * the UI needs ("N matching turns", drill-in on `bestTraceId`, etc.).
+ * `(trace_id, relevance_score)` candidates; we materialize that subquery
+ * via `fetchSearchCandidates` in its own roundtrip, then feed the candidate
+ * trace_ids and scores as parameter-bound arrays into a
+ * `trace_rollup → session_rollup` query so results collapse to one row per
+ * session with the matching-trace metadata the UI needs ("N matching
+ * turns", drill-in on `bestTraceId`, etc.). The split keeps the `traces`
+ * read on ClickHouse's PREWHERE-friendly path — see LAT-649 for the cliff
+ * the previous single-statement JOIN form hit on XL projects.
  *
  * The functions in this file are pure read-side: they own the search SQL and
  * the per-session row → domain mapping, but defer to a `fetchFullSessions`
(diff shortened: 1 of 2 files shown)


andresgutgon:
@claude be super strict review this fix carefully

andresgutgon:
Addressed in `b3919e79b`. Thanks both for the thorough pass.

### 🔴 Critical — candidate cap (@claude)

Added `MAX_SEARCH_CANDIDATES = 50_000` in `search-plan.ts` alongside `SEMANTIC_SCAN_LIMIT` and applied it as `LIMIT {candidateCap:UInt32}` in `fetchSearchCandidates`. JSDoc spells out why: semantic plans already cap server-side, lexical/hybrid don't, so a broad LIKE on an XL project could stream millions of rows into the Node worker. 50k keeps `Array(FixedString(32))` serialization under ~2 MB.

### 🟡 Dedupe in candidate fetch (@copilot)

Good catch on the pre-merge dup window in `trace_search_documents`. Candidate query now:

```sql
SELECT trace_id, max(relevance_score) AS relevance_score
FROM (${plan.subquery})
GROUP BY trace_id
LIMIT {candidateCap:UInt32}
```

`max(relevance_score)` keeps the best score per trace, which also matches what `mapFromArrays` would have silently picked on the duplicate-key path — but now the choice is explicit and deterministic instead of last-write-wins.

### 🟡 `Array(FixedString(32))` instead of `Array(String)` (@copilot)

Switched both list and count paths. The map keys, the `IN`-list, and `t.trace_id` are all `FixedString(32)` now — no more implicit cast on the PREWHERE-critical column. Dropped the `toString(t.trace_id)` from the map lookup too.

### 🟡 `any()` → `max()` (@claude)

Done. Lines up with the other `max()`/`min()` aggregates in the rollup, and removes the false "non-determinism" smell on a value that's actually a deterministic Map lookup.

### 🟡 Implicit `scoreByTrace` dependency on `TRACE_ROLLUP_BODY` (@claude)

Added a `**Coupling**:` section to the JSDoc explicitly stating the required outer `WITH` declaration. Doesn't make it compile-time-safe — TypeScript can't model SQL scope — but at least the next person who touches this won't get an unexplained CH parser error.

### 🔵 Deliberately not addressed — count path unused `relevanceScores` (@claude)

Keeping the parallel arrays as-is. With the 50k cap that's a worst case of ~400 KB extra payload per count request — under the threshold where splitting `TRACE_ROLLUP_BODY` into list/count variants pays for itself. Happy to revisit if the count path ever shows up in latency profiling.

neoxelox (approved):

PR #3382, opened 2026-06-03
feat(traces): add duration composition bar to trace and session drawers

csansoon:
## summary

Replaces the plain "Duration" key-value in the trace and session detail drawers with a **duration composition bar**: a breakdown of where wall-clock time went — AI generation, tool execution, retrieval, and idle (orchestration / network / waiting). All of it is derived from span timings we already ingest; no new telemetry.

The bar is rendered as a measured "ruler" (square segments, a `|` tick at each edge and between every segment) so it reads as *time* and stays visually distinct from the rounded token/cost bars above it. Idle is drawn as a hatched gap rather than a solid color, so "working vs waiting" is legible without a legend. It sits in the same usage group as Tokens and Cost, and the duration outlier badge moves onto it.

image

## how time is attributed

`computeDurationBreakdown(spans)` partitions the trace's wall-clock timeline:

- **Leaf spans only.** A span is a leaf if no other span lists it as parent. Container spans (`invoke_agent`, `chain`) are excluded so their children aren't double-counted; time a container holds beyond its children falls into idle.
- **Operation → category:** generation (`chat`, `text_completion`), tool (`execute_tool`), retrieval (`retrieval`, `reranker`), other (everything else).
- **Priority-partition sweep.** Because sibling spans can run in parallel, summing per-category durations would exceed wall-clock. Instead the timeline is swept and each instant attributed to one category by priority (generation > tool > retrieval > other), with uncovered instants counted as idle. Segments therefore sum exactly to the wall-clock total.

Times come from the spans' ISO `startTime`/`endTime`, so resolution is milliseconds — the nanosecond precision from ClickHouse doesn't survive serialization, which is fine at trace scale.

## sessions

Sessions span multiple traces with gaps between turns. `computeSessionDurationBreakdown` computes the breakdown **per trace and sums the categories**, so the gaps *between* traces (user think-time) are never counted as idle — only idle *within* each trace is. The bar's total then equals the sum of per-trace wall-clocks, matching how `session.durationNs` is already derived.

This needs spans for a whole session, which had no fetch path. Added `SpanRepository.listByTraceIds` (port + ClickHouse impl + test fake) — it filters `trace_id IN (...)` rather than `session_id`, which also covers orphan single-trace sessions whose spans carry no `session_id`. Same `LIMIT 1 BY span_id` dedupe and start-time partition pruning as `listByTraceId`. Exposed via `listSpansBySession` + `useSpansBySessionCollection`.

## data fetching

The trace tab's bar needs the span list, which previously only loaded when the Spans tab opened. Lifted `useSpansByTraceCollection` to `TraceDetailBody`; the Spans tab still fetches with the same key, so the cached collection dedupes both into one fetch (this also left the session panel's `SpansTab` caller untouched).

## tested

Unit tests for both helpers (`duration-composition.test.ts`): serial gaps → idle, parallel siblings → no double-count, container exclusion, segments-sum-to-wall-clock, and the session case asserting between-trace gaps are not idle. Typecheck + Biome clean across `@domain/spans`, `@platform/db-clickhouse`, and `web`.

Not yet validated against live data in the running app.

### DIFF
--- a/apps/web/src/domains/spans/spans.functions.ts
+++ b/apps/web/src/domains/spans/spans.functions.ts
@@ -170,6 +170,33 @@ export const listSpansByTrace = createServerFn({ method: "GET" })
     return spans.map(serializeSpan)
   })
 
+export const listSpansBySession = createServerFn({ method: "GET" })
+  .inputValidator(
+    z.object({
+      projectId: z.string(),
+      traceIds: z.array(z.string()),
+      startTimeFrom: dateTimeParamSchema.optional(),
+      startTimeTo: dateTimeParamSchema.optional(),
+    }),
+  )
+  .handler(async ({ data }): Promise<SpanRecord[]> => {
+    const { organizationId } = await requireSession()
+    const orgId = OrganizationId(organizationId)
+    const spans = await Effect.runPromise(
+      Effect.gen(function* () {
+        const repo = yield* SpanRepository
+        return yield* repo.listByTraceIds({
+          organizationId: orgId,
+          projectId: ProjectId(data.projectId),
+          traceIds: data.traceIds.map((id) => TraceId(id)),
+          ...(data.startTimeFrom ? { startTimeFrom: data.startTimeFrom } : {}),
+          ...(data.startTimeTo ? { startTimeTo: data.startTimeTo } : {}),
+        })
+      }).pipe(withClickHouse(SpanRepositoryLive, getClickhouseClient(), orgId), withTracing),
+    )
+    return spans.map(serializeSpan)
+  })
+
 export const mapConversationToSpans = createServerFn({ method: "GET" })
   .inputValidator(z.object({ projectId: z.string(), traceId: z.string() }))
   .handler(
--- a/apps/web/src/domains/spans/spans.functions.ts
+++ b/apps/web/src/domains/spans/spans.functions.ts
@@ -1,4 +1,4 @@
-import { OrganizationId, ProjectId, SpanId, TraceId } from "@domain/shared"
+import { OrganizationId, ProjectId, SessionId, SpanId, TraceId } from "@domain/shared"
 import type { Operation, Span, SpanDetail, SpanKind, SpanStatusCode } from "@domain/spans"
 import { buildConversationSpanMaps, SpanRepository, TraceRepository } from "@domain/spans"
 import { withAi } from "@platform/ai"
@@ -174,7 +174,7 @@ export const listSpansBySession = createServerFn({ method: "GET" })
   .inputValidator(
     z.object({
       projectId: z.string(),
-      traceIds: z.array(z.string()),
+      sessionId: z.string(),
       startTimeFrom: dateTimeParamSchema.optional(),
       startTimeTo: dateTimeParamSchema.optional(),
     }),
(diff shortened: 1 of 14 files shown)


apps/web/src/domains/spans/spans.functions.ts
```diff
@@ -170,6 +170,33 @@ export const listSpansByTrace = createServerFn({ method: "GET" })
...
+      startTimeTo: dateTimeParamSchema.optional(),
+    }),
+  )
+  .handler(async ({ data }): Promise<SpanRecord[]> => {
+    const { organizationId } = await requireSession()
+    const orgId = OrganizationId(organizationId)
+    const spans = await Effect.runPromise(
+      Effect.gen(function* () {
+        const repo = yield* SpanRepository
+        return yield* repo.listByTraceIds({
```
andresgutgon: Why not doing this in SQL? `repo.listBySessionId` and inside it finds single trace session or multi trace session.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/session-detail-drawer/metadata-tab.tsx
```diff
@@ -135,7 +141,16 @@ export function MetadataTab({
         </div>
       )}
 
-      <UsageSummary data={session} costBadges={costBadgesNode} />
+      {/* ── Duration composition + usage (tokens + cost) ── */}
```
andresgutgon: Yapping

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/trace-detail-drawer/tabs/trace-tab.tsx
```diff
@@ -77,12 +84,13 @@ export function TraceTab({
...
 
-  const durationValue = traceRecord ? (
-    <span className="flex items-center gap-1">
-      {renderBadge("durationNs", traceRecord.durationNs)}
-      {traceRecord.durationNs > 0 ? formatDuration(traceRecord.durationNs) : "-"}
-    </span>
-  ) : undefined
+  // Duration composition from the span list (leaf intervals partitioned by
+  // category). Falls back to the trace record's wall-clock total (ns → ms) until
+  // spans load or when none are available, so the total always renders.
```
andresgutgon: Yapping

apps/web/src/domains/spans/spans.functions.ts
csansoon: Done in `50e5a60` — switched to `repo.listBySessionId(sessionId)`. The membership predicate mirrors the `sessions_mv` grouping key exactly: `coalesce(nullIf(session_id, ''), toString(trace_id)) = {sessionId}`, so single-trace (orphan) and multi-trace sessions resolve in one query. Also drops the `traceIds` payload from the client and the 500-trace cap.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/session-detail-drawer/metadata-tab.tsx
csansoon: Removed in `50e5a60`.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/trace-detail-drawer/tabs/trace-tab.tsx
csansoon: Trimmed to one line in `50e5a60`.

andresgutgon (approved):

PR #3391, opened 2026-06-04
fix(analytics): make PostHog org-group attribution work end to end

cesr:
## Why

Org-based retention / funnels / "WAU per org" aggregate on PostHog's **`organization` group** (group type index 0), and that join uses the **materialized `$group_0` column**. The activation→return retention insight ([nexRDBsZ](https://eu.posthog.com/project/158550/insights/nexRDBsZ)) read **all zeros**.

Root-caused against **live PostHog data** (project 158550) — two independent gaps, one backend, one frontend. Both must ship together or retention stays broken.

## Backend: `FirstTraceReceived` was anonymous → no `$group_0`

PostHog only materializes `$group_0` for **identified** events. `FirstTraceReceived` fires off **API-key trace ingestion** (no acting user), so it was captured with `$process_person_profile: false` (anonymous) and got no group column.

Evidence (per-row resolution over 90 days, by `$process_person_profile`):

| events | person profile | n | with `$group_0` |
|---|---|---|---|
| ProjectCreated / ApiKeyCreated / UserSignedUp / OrganizationCreated | enabled | 1326 | **100%** |
| FirstTraceReceived + user-less variants | **false** | 223 | **0%** |

Same posthog-node `capture({ groups })` call in every case — the only variable that flips `$group_0` 100%→0% is the anonymous flag. So it was **not** a `groups=`-vs-`properties.$groups` SDK bug.

**Fix — attribute the milestone to the org owner (no synthetic person):** `checkFirstTrace` resolves the org owner via `MembershipRepository` (best-effort; falls back to earliest member, then to no actor) and stamps `actorUserId` on the event. It then flows through the existing real-user path → identified → `$group_0` materializes, attributed to a real human who already has frontend events. The lookup runs **once per project** (gated by `firstTraceAt`).

Rejected alternative: making the event identified under a synthetic `organization_<id>` distinct id. That works but creates one synthetic person per org (tens–hundreds across the product, inflating unique-user counts). Owner attribution avoids that entirely. `payload-mapping.ts` is therefore **unchanged** — genuinely user-less events stay cheap/anonymous.

## Frontend: the first authenticated pageview was stranded before `group()`

`opt_in_capturing()` fires the session's first `$pageview`. `identify()` / `group()` ran *after* it, so that pageview had no `$group_0` and no `organizationId`. A real session trace shows the opt-in `$pageview` landing **1 ms before `$identify`**, attributed to the anonymous distinct_id.

**Fix:**
- Run `register()` + `identify()` + `group()` **before** `opt_in_capturing()`.
- `register({ organizationId })` so events carry it as a property even before `group()` resolves.
- Add `organizationId` to `identify()`, and `slug` + `plan` to `group()`. (`$groupidentify` updates group props, so existing orgs backfill slug/plan on next login — no migration.)

## Scope / notes

- **Historical data can't backfill:** the 124 already-ingested anonymous `FirstTraceReceived` events stay without `$group_0`. The insight populates from **new** activation cohorts post-deploy.
- The insight query itself was already correct (`aggregation_group_type_index: 0`); only its description was corrected.
- Backend was otherwise already correct (events carry `actorUserId` → real-user distinctIds; `groups` set on every capture).

## Test plan

- `@platform/analytics-posthog` unit tests (13) + typecheck green; `@domain/events` typecheck green; `apps/workers/projects.ts` typechecks clean once workspace deps are built.
- Post-deploy verification (PostHog Live Events, project 158550):
  1. `FirstTraceReceived` (last hour) → `distinct_id` is the org owner's uuid, `$group_0` set, no `$process_person_profile:false`.
  2. `$pageview` from `console.latitude.so` → `$group_0` + `organizationId` set, incl. the session's first authenticated pageview.
  3. nexRDBsZ retention cells fill for orgs that activate after deploy.

### DIFF
--- a/apps/web/src/routes/_authenticated.tsx
+++ b/apps/web/src/routes/_authenticated.tsx
@@ -67,11 +67,25 @@ export const Route = createFileRoute("/_authenticated")({
 
     const supportIdentity = await getSupportUserIdentity()
 
+    // Plan tags the PostHog org group. Resolve it server-side here (rather than
+    // a client billing query) and skip it for sessions we don't track anyway
+    // (staff / impersonation), which never read it.
+    const excludeFromAnalytics = isLatitudeStaffEmail(session.user.email) || impersonatedBy != null
+    let organizationPlan: string | null = null
+    if (!excludeFromAnalytics) {
+      try {
+        organizationPlan = (await getBillingOverview()).planSlug
+      } catch {
+        organizationPlan = null
+      }
+    }
+
     return {
       user: session.user,
       organizationId,
       impersonatedBy,
       supportIdentity,
+      organizationPlan,
     }
   },
   component: AuthenticatedLayout,
@@ -356,19 +370,9 @@ function AuthenticatedLayout() {
     // `routeId` is the stable file-route id from `createFileRoute(...)`.
     select: (s) => s.matches.some((m) => m.routeId === projectOnboardingRouteId),
   })
+  const organizationPlan = Route.useLoaderData({ select: (data) => data.organizationPlan })
   const { data: allOrgs } = useOrganizationsCollection()
   const org = allOrgs?.find((o) => o.id === organizationId)
-  const excludeFromAnalytics = isLatitudeStaffEmail(user.email) || impersonatedBy != null
-  // Only used to tag the PostHog org group with its plan, so skip it for
-  // excluded sessions (staff / impersonation) where we opt out of capturing
-  // anyway. Shares BillingCreditCounter's query key, so when the header is
-  // mounted this dedupes rather than firing a second request.
-  const { data: billingOverview } = useQuery({
-    queryKey: ["billing", "overview", organizationId],
-    queryFn: () => getBillingOverview(),
-    staleTime: 30_000,
-    enabled: !excludeFromAnalytics,
-  })
 
   return (
     <IntercomProvider identity={supportIdentity} floatingButton="none">
@@ -382,8 +386,8 @@ function AuthenticatedLayout() {
             organizationId={organizationId}
             organizationName={org?.name}
             organizationSlug={org?.slug}
-            organizationPlan={billingOverview?.planSlug}
-            excludeFromAnalytics={excludeFromAnalytics}
+            organizationPlan={organizationPlan}
+            excludeFromAnalytics={isLatitudeStaffEmail(user.email) || impersonatedBy != null}
           />
           {impersonatedBy && <ImpersonationBanner impersonatedUserEmail={user.email} />}
           {isProjectOnboarding ? null : <NavHeader />}
--- a/apps/web/src/routes/_authenticated.tsx
+++ b/apps/web/src/routes/_authenticated.tsx
@@ -358,6 +358,13 @@ function AuthenticatedLayout() {
   })
   const { data: allOrgs } = useOrganizationsCollection()
   const org = allOrgs?.find((o) => o.id === organizationId)
+  // Shares the query key with BillingCreditCounter so this is deduped, not a
+  // second request. Used to tag the PostHog org group with its plan.
+  const { data: billingOverview } = useQuery({
+    queryKey: ["billing", "overview", organizationId],
+    queryFn: () => getBillingOverview(),
+    staleTime: 30_000,
+  })
 
   return (
     <IntercomProvider identity={supportIdentity} floatingButton="none">
@@ -370,6 +377,8 @@ function AuthenticatedLayout() {
             userName={user.name}
             organizationId={organizationId}
             organizationName={org?.name}
+            organizationSlug={org?.slug}
+            organizationPlan={billingOverview?.planSlug}
             excludeFromAnalytics={isLatitudeStaffEmail(user.email) || impersonatedBy != null}
           />
           {impersonatedBy && <ImpersonationBanner impersonatedUserEmail={user.email} />}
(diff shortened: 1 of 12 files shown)


apps/web/src/routes/_authenticated.tsx
```diff
@@ -358,6 +358,13 @@ function AuthenticatedLayout() {
   })
   const { data: allOrgs } = useOrganizationsCollection()
   const org = allOrgs?.find((o) => o.id === organizationId)
+  // Shares the query key with BillingCreditCounter so this is deduped, not a
+  // second request. Used to tag the PostHog org group with its plan.
+  const { data: billingOverview } = useQuery({
+    queryKey: ["billing", "overview", organizationId],
+    queryFn: () => getBillingOverview(),
+    staleTime: 30_000,
+  })
```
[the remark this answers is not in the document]
cesr: Good catch — fixed in 5e5e04e7b by gating the query on `!excludeFromAnalytics`, so it no longer fetches for staff/impersonation sessions (which opt out of capturing and never read the plan).

cesr:
Note for reviewers: the Copilot review summary above is **stale** — it was generated before a force-push and describes an earlier approach (removing `$process_person_profile:false` and renaming `org_`→`organization_` in `payload-mapping.ts`). That approach was **rejected** because it would create one synthetic person per org (tens–hundreds across the product, inflating unique-user counts).

The current PR instead **attributes `FirstTraceReceived` to the org owner** so it stays identified with no synthetic person, and `payload-mapping.ts` is **unchanged**. See the updated PR description for details. Its one actionable inline comment (billing query) has been addressed in 5e5e04e7b.

csansoon (dismissed):

apps/web/src/routes/_authenticated.tsx
```diff
@@ -358,6 +358,17 @@ function AuthenticatedLayout() {
   })
   const { data: allOrgs } = useOrganizationsCollection()
   const org = allOrgs?.find((o) => o.id === organizationId)
+  const excludeFromAnalytics = isLatitudeStaffEmail(user.email) || impersonatedBy != null
+  // Only used to tag the PostHog org group with its plan, so skip it for
+  // excluded sessions (staff / impersonation) where we opt out of capturing
+  // anyway. Shares BillingCreditCounter's query key, so when the header is
+  // mounted this dedupes rather than firing a second request.
+  const { data: billingOverview } = useQuery({
```
andresgutgon: Do this on backend on a beforeRoute and just pass the `billing.slugPlan` to the fronetend

apps/workers/src/workers/projects.ts
```diff
@@ -61,6 +68,20 @@ export const createProjectsWorker = ({ consumer, postgresClient }: ProjectsDeps)
...
+        // a lookup failure falls back to no actor rather than dropping the
+        // event. Prefer the owner, else the earliest member. This only runs on
+        // the first trace per project (the fast-path return above gates it).
+        const membershipRepo = yield* MembershipRepository
+        const memberships = yield* membershipRepo
+          .listByOrganizationId(OrganizationId(payload.organizationId))
+          .pipe(Effect.orElseSucceed(() => []))
+        const owner =
+          memberships.find((m) => m.role === "owner") ??
+          [...memberships].sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime())[0]
```
andresgutgon: Do a `memberships.findFirstOwner` method that can return null

apps/workers/src/workers/projects.ts
```diff
@@ -74,6 +95,7 @@ export const createProjectsWorker = ({ consumer, postgresClient }: ProjectsDeps)
             organizationId: payload.organizationId,
             projectId: payload.projectId,
             traceId: payload.traceId,
+            ...(owner ? { actorUserId: owner.userId } : {}),
```
andresgutgon: Wait. This is a hot path. Maybe better do the fetch of memberships near the call to posthog. That would be in the event handler of `FirstTraceReceived`. Other consumers of this don't need to know and wait for membership fetching

packages/domain/events/src/event-payloads.ts
```diff
@@ -298,6 +298,15 @@ export interface EventPayloads {
...
     readonly traceId: string
     readonly onboardingType?: "prod-traces" | "code-agents"
+    /**
+     * The org owner's user id, resolved when the milestone is emitted. There is
+     * no acting user on trace ingestion (it runs off an API key), but analytics
+     * attributes the event to the owner so it stays an IDENTIFIED PostHog event
+     * — anonymous events don't get a materialized `$group_0`, which org-level
+     * retention / funnels join on. Falling back to no actor keeps the event
+     * org-scoped (and anonymous) rather than failing.
+     */
```
andresgutgon: Yapping. Remove this

apps/web/src/routes/_authenticated.tsx
cesr: Done in 716d3764e — plan is now resolved in the route loader (server-side) and passed via loader data; dropped the client billing `useQuery`.

apps/workers/src/workers/projects.ts
cesr: Added `MembershipRepository.findFirstOwner(orgId)` (returns `null` when none, earliest by createdAt on ties) + fake impl, replacing the inline `listByOrganizationId().find(...)`. 716d3764e

apps/workers/src/workers/projects.ts
cesr: Moved it off the hot path — the membership lookup now lives in the posthog-analytics worker's `FirstTraceReceived` handling, not `checkFirstTrace`. The producer and the event payload are thin again (no `actorUserId`); the owner is injected at the PostHog boundary, best-effort. 716d3764e

packages/domain/events/src/event-payloads.ts
cesr: Removed — the `actorUserId` field is gone from the payload entirely now that the lookup moved to the analytics worker. 716d3764e

andresgutgon (approved):

PR #3396, opened 2026-06-04
fix(issues): close escalation on resolve/ignore and add cold-start exit

csansoon:
## what this fixes

Two reasons an `issue.escalating` incident could stay open long after the issue stopped firing — both observed on a real production issue that sat "Escalating" for ~2 days with ~21h of zero occurrences.

The "Escalating" badge is driven entirely by an open row in `alert_incidents` (`kind = issue.escalating`, `ended_at IS NULL`), surfaced through `deriveIssueLifecycleStates`'s `isEscalating` flag — it is not recomputed from the live occurrence rate. So an incident that never gets its `ended_at` written keeps the issue escalating regardless of current activity.

### 1. resolve / ignore never closed the escalation

`applyIssueLifecycleCommandUseCase` only set `resolvedAt` / `ignoredAt` and emitted nothing, so the open incident was never closed. An issue could show `escalating` and `resolved` at the same time. Ignored and resolved issues already stop driving lifecycle and alerting transitions (the escalation check bails on `ignoredAt`, regression is gated on `ignoredAt === null`), so a lingering open escalation is simply stale.

Now the use-case emits `IssueEscalationEnded` for every issue whose state actually changes on `resolve` / `ignore`. The alert-incidents worker closes the open row idempotently (no-op when none is open), so emitting on a state change is safe; `unresolve` / `unignore` emit nothing. `OutboxEventWriter` is wired into the use-case and its API + web call sites.

Closing an incident normally fans out a "Resolved: escalation" in-app + email notification. That recovery notification is meant for organic recovery, not a deliberate user action, so the `IncidentClosed` handler in `domain-events` skips the notification when the reason is `resolved` or `ignored`. Organic closes (`threshold` / `absolute-rate-drop` / `timeout`) still notify.

### 2. cold-start incidents had no fast exit

An issue with no seasonal history (`samplesCount === 0` — e.g. a normally-silent issue hit by a one-off burst) took the cold-start branch of `evaluateSeasonalEscalation`, whose only active-incident exits were the 72h hard timeout and the absolute-rate backstop (itself gated on a non-null entry snapshot). The fast band-shape + dwell exit was skipped, so a quiet incident could hang until the 72h ceiling instead of recovering shortly after going silent.

The exit path is now unified across all sample counts. Entry stays split (cold start keeps the `ESCALATION_MIN_OCCURRENCES_THRESHOLD` floor; the seasonal path keeps the multi-window band test), but once an incident is open the same pipeline runs regardless of history: absolute-rate backstop first, then band-shape + dwell. `sigmaEffective` floors σ at 1.0, so the exit band stays a small positive number even with zero history and a quiet incident de-escalates ~one dwell after going silent. The 72h timeout and backstop priority are unchanged.

## new exit reasons

`SeasonalEscalationExitReason` (and the `IssueEscalationEnded` / `IncidentClosed` payloads, the close use-case input, the worker handler, and the queue topic registry) gain two manual-lifecycle reasons, `resolved` and `ignored`, alongside the organic `threshold` / `absolute-rate-drop` / `timeout`. No DB migration — `alert_incidents` has no `reason` column; the reason travels on events only and is used to gate the close notification.

## decisions

- Close on both resolve and ignore — ignored issues stop generating incidents, so the open escalation is stale in both cases.
- Suppress the recovery notification for both manual reasons. Easy to flip for `resolved` (drop it from the gate) if we later want resolve to notify.
- Emit `IssueEscalationEnded` unconditionally on a state change rather than first checking for an open incident — `closeOpen` is idempotent, which avoids an extra query and dependency.

## testing

- `@domain/issues`, `@domain/events`, `@domain/alerts`, `@domain/queue`, `@app/api`, `@app/web`, `@app/workers` typecheck clean.
- `@domain/issues` unit tests pass (added: resolve/ignore/no-op/unresolve/unignore/bulk emission cases, and cold-start active-incident exits — dwell start with and without an entry snapshot, threshold close after dwell, dwell cleared while elevated, backstop priority).
- `domain-events.test.ts` passes, including the notification-suppression gate (organic close still notifies; `resolved`/`ignored` do not).

### DIFF
--- a/apps/api/src/routes/issues.ts
+++ b/apps/api/src/routes/issues.ts
@@ -193,7 +193,13 @@ const buildLifecycleEndpoint = ({
           })
         }).pipe(
           withPostgres(
-            Layer.mergeAll(ProjectRepositoryLive, IssueRepositoryLive, EvaluationRepositoryLive, SettingsReaderLive),
+            Layer.mergeAll(
+              ProjectRepositoryLive,
+              IssueRepositoryLive,
+              EvaluationRepositoryLive,
+              OutboxEventWriterLive,
+              SettingsReaderLive,
+            ),
             c.var.postgresClient,
             organizationId,
           ),
--- a/apps/web/src/domains/issues/issues.functions.ts
+++ b/apps/web/src/domains/issues/issues.functions.ts
@@ -33,6 +33,7 @@ import { ScoreAnalyticsRepositoryLive, TraceRepositoryLive, withClickHouse } fro
 import {
   EvaluationRepositoryLive,
   IssueRepositoryLive,
+  OutboxEventWriterLive,
   ScoreRepositoryLive,
   SettingsReaderLive,
   withPostgres,
@@ -619,7 +620,7 @@ export const applyIssueLifecycleAction = createServerFn({ method: "POST" })
         keepMonitoring: data.keepMonitoring,
       }).pipe(
         withPostgres(
-          Layer.mergeAll(IssueRepositoryLive, EvaluationRepositoryLive, SettingsReaderLive),
+          Layer.mergeAll(IssueRepositoryLive, EvaluationRepositoryLive, OutboxEventWriterLive, SettingsReaderLive),
           pgClient,
           orgId,
         ),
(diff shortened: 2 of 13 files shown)


csansoon:
@claude review this PR

andresgutgon (approved):

PR #3405, opened 2026-06-04
Fix silent Claude Code issues Linux <20 Node version

### LINKED ISSUE #3404: [Bug]: Claude code npx package not installing in Linux Node 18
### What happened?

Hey guys, having these two issues:

Issue 1: Silent crash on Node.js < 20.12

The hook uses @clack/prompts which requires node:util styleText (Node ≥ 20.12). On older Node it crashes at load time before any code runs, making it undetectable. Since the hook is async: true + fail-open (always exits 0), the failure is completely silent. No engines field or version check existed to warn users.

Issue 2: BUN_OPTIONS from settings.json doesn't reach the Claude process on Linux

Claude Code only forwards settings.json env vars to hook subprocesses, not to the main claude binary. On macOS the installer uses launchctl setenv to set BUN_OPTIONS at the OS level so all GUI apps inherit it. On Linux there was no equivalent — only settings.json was written, which isn't enough when Claude is launched from VS Code.

### Steps to reproduce

1. Install on a Linux machine with VS Code + Node 22 via nvm (so Node 18 is the system default)
2. Run npx -y @latitude-data/claude-code-telemetry install
3. Look at latitude.so, no traces will be there

### Area

None

### Version / commit

_No response_

### Relevant logs or screenshots

```shell

```

learningbizz:
## What does this PR do?

Fixes two silent failures in the @latitude-data/claude-code-telemetry Stop hook that caused traces to never arrive in Latitude with no visible error to the user:

Node.js version check — on systems where the system node is < v20.12 (common on Ubuntu LTS), the hook crashed at load time because @clack/prompts uses node:util styleText which doesn't exist in older Node. Since the hook is async: true + fail-open, the crash was invisible. Fixes this by introducing a thin entry.ts wrapper that checks the Node version before any imports load, and prints a clear error with the nvm fix if too old.

Linux BUN_OPTIONS not reaching Claude — settings.json env vars only reach hook subprocesses, not the main claude binary. On macOS this was already solved via launchctl setenv. On Linux there was no equivalent, so the intercept preload never loaded when Claude was launched from VS Code. Fixes this by adding a systemd user environment path during install: systemctl --user set-environment for the current session + ~/.config/environment.d/latitude-telemetry.conf for persistence across reboots.

Closes #3404 

## How was this tested?

For Fix 1 we ran two tests:

Happy path — piped {} into dist/entry.js on Node 22, verified it exited 0 with no crash.
Error path — monkey-patched process.versions.node to '18.19.1' before the dynamic import, verified the clear error message printed to stderr and the process still exited 0 (fail-open).

For Fix 2 we only verified the file-writing logic (writeEnvironmentDConf) in isolation — confirmed the correct content gets written to ~/.config/environment.d/latitude-telemetry.conf. The actual systemd integration (systemctl --user set-environment and VS Code picking it up) was not tested and needs verification on a real Linux desktop.

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/packages/telemetry/claude-code/src/entry.ts
+++ b/packages/telemetry/claude-code/src/entry.ts
@@ -11,4 +11,9 @@ if (error) {
   process.exit(0) // fail-open — do not block Claude Code
 }
 
-await import("./index.js")
+try {
+  await import("./index.js")
+} catch (err) {
+  process.stderr.write(`[latitude-claude-code] failed to load: ${String(err)}\n`)
+  process.exit(0)
+}
--- a/packages/telemetry/claude-code/src/entry.ts
+++ b/packages/telemetry/claude-code/src/entry.ts
@@ -3,18 +3,11 @@
 // In ESM the static imports at the top of index.ts are evaluated before
 // any module body runs, so this file dynamically imports index.js after
 // the check passes.
-const parts = process.versions.node.split(".")
-const major = parseInt(parts[0] ?? "0", 10)
-const minor = parseInt(parts[1] ?? "0", 10)
+import { buildVersionError } from "./node-version.ts"
 
-if (major < 20 || (major === 20 && minor < 12)) {
-  process.stderr.write(
-    `[latitude-claude-code] Node.js ${process.versions.node} is too old — requires >=20.12.\n` +
-      `The Stop hook is failing silently. To fix, ensure the "node" on your PATH is v20.12+\n` +
-      `before Claude Code runs this hook.\n` +
-      `With nvm, update the hook command in ~/.claude/settings.json to prefix the PATH:\n` +
-      `  "command": "PATH=\\"/home/<you>/.nvm/versions/node/v22.x.x/bin:$PATH\\" npx -y @latitude-data/claude-code-telemetry"\n`,
-  )
+const error = buildVersionError(process.versions.node)
+if (error) {
+  process.stderr.write(error)
   process.exit(0) // fail-open — do not block Claude Code
 }
 
--- a/packages/telemetry/claude-code/src/entry.ts
+++ b/packages/telemetry/claude-code/src/entry.ts
@@ -1,0 +1,21 @@
+// Version check must run before any other import — @clack/prompts uses
+// node:util styleText which only exists from Node.js 20.12 onward.
+// In ESM the static imports at the top of index.ts are evaluated before
+// any module body runs, so this file dynamically imports index.js after
+// the check passes.
+const parts = process.versions.node.split(".")
+const major = parseInt(parts[0] ?? "0", 10)
+const minor = parseInt(parts[1] ?? "0", 10)
+
+if (major < 20 || (major === 20 && minor < 12)) {
+  process.stderr.write(
+    `[latitude-claude-code] Node.js ${process.versions.node} is too old — requires >=20.12.\n` +
+      `The Stop hook is failing silently. To fix, ensure the "node" on your PATH is v20.12+\n` +
+      `before Claude Code runs this hook.\n` +
+      `With nvm, update the hook command in ~/.claude/settings.json to prefix the PATH:\n` +
+      `  "command": "PATH=\\"/home/<you>/.nvm/versions/node/v22.x.x/bin:$PATH\\" npx -y @latitude-data/claude-code-telemetry"\n`,
+  )
+  process.exit(0) // fail-open — do not block Claude Code
+}
+
+await import("./index.js")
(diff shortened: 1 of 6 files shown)


packages/telemetry/claude-code/tsdown.config.ts
```diff
@@ -2,15 +2,27 @@ import { defineConfig } from "tsdown"
...
 export default defineConfig([
   {
-    entry: ["src/index.ts"],
+    // Thin entry-point that checks the Node.js version before loading index.ts.
+    // Carries the shebang; index.ts is imported dynamically so the version check
+    // runs before any module with Node 20+ APIs is evaluated.
+    entry: ["src/entry.ts"],
     format: ["esm"],
-    dts: true,
+    dts: false,
```
andresgutgon, line 8: Why this?

packages/telemetry/claude-code/tsdown.config.ts
```diff
@@ -2,15 +2,27 @@ import { defineConfig } from "tsdown"
 
 export default defineConfig([
   {
-    entry: ["src/index.ts"],
+    // Thin entry-point that checks the Node.js version before loading index.ts.
+    // Carries the shebang; index.ts is imported dynamically so the version check
+    // runs before any module with Node 20+ APIs is evaluated.
```
andresgutgon: yapping

packages/telemetry/claude-code/tsdown.config.ts
```diff
@@ -2,15 +2,27 @@ import { defineConfig } from "tsdown"
...
-    dts: true,
+    dts: false,
     sourcemap: true,
     clean: true,
     target: "node20",
     fixedExtension: false,
     banner: { js: "#!/usr/bin/env node" },
   },
+  {
+    entry: ["src/index.ts"],
```
andresgutgon, line 16: Why another esm entry?

packages/telemetry/claude-code/tsdown.config.ts
learningbizz: entry.ts is just the bin executable — nobody imports it programmatically so generating type declarations for it is pointless.

packages/telemetry/claude-code/tsdown.config.ts
learningbizz: Trimmed to one line, fixed.

packages/telemetry/claude-code/tsdown.config.ts
learningbizz: Before this PR, index.ts was doing double duty as both the bin and the library export. Now that the bin is entry.ts, index.ts needs its own explicit entry to keep producing dist/index.js with dts: true for the package's programmatic exports ("main" and "exports" in package.json).

packages/telemetry/claude-code/src/setup.ts
```diff
@@ -437,6 +477,40 @@ function writePlist(): void {
...
+    process.stderr.write(
+      `[latitude-claude-code] systemctl set-environment failed (exit ${res.status})${details ? `: ${details}` : ""}. Continuing.\n`,
+    )
+  }
+}
+
+function writeEnvironmentDConf(): void {
+  mkdirSync(dirname(ENVIRONMENT_D_CONF_PATH), { recursive: true })
+  writeFileSync(ENVIRONMENT_D_CONF_PATH, `BUN_OPTIONS=--preload=${INTERCEPT_INSTALL_PATH}\n`, "utf-8")
+}
```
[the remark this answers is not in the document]
learningbizz: Good catch — fixed. The value is now written as `BUN_OPTIONS="--preload=..."` so paths with spaces are handled correctly.

packages/telemetry/claude-code/src/entry.ts
```diff
@@ -0,0 +1,21 @@
...
+    `[latitude-claude-code] Node.js ${process.versions.node} is too old — requires >=20.12.\n` +
+      `The Stop hook is failing silently. To fix, ensure the "node" on your PATH is v20.12+\n` +
+      `before Claude Code runs this hook.\n` +
+      `With nvm, update the hook command in ~/.claude/settings.json to prefix the PATH:\n` +
+      `  "command": "PATH=\\"/home/<you>/.nvm/versions/node/v22.x.x/bin:$PATH\\" npx -y @latitude-data/claude-code-telemetry"\n`,
+  )
+  process.exit(0) // fail-open — do not block Claude Code
+}
+
+await import("./index.js")
```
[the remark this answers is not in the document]
learningbizz: Good catch — fixed. The dynamic import is now wrapped in try/catch that logs to stderr and exits 0, keeping the hook fail-open.

andresgutgon (approved):

PR #3406, opened 2026-06-04
feat(notifications): align incident-trend chart with the issue-detail drawer

csansoon:
Two related changes to the escalating-incident notifications surface.

## incident-trend chart parity

The escalating-incident Slack message and email embed a trend PNG (`/api/notifications/{id}/incident-trend.png`). It used a 3h window at 10-min buckets with its own visual style, so it looked nothing like the in-app issue-detail drawer trend and didn't match what someone sees after clicking through. Both render real ClickHouse occurrence data — they just disagreed on window, granularity, and design.

dyn-7afb87d0e994af941ecfd3157c132ee6

The snapshot now mirrors the drawer: a 14-day window of UTC-aligned 12h buckets (~28 points), the seasonal-expectation threshold series, gray occurrence bars, a dashed expectation curve, and a severity band + start dot for the triggering incident. The window is frozen at incident time (the PNG is immutable/cached, so it can't drift like the live drawer).

Notable decisions:

- **Dropped Satori for the chart, render via Resvg.** Satori can't render SVG `stroke-dasharray`, which the dashed expectation curve needs; the prior renderer faked it with positioned divs and omitted the smooth curve. The chart is now a hand-authored SVG string rasterized by Resvg. resvg-js 2.6.2 has no in-memory `fontBuffers`, so the chart font is materialized to a temp file and passed via `fontFiles`.
- **Shared geometry.** The drawer's pure geometry (Catmull-Rom path, height scaling, threshold segmentation) is extracted into a shared module used by both the live `IssueTrendBar` and the server renderer, so the two stay in sync. The triplicated histogram bucket scaffold/fill helper is consolidated into `@domain/issues`.
- **Payload.** Added an optional `marker` (`startedAt`/`endedAt`/`severity`) to the trend payload for the band + dot. Optional, so notifications written before this change still parse — no migration.
- **Breach copy decoupled.** The "rate climbed to X/hr" line is derived from a separate fine-grained recent peak rather than the display trend, since 12h buckets are too coarse to read an hourly rate from.
- Brand-new issues with no seasonal history render bars but no dashed curve (no contributing prior weeks) — matching the drawer.

## Button asChild label fix

The filled `Button` variants (default/destructive/secondary) draw their surface with an `::after` overlay. The non-`asChild` path lifts the label into a `relative z-1` wrapper so it paints above that overlay; the `asChild`/`Slot` path rendered the label as a bare child with no such wrapper, so the fill covered it. The "Connect Slack" buttons (Integrations settings + Slack onboarding) are the only `asChild` usages on a filled variant, so they showed as blue buttons with no text. Fixed by pushing the overlay behind the slotted content with `after:-z-10` in the `asChild` branch.

## testing

- Unit tests for the extracted geometry helpers and the SVG builder (including the no-history and empty-trend cases); updated the producer snapshot test for the 14d/12h window + marker.
- Rendered a real PNG through the Resvg + font pipeline and eyeballed it against the drawer; drove a full escalation end to end locally (issue + a month of synthetic history → real detector → incident → notification) to confirm the workflow path.
- Button fix verified live in the running app.

### DIFF
--- a/apps/web/src/domains/alerts/incident-markers.ts
+++ b/apps/web/src/domains/alerts/incident-markers.ts
@@ -17,6 +17,18 @@ export const INCIDENT_SEVERITY_COLOR: Record<AlertSeverity, string> = {
   high: "hsl(0 84% 60%)",
 }
 
+/**
+ * Concrete hex equivalents of {@link INCIDENT_SEVERITY_COLOR}, for renderers that can't rely on
+ * CSS `hsl()` parsing. usvg (the SVG parser behind the server-side Resvg incident-trend PNG)
+ * does not reliably parse the space-separated `hsl(H S% L%)` syntax used above, so the hand-built
+ * SVG markup references these instead. Values are the same Tailwind-500 colors (blue/amber/red).
+ */
+export const INCIDENT_SEVERITY_HEX: Record<AlertSeverity, string> = {
+  low: "#3b82f6",
+  medium: "#f59e0b",
+  high: "#ef4444",
+}
+
 type TopSymbol = NonNullable<BarChartOverlayLine["topSymbol"]>
 
 // Markers are paint-only — interactivity lives at the bucket level via the histogram's hover
--- a/apps/web/src/domains/issues/issues.functions.ts
+++ b/apps/web/src/domains/issues/issues.functions.ts
@@ -3,9 +3,11 @@ import { exportSelectionSchema } from "@domain/exports"
 import {
   type ApplyIssueLifecycleCommandResult,
   applyIssueLifecycleCommandUseCase,
+  buildHistogramBucketScaffold,
   DEFAULT_ESCALATION_SENSITIVITY_K,
   deriveIssueLifecycleStates,
   embedIssueSearchQueryUseCase,
+  fillBuckets,
   getEscalationOccurrenceThreshold,
   type Issue,
   type IssueListItem,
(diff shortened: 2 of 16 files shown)


packages/ui/src/components/button/button.tsx
```diff
@@ -189,6 +192,7 @@ function Button({
           buttonVariantsConfig({ variant, size }),
           className,
           isLoading && "animate-pulse",
+          "after:-z-10",
```
andresgutgon: This is already fixed in prod

packages/ui/src/components/button/button.tsx
```diff
@@ -180,7 +180,10 @@ function Button({
   const visibleChildren = isLoading ? stripLeadingIconChild(children) : children
 
   if (asChild) {
-    // Used in combobox.ts <ComboboxInput>
+    // The filled variants (default/destructive/secondary) draw their surface with an `::after`
+    // overlay; the non-asChild path lifts its label into a `relative z-1` wrapper so it paints
+    // above that overlay. Slot has a single child we don't control, so instead push the overlay
+    // behind the slotted content with `after:-z-10` (twMerge overrides the variant's `after:z-0`).
```
andresgutgon: Yapping

packages/ui/src/components/button/button.tsx
csansoon: Good catch — `development` already fixes this in the compound variant (`after:-z-10` on `buttonVariantsConfig`), which covers the `asChild` path too. Rebased onto latest `development` and dropped my button commit; this PR is now just the incident-trend chart change.

packages/ui/src/components/button/button.tsx
csansoon: Gone — the whole commit was dropped (fix is already in prod), so the comment went with it.

andresgutgon (approved):

PR #3408, opened 2026-06-04
feat(test-mode): block outbound notifications for sandbox orgs [AGE-113]

andresgutgon:
## Summary

Test Mode hard guarantee: **a sandbox org never reaches an external notification channel.** A developer's experiments must never page teammates or hit customers' alert channels. In-app notification surfacing is unaffected — only outbound transports are blocked.

Each channel adapter resolves its target org and short-circuits dispatch when `isSandbox(org)` (the shared predicate from `@domain/organizations`, where a non-null `organizations.parent_org_id` ⇒ sandbox), **before** any send. The guard is placed in each channel's dispatch use case so it's a single, unit-testable point per channel, and it must hold even once LLM-driven feedback is enabled per-sandbox in the future.

### Channels guarded

| Channel | Guard point | Behavior for a sandbox org |
| --- | --- | --- |
| **Email** | `sendNotificationEmailUseCase` (`@domain/notifications`) | Guards **before** the `markEmailed` claim — the notification row stays unstamped (`emailed_at` null) and nothing is rendered or sent. Returns `{ sent: false }`. |
| **Slack** | `dispatchSlackNotificationUseCase` (`@domain/integrations`) | Guards **before** the `slack_deliveries` claim and the post — no delivery row, no message. Returns the new `{ status: "skipped-sandbox" }` outcome. |

Email and Slack are the only outbound notification channels implemented today; webhook/SMS/push channels do not exist yet, so there is nothing to guard for them. When one is added, it should resolve its target org and apply the same `isSandbox` guard at its dispatch point.

In-app rows (the `notifications` table + bell feed) are written by the upstream creator step and are intentionally **not** touched — only the outbound transports short-circuit.

## Test plan

- `pnpm --filter @domain/notifications --filter @domain/integrations --filter @domain/organizations typecheck` ✅
- `pnpm --filter @domain/notifications test` ✅ (34 passed)
- `pnpm --filter @domain/integrations test` ✅ (21 passed)
- `pnpm --filter ... check` (Biome) ✅

New tests, per channel, exercising sandbox vs. live:
- **Email** — `send-notification-email.test.ts`: sandbox org → no claim, no render, no send, `emailed_at` stays null; live org cases (send / idempotent / project lookup) still pass.
- **Slack** — `dispatch-slack-notification.test.ts`: sandbox org → `skipped-sandbox`, messenger never called; live org cases (delivered / dedupe / render error) still pass.

## Linear

[AGE-113 — Test Mode: block all outbound notifications for sandbox orgs](https://linear.app/latitude/issue/AGE-113)

Spec: `specs/test-mode.md` → "A sandbox never sends external notifications".

### DIFF
--- a/packages/ui/src/components/genai-conversation/parts/remark-code-content-positions.ts
+++ b/packages/ui/src/components/genai-conversation/parts/remark-code-content-positions.ts
@@ -1,26 +1,16 @@
-// Attaches `data-code-content-start/end` to HAST <code> elements so that
-// sourceMappedTextPlugin can map code-block text to source offsets.
-// remark-to-hast doesn't propagate positions for code-fence text nodes, so we
-// find the content verbatim in the source string and store the character range.
-
-function visitCode(node: any, callback: (node: any) => void): void {
-  if (!node) return
-  if (node.type === "code") callback(node)
-  if (Array.isArray(node.children)) {
-    for (const child of node.children) visitCode(child, callback)
-  }
-}
+import type { Code, Root } from "mdast"
+import { visit } from "unist-util-visit"
 
 export function remarkCodeContentPositions() {
-  return (tree: any, file: any) => {
-    const source: string = String(file)
+  return (tree: Root, file: unknown) => {
+    const source = String(file)
     if (!source) return
 
-    visitCode(tree, (node) => {
-      if (!node.position || typeof node.value !== "string" || node.value === "") return
+    visit(tree, "code", (node: Code) => {
+      if (!node.position || node.value === "") return
 
-      const fenceStart: number | undefined = node.position.start?.offset
-      const fenceEnd: number | undefined = node.position.end?.offset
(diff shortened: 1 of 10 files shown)


packages/ui/src/components/genai-conversation/parts/remark-code-content-positions.ts
```diff
@@ -1,34 +1,24 @@
...
-
-function visitCode(node: any, callback: (node: any) => void): void {
-  if (!node) return
-  if (node.type === "code") callback(node)
-  if (Array.isArray(node.children)) {
-    for (const child of node.children) visitCode(child, callback)
-  }
-}
+import type { Code, Root } from "mdast"
+import { visit } from "unist-util-visit"
```
andresgutgon, line 2: Used this helper to improve typescript types. This works in the same way. You can keep annotating in a code block

Image

PR #3410, opened 2026-06-04
feat(test-mode): sandbox routing + middleware scaffold + seed fixtures [AGE-102, AGE-126]

andresgutgon:
Bundles [AGE-102](https://linear.app/latitude/issue/AGE-102/test-mode-sandbox-routing-middleware-scaffold) (sandbox routing + middleware) and [AGE-126](https://linear.app/latitude/issue/AGE-126/test-mode-seed-data-second-org-sandboxes-for-the-acme-org) (seed fixtures) so the new `/sandbox/:sandboxOrgId` page is QA-able locally without hand-rolling rows.

Spec: `specs/test-mode.md` → *How isolation works*, *Access without membership*, *What a sandbox is*.

## Summary

**Routing & middleware (AGE-102)**
- **`sandboxMiddleware`** (`apps/web/src/server/sandbox-middleware.ts`) — guard for `createServerFn` handlers under the `/sandbox/:sandboxOrgId` namespace. Authorizes the caller iff the resolved sandbox's `parent_org_id` is an org they're a member of (membership check is an ordinary RLS-scoped read against the **parent** — no bypass), then injects a sandbox-scoped `context.sandbox` (organizationId / name / parentOrgId) + `userId`. The browser session's `activeOrganizationId` is never seen by the handler, structurally preventing it from acting on the live parent by accident.
- **`createSandboxServerFn` is realised as the middleware**, not a factory. The TanStack Start Vite plugin only recognises the literal `createServerFn(...).handler(...)` chain at the call site — hiding it behind a factory leaks Node-only imports into the browser bundle (see `.agents/skills/backoffice/SKILL.md`). Same shape as the admin pattern: there is no `createAdminServerFn` either.
- **`authorizeSandboxAccess`** (`apps/web/src/domains/sandbox/sandbox-access.functions.ts`) — server fn that owns the DB-heavy auth path so its Node-only imports are stripped by the compiler on the client (the original draft had them in the middleware file directly and CI flagged the leak; mirrors how `adminMiddleware` delegates to `getFreshSession`).
- **`getSandbox` server fn** (`apps/web/src/domains/sandbox/sandbox.functions.ts`) — first consumer of `sandboxMiddleware`, returns the sandbox's display attributes.
- **Stub page** at `/sandbox/:sandboxOrgId` — loader fetches via `getSandbox`, renders `Sandbox: $name` + a *Back to live* `Link`. Loader maps `NotFoundError` → `notFound()` and `UnauthorizedError` → `/login` redirect via `parseServerError`.
- **`isSandbox` is now a type predicate** so callers don't need a redundant `parentOrgId !== null` check.

**Seed fixtures (AGE-126)**
- **Beta Co.** — second top-level org (owner@beta.com, alex@beta.com, default project) with **no sandboxes**, exercising the empty-state branch for the Wave 2 sandbox UI.
- **Two Acme sandboxes** — one `active` ("Dev sandbox") and one `archived` ("Old sandbox", last activity ~30 days ago). Each sandbox is a sibling-tenant `organizations` row with `parent_org_id = SEED_ORG_ID`, a 1:1 `sandboxes` attrs row (status / last_activity_at / created_by = Acme owner), and one linked sandbox project pointing at `SEED_PROJECT_ID` via `projects.linked_project_id`. **No `members` rows** for sandbox orgs (by design — spec: *Access without membership*).
- Idempotent: deterministic ids + `onConflictDoNothing` on direct inserts so `pnpm db:seed` is safely rerunnable.

## Non-goals (deferred to follow-up tickets)

- No traces UI, settings UI, or Live↔Sandbox switcher.
- No `members` rows for sandbox orgs — by design.
- No span seeding inside the sandboxes — comes with the ingestion path (AGE-105+).
- **`resolveEffectivePlan` parent-follow** lands with AGE-127 (sandbox ingest pipeline running in parallel), where it becomes load-bearing.

## QA recipe

```bash
pnpm db:seed     # idempotent — safe to re-run
# log in as owner@acme.com (mailpit magic-link)
```

Then visit:

- `/sandbox/acmesbx1act000000000000a` → renders `Sandbox: Dev sandbox` + *Back to live*
- `/sandbox/acmesbx2arc000000000000a` → renders `Sandbox: Old sandbox`
- `/sandbox/notarealsandboxidaaaaaaa` → 404
- `/sandbox/iapkf6osmlm7mbw9kulosua4` (the live Acme org id, not a sandbox) → 404

Then sign out and back in as `owner@beta.com`:

- `/sandbox/acmesbx1act000000000000a` → 404 (Beta is not a parent of Acme's sandbox — no fingerprint of existence)

## Test plan

- [x] `pnpm --filter @app/web typecheck` (clean)
- [x] `pnpm --filter @app/web build` (clean — reproduces the original CI MISSING_EXPORT before the `.functions.ts` refactor, succeeds after)
- [x] `pnpm --filter @platform/db-postgres --filter @domain/shared typecheck` (clean)
- [x] `pnpm knip` (clean)
- [ ] Dev-server smoke (reviewer to verify after `pnpm db:seed`):
  - [ ] Routes resolve per the QA recipe above
  - [ ] Re-running `pnpm db:seed` is idempotent (no duplicate-key errors)

### DIFF
--- a/apps/web/src/domains/sandbox/sandbox-access.functions.ts
+++ b/apps/web/src/domains/sandbox/sandbox-access.functions.ts
@@ -1,0 +1,94 @@
+import { isSandbox, MembershipRepository, OrganizationRepository } from "@domain/organizations"
+import { NotFoundError, type OrganizationId, organizationIdSchema, UnauthorizedError, UserId } from "@domain/shared"
+import { MembershipRepositoryLive, OrganizationRepositoryLive, withPostgres } from "@platform/db-postgres"
+import { withTracing } from "@repo/observability"
+import { createServerFn } from "@tanstack/react-start"
+import { Effect } from "effect"
+import { z } from "zod"
+import { getAdminPostgresClient, getPostgresClient } from "../../server/clients.ts"
+import { getFreshSession } from "../sessions/session.functions.ts"
+
+/**
+ * Server-side resolution of "is the caller allowed to act on this sandbox?".
+ *
+ * Lives in a `.functions.ts` so the TanStack Start Vite compiler can strip
+ * its Node-only transitive imports (`@platform/db-postgres`,
+ * `@repo/observability`, `server/clients.ts`) from the browser bundle.
+ * `sandbox-middleware.ts` would leak them if it owned this logic directly,
+ * which is why `adminMiddleware` similarly delegates session work to the
+ * `getFreshSession` server fn instead of importing DB modules itself.
+ */
+
+interface SandboxAccessDto {
+  readonly userId: string
+  readonly sandbox: {
+    readonly organizationId: string
+    readonly name: string
+    readonly parentOrgId: string
+  }
+}
+
+const inputSchema = z.object({ sandboxOrgId: z.unknown() })
+
+export const authorizeSandboxAccess = createServerFn({ method: "GET" })
+  .inputValidator(inputSchema)
+  .handler(async ({ data }): Promise<SandboxAccessDto> => {
+    const session = await getFreshSession()
+    if (!session) {
+      throw new UnauthorizedError({ message: "Unauthorized" })
+    }
+    const userId = UserId(session.user.id)
+
+    const parsed = organizationIdSchema.safeParse(data.sandboxOrgId)
+    if (!parsed.success) {
+      throw new NotFoundError({ entity: "Sandbox", id: String(data.sandboxOrgId ?? "") })
+    }
+    const sandboxOrgId: OrganizationId = parsed.data
+
+    // The org lookup uses the admin client so it works at any RLS scope —
+    // `organizations` has no policy, but `withPostgres`'s default "system"
+    // scope is only sanctioned for the admin connection.
+    const sandbox = await Effect.runPromise(
+      Effect.gen(function* () {
+        const orgRepo = yield* OrganizationRepository
+        const org = yield* orgRepo
+          .findById(sandboxOrgId)
+          .pipe(
+            Effect.catchTag("NotFoundError", () =>
+              Effect.fail(new NotFoundError({ entity: "Sandbox", id: sandboxOrgId })),
+            ),
+          )
+        if (!isSandbox(org)) {
+          return yield* Effect.fail(new NotFoundError({ entity: "Sandbox", id: sandboxOrgId }))
+        }
+        return {
+          organizationId: org.id,
+          name: org.name,
+          parentOrgId: org.parentOrgId,
+        }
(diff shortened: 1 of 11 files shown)


andresgutgon:
@claude review please

neoxelox (dismissed):

andresgutgon:
Thanks for the review! Addressed in `087730e31`:

**Applied**
- **Cached session** (`authorizeSandboxAccess`) → switched to `getFreshSession()` for parity with `adminMiddleware`. Parent-org membership removed via the DB now takes effect on the next request, not five minutes later.
- **`rawArgs` cast** in `sandboxMiddleware` → added a comment explaining the cast (mirrors `tracingFnMiddleware` in `start.ts`) and the safe-failure path: `undefined data → invalid sandboxOrgId → NotFoundError → 404`. No silent open access.
- **`archivedSince` / `lastActivityAt`** in the sandbox seed → now anchor on `ctx.scope.dateDaysAgo(...)` so Postgres and ClickHouse seeds agree and re-runs don't drift the timestamps.
- **`onConflictDoUpdate` vs `onConflictDoNothing`** → added a comment making the split deliberate: users are upserted (keeps email/name aligned with `@domain/shared/seeding`), sandbox attrs and linked projects use do-nothing so local QA tweaks survive a re-seed.
- **Nits**: `return yield* new NotFoundError(...)` → `yield* Effect.fail(...)` for readability; `SandboxSummaryDto.parentOrganizationId` → `parentOrgId` so the DTO matches the context shape.

**Kept as-is (by design)**
- **`RepositoryError` from `isMember` surfaces as a 500** — agreed it's opaque on the client, but for an unexpected DB error that's the right shape: the route loader's catch block re-throws unknown tags to the router error boundary (only `NotFoundError` / `UnauthorizedError` are mapped), so it gets a real error page rather than a misleading 404. Adding bespoke handling here would hide real infra failures.
- **Two `Effect.runPromise` calls** — structural: we can only scope the membership read to `parentOrgId` *after* the org load tells us what it is, and the two reads use different `withPostgres` scopes (admin client + system scope for the orgs table; regular client + `parentOrgId` scope for the RLS-protected `members` table). Folding into one composition would require either committing to one scope (wrong for one of the two) or threading the parent id through a layered effect — both worse than the two-pool-lease cost at current scale.

**Copilot's `resolve-effective-plan-cached.ts` thread** — outdated. The parent-follow change was reverted in `ea321def2` and now lives with the sandbox ingest path in AGE-127. The file is no longer modified in this PR. Thread resolved.

PR #3412, opened 2026-06-04
feat(test-mode): sandbox ingest pipeline [AGE-127]

andresgutgon:
# 🏖️ Test Mode: the sandbox ingest pipeline — AGE-127

> Like Stripe's test mode, but for traces. Point your dev app at a **sandbox**, watch your traces stream in live, and never page a teammate or burn a credit doing it.

## The whole thing in one sentence

When traces arrive we ask **one** question — *"is this org a sandbox?"* (`parent_org_id != null`) — answer it **once** at the front door, stamp the answer on the event, and let every step downstream branch on that single bit. No flag threaded through ten functions.

## A sandbox stores & shows your traces. Everything else is off.

Five things change the moment the sandbox bit is `true`. Here's the before/after:

| # | Concern | 🟢 Live org | 🏖️ Sandbox |
|---|---------|------------|-----------|
| 1 | **LLM fan-out** (flaggers, evals, issue-clustering) | runs | **skipped** — none of it runs |
| 2 | **Archived (asleep) sandbox** | n/a | **403 `SandboxArchived`** before a single span is stored |
| 3 | **Per-period span quota** | n/a | Redis counter; over the line → **403 `SandboxQuotaExceeded`** |
| 4 | **`last_activity_at`** | n/a | stamped on ingest (debounced, ~once / 5 min) to power idle-sleep |
| 5 | **Realtime** | n/a | spans pulse to the UI over Redis Pub/Sub |
| 💸 | **Billing** | metered & charged | **never** — sandboxes are free |

Both refusals (2 & 3) mirror billing's existing `402 NoCreditsRemaining`: a **loud, pre-persist 4xx** (and *not* a 429, so OTLP exporters drop the batch instead of retry-looping). Each one drops a tiny `last_rejected_ingest` marker in Redis so the sandbox UI can say *"hey, your traces are bouncing"* (AGE-123).

## Realtime, the part that makes it feel alive ✨

You type a message in your dev app → you want to see *something* in the sandbox UI **now**. So there are **two** signals, not one:

```
span arrives ──▶ ⚡ "liveness" pulse   (at the HTTP door, BEFORE persist — instant)
             │
             └──▶ 💾 persist to ClickHouse
                        └──▶ ✅ "upsert" pulse   (AFTER persist — safe to refetch)
```

- **liveness** = *"activity! something's arriving"* → UI lights up immediately. Fired before the span is even stored, so it skips the entire queue + worker delay.
- **upsert** = *"the row exists now, go refetch the list"* → only after persist, so a refetch is guaranteed to actually find the data (read-after-write).

Both are **id-only** (`{ kind, sandboxId, traceId, sessionId }` — never trace content) and coalesced per trace so a chatty trace pulses ~once every few hundred ms instead of once per span.

> 📝 Heads-up for the "why isn't it *instant*?" question: the dominant latency is your app's OTLP `BatchSpanProcessor` (5s default flush), which lives client-side — no server trick beats it. The fix is a realtime-tuned exporter preset for sandbox keys (short flush / `SimpleSpanProcessor`), shipped via the onboarding snippet. That's SDK/onboarding work (AGE-104), **out of scope here** — this PR makes *our* half near-zero-latency.

## Who decides what? (the design rule)

The event carries **facts** (`isSandbox` + a plan snapshot); the **consumer** owns the policy. The producer stays dumb and emits the same full `TracesIngested` for everyone — then the `domain-events` worker decides, in one place, to skip both the LLM fan-out **and** billing for sandbox events. (Matches the repo's "payload carries facts, not routing hints" convention.)

## Under the hood

- 🆕 **`@domain/sandboxes`** package — `Sandbox` entity, `SandboxRepository` (Postgres) + `SandboxSignals` (Redis) ports, the two refusal errors, constants/keys, the debounced activity-stamp use-case, and test fakes.
- 🔌 Adapters: `SandboxRepositoryLive` (`@platform/db-postgres`), `SandboxSignalsLive` (`@platform/cache-redis`) — every Redis op **fails open**, so a cache hiccup never drops a real trace.
- 🧭 One resolution point: a single scoped read of the `sandboxes` row at the top of ingest (its presence ⇔ `parent_org_id != null`). The bit then rides the queue message and the `TracesIngested` event.
- 🧱 `spanQuotaPerPeriod` added to each `PlanConfig`, read via `resolveEffectivePlan` (Free/Pro finite, Enterprise ∞).

## Test plan ✅

- **Typecheck** clean across all touched packages.
- `@domain/spans` (641) · `@domain/billing` (29) · `domain-events` worker (22) — all green. They cover the four headline scenarios:
  - 🟢 **Live** → full fan-out, billing checked, quota untouched, no pulse, queued without the bit.
  - 🏖️ **Active sandbox** → no billing, quota++, activity stamped, **liveness pulse fires**, spans persisted + **upsert pulse**, queued with the bit.
  - 😴 **Archived** → `403 SandboxArchived` before persist, marker stored, nothing else fires (no pulse).
  - 🚧 **Over quota** → `403 SandboxQuotaExceeded`, marker stored, no pulse.
- ClickHouse-backed integration tests are behavior-unchanged on the live path (they need the native `chdb` binary → CI).

📋 Linear: https://linear.app/latitude/issue/AGE-127

### DIFF
--- a/apps/ingest/package.json
+++ b/apps/ingest/package.json
@@ -15,6 +15,7 @@
     "@domain/api-keys": "workspace:*",
     "@domain/billing": "workspace:*",
     "@domain/queue": "workspace:*",
+    "@domain/sandboxes": "workspace:*",
     "@domain/shared": "workspace:*",
     "@domain/spans": "workspace:*",
     "@hono/node-server": "catalog:",
--- a/apps/ingest/src/middleware/auth.ts
+++ b/apps/ingest/src/middleware/auth.ts
@@ -30,5 +30,6 @@ export const authMiddleware: MiddlewareHandler<IngestEnv> = async (c, next) => {
 
   c.set("organizationId", result.organizationId)
   c.set("apiKeyId", result.keyId)
+  c.set("isSandbox", result.isSandbox)
   await next()
 }
--- a/apps/ingest/src/routes/traces.ts
+++ b/apps/ingest/src/routes/traces.ts
@@ -1,10 +1,13 @@
 import { NoCreditsRemainingError } from "@domain/billing"
+import { SandboxArchivedError, SandboxQuotaExceededError } from "@domain/sandboxes"
 import { OrganizationId } from "@domain/shared"
 import { ingestSpansWithBillingUseCase } from "@domain/spans"
+import { SandboxSignalsLive } from "@platform/cache-redis"
 import {
   BillingOverrideRepositoryLive,
   BillingUsagePeriodRepositoryLive,
   ProjectRepositoryLive,
+  SandboxRepositoryLive,
   SettingsReaderLive,
   StripeSubscriptionLookupLive,
   withPostgres,
@@ -28,6 +31,7 @@ const traceIngestionBillingLayers = Layer.mergeAll(
   BillingOverrideRepositoryLive,
   BillingUsagePeriodRepositoryLive,
   ProjectRepositoryLive,
+  SandboxRepositoryLive,
   SettingsReaderLive,
   StripeSubscriptionLookupLive,
 )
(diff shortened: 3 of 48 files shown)


andresgutgon:
@claude review please

andresgutgon:
Claude review addressed

neoxelox (approved):

PR #3414, opened 2026-06-04
feat(ingest): auto-create project for unrecognized span slug

csansoon:
## what this does

Spans ingested with a project slug that doesn't exist in the org are no longer rejected. The slug is provisioned as a new project (using the slug as the project name) and the spans are ingested against it.

Previously, an unrecognized slug meant `findBySlug` missed, the slug was swallowed to `null`, and every span carrying it was counted as rejected. A batch where *all* spans carried an unknown slug returned HTTP 400 and the data was lost; mixed batches returned a `partialSuccess` accounting for the dropped spans. Now those slugs resolve to a freshly created project.

## how it works

- **`createProjectUseCase` gains an optional `slug`.** When supplied, it's normalized with `toSlug` and used verbatim — the name-derived `generateSlug` (which lowercases, collapses, and adds a random collision suffix) is skipped. This keeps the slug stable, which is the whole point: the same slug has to resolve the same project on the next export.
- **`findOrCreateProjectBySlugUseCase`** (new) normalizes the slug, looks it up, and on a miss creates the project — name = the raw slug, slug = the normalized form — by reusing `createProjectUseCase`, so the `ProjectCreated` event still fires for downstream provisioning. It's idempotent, and safe under concurrency: two ingests racing on the same new slug converge to one project, because the loser of the org-scoped unique constraint (Postgres `23505`) re-fetches the winner.
- **`ingestSpansUseCase`** resolves slugs through that use-case. A slug that can't become a usable project (empty after normalization) still rejects its span gracefully — only a genuine `RepositoryError` aborts the batch, same as before. `OutboxEventWriter` is added to its requirements, and the ingest route provides `OutboxEventWriterLive`.

## decisions worth a look

- **Normalized, not raw.** The stored slug is `toSlug(incoming)`, matching how `updateProjectUseCase` already handles explicit slugs. Storing raw would let spaces/uppercase into the slug column and break URL routing; and since every existing project slug is already normalized, normalizing the lookup also lets a malformed incoming slug correctly match a pre-existing project. The project's display **name** keeps the raw slug.
- **No creation cap.** Every novel slug creates a project, so a misbehaving exporter spraying random slugs could create unbounded projects in an org. Shipping without a guardrail for now; if we want one (per-org cap or slug allowlist) it slots into `findOrCreateProjectBySlugUseCase`.

## tests

- New unit suite for `findOrCreateProjectBySlugUseCase`: existing/unknown/normalize/idempotent/invalid-slug, plus the concurrent unique-constraint re-fetch path.
- Updated `ingest-spans` suite: auto-creation accepts the span, idempotency across two exports, empty-normalized slug still rejects, and the same slug under two orgs yields two distinct projects.
- Typecheck + Biome clean across `@domain/projects`, `@domain/spans`, `@app/ingest`. Full suites green (projects 12, spans 638).

### DIFF
--- a/apps/ingest/src/routes/traces.ts
+++ b/apps/ingest/src/routes/traces.ts
@@ -4,6 +4,7 @@ import { ingestSpansWithBillingUseCase } from "@domain/spans"
 import {
   BillingOverrideRepositoryLive,
   BillingUsagePeriodRepositoryLive,
+  OutboxEventWriterLive,
   ProjectRepositoryLive,
   SettingsReaderLive,
   StripeSubscriptionLookupLive,
@@ -27,6 +28,7 @@ interface TracesRouteContext {
 const traceIngestionBillingLayers = Layer.mergeAll(
   BillingOverrideRepositoryLive,
   BillingUsagePeriodRepositoryLive,
+  OutboxEventWriterLive,
   ProjectRepositoryLive,
   SettingsReaderLive,
   StripeSubscriptionLookupLive,
--- a/packages/domain/projects/src/index.ts
+++ b/packages/domain/projects/src/index.ts
@@ -18,6 +18,10 @@ export {
   type CreateProjectInput,
   createProjectUseCase,
 } from "./use-cases/create-project.ts"
+export {
+  type FindOrCreateProjectBySlugInput,
+  findOrCreateProjectBySlugUseCase,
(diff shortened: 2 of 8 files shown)


geclos:
been discussed offline, project creation needs to be a proactive action and the process is already super simple enough thanks to mcp integration

PR #3428, opened 2026-06-05
Add session_id in UserSignedUp event

learningbizz:
The server-side UserSignedUp event now carries the browser's PostHog $session_id (plus referrer/UTM), so PostHog can bind it to the session and $entry_current_url filters resolve. 

This fix only applies for magic links, with OAuth we need a diff solution.


apps/web/src/server/clients.ts
```diff
@@ -229,14 +235,29 @@ export const getBetterAuth = () => {
...
+        // the payload; the worker forwards it to PostHog.
+        let attribution: MarketingAttribution = {}
+        try {
+          const redis = getRedisClient()
+          const key = signupAttributionKey(user.email)
+          const raw = await redis.get(key)
+          if (raw) {
+            await redis.del(key)
+            attribution = toMarketingAttribution(JSON.parse(raw) as SignupAttributionInput)
+          }
```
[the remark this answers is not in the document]
learningbizz: Good catch — fixed in 8294641. Now we parse/map first and only `del` the key on success, so a parse failure leaves it for a retry (and the TTL self-expires it anyway).

apps/web/src/lib/analytics/signup-attribution.ts
```diff
@@ -0,0 +1,43 @@
+import type { MarketingAttribution } from "@domain/events"
+
+// Signup is passwordless: the account is created in a later (maybe cross-device)
+// request where `onUserCreated` has no request context. So stash browser
+// attribution in Redis keyed by email at request time and read it back on create.
+// Pre-org scope, so the `org:` key convention doesn't apply.
+export const signupAttributionKey = (email: string): string => `signup-attr:${email.toLowerCase()}`
+
+// Must comfortably outlast the magic-link click window.
+export const SIGNUP_ATTRIBUTION_TTL_SECONDS = 60 * 30
```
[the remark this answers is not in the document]
learningbizz: Correct — bumped `SIGNUP_ATTRIBUTION_TTL_SECONDS` from 30m to 2h in 8294641 so it comfortably outlasts the 1h magic-link window.

packages/domain/events/src/event-payloads.ts
```diff
@@ -8,6 +8,28 @@ import type { AlertIncidentKind, AlertIncidentSourceType } from "@domain/shared"
  * Acceptable as pragmatic exceptions at the auth boundary, but do not use this
  * naming pattern for domain-owned aggregates.
  */
+/**
+ * Signup attribution captured in the browser, spread verbatim into the PostHog
+ * `UserSignedUp` event. `$session_id` links the server event to the browser
+ * session (unlocks `$entry_current_url`); UTM/click ids are the fallback.
+ */
+// `type` not `interface`: keeps the implicit index signature so the `UserSignedUp`
+// intersection stays assignable to the worker's `Record<string, unknown>` constraint.
```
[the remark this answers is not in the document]
learningbizz: This one's a false positive: object **type aliases** do get an implicit index signature for index-signature assignability, whereas interfaces don't (TS#15300). Empirically, as an `interface` the `apps/workers` typecheck fails with "Index signature for type 'string' is missing in type '{ userId; email } & MarketingAttribution'"; as a `type` it passes. So the distinction is load-bearing here. I reworded the comment in 8294641 to cite the TS issue and the concrete failure rather than remove the rationale.

packages/domain/events/src/event-payloads.ts
```diff
@@ -8,6 +8,30 @@ import type { AlertIncidentKind, AlertIncidentSourceType } from "@domain/shared"
...
+  readonly utm_medium?: string
+  readonly utm_campaign?: string
+  readonly utm_term?: string
+  readonly utm_content?: string
+  readonly gclid?: string
+  readonly fbclid?: string
+  readonly ttclid?: string
+  readonly li_fat_id?: string
+  readonly msclkid?: string
+}
```
andresgutgon: All redis management, attribution parsing and all of this should go into `packages/domain/marketing` we're spreading marketing concerns accross the web code

andresgutgon (changes requested):
Tell claude to move almost all of the logic to marketing package

packages/domain/events/src/event-payloads.ts
learningbizz: Done in 6fdf90b. Moved all the marketing logic into `@domain/marketing`:

- `stashSignupAttribution` / `consumeSignupAttribution` use-cases behind the shared `CacheStore` port (`@domain/shared`) — no direct Redis client in the web app anymore.
- The Redis key scheme, TTL, UTM/click-id whitelist, and PostHog property mapping all live in `packages/domain/marketing/src/signup-attribution.ts`.
- The browser-input contract is a Zod schema (`signupAttributionInputSchema`) exported from the domain package; the web `sendMagicLink` boundary just reuses it to validate.

`apps/web` now only does what's intrinsically web: capture the browser session/referrer (login.tsx), validate at the HTTP boundary, and invoke the use-cases with `RedisCacheStoreLive(getRedisClient())`. Tests moved to the marketing package too.

apps/web/src/server/clients.ts
```diff
@@ -229,14 +231,28 @@ export const getBetterAuth = () => {
...
+        try {
+          attribution = await Effect.runPromise(
+            consumeSignupAttribution({ email: user.email }).pipe(
+              Effect.provide(RedisCacheStoreLive(getRedisClient())),
+              withTracing,
+            ),
+          )
+        } catch {
+          // Never block signup on attribution.
+        }
```
andresgutgon: Do the consumption of the attribution on `UserSignedUp` event handler because is stored in redis. This file is the entry point and we do the work on the event handlers

learningbizz: Done in 01f25c6. `onUserCreated` is back to emitting just `{ userId, email }` — the consumption now happens in the `UserSignedUp` event handler (posthog-analytics worker), which reads the Redis-stashed attribution by email and merges it onto the capture. It mirrors the existing `FirstTraceReceived` enrichment right above it. Also moved `MarketingAttribution` out of `@domain/events` into `@domain/marketing` (the event payload is back to `{ userId, email }`) and dropped marketing's `@domain/events` dep.

packages/domain/events/src/event-payloads.ts
learningbizz: Re: cookie / 'why infra' — a cookie can't carry the PostHog `$session_id` to the *server-side* UserSignedUp event in this flow, for two reasons: (1) signup is passwordless, so the account is created when the magic link is clicked, often on a different device than where the link was requested — a cookie lives only on the requesting device, whereas the Redis key is keyed by email and survives the device switch; (2) `onUserCreated` has no request access, so we couldn't read a cookie at emit time without threading it through BA internals. Reading a cookie client-side at /welcome would just be a client-side capture, which doesn't enrich the canonical server event and (because `opt_out_capturing_by_default` starts a fresh session post-auth) lands the app URL as the entry, not the landing page. So it's a single transient ~2h key on Redis we already run, not new infra. Happy to discuss if you'd rather trade away cross-device attribution.

andresgutgon (approved):

PR #3434, opened 2026-06-05
feat(sso): enterprise SSO (SAML 2.0 + OIDC) via @better-auth/sso

geclos:
## what

Enterprise customers can now sign in through their own identity provider. Org owners/admins configure a SAML 2.0 or OIDC provider self-serve from a new **Settings → Single sign-on** page (gated by a new opt-in `sso` feature flag), verify their email domain with a DNS TXT record, and optionally **require SSO** — blocking magic-link and Google/GitHub sign-in for that domain.

On `/login`, an email whose domain has a verified SSO provider is redirected to the IdP instead of receiving a magic link. First SSO sign-in JIT-provisions the user and their org membership (`member` role) into the provider's bound organization.

Spec with the full plan, decisions, and remaining work: `specs/enterprise-sso.md`.

## how it's put together

- **Plugin**: `@better-auth/sso@1.6.9` (pinned to core), registered in `createBetterAuth` with `domainVerification`, org provisioning, and strict SAML options: SP-initiated only (`allowIdpInitiated: false`), InResponseTo validation, SAML2Int timestamps required.
- **Schema**: new `latitude.sso_providers` table (ported from the BA reference into `schema/better-auth.ts`), RLS-scoped by `organization_id`, plus an app-extended `enforced` boolean. RLS is safe here because Better Auth runs on the admin client; tenant reads stay org-scoped.
- **Domain**: new `@domain/sso` package. The `SsoProvider` entity is a non-secret projection — the `saml_config`/`oidc_config` blobs (IdP certs, SP keys, OIDC client secret) never leave the repository layer, and DTOs returned to the browser only carry computed URLs (ACS, SP EntityID, SP metadata, OIDC callback).
- **Server fns** (`apps/web/src/domains/sso`): register / read / verify-domain / toggle-enforcement / delete, all gated on the `sso` flag + owner/admin, plus an unauthenticated `lookupSsoForEmail` for the login page (admin client — the tenant role sees nothing pre-auth under RLS).

## security decisions worth knowing

- **Registration is server-side only.** Two reasons found while reading BA 1.6.9 source: its `registerSSOProvider` only checks org *membership*, not owner/admin role; and its HTTP mutation endpoints would otherwise be browser-callable. The five `/sso/*` mutation paths are 404'd via `disabledPaths` (router-only — `auth.api.*` keeps working), and the server fns are the authorization layer.
- **Enforcement is keyed by verified email domain, not org membership.** The check runs pre-auth where no org context exists, and users belong to multiple orgs. Two choke points share one tested predicate: `sendMagicLink` refuses issuance, and a `session.create.before` hook blocks `/callback/*` (social) and `/magic-link/verify` (stale links) while exempting SSO callbacks, impersonation, and token flows. Enforcement can only be switched on for a verified domain.
- A provider only becomes active for sign-in after DNS TXT domain verification, so an unverified (squatted) domain claim can't capture logins or lock anyone out.

## validation

- `@domain/sso`: 16 unit tests (domain matching, malformed emails, enforcement predicate, verified-domain invariant).
- `@platform/db-postgres`: 10 PGlite tests pinning RLS behavior (tenant client can't read other orgs; cross-org domain lookup only via admin client) and 6 tests pinning the enforcement hook + `disabledPaths` posture.
- Workspace `typecheck`, `format`, `knip` green; boot smoke test confirms the auth instance starts with the plugin and the dev DB has the table + RLS policy.

## not in this PR

- Live SAML round-trip against a real IdP (mocksaml/Okta dev tenant) — needed to confirm ACS POST acceptance, `activeOrganizationId` on the post-callback session, and the cookieCache `compact` interaction (tracked as P3-2/P5-1 in the spec).
- Pilot rollout: enable the `sso` flag for one org from backoffice, then GA.
- Promotion of durable knowledge into `dev-docs/` and spec retirement.


apps/web/src/domains/sso/sso.functions.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  const provider = await findOrgProvider(organizationId)
+  return provider ? toDto(provider) : null
+})
+
+const domainSchema = z
+  .string()
+  .trim()
+  .toLowerCase()
+  .regex(/^(?!-)[a-z0-9-]{1,63}(?<!-)(\.(?!-)[a-z0-9-]{1,63}(?<!-))+$/, "Enter a valid domain, e.g. acme.com")
```
andresgutgon, line 145: Zod already have email validation:
https://zod.dev/v4?id=custom-email-regex#custom-email-regex

geclos:
@claude code review and address comments

apps/web/src/domains/sso/sso.functions.ts
geclos: this check domains not emails.

geclos:
@claude we've fixed issues 1/ and 6/ from your previous review, fix the rest of issues and nitpicks now.

PR #3464, opened 2026-06-08
Fix taxonomy lone-surrogate crash and add the fix-datadog-issues skill

csansoon:
This came out of a sweep of Datadog Error Tracking. It does two independent things, kept together because the bug is the worked example the skill is built around:

1. fixes a production crash in taxonomy session observation, and
2. adds a skill that codifies the find → triage → fix → PR workflow used to find it.

Happy to split into two PRs if you'd rather review them separately.

## taxonomy: sanitize lone UTF-16 surrogates

Two Datadog issues in `workers` shared one root cause:

- `AIError` — `Embedding failed (voyage-4-large): 400 … input must be valid UTF-8 …`
- `RepositoryError` — ClickHouse `Cannot parse escape sequence: missing second part of surrogate pair` while inserting `summary`

Session conversation text (arbitrary LLM I/O, and length-sliced previews that can split an emoji's surrogate pair) was sent unsanitized to two sinks that both reject lone UTF-16 surrogates: the Voyage embeddings API and the ClickHouse JSON insert.

The fix canonicalizes `summary` and `textToEmbed` with the existing `stripLoneSurrogates` helper (`@domain/spans`, already used by `build-trace-search-document` for the same ClickHouse constraint) in `record-session-observation`, covering both the short- and long-session paths. Fixing at the source closes both symptoms rather than patching each sink.

A regression test feeds lone surrogates and asserts both the embed input and the persisted summary are sanitized; it fails without the fix.

## new skill: fix-datadog-issues

End-to-end playbook for finding, triaging, and fixing production errors from Datadog Error Tracking and opening a PR. It's written to be run repeatedly across separate, context-free sessions.

The pipeline: navigate Datadog → pick an issue → root-cause in code → **confirm the plan with the user** (hard gate before any code is written) → reproduce with tests + fix → comment on the issue → PR to `development`.

It front-loads the things that cost time the first time through:

- disambiguations: "incidents" means Error Tracking issues (Incident Management is empty here); only v2 services (`api/ingest/web/workers/workflows`) are in scope, not the legacy `latitude-llm-*` v1 services; production only; the `mcp__latitude__*` tools are a different product surface.
- a classify-before-rank funnel, because occurrence count alone is misleading — most high-count "errors" are infra, deploy version-skew, or expected control-flow noise, not code bugs.
- prioritization by impact × trend × fix-confidence, with a recency premium for fresh regressions.
- how Error Tracking issues are read (the `error-tracking` toolset, plus a span-aggregation fallback that works without it) and recorded (comment on PR creation; never auto-resolve — a fixed error simply stops getting occurrences).

Registered in the skill glossary in `AGENTS.md`.

## testing

- `pnpm --filter @domain/taxonomy test` — 58 passing, including the new regression test (verified it fails when the fix is reverted).
- `pnpm --filter @domain/taxonomy typecheck` (tsgo) and `biome check` on changed files — clean.

## scope / follow-ups

- The `process billing` ClickHouse `deadlock detected` issue (~68 occ/7d) surfaced in the same sweep but is not addressed here.

<a href="https://capy.ai/project/7946b56a-7614-456d-9fb8-62ecca810878/pull/3464"><picture><source media="(prefers-color-scheme: dark)" srcset="https://capy.ai/api/badge/open.svg?theme=dark"><source media="(prefers-color-scheme: light)" srcset="https://capy.ai/api/badge/open.svg">Open in Capy</picture></a>

### DIFF
--- a/.agents/skills/fix-datadog-issues/SKILL.md
+++ b/.agents/skills/fix-datadog-issues/SKILL.md
@@ -116,6 +116,11 @@ search_datadog_spans(
   custom_attributes = ["error.*","http.*","issue.*"], max_tokens = 7000)
 ```
 
+**Link to an issue** (needed in the §4 report and the §7 PR). Prefer the canonical URL the error-tracking
+tool returns. Otherwise build it from the org base + issue id and open it to confirm it resolves:
+`https://app.datadoghq.eu/error-tracking/issues/<issue.id>`. (Span search responses also return a
+`base_url` and a `traces_explorer_url` you can fall back to.)
+
 ### 1.4 Query pitfalls (these bit us)
 - **`@error.message` is not reliably wildcard/full-text searchable.** `@error.message:"foo*"` may return 0.
   Instead **group by `@error.message`** in `aggregate_spans`, or filter by `@error.type` + `resource_name`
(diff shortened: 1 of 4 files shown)


csansoon:
Splitting into separate PRs. Skill: #3465. Taxonomy surrogate fix will follow in its own PR.

PR #3489, opened 2026-06-09
fix(web): apply search query to traces/sessions analytics

neoxelox:
On the traces/sessions dashboard, typing a search query filtered the table rows but left everything else showing project-wide numbers: the histogram, the analytics cards (sessions/cost/duration/tokens/ttft/traces/spans), and the table column rollups (avg/min/max/median on duration, ttft, cost, spans). This makes those surfaces reflect the matched set instead.

## Trace tab

UI wiring only — the trace repository already accepted `searchQuery` for metrics and histogram. The page never threaded the query into `TraceAggregationsPanel`, and `GeneralAggregations` / `Histogram` didn't accept it. They do now. The trace table's column rollups already worked (`TracesView` passed `searchQuery` to `useTraceMetrics`).

## Session tab

The session side had no search support below the table at all, so it gained it end to end:

- `SessionRepository` port: `searchQuery` added to `aggregateMetricsByProjectId` and `histogramByProjectId`.
- ClickHouse repo: when a search is active, both dispatch to new `aggregateMetricsBySearchQuery` / `histogramBySearchQuery` in `search-by-project.ts`. These reuse the existing `trace_rollup -> session_rollup` candidate machinery that `listSessionsBySearchQuery` / `countSessionsBySearchQuery` already use, so the panels aggregate over the same matched set the list shows — non-matching sessions drop out, per-session numerics reflect only the matched traces, and orphan-trace sessions are included.
- Server functions `getSessionMetricsByProject` / `getSessionTimeHistogramByProject`: accept `searchQuery`, and now run under the `withAi(AIEmbedLive, ...)` layer the semantic-search path needs (the list/count handlers already did; these two didn't).
- Hooks `useSessionMetrics` / `useSessionTimeHistogram`: `searchQuery` added to params and query keys.
- `SessionsView` now passes `searchQuery` to `useSessionMetrics`, which feeds the table column subheaders.

## Semantics

In search mode the session cards, histogram, and column rollups reflect **only the matched traces per session** (cost summed over matching traces, "traces" = matching-trace count), matching what the searched list already displays — not the sessions' full all-trace totals.

## Tests

Added three ClickHouse repo tests (search-mode metrics, histogram, empty-match) asserting the panels agree with `countByProjectId`'s `totalCount` / `matchingTraceCount`. `@platform/db-clickhouse` (228) and `@app/web` (413) suites pass; typecheck clean on the three affected packages.

### DIFF
--- a/apps/web/src/domains/sessions/sessions.collection.ts
+++ b/apps/web/src/domains/sessions/sessions.collection.ts
@@ -253,21 +253,24 @@ export function useSessionCohortSummary({ projectId }: { readonly projectId: str
 export function useSessionMetrics({
   projectId,
   filters,
+  searchQuery,
 }: {
   readonly projectId: string
   readonly filters?: FilterSet
+  readonly searchQuery?: string
 }) {
   const scope = use(TraceScopeContext)
   const effectiveFilters = useMemo(() => withSessionDefaults(filters), [filters])
 
   return useQuery({
-    queryKey: [...traceScopeKey(scope), "sessions-metrics", projectId, effectiveFilters],
+    queryKey: [...traceScopeKey(scope), "sessions-metrics", projectId, effectiveFilters, searchQuery],
     queryFn: () =>
       getSessionMetricsByProject({
         data: {
           ...traceScopeData(scope),
           projectId,
           filters: effectiveFilters,
+          ...(searchQuery ? { searchQuery } : {}),
         },
       }),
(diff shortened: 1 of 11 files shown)


geclos:
Wait don't merge i'm redoing the search, will roll it up with my changes.

neoxelox:
closing as @geclos will merge this changes with his changes

PR #3491, opened 2026-06-09
chore(oss): Phase 2 — house-cleaning & foundational fixes [LAT-655]

neoxelox:
## Phase 2 — House-cleaning & foundational fixes

Phase 2 of the OSS self-hosting initiative ([`specs/oss.md`](../blob/development/specs/oss.md)). Strips cruft and lands the audit-driven fixes so later tiers build on a clean, fully-permissive, coexistence-ready base. Shipped as a single milestone PR.

### What's in here

**P2-1 — `ua-parser-js` (AGPL-3.0) → `bowser` (MIT)**
Server-side User-Agent parsing in the web app now uses `Bowser.parse(ua)` / `platform.type` (admin user-details + account sessions). The shipped web bundle is now 100% permissive. `bowser@2.14.1` was already in the lockfile (transitive via `@aws-sdk/util-user-agent-browser`), so no net-new package. Device-icon map trimmed to Bowser's value set (`desktop|mobile|tablet|tv`).

**P2-2 — Remove 5 stale leftover package dirs**
`db-weaviate`, `events-outbox`, `domain/{auth,subscriptions,onboarding}` each held only a gitignored `.tsbuildinfo` (no `package.json`/`src`, zero importers). Removed locally — **no diff here** because git never tracked them; confirmed via `git ls-files`.

**P2-3 — knip sweep**
`knip` reports zero unused files/exports/deps after P2-1/P2-2, and a bidirectional `.env.example`↔source check shows no `LAT_*`/`VITE_LAT_*` drift. Nothing to prune — verification only.

**P2-4 — Namespace cache + BullMQ on a shared Redis**
Hardcoded shared prefix (no env var, mirrors v1):
- `REDIS_KEY_PREFIX = "latitude:"` → ioredis `keyPrefix` on the single shared cache client (single-node + cluster). All cache/lock/rate-limiter/budget keys land under `latitude:`, including Lua `eval` KEYS (ioredis prefixes them by `numkeys` — verified in `@ioredis/commands` `getKeyIndexes`).
- `BULLMQ_PREFIX = "latitude:{bull}"` (derived from the constant) → keeps the `{bull}` cluster hash-tag and avoids the v1 double-colon (`latitude:{bull}:<queue>:<type>`, never `latitude::…`).
- New `constants.test.ts` guards the no-double-colon / hash-tag / shared-prefix invariants; `client.test.ts` asserts the `keyPrefix`.

**P2-5 — De-bake web deployment URLs (one image serves any domain)**
- `VITE_LAT_API_URL` deleted everywhere (had zero readers — web calls the API via same-origin server-fn RPC).
- `VITE_LAT_WEB_URL` removed from the bundle: `auth-client` drops `baseURL` (Better Auth resolves `window.location.origin` / relative `/api/auth`), `login` uses a relative OAuth-start path, and the Wrapped `og:image` reads runtime `LAT_WEB_URL` via `getWrappedPageData`.
- Datadog source-map `--minified-path-prefix` now reads `${{ vars.DD_SOURCEMAP_URL_PREFIX }}`.

### Verification
- Typecheck green: `cache-redis`, `queue-bullmq`, `web`, `infra`, and the `api`/`ingest`/`workers`/`workflows` consumers.
- Tests green: cache-redis (25), queue-bullmq (34, incl. the new double-colon guard), web admin+sessions (26).
- biome-formatted; knip clean.

### ⚠️ Manual coordination before merge / on cutover

1. **Repo settings (P2-5, out of band):** create GitHub Actions repo variable **`DD_SOURCEMAP_URL_PREFIX`** (= public web origin, the old `VITE_LAT_WEB_URL` value) and remove the now-unused `VITE_LAT_API_URL` / `VITE_LAT_WEB_URL` Actions vars. Code refs are updated; without the new var the prod source-map upload passes an empty prefix.
2. **BullMQ migration (P2-4):** the queue prefix moves `{bull}` → `latitude:{bull}`, orphaning in-flight/delayed/repeatable jobs under the old prefix on the cutover deploy (cache is ephemeral, so it just cold-starts). Drain the queues or accept the one-time loss.
3. P2-5 also touches `deploy.yml` (prod pipeline) and `infra/lib/ecs.ts` (normally untouched) — dead/no-op cleanup only, flagged in the spec to coordinate before landing.

Spec tasklist (`specs/oss.md`) updated: P2-1…P2-5 → `[x]` with per-task `**Result:**` notes; Appendix B refreshed.


Dockerfile
```diff
@@ -76,8 +76,11 @@ RUN pnpm turbo run build --filter @app/workflows
 # ---------------------------------------------------------------------------
 FROM source AS build-web
 
-ARG VITE_LAT_API_URL
-ARG VITE_LAT_WEB_URL
+# No deployment URLs are baked: frontend URLs are relative and the one
+# SSR-absolute case (og:image) reads runtime LAT_WEB_URL, so one image serves
+# any domain. The optional VITE_* below are Latitude-cloud-only telemetry —
+# left unset on the public image, they simply no-op (nothing deployment-specific
+# is baked).
```
andresgutgon, line 83: Yapping

.github/workflows/deploy.yml
```diff
@@ -489,7 +487,7 @@ jobs:
           pnpm dlx @datadog/datadog-ci sourcemaps upload apps/web/.output/public \
             --service web \
             --release-version "$RELEASE" \
-            --minified-path-prefix "${{ vars.VITE_LAT_WEB_URL }}"
+            --minified-path-prefix "${{ vars.DD_SOURCEMAP_URL_PREFIX }}"
```
andresgutgon, line 490: Are these env vars the same?

andresgutgon (approved):

PR #3492, opened 2026-06-09
feat(sandbox): one sandbox per org with a sidebar toggle

andresgutgon:
## Summary

Treats Test Mode as **one sandbox per organization** for now, without touching the data model (which still supports multiple sandboxes per plan). All behavior and surfaces are simplified around a single sandbox.

## Changes

- **Cap all plans to one active sandbox** (`packages/domain/billing/src/constants.ts`) as a server-side backstop; the UI never offers a second one.
- **Sidebar dropdown → context-aware toggle** (`sandbox-switcher.tsx`): flipping it find-or-creates the org's sandbox **and** a sandbox project mirroring the live project you're currently in, then navigates there. Re-toggling just navigates.
- **"Switch to live"** returns to the linked live project you came from, not the org default.
- **Removed the settings → Sandboxes list** and the create modal. Manual archive/delete are dropped from the UI (auto idle-archive is unchanged).
- **Archived header**: gray strip with an **Activate** button (default variant) inline with the message, a **disabled** "View configuration", and a click-blocking veil over the interface so the only live action is Activate.
- **No FOUC**: sandbox `status` is threaded through route context (`resolveSandboxAccess` → middleware → `getSandbox`) so the strip knows it on first paint; `router.invalidate()` flips it back after Activate.
- **Add-project chooser** collapses to just the new-project name field when every production project is already attached.
- **Docs** (`docs/more/sandbox.mdx`): describe a single sandbox + the toggle; reactivation via the banner; dropped per-plan sandbox table and manual archive/delete language.

## Verification

- `@app/web` + `@domain/billing` typecheck clean; Biome clean.
- `@domain/sandboxes` (8) and `@domain/billing` (29) tests pass with caps = 1.
- Manual: toggle from a live project creates/enters the sandbox twin; re-toggling navigates; archived sandbox shows gray + Activate and reactivates back to blue; settings Sandboxes page is gone.


andresgutgon:
@claude review please

PR #3496, opened 2026-06-10
feat: share message embeddings across trace search and CI

geclos:
## What changed

This PR now contains the full shared message embeddings epic from `specs/shared-message-embeddings.md`, not just the initial store setup.

It moves trace search and conversation intelligence onto a shared, message-level embedding model. Both pipelines canonicalize messages as `"{role}: {text}"`, hash that canonical text through `@domain/spans`, look up existing vectors in `message_embeddings`, and embed only cache misses under the existing organization embedding budget.

## Shared embedding model

- Adds the `message_embeddings` ClickHouse table and repository for content-addressed vectors scoped by organization, project, and embedding model.
- Adds `trace_message_occurrences`, which records where each hashed message appears in each trace without duplicating the vector.
- Keeps trace identity out of the vector table so CI-written embeddings cannot create phantom search results; search only enters through occurrence rows.
- Adds cross-pipeline tests for the canonicalization/hash invariant.

## Conversation intelligence

Conversation analysis now resolves turn embeddings from the shared store before running the existing segmentation, label detection, centroid, and taxonomy logic. Static moment-label anchors are cached per detector/model configuration, and unchanged taxonomy projections reuse the previous embedding instead of re-embedding the same projection text.

Session reconstruction also switches to an occurrence-derived spine when occurrence rows exist: distinct message hashes are ordered by first seen position across the session, with a legacy session-detail fallback for older data. This preserves compacted history, marks compaction-summary candidates, excludes those summaries from anchor-label scoring, and bumps the detector version to `ci-v5` for the index-shifting behavior change.

## Trace search

The trace-search worker now writes shared message embeddings and per-trace occurrence rows while leaving the lexical document path in place. Semantic reads can use the shared model behind `LAT_TRACE_SEARCH_SHARED_MESSAGE_EMBEDDINGS_READS`: query vectors search distinct message embeddings, fan out through occurrences, and max-pool back to trace relevance/highlight metadata.

The shared semantic path also:

- uses ClickHouse vector-search settings for the ANN-shaped top-K read;
- caps candidates after occurrence fan-out;
- filters high-frequency boilerplate messages from semantic scoring;
- recalibrates the semantic relevance floor to `0.35` for per-message vectors;
- keeps the legacy chunk table/read path available while old traces age out instead of backfilling them.

The web search URL state was adjusted so active searches default to relevance sorting unless the user explicitly chooses another sort.

## Validation

Validated across the merged PRs with focused domain, worker, workflow, and ClickHouse repository tests plus package typechecks/checks, including:

- `@domain/spans` canonicalization and repository coverage
- `@domain/conversation-intelligence` analysis tests and typecheck
- `@platform/db-clickhouse` message embedding, trace search, session repository, and search-plan tests
- `@app/workflows` analyze-session workflow tests and typecheck
- `@app/workers` trace-search tests and typecheck
- commit hooks for workspace format and `knip`


packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00029_message_embeddings.sql
```diff
@@ -0,0 +1,21 @@
+-- +goose NO TRANSACTION
+-- +goose Up
+
+CREATE TABLE IF NOT EXISTS message_embeddings
+(
+    organization_id    LowCardinality(String)                CODEC(ZSTD(1)),
+    project_id         LowCardinality(String)                CODEC(ZSTD(1)),
+    content_hash       String                                CODEC(ZSTD(1)),
+    embedding          Array(Float32)                        CODEC(ZSTD(1)),
+    embedding_model    LowCardinality(String)                CODEC(ZSTD(1)),
```
andresgutgon, line 12: Is there something else other than embedding model that can change?

geclos: nope, the dimension on the embeddings but that's a can of worms

geclos:
@cursoragent review

geclos:
@claude review

PR #3503, opened 2026-06-10
feat(issues): Related issues section — semantic similarity + session co-occurrence

csansoon:
## what this does

Adds a **Related issues** section to the issue page (Phase 4 of the issue-details-page spec, minus the benchmark): a card grid at the bottom of the page listing the project's most related issues, each linking to its own page.

"Related" merges two independent signals into one ranking:

- **Semantic similarity** — this issue's failure pattern *means* the same thing as another's. Computed as cosine similarity between the issues' `centroid_embedding` vectors (already maintained by the clustering pipeline; no new storage). Surfaces near-duplicate clusters and "a similar issue was already resolved — see how".
- **Session co-occurrence** — the two issues *travel together*: they have occurrences in the same conversations (trailing 30 days). Surfaces correlated/cascading failures that share a root cause.

Either signal alone carries a card; a card scoring on both is the "possibly the same issue" case and ranks highest.

## scoring

Each signal is normalized to `[0, 1]` and fused; the math lives in a pure `@domain/issues` module (`related-issues.ts`) so it's unit-tested without infra:

- **Semantic**: linear rescale of cosine over its useful band (`0.55–0.85`, named constants). Above the ceiling, discovery would have merged the clusters; below the floor it's noise. Surviving issue pairs sit below the discovery merge threshold by construction, so the band is where the signal lives. The band is provisional — calibrate against real data.
- **Co-occurrence**: clamped **NPMI** over session sets, gated on ≥ 3 shared sessions. NPMI instead of raw lift (unbounded, explodes for rare pairs) or overlap percent (inflated by big issues that overlap everything by chance): its numerator is log-lift and its denominator normalizes by joint-event rarity, killing both failure modes. The probability universe is sessions carrying at least one issue occurrence — unscored sessions carry no information about issue association.
- **Fusion**: noisy-OR (`1 − (1−a)(1−b)`), min-relatedness gate, top 8.

The combined score and raw cosine are never displayed — they rank, the reason chips explain (`Similar topic` / `Very similar topic` vs `Same conversations · 64%`, each with a tooltip).

## pieces

- `IssueRepository.findSimilarByCentroid` (Postgres): exact pgvector cosine scan over the project's other issues, same no-ANN-index trade-off as `hybridSearch`. Empty when the source has no embedding. Resolved/ignored neighbors included on purpose.
- `ScoreAnalyticsRepository.coOccurrenceByIssue` (ClickHouse): per-candidate shared/total session counts + the probability universe, one windowed scan over issue-carrying scores, self-excluded, capped candidate pool.
- `getRelatedIssuesUseCase` (`@domain/issues`): runs both reads in parallel, ranks with the pure scorer, hydrates rows from Postgres (`findByIds`) with lifecycle states, plus a batched `aggregateByIssues` read for each card's occurrences / last-seen footer. Candidates that vanished between reads are dropped.
- Web: `getRelatedIssues` server fn, `useRelatedIssues` hook, `IssueRelated` card grid (name + lifecycle badges, 2-line description so the reader can judge the similarity, reason chips, activity footer). Rendered through a new `append` slot on `IssueDetailBody` so it sits after Traces as a page footer — it's a "where to go next" section, not evidence about the issue itself.

## scope cuts / follow-ups

- The project **benchmark** (occurrence/user-reach percentile, original P4-2) is not in this PR — tracked as P4-5 in the spec.
- Semantic band constants (`ISSUE_RELATED_SEMANTIC_FLOOR/CEILING`) need calibration against production data once real clusters exist.
- Co-occurrence cards require occurrences with session ids, ≥ 3 shared sessions in 30 days, and above-chance overlap — sparse projects will mostly see semantic cards, by design.

## testing

- PGlite: `findSimilarByCentroid` ranking, self/project exclusion, resolved/ignored inclusion, missing-embedding degradation.
- chdb: `coOccurrenceByIssue` shared/total counts, windowing, self-exclusion, candidate cap.
- Unit: scorer math, including the chance-overlap (lift = 1 → 0) and big-neighbor traps, perfect-rare-pair = 1, gates, ordering, top-N.
- Use case: signal merge, hydration, lifecycle states, 30-day window, vanished-candidate drop, missing-analytics degradation.

Spec updated in `specs/issue-details-page.md` (Data model #3, glossary, layout, task status).

### DIFF
--- a/packages/platform/db-postgres/src/repositories/issue-repository.ts
+++ b/packages/platform/db-postgres/src/repositories/issue-repository.ts
@@ -343,6 +343,59 @@ const issueRepositoryCoreLive = Layer.effect(
           }))
         }),
 
+      findSimilarByCentroid: ({ projectId, issueId, limit }) =>
+        Effect.gen(function* () {
+          const sqlClient = (yield* SqlClient) as SqlClientShape<Operator>
+
+          // Two round-trips instead of a self-join: the source embedding is read
+          // first so the second query can inline it as a pgvector literal (the
+          // same wire-format constraint `hybridSearch` documents — a JS number[]
+          // binds as a Postgres array, which cannot be cast to `vector`).
+          const [source] = yield* sqlClient.query((db, organizationId) =>
+            db
+              .select({ centroidEmbedding: issues.centroidEmbedding })
+              .from(issues)
+              .where(
+                and(eq(issues.organizationId, organizationId), eq(issues.projectId, projectId), eq(issues.id, issueId)),
+              )
+              .limit(1),
+          )
+
+          // Missing issue or zero-mass centroid (no embedding persisted): the
+          // semantic signal degrades to nothing rather than failing the read.
+          const embedding = source?.centroidEmbedding ?? null
+          if (embedding === null) return []
+          const vector = yield* validateVector(embedding, "IssueRepository.findSimilarByCentroid")
+          const queryVector = sql.raw(`'[${vector.join(",")}]'::vector`)
+
+          // Exact cosine scan over the project's other issues — no ANN index by
+          // design (see the schema comment on `centroidEmbedding`). Resolved and
+          // ignored issues are deliberately included. `save()` only persists
+          // embeddings for CENTROID_EMBEDDING_MODEL, so every non-null row is in
+          // the same embedding space by construction.
+          const similarity = sql<number>`(1::double precision - (${issues.centroidEmbedding} <=> ${queryVector}))`
+          const rows = yield* sqlClient.query((db, organizationId) =>
+            db
+              .select({ issueId: issues.id, similarity })
+              .from(issues)
+              .where(
+                and(
+                  eq(issues.organizationId, organizationId),
+                  eq(issues.projectId, projectId),
+                  ne(issues.id, issueId),
+                  isNotNull(issues.centroidEmbedding),
+                ),
+              )
+              .orderBy(desc(similarity), desc(issues.updatedAt), asc(issues.id))
+              .limit(limit),
+          )
+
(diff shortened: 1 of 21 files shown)


packages/platform/db-postgres/src/repositories/issue-repository.ts
```diff
@@ -343,6 +343,59 @@ const issueRepositoryCoreLive = Layer.effect(
...
+                and(eq(issues.organizationId, organizationId), eq(issues.projectId, projectId), eq(issues.id, issueId)),
+              )
+              .limit(1),
+          )
+
+          // Missing issue or zero-mass centroid (no embedding persisted): the
+          // semantic signal degrades to nothing rather than failing the read.
+          const embedding = source?.centroidEmbedding ?? null
+          if (embedding === null) return []
+          const vector = yield* validateVector(embedding, "IssueRepository.findSimilarByCentroid")
```
andresgutgon, line 368: Is this calling any service or is pure postgresql?

csansoon: Pure PostgreSQL — no external service and no model call. `<=>` is the pgvector cosine-distance operator, so this is an exact scan over the project's `centroid_embedding` column (same pattern as `hybridSearch`/`searchOrgWide`, deliberately no ANN index — exact scan wins below ~10k issues/project, see the schema comment on `centroidEmbedding`).

No embeddings are computed at read time either: the query vector is the *source issue's own* stored `centroid_embedding` (read in the first query above and inlined as a vector literal), which the clustering pipeline already maintains on every score assignment. So the whole read is two Postgres round-trips.

andresgutgon (approved):

PR #3505, opened 2026-06-10
feat: issue assignee and priority across the list, palette, and notifications

csansoon:
Issue `assigneeId` and `priority` landed with the issue details page (`specs/issue-details-page.md`) but nothing read them — they were set on the page and displayed nowhere else. This PR makes both fields functional across the issues list, the command palette, and the notification pipeline.

## issues list: grouped by priority, filterable by assignee

The list is now always grouped by triage priority, Linear-style: `Urgent → High → Medium → Low → No priority`, with full-width group headers showing each group's count over the whole filtered set (not just loaded pages). There is no group-by toggle; the user-selected sort (last seen / occurrences / state) applies within each group.

The grouping is implemented as an **unconditional primary sort key** in `listIssuesUseCase` rather than a UI-side transform. Since the use case sorts the full candidate set in memory and paginates by slice, this one change keeps exports, bulk select-all pagination, and the issue page's prev/next navigation in the same order the table shows. `InfiniteTable` gained a `getRowGroup`/`renderGroupHeader` prop pair that injects virtualized header rows at group boundaries — headers live outside the data array, so selection, row links, and active-row scrolling never see them.

New on the toolbar:

- an **assignee filter** (multi-select over org members plus an `Unassigned` option), stored in the `issuesAssignees` URL param and forwarded to bulk lifecycle actions and CSV exports so select-all targets the visible set
- a **My issues** toggle with a count badge. The count respects the current tab/time/search filters but not the assignee filter itself — `assigneeCounts` is computed before that filter is applied, so activating the button can't zero its own badge
- an **Assignee column** (avatar + name), hydrated client-side from the members collection; the list payload carries only `assigneeId`

There is no priority column — the group headers already say it. Priority icons/labels moved to a shared `issue-priority-meta` module so the triage picker, group headers, and palette can't drift.

## command palette

While an issue page is open, ⌘K offers `Assign to…` (Me / Unassigned / org members, current marked) and `Set priority…` drill-downs. Both run the same `updateIssueTriage` mutation as the right-rail pickers, which now also invalidates the issues list so triage edits regroup it.

## incident notifications carry the triage snapshot

The incident-notification producer loads the issue row for issue-sourced incidents and snapshots `assigneeId` + `priority` onto all three payload kinds. Fields are optional, so stored pre-change rows and savedSearch-sourced incidents keep parsing. The assignee's *name* is deliberately not snapshotted — renderers resolve it live from the id (email/Slack via `UserRepository`, in-app via the members collection):

- **email**: a `PriorityBadge` pill (same shape as `SeverityBadge`) and an "Assigned to" row in the metadata table
- **Slack**: a priority emoji + "Assigned to X" suffix on the incident context line
- **in-app**: priority icon next to the issue name and a muted assigned-to line on the summary card

Incident `severity` is untouched: it stays a pure function of kind (`SEVERITY_FOR_KIND`). Priority is a manual judgment about the issue, severity describes the event — they render side by side instead of one overwriting the other.

## `issue.assigned` notification, new `personal` group

Changing an issue's assignee notifies the new assignee ("Carlos assigned you to *<issue>*") in-app and by email:

- `updateIssueTriageUseCase` now requires `actorUserId` and emits an `IssueAssigneeChanged` outbox event from its transaction whenever the assignee actually changes (priority-only edits stay silent)
- the domain-events router publishes `request-issue-assigned-notifications`, skipping cleared assignments and self-assignments; a new single-recipient producer re-checks both rules and anchors the row to the issue's project for the `ProjectDeleted` cascade
- **idempotency** keys on `issueId` + the transaction-frozen `assignedAt`. The `(org, user, key)` unique index is permanent, so keying on `issueId:assigneeId` would suppress a legitimate later re-assignment forever, and a generated id would duplicate on outbox redelivery. The frozen timestamp gives both properties: redelivery replays the same key, A→B→A mints a new one

The kind lives in a new `personal` notification group ("notifications addressed directly to you"). Email defaults on; the account-settings toggle appears automatically. The group is **not Slack-routable** — group meta gained a `slackRoutable` flag enforced in three places (routes settings UI, route-config server fns, worker fan-out). The exhaustive Slack renderer registry keeps a never-called stub so future kinds still have to make an explicit Slack decision.

## scope cuts

- No assignee-only incident routing: most issues are unassigned when an incident fires, and gating delivery on the assignee would drop notifications exactly when awareness matters. A per-user "only my issues" preference is a possible follow-up.
- No collapsible group headers (v1 keeps them static).
- No project-level gate for `personal` — a notification addressed at one user shouldn't be silenced by a project's broadcast mute.

## testing

Backend tests per repo convention: grouping order / within-group sort / pagination stability / assignee + unassigned filters / count semantics in `list-issues`, export narrowing, outbox emission (assign/clear/reassign, silent on priority-only and no-ops), the single-recipient producer (skips, redelivery vs re-assignment key behavior), unique-index dedupe at the create step, payload backwards-compat for stored incident rows, idempotency-key shapes, domain-events routing, and the Slack triage suffix. ~470 tests pass across the touched packages; the only failures anywhere are pre-existing `@platform/latitude-api` issues verified present without this branch.


packages/ui/src/components/infinite-table/infinite-table.tsx
```diff
@@ -296,8 +335,26 @@ export function InfiniteTable<T>({
...
+                // DataRow, no row link/click — so selection and navigation
+                // semantics never see it.
+                return (
+                  <tbody key={`group-${displayRow.groupKey}`} ref={virtualizer.measureElement} data-index={index}>
+                    <tr>
+                      <td colSpan={colCount} className="p-0 border-none">
+                        {renderGroupHeader?.(displayRow.groupKey)}
+                      </td>
+                    </tr>
+                  </tbody>
```
andresgutgon, line 415: Buah enterprise level now 😂 tables with grouped sections

andresgutgon (dismissed):

PR #3508, opened 2026-06-10
feat(tools): Tools dashboard — every defined and called tool with usage, failure and latency analytics

csansoon:
## what this does

Adds a project-level **Tools** dashboard (behind a new `tools` feature flag): every tool your agents *define* or *call*, with usage, failure and latency analytics — plus a full-page detail view per tool, a page-wide errors-only analysis mode, a `tools` filter on the traces/sessions surfaces, and a "View tool analytics" link from `execute_tool` span details.

Tool telemetry has two independent sides and a tool can exist on either alone, so the inventory is their union:

- **Defined** — tool definitions ride chat spans (`tool_definitions` JSON blob, already populated from the GenAI / Vercel / OpenInference / Claude Code SDK conventions). A defined-but-never-called tool shows an `Unused` badge and its offered count.
- **Called** — `execute_tool` spans with their own `tool_name` / `tool_input` / `tool_output` / status columns. A called-but-never-defined tool shows `No definition`.

https://github.com/user-attachments/assets/1dba5c05-9e1c-42c5-8cff-6352d402ef66

## storage

Two migrations, no new tables:

- `00023` adds `tool_names Array(LowCardinality(String))` to `spans`, MATERIALIZED from the `tool_definitions` blob, plus bloom-filter indexes on `tool_name` and `tool_names`. Writers are untouched (materialized columns can't be inserted).
- `00024` backfills old parts (`MATERIALIZE COLUMN/INDEX`, background mutation). Until it completes, reads over historical ranges compute `tool_names` on the fly from the ZSTD(3) blob — ship both together.

Why not the ClickHouse `JSON` type: verified against 26.2 that an in-place `ALTER` fails (our blobs are top-level *arrays*, which the JSON type rejects), and the type is lossy — it drops empty objects like `"parameters": {}` and normalizes key order. The String blob stays the display source of truth; the latest definition is extracted losslessly with `argMax` + `arrayFirst` over the raw JSON text.

No rollup MV for v1: every query is bounded by the spans primary-key prefix `(org, project, start_time)` plus the new bloom indexes. Revisit if list latency demands it.

## reads

New `ToolAnalyticsRepository` (`@domain/spans` port, ClickHouse adapter):

- list: usage aggregate `FULL OUTER JOIN` defined aggregate on tool name (`join_use_nulls` expresses the "No calls" / "Definition not found" sides), with project totals and per-tool trend buckets fetched in parallel — sparklines without N+1 queries
- per-tool: usage summary, time histogram, parameter stats over a bounded recent sample (10k calls, `topK` values), tag breakdown, co-occurrence (tools sharing the anchor tool's traces), cursor-paginated recent calls (deduped `LIMIT 1 BY span_id`, 4 KB payload previews)
- count semantics match `traces_mv` (raw `count()`, re-ingestion can transiently double-count until merges); trace/session ratios use `uniqExact`

Two derived metrics worth defining: **calls per offer** = execute_tool count ÷ chat spans offering the tool (how often the model picks it when available; >100% means multi-call turns), and **% of traces/sessions** = share of the window's traces/sessions with at least one call.

## the pages

**List** (`/projects/:slug/tools`): stats panel (tools, calls, error rate, traces/sessions using tools, unused count + stacked ok/failed chart with an error-rate line), sortable table with trend sparklines and `Unused` / `Failing` / `No definition` badges, All | Unused | Failing tabs, search, 30-day default window, configurable columns.

**Detail** (`/projects/:slug/tools/:toolName`): usage tile band; calls-over-time + latency-over-time charts; a **Parameters** explorer listing schema-defined ∪ observed input keys (type badge, description tooltip; selecting one shows top values as shares of all sampled calls, with a "Not included" entry — defined-but-never-sent params read as `Not included · 100%`, and the section renders for never-called tools so dead tools keep a useful page); "Where it's used" (tags + co-occurring tools linking to their own pages); recent calls opening the trace sheet on the Spans tab with the call's span preselected; J/K prev/next-tool nav; "View traces" deep link.

**Errors only**: a header switch (URL param) scopes the charts, parameter sample, context panel and recent calls to failed calls. The usage band switches to a relabeled variant (Failed calls, Affected traces/sessions, p50/p95 of failures, Last failed); error rate deliberately stays global so it never reads 100%.

## tools filter on traces & sessions

`tools` is a new multiselect filter field on both surfaces, resolved through a synthetic spans subquery (`trace_id/session_id IN (SELECT … execute_tool …)`) because the traces/sessions rollup MVs have no tool column and can't be backfilled. The options list is the called ∪ defined union, and is deliberately not session-gated — the traces page defaults to the Sessions tab and most spans carry no session id, so a session-scoped list hid most known tools.

## also in here

- `dump-schema.sh` strips `TYPE text` indexes (26.2-only syntax the chdb test engine can't parse) — before this, any fresh schema dump broke every chdb suite
- seed pattern generators fail ~6% of tool calls so dashboards show realistic error rates
- the trace detail drawer's local-tabs variant gains `initialSpanId` for span preselection

## scope cuts / follow-ups

- duration *distribution* chart skipped — the latency-over-time p50 chart covers the need with data we already bucket
- model/provider context breakdowns exist in the repository but aren't rendered (single-model projects made them noise); the traces filter answers that question instead
- the filter subquery is unbounded in time, matching the score-filter subquery convention

## testing

- chdb: tool-analytics suite (union semantics, lossless `"parameters": {}` round-trip, latest-definition-wins, metrics/rates math, histogram bucketing, parameter sampling, co-occurrence, cursor pagination, re-ingestion dedup, payload truncation, errors-only scoping) plus trace/session regressions for the filter clause and the options union
- query shapes that motivated design decisions (JSON-type failures, FULL OUTER JOIN null sides, tuple cursor, window + `LIMIT BY` parameter stats) were verified against a real ClickHouse 26.2 before landing
- end-to-end against seeded data: 65-tool inventory with called-only, defined-only and failing tools across list, detail, filters and deep links


apps/web/src/routes/_authenticated/projects/$projectSlug/tools/-components/tools-view.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+/** Client-side sort — the whole list is loaded in one query. */
+export function sortTools(
+  tools: readonly ToolSummaryRecord[],
+  sorting: ToolsTableSorting,
+): readonly ToolSummaryRecord[] {
+  const getValue = SORT_VALUE[sorting.column]
+  const sign = sorting.direction === "asc" ? 1 : -1
+  return [...tools].sort((a, b) => {
+    const diff = (getValue(a) - getValue(b)) * sign
```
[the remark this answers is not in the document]
csansoon: Fixed in 228365035 — `SORT_VALUE` is now a `Map` with an explicit fallback getter, so the lookup can never dispatch to an `Object.prototype` member even if a column value escaped `parseSorting` validation.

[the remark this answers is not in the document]
csansoon: Same fix as the sibling alert, in 228365035 — object indexing replaced with a `Map` lookup + fallback.

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00023_tool_analytics.sql
```diff
@@ -0,0 +1,21 @@
...
+-- tool_definitions JSON blob so tool analytics can aggregate without
+-- decompressing the full ZSTD(3) payload. '' parses to [] (JSONExtractArrayRaw('') = []).
+-- MATERIALIZED: old parts compute it on the fly at read until 00024 backfills.
+-- Keep one ALTER per table to reduce DDL metadata churn on replicated clusters.
+ALTER TABLE spans ON CLUSTER default
+    ADD COLUMN IF NOT EXISTS tool_names Array(LowCardinality(String))
+        MATERIALIZED arrayMap(x -> JSONExtractString(x, 'name'), JSONExtractArrayRaw(tool_definitions))
+        CODEC(ZSTD(1)) AFTER tool_output,
+    ADD INDEX IF NOT EXISTS idx_tool_name  tool_name  TYPE bloom_filter(0.01) GRANULARITY 4,
+    ADD INDEX IF NOT EXISTS idx_tool_names tool_names TYPE bloom_filter(0.01) GRANULARITY 4;
```
andresgutgon: How much storage means storing tools this way?

geclos: why granularity 4?

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00023_tool_analytics.sql
```diff
@@ -0,0 +1,20 @@
...
+
+-- Tool names offered to the LLM on chat spans, extracted from the
+-- tool_definitions JSON blob so tool analytics can aggregate without
+-- decompressing the full ZSTD(3) payload. '' parses to [] (JSONExtractArrayRaw('') = []).
+-- MATERIALIZED: old parts compute it on the fly at read until 00024 backfills.
+ALTER TABLE spans
+    ADD COLUMN IF NOT EXISTS tool_names Array(LowCardinality(String))
+        MATERIALIZED arrayMap(x -> JSONExtractString(x, 'name'), JSONExtractArrayRaw(tool_definitions))
+        CODEC(ZSTD(1)) AFTER tool_output,
+    ADD INDEX IF NOT EXISTS idx_tool_name  tool_name  TYPE bloom_filter(0.01) GRANULARITY 4,
```
geclos: tool_name column  existed already or is this a typo?

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00024_materialize_tool_names.sql
```diff
@@ -0,0 +1,14 @@
...
+-- stop computing the column on the fly from the ZSTD(3) tool_definitions blob.
+-- These submit background mutations; they do not block.
+ALTER TABLE spans ON CLUSTER default MATERIALIZE COLUMN tool_names;
+ALTER TABLE spans ON CLUSTER default MATERIALIZE INDEX idx_tool_name;
+ALTER TABLE spans ON CLUSTER default MATERIALIZE INDEX idx_tool_names;
+
+-- +goose Down
+
+-- No-op: materialization cannot (and need not) be reversed.
+SELECT 1;
```
geclos, line 14: can we do the migration in a single migration intead of 2?

packages/platform/testkit/src/clickhouse/schema.sql
```diff
@@ -187,6 +92,74 @@ GROUP BY
...
+    `coherence_score` Float32 DEFAULT 0. CODEC(ZSTD(1)),
+    `retention_days` UInt16 DEFAULT 90 CODEC(T64, ZSTD(1)),
+    `indexed_at` DateTime64(3, 'UTC') DEFAULT now64(3) CODEC(Delta(8), LZ4)
+)
+ENGINE = ReplacingMergeTree(indexed_at)
+PARTITION BY toYYYYMM(start_time)
+PRIMARY KEY (organization_id, project_id, session_id, analysis_hash, moment_id)
+ORDER BY (organization_id, project_id, session_id, analysis_hash, moment_id)
+SETTINGS index_granularity = 8192;
+
```
geclos, line 162: why are you moving this

packages/platform/testkit/src/clickhouse/schema.sql
```diff
@@ -386,14 +360,47 @@ CREATE TABLE spans
...
+    INDEX idx_taxonomy_observations_analysis_hash analysis_hash TYPE bloom_filter(0.01) GRANULARITY 4,
+    INDEX idx_taxonomy_observations_observation_id observation_id TYPE bloom_filter(0.01) GRANULARITY 4,
+    INDEX idx_taxonomy_observations_cluster_id assigned_cluster_id TYPE bloom_filter(0.01) GRANULARITY 4
+)
+ENGINE = ReplacingMergeTree(indexed_at)
+PARTITION BY toYYYYMM(start_time)
+PRIMARY KEY (organization_id, project_id, observation_id)
+ORDER BY (organization_id, project_id, observation_id)
+SETTINGS index_granularity = 8192;
+
```
geclos, line 406: wat is this doing here

packages/platform/db-clickhouse/src/repositories/session-repository.ts
```diff
@@ -999,24 +999,50 @@ export const SessionRepositoryLive = Layer.effect(
...
+                          AND operation = 'execute_tool'
+                        UNION ALL
+                        SELECT arrayJoin(tool_names) AS val
+                        FROM spans
+                        WHERE organization_id = {organizationId:String}
+                          AND project_id = {projectId:String}
+                      )
+                      WHERE val != ''${searchClause}
+                      ORDER BY val
+                      LIMIT {limit:UInt32}`
```
geclos: could we persist it in sessions instead of doing this subquery in another table? trhis is fine but it adds special cases to read time queries that i'd like to avoid if adding tools to sessions is easy enough

packages/platform/db-clickhouse/src/repositories/trace-repository.ts
```diff
@@ -1440,24 +1440,46 @@ export const TraceRepositoryLive = Layer.effect(
...
-          return yield* chSqlClient
-            .query(async (client) => {
-              const result = await client.query({
-                query: `SELECT DISTINCT val FROM (
-                        SELECT ${expr} AS val
+          // Tools have no traces column — union called tools (execute_tool
+          // spans) with defined-but-never-called ones (tool_names on chat
+          // spans) so the options match the Tools page inventory.
+          const query =
+            column === "tools"
```
geclos: ditto

packages/platform/db-clickhouse/src/registries/helpers.ts
```diff
@@ -56,6 +56,52 @@ export function buildStatusClause(
   }
 }
 
+function collectToolNames(value: FilterCondition["value"]): readonly string[] {
+  const raw = Array.isArray(value) ? value : [value]
+  return raw.filter((v): v is string => typeof v === "string" && v.length > 0)
+}
+
+/**
+ * Tool filters have no column on the traces/sessions rollup tables — tools
```
geclos: more special case handling, persisting tools in traces/sessions would alleviate most of this?

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00023_tool_analytics.sql
csansoon: @andresgutgon 
Measured against prod: only 4.5% of rows have tools, there are just 1,956 distinct tool lists and 5,335 distinct names. As Array(LowCardinality) that’s ~73 MB raw (mostly empty-array offsets) which ZSTD will crush to an estimated ~2–8 MiB. This means ~0.01% of the table.

csansoon: @geclos 
Actually, looking at prod's data, GRANULARITY 2 makes more sense for this. There are ~5.3k distinct defined tools, on ~5% of the rows. Structurally, it's really similar to the `tags` array!

GRANULARITY 2 halves the smallest skippable unit (16k rows vs 32k), which is what pays off when filtering one tool inside a large project’s range; full GRANULARITY 1 wouldn’t add much since a tool a project actively uses saturates its own time range. Cost is a ~2× larger bloom index, which at idx_tags scale means tens of KiB (negligible).

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00023_tool_analytics.sql
csansoon: Not a typo. They're two different things.
`tool_name` is reserved for tool call spans, which contains the tool call.
`tool_names` is reserved for LLM spans, and contains the names of the defined tool calls the LLM has access to.

packages/platform/db-clickhouse/clickhouse/migrations/clustered/00024_materialize_tool_names.sql
csansoon: Yes, we can, but it is deliberate, as per our own ClickHouse skill:
> if statement B depends on metadata introduced by statement A, prefer splitting them into separate migration files

The first migration is instant, while the second one is a long-running (expected to be about 12s lol) background mutation, containing the backfilling process. If the backfilling were to fail somehow, the schema change is already commited on a cheaper migration and we would just need to re-run the backfilling process.

packages/platform/testkit/src/clickhouse/schema.sql
csansoon: Oh that was autogenerated by the `pnpm ch:schema:dump` script. Didn't move anything by hand

packages/platform/testkit/src/clickhouse/schema.sql
csansoon: ditto

packages/platform/db-clickhouse/src/repositories/session-repository.ts
csansoon: actually, it's easier than expected at first. Taking a look!

csansoon: you're absolutely right! — i have added a migration — moving `tools` as a new column on traces and sessions — and change the queries to use those — instead of using subqueries

geclos (approved):

PR #3514, opened 2026-06-11
feat(tools): error view breakdown and dashboard feedback fixes

csansoon:
Follow-up to the tools dashboard (#3508), addressing the first round of internal feedback.

https://github.com/user-attachments/assets/55016022-17d8-4f82-ad5a-4ed451f53453

## error view and common errors

The detail page's "Errors only" switch is renamed to "Error view", and it now answers the question it exists for — what is failing. In error view, the latency chart (whose failed-call latency already appears in the Usage row's Duration tile) is replaced by a **Common errors** panel: failed-call outputs clustered by a normalized form, each cluster showing a verbatim sample (click to copy), its `error_type` badge, count, and share of all failed calls. An "Other" remainder row keeps the bars accounting for every failure even past the server's top-10. Normal view is unchanged.

Clustering happens in ClickHouse in a new `getToolErrorBreakdown` repository method. The error text is `status_message`, falling back to `tool_output`; UUIDs, 16+-char hex runs, and numbers are collapsed to `<id>` / `<hex>` / `<n>` (in that order — they nest) before grouping, so "Timeout after 5023ms" and "Timeout after 4998ms" form one cluster instead of two. No migration — it reads existing columns. Covered by chdb tests: clustering, output fallback, the empty-output bucket, exclusion of ok calls and other tools, and the limit cap.

A known follow-up, deliberately out of scope: clicking a cluster to filter Recent calls down to it — that needs the normalized key to round-trip through the query layer.

## layout and visual fixes

- The detail page's two-column rows (charts, parameters/context) each hard-coded a different right-rail width (420px vs 340px). Both rows now share one grid template and the panels share a height cap, so the rail is straight and columns of a row end at the same height.
- The selected parameter in the Parameters explorer rendered with border + background + shadow — the raised-button recipe — so it read as clickable rather than selected. It now uses a selection treatment (fill + left accent bar), and a tool with a single parameter skips the master–detail split entirely.
- The tools list's error rate column shows `rate · count` (matching the "% of traces" idiom) with a "N of M calls failed" tooltip.
- Tool descriptions are prompt content with no length bound and could fill the screen. The header clamps them to two lines with a "Show more" modal; the trigger only appears when the text actually overflows, measured on the rendered element.
- The "Definition not found…" / "No definition or calls…" placeholders render in italics so they don't read as a tool whose description literally says that.

## navigation

- The Previous/Next tool buttons are real links (href, cmd/middle-click) and carry the current search params over, so error view and a custom time range survive paging between tools. The J/K hotkeys keep the same search-preserving behavior through `navigate`.
- "View traces" opens the traces page with the filter sidebar already open (`filtersOpen=true`), so the applied tool/time filters are visible on arrival instead of an unexplained scoped view.

## ui kit

`Button`'s `asChild` branch skipped the inner wrapper that gives normal buttons their icon–label gap, and landed the face's `w-full` directly on the child element — "View traces" rendered with its icon glued to the label, stretched across the header's free space. The Slot now carries `gap-x-1.5` itself (it can't add wrapper elements); the width fix stays at the call site (`w-auto`) since other `asChild` consumers may rely on the full-width base.

### DIFF
--- a/packages/ui/src/components/button/button.tsx
+++ b/packages/ui/src/components/button/button.tsx
@@ -186,9 +186,7 @@ function Button({
         className={cn(
           buttonContainerVariants({ variant }),
           buttonVariantsConfig({ variant, size }),
-          // The non-asChild branch spaces icon and label through an inner
-          // wrapper div; Slot can't add wrappers, so the gap goes on the
-          // child element itself.
+          // Slot can't add the inner wrapper that normally spaces icon and label.
           "gap-x-1.5",
           className,
           isLoading && "animate-pulse",
--- a/packages/ui/src/components/button/button.tsx
+++ b/packages/ui/src/components/button/button.tsx
@@ -180,13 +180,16 @@ function Button({
   const visibleChildren = isLoading ? stripLeadingIconChild(children) : children
 
   if (asChild) {
-    // Used in combobox.ts <ComboboxInput>
     return (
       <Slot
         ref={ref}
         className={cn(
           buttonContainerVariants({ variant }),
           buttonVariantsConfig({ variant, size }),
+          // The non-asChild branch spaces icon and label through an inner
+          // wrapper div; Slot can't add wrappers, so the gap goes on the
+          // child element itself.
+          "gap-x-1.5",
           className,
           isLoading && "animate-pulse",
         )}
--- a/apps/web/src/domains/tools/tools.collection.ts
+++ b/apps/web/src/domains/tools/tools.collection.ts
@@ -8,6 +8,7 @@ import {
   getToolCallHistogram,
   getToolContextBreakdown,
   getToolCoOccurrence,
+  getToolErrorBreakdown,
   getToolParameterStats,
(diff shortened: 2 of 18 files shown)


packages/ui/src/components/button/button.tsx
```diff
@@ -180,13 +180,16 @@ function Button({
...
     return (
       <Slot
         ref={ref}
         className={cn(
           buttonContainerVariants({ variant }),
           buttonVariantsConfig({ variant, size }),
+          // The non-asChild branch spaces icon and label through an inner
+          // wrapper div; Slot can't add wrappers, so the gap goes on the
+          // child element itself.
+          "gap-x-1.5",
```
andresgutgon, line 190: Por encima de mi cadáver. Remove this yapping ya

geclos:
looks lovely!

my only complain is that there's too many arrows 😂 
image

andresgutgon (approved):

PR #3515, opened 2026-06-11
docs(agents): add repo-wide code comment policy

csansoon:
## What

Adds a code comment policy to the repo-wide conventions in `AGENTS.md`, and aligns the existing one-liner in the code-style skill with it.

## Why

Coding agents over-comment: narrating edits, restating what the next line does, and writing multi-line rationale for self-explanatory code. The only existing guidance was a single line in `.agents/skills/code-style/SKILL.md`, which agents only load for style-specific tasks — so most code-writing sessions never saw it. Putting the rule in the repo-wide conventions section makes it part of every agent's base context.

## The policy

- Default to no comment; naming and structure carry the meaning. A comment explaining *what* code does is a signal to rewrite the code.
- A comment (one line) is earned only by something the code cannot show: a trap (code that looks wrong, removable, or simplifiable but breaks something), a non-obvious external constraint, or contract docs where the file's convention documents every sibling (e.g. port JSDoc).
- Never narrate the change just made, paraphrase a well-named identifier, add section headers, or put design rationale in code — rationale goes in the commit or PR description.
- When editing, don't comment the edit itself, and delete comments the change made stale.

`.agents/skills/code-style/SKILL.md` now points at the repo-wide rule instead of carrying its own divergent copy.


andresgutgon (approved):
<3

PR #3517, opened 2026-06-11
fix(taxonomy): reduce gardening sample memory pressure

geclos:
## What does this PR do?

Reduces taxonomy gardening memory pressure in the workflow worker by shrinking the clustering sample and avoiding full observation round-trips during the build step.

The gardening build now samples up to 1,500 observations per project instead of 10,000. The sample remains day-stratified across the 7-day lookback window, but the ClickHouse read used by clustering now returns only the observation id, start time, and embedding. Projection metadata and other full observation columns stay out of the workflow activity heap.

After clustering, leaf reassignments now use a ClickHouse-side `INSERT ... SELECT` keyed by observation id. The worker sends only the assignment map (observation id → cluster id/confidence), and ClickHouse copies unchanged columns from the latest observation rows while overwriting the assignment fields.

## Why

A production taxonomy gardening workflow was repeatedly timing out on the largest project. Its build activity scanned ~4k 2048-dimensional taxonomy observations, materialized ~130MB of ClickHouse JSON before JS object overhead, and blocked the workflows worker. This PR bounds the immediate sample size and removes the largest unnecessary data movement from the clustering path.

## Validation

- `pnpm --filter @domain/taxonomy typecheck`
- `pnpm --filter @domain/taxonomy test -- build-hierarchical-taxonomy`
- `pnpm --filter @domain/taxonomy check`
- `pnpm --filter @platform/db-clickhouse check`
- Verified the new ClickHouse assignment-array expression against production with a read-only SELECT.

`pnpm --filter @app/workflows typecheck` still fails on pre-existing workspace dependency/type errors in integrations/sandboxes/db-postgres imports, unrelated to this change.

### DIFF
--- a/dev-docs/taxonomy.md
+++ b/dev-docs/taxonomy.md
@@ -54,7 +54,7 @@ ClickHouse:
 Two distinct bounded reads sit over the observations table:
 
 - **`latestProjectWindow`** — the newest `TAXONOMY_GARDENING_OBSERVATION_WINDOW_MAX` (10k) project rows, newest-first. Used by the read/aggregation paths (`listByCluster`, `listAllByCluster`, `listNoise`).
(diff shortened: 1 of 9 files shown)


geclos:
@cursoragent review

PR #3518, opened 2026-06-11
Spec: data destinations (only spec)

andresgutgon:
[LAT-665](https://linear.app/latitude/issue/LAT-665/posthog-integration)

## what this is

A spec (`specs/data-destinations.md`) for letting customers continuously sync their Latitude telemetry into systems they own. The triggering ask was a PostHog integration, but the issue explicitly requires extensibility — to other destinations and to other data structures than spans/traces/sessions — so this spec designs the general capability and treats PostHog as the first destination. Implementation will follow on this same branch, phase by phase.

## research summary

Surveyed Langfuse, LangSmith, Braintrust, Helicone, Arize Phoenix, Datadog LLM Obs, and PostHog's own CDP. The market ships exactly two destination families: **event-push** (scheduled batch worker maps records to the destination's event schema and POSTs with the customer's write key) and **file-drop** (Parquet/JSONL windows into a customer-owned bucket — the actual "data lake" answer). Nobody forwards stored traces over OTLP (GenAI semconv still pre-stable) and nobody does zero-copy warehouse shares. The universal delivery contract is at-least-once plus idempotent dedup at the destination.

## design highlights

- **Scheduled micro-batch pull from ClickHouse** with an `ingested_at` watermark cursor and a safety lag, instead of tapping the ingest path. The hot path stays untouched, ReplacingMergeTree settling is respected, and re-running a window doubles as the backfill primitive.
- **Spans are the only source read.** PostHog's `$ai_generation` / `$ai_span` / `$ai_trace` events carry trace and session structure as properties, so traces and sessions are never read from the aggregating tables. This resolves the issue's open question about syncing three entities.
- **Deterministic event UUIDs** (UUIDv5 of destination + span + event name) make retries and window re-runs duplicate-free — PostHog dedups on event uuid.
- **Project-scoped `destinations` table** with kind-discriminated config and AES-256-GCM-encrypted credentials. Deliberately not reusing the `integrations` parent table: its invariants (one active per org+kind, vendor-account exclusivity, OAuth token lifecycle) don't apply to write-key destinations.
- **Port/adapter split** per the established pattern: `@domain/destinations` (entity, ports, pure mappers, sync use case) + `@platform/posthog-destination` (chunked `/batch/` delivery, retryable vs non-retryable error mapping). Kept separate from `@platform/analytics-posthog`, which remains internal product analytics.
- **Failure policy copied from PostHog CDP**: retryable errors retry the window with the cursor untouched; auth errors quarantine the destination after N consecutive failures. Per-run throughput caps with graceful catch-up (Langfuse once flooded customer PostHog instances at ~18k events/sec; see langfuse#12786).
- **Payload redaction toggle from day one** — prompts/completions in `$ai_input` are a compliance blocker for some customers, and PostHog bills the customer per event.

## phasing

1. Sync engine + PostHog adapter, feature-flag gated, no UI.
2. Settings UI, test-connection, quarantine surfacing.
3. Backfill (`historical_migration: true`) + hardening.
4. (Future, unscheduled) webhook destination, S3/GCS Parquet file-drop, scores as a second source, OTLP forwarding once GenAI semconv stabilizes.

### DIFF
--- a/specs/data-destinations.md
+++ b/specs/data-destinations.md
@@ -1,0 +1,247 @@
+# Data destinations
+
+> **Documentation (future)**: `dev-docs/data-destinations.md` once Phase 2 stabilizes.
+> **Linear**: [LAT-665](https://linear.app/latitude/issue/LAT-665/posthog-integration)
+
+## Context
+
+A customer asked for a PostHog integration: they want their Latitude traces inside their own PostHog project. The real capability behind that ask is bigger than PostHog — it is **outbound synchronization of customer telemetry into customer-owned systems** (analytics tools, object storage, data lakes). PostHog is the first destination; the system must be extensible to others, and to data structures beyond spans/traces/sessions (scores, issues, …) later.
+
+Today nothing pushes telemetry out of the platform. Exports are user-initiated CSV emails (`exports:generate`), the Slack integration delivers notifications (not data), and `@platform/analytics-posthog` is our own internal product analytics (write-only, env-configured, single-tenant — explicitly **not** the pattern to follow for customer-facing integrations).
+
+### Industry survey (how the market ships this)
+
+Researched Langfuse, LangSmith, Braintrust, Helicone, Arize Phoenix, Datadog LLM Obs, and PostHog's own CDP. The market splits cleanly into **two destination families**:
+
+| Family | Who ships it | Mechanism |
+| --- | --- | --- |
+| **Event-push** | Langfuse→PostHog (hourly batch), Helicone→PostHog (per-request), PostHog CDP realtime destinations | Scheduled or streaming worker maps records to the destination's event schema and POSTs batches with the customer's write key. Retry + quarantine-on-chronic-failure. |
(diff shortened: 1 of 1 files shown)


specs/data-destinations.md
```diff
@@ -0,0 +1,241 @@
...
+- **Config vs credentials split.** Non-secret config (`host`, redaction flags, interval, event filters) is plain jsonb validated by a Zod discriminated union on `kind` in `@domain/destinations`. Numeric knobs are bounds-validated — `intervalMs` 1–60 min, `maxSpansPerRun` 1k–50k — they are user input, not trusted constants. Secrets (`apiKey` for PostHog) live in a single `credentials` text column: AES-256-GCM-encrypted JSON, decrypted at the repository boundary (same pattern as `slack_integration_details.bot_access_token`).
```
andresgutgon: Is this credentials a text column? What if other destination requires 2 fields as secrets or a json file?

PR #3519, opened 2026-06-11
feat(taxonomy): split gardening build writes

geclos:
## What does this PR do?

Follow-up to #3517. This implements the remaining highest-ROI taxonomy gardening reliability work:

- Moves synchronous hierarchical clustering into a Node `worker_threads` worker so k-means CPU does not block the Temporal worker event loop.
- Splits the gardening build path into separate activities:
  - plan tree/read sample/run worker-thread clustering
  - save clusters
  - reassign observations in ClickHouse
  - deprecate old clusters
- Adds a Temporal `patched()` marker so executions that already scheduled the legacy single build activity remain replay-compatible.
- Adds adaptive sample sizing: small corpora are read whole up to 1.5k; larger corpora target 40% with a 500-row floor and a 1.5k system cap.
- Adds run-level sample metrics to `taxonomy_runs`:
  - `observations_available`
  - `observations_sampled`
  - `sample_strategy`
  - `sample_cap`

## Why

The previous PR reduced payload size and moved observation reassignment into ClickHouse. This PR further reduces operational risk: worker-thread clustering protects health checks/polling from CPU stalls, while the activity split prevents a crash after one write operation from retrying the entire build over partially persisted state.

## Validation

Passed:

- `pnpm --filter @domain/taxonomy typecheck`
- `pnpm --filter @domain/taxonomy test -- build-hierarchical-taxonomy`
- `pnpm --filter @domain/taxonomy check`
- `pnpm --filter @app/workflows exec vitest run src/workflows/taxonomy-gardening-workflow.test.ts`
- `pnpm --filter @platform/db-postgres check`
- `pnpm --filter @platform/db-postgres exec vitest run src/repositories/taxonomy-repositories.test.ts`
- `pnpm --filter @app/workers check`
- `pnpm --filter @app/workflows build` (passes; emits existing unresolved-import warnings for unrelated workspace deps)

Known unrelated failures still present:

- `@app/workflows typecheck` and `@platform/db-postgres typecheck` fail on pre-existing missing workspace deps/type errors (`@domain/users`, `@domain/api-keys`, `@better-auth/sso`, etc.).
- `@app/workflows test -- ...` through the package script over-selects unrelated activity tests and fails on the same missing `@domain/api-keys`; direct vitest for the workflow test passes.


apps/workflows/src/activities/taxonomy-gardening-activities.ts
```diff
@@ -248,6 +322,73 @@ export const buildHierarchicalGardenTaxonomyActivity = (input: GardenTaxonomySte
...
+    input,
+    planHierarchicalTaxonomyUseCase({
+      organizationId: OrganizationId(input.organizationId),
+      projectId: ProjectId(input.projectId),
+      runId: TaxonomyRunId(input.runId),
+      dimension: input.dimension,
+      now: new Date(input.now),
+      clusterBuilder: (builderInput) =>
+        Effect.tryPromise({
+          try: () => buildHierarchicalClustersInWorker(builderInput),
```
andresgutgon, line 362: So is this idempotent? If this runs, it knows if the worker has finished or is still running? I guess re-running does not happen so fast, and this is not a problem, but curious if this would be an issue

andresgutgon (dismissed):

geclos:
@cursoragent review

PR #3531, opened 2026-06-12
Add user profiles: end-user telemetry attributes, users dashboard, and user detail page

geclos:
## what this is

End-users of customers' agents become a first-class observability surface. The telemetry SDKs can now attach a user identity to traces, and a new **Users** project section shows who is using the product, what they do, and what breaks for them.

## telemetry + storage

The TypeScript SDK context accepts `userId` and `userEmail`, and the Python SDK accepts `user_id` and `user_email`, emitted as `user.id` / `user.email` span attributes with the same last-write-wins merge semantics as `sessionId`. Ingestion resolves `user.id` and `user.email` from a fallback chain covering our SDKs plus OpenInference, OpenLIT, Traceloop, LangSmith, Langfuse, and OTEL enduser semconv. Migration 00028 adds `user_email` to `spans` (bloom-filter indexed) and as an `argMaxIf` aggregate state on the `traces` and `sessions` MVs, mirroring how `user_id` already worked.

Both telemetry SDKs are bumped to **3.1.0** with changelog entries, so merging this PR publishes the new `userEmail` / `user_email` option to npm and PyPI.

## users dashboard (`/projects/:slug/users`)

Defaults to the last 30 days, like the tools section. A stats panel shows unique users, new users (first trace ever inside the window), identified-traces coverage (tooltip explains it), and traces per user, plus an active-users histogram whose drag-selection narrows the time filter. The table lists each user with id + email, last/first seen, an activity sparkline bucketed to the selected range, session count, session-based error rate (`rate · count`), and cost. The cost column's subheader pill cycles SUM/AVG/MED — it changes how each user's cost is computed, shows the column-wide aggregate for that mode, and re-targets an active sort. Search matches id or email case-insensitively; sorting, filters, and search all run server-side.

## user detail page (`/projects/:slug/users/:userId`)

Profile header (identity, copyable id, first/last seen), a stat strip (traces, sessions, errored sessions with rate, cost, tokens, avg duration, active days), a 30-day activity chart, issues affecting the user, observed behaviour clusters linking into the behaviours tree, model/provider/tool usage breakdowns, and the user's sessions with an error-view switch, opening the session panel in the side drawer.

## how the reads work

- A new `UserAnalyticsRepository` port in `@domain/spans` with a ClickHouse adapter that finalizes the `traces` MV per trace (`argMaxIfMerge`) and groups by user. Error rates count sessions with at least one errored trace, using the same coalesced session key as the sessions MV.
- Scores carry no user column, so `ScoreAnalyticsRepository.listIssuesByUser` resolves the user's trace set inside ClickHouse and rolls up issue occurrences over it; `listUserIssuesUseCase` joins issue rows + lifecycle states from Postgres.
- Behaviour observations are session-scoped, so `TaxonomyObservationRepository.getClusterCountsByUser` joins through the sessions MV's finalized `user_id`; `listUserBehavioursUseCase` maps cluster ids to the active taxonomy.

## seeds

Seed pools become weighted `SeedUser` identities (two without email, simulating id-only integrations). tau2 trajectories and the lifecycle traces get deterministic users via a prime-stride weighted assignment, so local data shows a power-law usage curve and issue occurrences land on stable profiles across re-seeds. `ch:seed:reset` to pick it up.

## testing

chdb tests cover the new repository (aggregation, sorting, pagination, search, time-range scoping, activity buckets, overview, profile, usage breakdowns, cost rollup). The chdb test schema was hand-synced with migration 00028 — worth re-running `ch:schema:dump` next time migrations run to confirm parity. Python SDK tests cover `user_email` merging and context propagation. QA'd manually against reseeded local data.

## known follow-ups

- Behaviours section is empty until the taxonomy pipeline has run (seeds don't populate `taxonomy_observations`).
- Web-only surface for now; no public API/MCP endpoints yet.


geclos:
@cursoragent review this pr

geclos:
## Code review

Reviewed the full diff and cross-checked the load-bearing details against the codebase (histogram bucket alignment, sessions `userId` filter support, migration precedent, docs link targets, existing collection patterns).

### What holds up

- **Bucket alignment**: CH `intDiv(toUnixTimestamp(...), bucketSeconds) * bucketSeconds` and `buildHistogramBucketScaffold` are both epoch-aligned with matching `.000Z` ISO keys, so `fillBuckets` joins correctly.
- **MV drop/recreate** is the established pattern (00002, 00004, 00005, 00008… did the same), the Down migration drops views before columns and restores the old definitions, and the migration lock was regenerated.
- **Tenancy**: every adapter query couples `organization_id` + `project_id`; `userIds`/`searchQuery` inputs are parameterized, no injection paths.
- **Tests**: the chdb suite covers aggregation, sorting, pagination, search, range scoping, overview, profile, usage, and the cost rollup.

### Findings

#### 1. Search by email skews the matched user's metrics (correctness — fix before merge)

In `user-analytics-repository.ts`, `buildUserScopeClauses` applies `user_id ILIKE ... OR user_email ILIKE ...` **inside** the per-trace rollup, before the `GROUP BY user_id`. A user's displayed email is "latest non-empty across traces", so their older traces typically carry an empty `user_email`. When searching by email, only the traces that themselves carry a matching email survive the filter — so the matched user's trace count, sessions, cost, and first-seen are computed over a subset of their traces. The test fixture demonstrates the setup (USER_A has the email on 1 of 3 traces) but only asserts membership, not metrics, so it doesn't catch it.

**Fix**: apply the search predicate as `HAVING` after `GROUP BY user_id` (the `argMaxIf(user_email, ...)` alias is legal there, which sidesteps the WHERE-alias problem the code comment mentions), and add a metrics assertion under search. Related nit: `%`/`_` in the search input aren't escaped, so searching `100%` behaves as a wildcard.

#### 2. Stat-strip session counts vs. the sessions table can disagree (minor)

`errorSessionCount`/`sessionCount` in the profile count sessions where *this user has ≥1 trace* (grouped from the traces rollup), while the sessions table below filters on the sessions MV's finalized `user_id` — the *latest non-empty* user of the session. A session with mixed/multiple users (or identified + anonymous traces) is counted in the strip but may not appear in the table. Probably acceptable; worth a one-line note in the port JSDoc so the next person doesn't chase it as a bug.

#### 3. Query cost per page fetch (performance)

- `listByProjectId` always runs the count + cost-rollup aggregation (a full per-user pass over the project's finalized traces) alongside the page query, and the collection eagerly prefetches the next page — so each scroll step triggers the heavy rollup twice. Consider computing total/rollup only for `offset === 0`, or dropping the prefetch.
- The user detail page fires ~7 server functions on mount (profile, activity, 3 usage dimensions, issues, behaviours), each finalizing every trace row in the project with no time bound (`findByUserId` is documented lifetime, but the usage/issue queries are too). Fine at current scale; flag for follow-up if a large tenant lands on it.
- `useProjectUsers` diverges from `sessions.collection.ts`/`traces.collection.ts`: those use plain `useInfiniteQuery`, whereas this adds per-offset `fetchQuery` entries + prefetch, double-caching every page in the query cache. Unless the prefetch is deliberate, the simpler established pattern would do.

#### 4. `isReloading` blanks the table (UX nit)

`users = isReloading ? EMPTY_USERS : usersData` in `users/index.tsx` throws away `keepPreviousData` — every sort/filter/search change flashes skeletons instead of keeping rows under the reload. If intentional, fine; otherwise just pass `usersData` through.

#### 5. Docs: Python example is unverified

`docs/observability/users.mdx` shows a Python tab passing `user_email`, but this PR only changes the TypeScript SDK. The existing `user-tracking.mdx` shows the same shape for `user_id`, so the convention is fine — but confirm the Python telemetry package actually accepts `user_email` before this page ships, or trim the tab. (The `./features/user-tracking` link target exists; no broken links.)

#### 6. Smaller notes

- **PII**: `user_email` now lives in `spans` rows and as `argMaxIf` aggregate states on two MVs. Retention via `retention_days` applies, but targeted erasure (GDPR delete-this-user) against aggregate states is effectively impossible without rewrites — worth capturing as a known constraint in dev-docs, not a blocker.
- `userEmailCandidates` covers Latitude/OpenInference, OTEL enduser, Langfuse — narrower than the `userId` chain; the PR body slightly oversells the email coverage (OpenLIT/Traceloop/LangSmith are id-only).
- The PR body already flags the hand-synced chdb `schema.sql` — agreed that re-running `ch:schema:dump` after the migration runs is worth doing before merge or as immediate follow-up.
- `seedUserAt`'s stride-37 vs. pool weight 100 are coprime as required; the degradation comment covers the trap. 👍

### Verdict

Solid, well-layered feature with good test coverage; the MV migration follows precedent and tenancy is handled correctly throughout. **One correctness fix requested before merge: the search-by-email metrics skew (finding 1).** Everything else is minor or follow-up material.

geclos:
Addressed the review feedback — triage below.

**Fixed**
- **Search-by-email metrics skew** (91de7d264): the search predicate now runs as `HAVING` against the finalized per-user row instead of filtering traces pre-`GROUP BY`, so a user matched by email keeps metrics from all their traces. `%`/`_` in the input are escaped as literals. Added metrics-under-search and wildcard-literal test assertions. Note for posterity: the HAVING must reference the `user_email` alias — spelling out `argMaxIf(...)` trips `ILLEGAL_AGGREGATION` via alias substitution.
- **`enduser.id` OTLP candidate** (08d5eba55), mirroring `enduser.email`.
- **Collection divergence + double-fetch** (bc859bddb): replaced the per-offset `fetchQuery` + eager prefetch with the plain `useInfiniteQuery` shape sessions/traces use — halves the per-scroll ClickHouse load and removes the double-caching. Same commit stops blanking the table to skeletons on sort/filter changes (`keepPreviousData` now actually shows previous rows).
- **First-seen semantics**: relabelled the seen-at tooltips to "First/Last seen in selected period" (bc859bddb). Deliberately did *not* compute lifetime first-seen in the list — it would add another full per-user pass to an already heavy query, against the cost concerns raised in both reviews. The detail page already shows lifetime values.
- **Session-count vs sessions-table mismatch**: documented on the port's `sessionCount` JSDoc (91de7d264).
- **chdb `schema.sql`**: regenerated via `ch:schema:dump` against the migrated dev DB (082ddc42a); only drift was index ordering.
- **Python SDK docs gap / unstored `user.plan`+name fields**: Python SDK now accepts `user_email` (39d6cc892) and the unstored fields were dropped from the TS SDK context (2aa165a73).

**Declined / follow-up**
- Lifetime first-seen subquery — declined for query cost (relabel instead, see above).
- Deeper query-cost work (count/rollup only at offset 0, pre-aggregated per-user rollup) — follow-up; watch-and-optimize as both reviews suggested.
- Shared histogram-scaffold helper — premature with two consumers.
- GDPR erasure of `user_email` in MV aggregate states — acknowledged as a known constraint; not blocking v1.
- Historical `user_email` backfill — accepted as documented v1 limitation.

PR #3536, opened 2026-06-12
feat(oss): Tier 3 cluster deployment via cloud-agnostic Helm chart

neoxelox:
Phase 5 of the OSS self-hosting spec (`specs/oss.md`, LAT-655): the Cluster tier. A third party can now deploy a scalable Latitude on any conformant Kubernetes cluster with `helm install`, using the published `latitudedata/<service>` images — no Latitude-account-specific assumptions, `infra/` untouched (except one cleanup, below).

## the chart (`charts/latitude/`)

Dependency-free and cloud-agnostic — no external subcharts. It deploys:

- **App services**: Deployments for `web`/`api`/`ingest`/`workers`/`workflows` rendered through one shared helper, with HPA-ready resource requests and probes on each service's health route. `web`/`api`/`ingest` get ClusterIP Services plus a three-host Ingress whose hosts derive from the required `config.{web,api,ingest}Url`; `workers`/`workflows` are internal-only.
- **Migrations** as a `post-install,pre-upgrade` hook Job: on install it retries until the bundled databases accept connections; on upgrade the schema rolls before new pods (migrations are append-only/backward-compatible).
- **Bundled infrastructure**, each an independent toggle (`<dep>.enabled` + `external:` block for bring-your-own): Postgres (pgvector, Tier-2 `init-db.sh` runtime-user bootstrap), ClickHouse (Tier-2 tiered `storage.xml`), Redis cache (ephemeral) + BullMQ Redis (appendonly/noeviction, PVC), SeaweedFS as the same single-container `server -s3` recipe as `docker-stack.yml` (bucket-bootstrap wrapper included), and Temporal.
- **Temporal** uses the plain `temporalio/server` image (not `auto-setup`), Postgres-backed default + visibility stores, no Elasticsearch/Cassandra. Schema setup and namespace registration run as one-shot `admin-tools` hook jobs; the schema script is idempotent so the same job serves install and upgrades. The CLI surface (`temporal-sql-tool` flags, schema dirs, `temporal operator namespace`) was verified against the real image — note `admin-tools` publishes no patch tags, so it pins `:1.27`.

### env contract and secrets

All services and jobs share one contract mirroring Tier 2's `.env.production`: a ConfigMap for non-secrets and a Secret for connection URLs, passwords, and app secrets, consumed via `envFrom` everywhere (the bundled Postgres/ClickHouse read their image-bootstrap vars from the same sources). Optional settings go in `config.extraEnv`; optional **secret** values (AI provider keys, email passwords, OAuth client secrets) go in `secrets.extra`, merged into the chart-managed Secret. Operators with their own secret management replace the whole thing via `secrets.existingSecret` (key list in the chart README). There are no default passwords — required values fail fast with actionable messages.

### versioning

The chart's `version`/`appVersion` track the Latitude release, and `image.tag` defaults to `appVersion`, so the chart always deploys the release it shipped with. The release flow owns the bump: the production-release skill sets both fields in the `release: vX.Y.Z` commit, and `scripts/release.sh` now refuses to tag a release whose chart version at `origin/development` doesn't match.

## also in this PR

- **Bull board is now opt-in at the code level** (`apps/workers`): the dashboard mounts only when both `LAT_BULL_BOARD_*` vars are set — previously they were required env and the dashboard was always on. The self-host tiers no longer ship it (removed from the chart entirely; `.env.example` marks the vars "Remove for production", dev keeps them). Latitude's own ECS deploy still injects them, so the cloud dashboard is unaffected.
- **`infra/lib/ecs.ts` cleanup**: the legacy `CLICKHOUSE_*` env names kept during the `LAT_CLICKHOUSE_*` rename transition are removed — the renamed release is live in production, closing P6-7.
- **Docs**: new `deployment/cluster` page (quick start, BYO, scaling, backups, upgrades, health) wired into the nav, overview card, single-host note, and README; the "coming soon" placeholders are gone.

## validation

`helm lint` clean; the default render (29 resources) is kubeconform-valid; the BYO-everything variant renders only the app workloads with every `external.*` value landing in the right `LAT_*` var; missing-required variants fail with the intended messages. Workers typecheck + 89 tests pass; infra typechecks. The remaining P5-6 item — a real `helm install` smoke test against a live cluster — is tracked in the spec as in progress.


charts/latitude/templates/seaweedfs.yaml
andresgutgon, line 1: All the helm charts is part of infra. Put inside `infra` folder

scripts/release.sh
```diff
@@ -119,6 +119,18 @@ if git rev-parse --verify "refs/tags/${version}" >/dev/null 2>&1; then
   exit 1
 fi
 
+# The Helm chart version tracks the Latitude release; the bump lands in the
+# release commit (with CHANGELOG.md) so the tagged commit ships a matching chart.
+release_semver="${version#v}"
```
andresgutgon, line 124: Do we have to use this release script for helm charts? Can't we have another and call this one cloud-release.sh or something like this?

charts/latitude/templates/seaweedfs.yaml
neoxelox: no

scripts/release.sh
neoxelox: its not releasing the helm chart, its just updating the version of the helm chart with the release version, so we maintain the same versions, we do the same for docker images

PR #3542, opened 2026-06-12
feat(oss): wait-for-deps init containers and install feedback for the Helm chart

neoxelox:
Follow-up to #3536, driven by the first real `helm install` smoke test (k3d, v0.3.1). The install worked end to end, but it *looked* broken for minutes: app pods started before their dependencies, crashed, and Kubernetes backed their restarts off to ~5 minutes — so `api`/`workers`/`workflows` sat in `CrashLoopBackOff` long after Temporal and the databases were actually ready (the Temporal server itself restarted 7 times waiting for its schema). This PR makes the rollout ordered and self-explanatory.

## wait-for-deps init containers

Every app Deployment gets a `wait-for-deps` init container that TCP-polls the **full infrastructure set — identical for all five services**: Postgres, both Redis, ClickHouse, Temporal, and the bundled SeaweedFS (skipped when using external S3, which has no derivable host:port). No per-service special-casing. Pods wait in `Init:0/1` — a state that says what it means — and start exactly once, with no restart-backoff penalty.

The init container reuses the service's own image (a bash `/dev/tcp` loop), so it adds no extra pull. The endpoints are toggle-aware: bundled deps resolve to in-cluster Service DNS; `external:` endpoints are derived from the same values the env contract uses (Postgres host:port parsed from the database URL, Temporal Cloud address as-is, ClickHouse URL with scheme-based port defaulting).

The migrations job waits for the two stores it migrates (Postgres + ClickHouse native port). The Temporal server's blocker is the schema rather than a port, so it gets a `wait-for-schema` init container instead: a `psql` poll (postgres image) of the `schema_version` table in both stores, releasing the server only after the schema job has done its work. A fresh install now brings every pod up with zero restarts.

## install feedback

- `NOTES.txt` now explains the rollout model (pods wait in Init until dependencies accept connections), and gives the watch and job-inspection commands.
- The cluster guide's Install step sets the same expectation, and a new **Troubleshooting** section covers the states an operator actually meets: `Init:0/1` (and `kubectl logs <pod> -c wait-for-deps` to see what it's polling), failed hook jobs (logs + idempotent re-run), `Pending` (StorageClass/PVC), and `ImagePullBackOff`.

## spec

P5-6 (real-cluster smoke test) flips to done: full install of the v0.3.1 chart on k3d — bundled infra healthy, all three hook jobs complete, sign-in via in-cluster Mailpit, OpenAI generation + Voyage embeddings wired through `config.extraEnv`/`secrets.extra` — with the backoff finding recorded and resolved here.

## validation

`helm lint` clean; default render shows the identical dependency list on all five services and kubeconform-validates 29/29; the BYO-everything render derives every external endpoint correctly. Renders verified for default, BYO, and missing-required variants.


neoxelox:
@claude review

PR #3544, opened 2026-06-12
feat(destinations): @platform/data-destinations PostHog deliverer (LAT-670)

andresgutgon:
P1-5 of the data-destinations spec ([LAT-670](https://linear.app/latitude/issue/LAT-670)) — the first vendor adapter for the destinations sync engine. New package `@platform/data-destinations` implementing the `DestinationDeliverer` port from #3538 (email-transport pattern: port in domain, adapter in platform). Deliberately separate from `@platform/analytics-posthog`, which is our own single-tenant product analytics.

## what the deliverer does

`createPosthogDeliverer` POSTs mapped `$ai_*` events to `{host}/batch/` with the customer's `phc_` key, via raw `fetch` (no SDK — multi-tenant, one key per call).

- **Span-atomic chunking**: events are serialized once, grouped by `spanId`, and groups are packed whole into chunks under the 500-event and 20 MB limits — a root span's `$ai_trace` can never land in a different chunk than its sibling event. Events over 1 MiB are dropped pre-chunking and reported in the result's `dropped` count.
- **`historical_migration`** is set when the delivery-context window ends more than 48 h ago — this adapter's rule, derived from the window the engine passes; the engine never knows the word "backfill".
- **SSRF guard**: custom hosts must be https and are re-resolved before every chunk POST with all addresses required to be public unicast (IPv4 special ranges incl. CGNAT/link-local; IPv6 global unicast only); redirects are refused (`redirect: "manual"`, any 3xx fails non-retryably). US/EU presets pin the official ingestion hosts and skip resolution. Known residual: our lookup and fetch's lookup are separate resolutions, so a sub-TTL DNS-rebind window remains — closing it fully needs a custom dispatcher or egress proxy (the spec's stated alternative).
- **Error taxonomy** (sanitized, response bodies never read): transport/5xx → retryable, 429 → retryable `rate_limited`, DNS failure → retryable; 401/403 → non-retryable `invalid_api_key`, other 4xx → non-retryable `request_rejected`, private-IP host → non-retryable.

## findings that affect the spec (need confirmation on a real PostHog project)

- **The spec's 401 assumption is partially wrong.** Per [posthog#54670](https://github.com/PostHog/posthog/issues/54670), `/batch/` returns `200 OK` for a well-formed but non-existent `phc_` key and silently drops the events — only malformed tokens get 401. Quarantine therefore won't fire for a wrong-project or revoked key, and P2-1's test-connection can't rely on `/batch/` status alone (`/flags/` does 401 on unknown keys and is the likely probe).
- **The per-event size cap is undocumented**: effectively Kafka's 1 MB message limit on PostHog Cloud (HTTP 413 on oversize, [posthog#23703](https://github.com/PostHog/posthog/issues/23703)). Encoded as `POSTHOG_EVENT_MAX_BYTES = 1 MiB`; self-hosted brokers may differ.

## testing

Fetch-mocked: happy path, 401 non-retryable, 429 retryable, oversized chunk splitting, span-atomicity across chunk boundaries, >48 h window setting `historical_migration`, private-IP and http hosts rejected, redirect refusal; plus a 37-case table for the public-unicast IP check. 55 tests, typecheck clean.


packages/platform/data-destinations/vitest.config.ts
```diff
@@ -0,0 +1,4 @@
+import base from "@repo/vitest-config"
+import { defineConfig, mergeConfig } from "vitest/config"
+
+export default mergeConfig(base, defineConfig({}))
```
andresgutgon: This file is not needed. Delete

packages/platform/data-destinations/src/posthog/deliverer.ts
```diff
@@ -0,0 +1,196 @@
...
+          redirect: "manual",
+        }),
+      catch: () => new RetryableDeliveryError({ kind: KIND, reason: "transport_error" }),
+    })
+    const error = responseError(response.status)
+    if (error) return yield* error
+  })
+
+export const createPosthogDeliverer = (options: PosthogDelivererOptions = {}): DestinationDeliverer => {
+  const fetchFn = options.fetchFn ?? fetch
```
andresgutgon, line 167: Why not use Posthog nodejs client?

andresgutgon: By design, not for lack of considering it. The deliverer is multi-tenant (a customer write key per call) and the engine already owns batching, retries, and the cursor — so the SDK's headline feature (async queue + auto-flush) is exactly what we *don't* want here. We need a synchronous deliver-and-report: a failed POST must leave the cursor unmoved and feed the quarantine/backoff state machine. On top of that the SDK can't express `historical_migration: true` (needed for backfill past spike detection), gives us no hook to validate the customer-supplied host against SSRF, and hides the HTTP status we map to retryable/non-retryable.

Worth noting: Langfuse *does* use `posthog-node` for their multi-tenant sync — and that is exactly how they hit langfuse#12786 (unbounded ~18k events/sec burst: concurrent clients, `flushAt: 1000`, no backpressure). Our raw-`fetch` + per-run cap + cursor-paced, one-run-at-a-time delivery is the deliberate countermeasure. The single-tenant `@platform/analytics-posthog` (our own key) is where the SDK is the right fit.

PR #3545, opened 2026-06-12
feat(destinations): PostHog mapper — spans → $ai_* events (LAT-668)

andresgutgon:
Second slice of the data-destinations effort (`specs/data-destinations.md`, P1-2): the pure mapping layer from Latitude spans to PostHog LLM-analytics events. No I/O — the sync engine (P1-6) and the PostHog deliverer adapter (P1-5) plug in around it.

## what it does

`mapSpansToPosthogEvents` in `@domain/destinations/mappers/posthog.ts` turns a window of `SpanDetail`s into deliverable `$ai_*` events:

- **Event kind per span**: `$ai_embedding` for embedding operations, `$ai_generation` when the span has a model, `$ai_span` otherwise. Root spans (`parentSpanId` empty) additionally emit an `$ai_trace`. Traces whose root never arrives emit nothing extra — PostHog's pseudo-trace grouping covers them.
- **Deterministic identity**: `uuid = UUIDv5(namespace, destinationId:spanId:eventName)`, with `timestamp` = span `end_time` and stable `distinct_id`/`event` across retries. PostHog dedups on `(date, event, distinct_id, uuid)`, so at-least-once delivery (retries, window re-runs, compound-cursor resumes) becomes effectively-once.
- **Field mapping**: microcents → USD costs, `$ai_latency` in seconds, `$ai_session_id = coalesce(session_id, trace_id)` (parity with `sessions_mv`), `distinct_id` = span user id with fallback to `trace_id` + `$process_person_profile: false` for anonymous traffic. `latitude_project_id` and `latitude_span_url` (via an injected pure `buildSpanUrl`) ride along for cross-linking.
- **Redaction**: `excludePayloads` derives a redaction set over the content-bearing properties (`$ai_input`, `$ai_output_choices`, `$ai_input_state`, `$ai_output_state`, `$ai_tools`, `$ai_error` → `error_type` only), nulled in one pass. Per-field granularity later is a config-only change.
- **Oversized events**: over the per-event limit, content properties are truncated with `latitude_truncated: true`; if still oversized the event is dropped and counted (`dropped` → `events_dropped` on the sync run). A giant span can never wedge the cursor. The mapper takes `maxEventBytes` as input — the deliverer adapter owns the live vendor cap; a 1 MiB domain default covers standalone use.

## the UUID namespace

`DESTINATION_EVENT_UUID_NAMESPACE` is not arbitrary: it's `UUIDv5(RFC 4122 DNS namespace, "destinations.latitude.so")`, so its provenance is recomputable, and a test pins the derivation. It's a compile-time constant by design — it's part of the wire contract, the same for every deployment (incl. self-hosted). Isolation across orgs/projects/deployments comes from `destinationId` inside the hashed name, not from the namespace; making it configurable would only create a way to break dedup. The constant carries a NEVER CHANGE warning and the pinning test doubles as a tripwire.

`uuidV5` itself is a small helper over web-standard `crypto.subtle` (no new dependency), verified against the RFC 9562 reference vector.

## tests

29 unit tests: helper reference vectors + fixture spans covering LLM call, embedding, tool span, root (`$ai_trace`), error span, session-less trace, anonymous user, redacted config, and oversized events (truncated and dropped).

Closes [LAT-668](https://linear.app/latitude/issue/LAT-668/p1-2-posthog-mapper-spans-dollarai-events-pure)


packages/domain/destinations/src/mappers/posthog.ts
```diff
@@ -0,0 +1,206 @@
...
+        errorType: params.errorType,
+        names: CONTENT_PROPERTY_SET,
+      }),
+      latitude_truncated: true,
+    },
+  }
+  return eventByteSize(truncated) <= params.maxEventBytes ? truncated : null
+}
+
+export const mapSpansToPosthogEvents = async (
```
andresgutgon, line 171: Is the mapper accept configurable data fields redacted option?

PR #3550, opened 2026-06-13
fix(ai-voyage): declare voyageai in api/workers so the bundled image can load it

geclos:
## Problem

Staging trace search silently degraded to **lexical-only** and the backfill wrote **zero embeddings**. Root cause in the deployed workers service:

```
AIError: Voyage client creation failed:
  Cannot find module '@platform/ai-voyage/package.json'  (MODULE_NOT_FOUND)
```

`@platform/ai-voyage`'s client loads the `voyageai` SDK at runtime (it's CJS-only with an ESM bug) via `createRequire(require.resolve("@platform/ai-voyage/package.json"))`. That fallback exists because, under pnpm's strict layout, `voyageai` is only resolvable from *within* the `@platform/ai-voyage` package.

But tsdown inlines `@platform/ai-voyage` **into each app bundle** (`alwaysBundle: /@platform/.*/`). In the built image that package is no longer a resolvable module, so:
- primary `require("voyageai")` fails (not a direct dep of the app), and
- the fallback `require.resolve("@platform/ai-voyage/package.json")` throws `MODULE_NOT_FOUND`.

Result: no Voyage client → no embeddings, on every app that bundles ai-voyage and reaches the Voyage path.

## Fix

Declare `voyageai` as a **direct dependency** of the apps that bundle ai-voyage and embed/rerank — `apps/api` and `apps/workers` — mirroring `apps/workflows`, which already does exactly this. The primary `require("voyageai")` then resolves from the app's own `node_modules` and the fragile fallback is never reached. (`apps/ingest` doesn't touch Voyage, so it's left alone.)

`voyageai` is only required dynamically, so it's added to each app's knip `ignoreDependencies` (again matching workflows).

## Verification

Reproduced the prod bundling locally: built `apps/workers/dist` and ran the **compiled** `backfill-trace-search.cjs`. Before: `MODULE_NOT_FOUND`. After: `require.resolve("voyageai")` resolves from the bundle, the client constructs, and forcing cache misses produced real embeddings — no errors.

## Notes

- Lockfile was regenerated with the repo's `minimumReleaseAge` gate bypassed for this single update; it preserves `development`'s existing versions (incl. esbuild@0.28.1 / @types/node@25.9.3, which already predate this branch) and only adds the `voyageai` importer entries. The remaining lock churn is cosmetic pnpm peer-id normalization, no version bumps.
- Follow-up worth considering: `apps/web` also reaches the Voyage path via server functions but uses a different bundler (not tsdown); it wasn't observed failing, but worth confirming its prod build resolves `voyageai` too.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -82,6 +82,7 @@
     "react-dom": "catalog:",
     "rosetta-ai": "catalog:",
     "satori": "0.18.4",
+    "voyageai": "catalog:",
     "zod": "catalog:"
   },
   "devDependencies": {
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -54,6 +54,7 @@
     "effect": "catalog:",
     "hono": "catalog:",
     "tsx": "catalog:",
+    "voyageai": "catalog:",
     "zod": "catalog:"
   },
   "devDependencies": {
--- a/apps/workers/package.json
+++ b/apps/workers/package.json
@@ -70,7 +70,8 @@
     "@repo/utils": "workspace:*",
     "effect": "catalog:",
     "fflate": "0.8.2",
-    "tsx": "catalog:"
+    "tsx": "catalog:",
+    "voyageai": "catalog:"
   },
   "devDependencies": {
     "@platform/testkit": "workspace:*",
--- a/knip.json
+++ b/knip.json
@@ -56,7 +56,8 @@
         "src/domains/feature-flags/feature-flags.collection.ts"
       ],
       "ignoreDependencies": [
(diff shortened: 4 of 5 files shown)


apps/web/package.json
```diff
@@ -82,6 +82,7 @@
     "react-dom": "catalog:",
     "rosetta-ai": "catalog:",
     "satori": "0.18.4",
+    "voyageai": "catalog:",
```
andresgutgon, line 85: This sucks a bit tbh. We're now coupled with this dependency. I guess there is no other easy way

PR #3555, opened 2026-06-14
refactor(monitors): generalize saved-search reader into MetricSeriesReader

geclos:
## What

First step toward the unified **(stream, FilterSet, metric)** monitor-target model from `specs/signals.md` (LAT-664). This PR is a **strictly behavior-preserving refactor**: it generalizes the saved-search-only ClickHouse reader into a stream/metric-generic `MetricSeriesReader`, with the `traces` + `count` path **byte-identical** to the reader it replaces.

**No** alert-kind renames, **no** DB schema changes, **no** firing-behavior changes. Saved-search monitors fire exactly as before.

## Why

Today the monitor system is saved-search-centric: the only target reader counts matching traces, and conditions are count-only. To monitor tools/users/raw-streams with real metrics, the target needs to be modeled as `(stream, FilterSet, metric)`. This PR lays that foundation under the existing firing pipeline without changing any observable behavior, so the riskier follow-ups (agnostic kinds, spans/scores streams, aggregate metrics, in-context UI) build on a verified base.

## Changes

- **`@domain/shared`**: add `MonitorStream`/`MONITOR_STREAMS` and `MonitorMetric` (the target axes the next PRs consume).
- **`@domain/monitors`**: replace the `SavedSearchMatchReader` port with a stream/metric-generic **`MetricSeriesReader`** (`valueInWindow` / `firstEventAt` / `lastEventAt` / `seriesPerBucket`).
- **`@platform/db-clickhouse`**: `MetricSeriesReaderLive` — the `traces` branch is lifted verbatim from the old reader (same SQL, params, window/bucket math, null-guards); clear extension points for spans/sessions and non-count metrics.
- **Repoint** the evaluator, the three saved-search alert state machines, the firing orchestrator, and the workers Live layer onto the new reader. Saved searches resolve to `{stream:'traces', filterSet, query, metric:{kind:'count'}}` via `toCountTarget`.
- **Tests**: port the platform reader test and the in-memory test double to the new port.

## Verification

- `@domain/shared`, `@domain/monitors`, `@platform/db-clickhouse`, `@app/workers` typecheck clean.
- `@domain/monitors`: 115/115 unit tests pass. `@platform/db-clickhouse`: full suite (incl. the ported reader test) passes.
- `pnpm knip` and `pnpm format` clean.
- Adversarial review subagent confirmed the `traces`+`count` SQL is byte-identical to the deleted reader, all callers repointed, and test-coverage parity — no blockers/majors.

## Follow-ups (subsequent PRs)

[...]

PR #3556, opened 2026-06-14
feat(monitors): aggregate metrics (errorRate/avg/p95/sum) for the traces stream

geclos:
> **Stacked on #3555** (`feat/general-monitors-stream-model`). Review/merge that first; this PR's base will retarget to `development` once it lands.

## What

Extends `MetricSeriesReader` beyond `count` with the aggregate metrics from the unified monitor model (`specs/signals.md`): **errorRate**, **avg**, **p95**, **sum** over a numeric field — for the `traces` stream.

The traces inner query already exposes the per-trace columns (`duration_ns`, `cost_total_microcents`, `tokens_total`, `error_count`), so the metrics map straight onto it:
- `errorRate` → `countIf(error_count > 0) / count()`
- `avg|p95|sum` → `avg|quantile(0.95)|sum(<field column>)`

Ratios/averages guard the empty group (`if(count() = 0, 0, …)`) so a metric over an empty window/bucket reads `0`, not `nan` — densified empty buckets stay numeric.

## Why

This is purely **additive and behavior-preserving**: saved-search monitors still use `count`, and nothing yet constructs a non-count target. It unblocks **user-monitor metrics** (error rate / cost / p95 over a user's traces, all on the `traces` stream). The `spans` stream for per-call **tool** metrics follows as its own PR.

## Verification

- `@platform/db-clickhouse` typecheck clean; reader test green (13/13).
- New chdb tests cover `errorRate`/`avg`/`sum`/`p95` over isolated windows + the empty-window guard (incl. `p95`). p95 is asserted only on uniform data (exact regardless of the approximate estimator).
- `knip` + `format` clean.
- Adversarial review subagent: **APPROVE** — verified column mappings resolve to real `LIST_SELECT` aliases, `error_count > 0` matches the traces MV's `status_code = 2` error definition, empty-group guards prevent nan, no SQL injection (fixed column map indexed by a Zod enum), switch is exhaustive. Acted on its nit (added `p95` to the empty-window test).

## Follow-ups

- `spans` stream adapter (`SPAN_FIELD_REGISTRY` + `buildSpanFilterClauses`) → tool per-call metrics.
- Target-on-monitor model + create/update wiring.
- In-context tool & user monitor UI.

### DIFF
--- a/packages/platform/db-clickhouse/src/repositories/metric-series-reader.test.ts
+++ b/packages/platform/db-clickhouse/src/repositories/metric-series-reader.test.ts
@@ -79,6 +79,11 @@ const t0930 = new Date("2026-06-01T09:30:00.000Z")
 const t10 = new Date("2026-06-01T10:00:00.000Z")
 const t1030 = new Date("2026-06-01T10:30:00.000Z")
 const t11 = new Date("2026-06-01T11:00:00.000Z")
+// Aggregate-metric fixtures live in their own windows so they can't perturb the
+// count/bucket assertions above regardless of test ordering.
+const t12 = new Date("2026-06-01T12:00:00.000Z")
+const t13 = new Date("2026-06-01T13:00:00.000Z")
+const t14 = new Date("2026-06-01T14:00:00.000Z")
 
 /** A `traces` + `count` target — the saved-search/match shape this reader supersedes. */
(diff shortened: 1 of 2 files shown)


geclos:
@claude review

PR #3557, opened 2026-06-14
feat(monitors): spans stream for MetricSeriesReader (per tool-call metrics)

geclos:
> **Stacked on #3556 → #3555.** Merge those first; base retargets to `development` as they land.

## What

Adds the `spans` target stream to `MetricSeriesReader`, so a monitor can measure **individual tool calls** rather than the traces that contain them — the per-call metric source for **tool monitors**.

- New span filter registry (`operation`, `tool_name`, `model`/`provider`, `user`/`session`/`trace`, `tags`, `duration`, `cost`, `tokens`) drives a row-local `WHERE` via the shared `buildClickHouseWhere`.
- The reader's spans branch windows on each span's own `start_time` (plain `WHERE`, no GROUP BY) and aggregates `count` / `errorRate` (`status_code = 2`) / `avg`|`p95`|`sum` of `duration`|`cost`|`tokens`.
- Matches how `tool-analytics-repository` aggregates `execute_tool` spans (no `span_id` dedup — same convention).

## Why

Tool monitors need per-call metrics (error rate / p95 latency of `search`'s calls), which are span-level — the trace `tools` filter only selects *traces containing* a tool. This is **additive and behavior-preserving**: the traces path is byte-for-byte unchanged, and nothing constructs a `spans` target yet.

## Verification

- `@platform/db-clickhouse` typecheck clean; reader test green (17/17).
- New chdb tests: tool-name scoping, `errorRate`/`avg`/`sum`, `p95` (uniform → exact), newest-first bucketing over the spans stream.
- `knip` + `format` clean (`SPAN_FIELD_REGISTRY` kept module-private until the UI PR needs it).
- Adversarial review: **Approve with minor changes** — confirmed all spans columns exist, `status_code = 2` matches tool-analytics' error predicate, no-dedup matches convention, `gtePercentile` on spans fails as an isolated defect (never mis-filters), traces path untouched, switch exhaustive, test ids collision-safe. Acted on its notes: corrected the `duration_ns` ALIAS comment, documented that `cost`/`tokens` are inert on `execute_tool` spans, and noted the intentional `gtePercentile` gap.

## Follow-ups

- Target-on-monitor model (`{stream, filterSet, metric}` on the monitor) + create/update wiring.

### DIFF
--- a/packages/platform/db-clickhouse/src/registries/span-fields.ts
+++ b/packages/platform/db-clickhouse/src/registries/span-fields.ts
@@ -1,0 +1,34 @@
+import type { FilterSet } from "@domain/shared"
+import { buildClickHouseWhere, type ChFieldRegistry } from "../filter-builder.ts"
+
+/**
+ * Span-level filter fields for the `spans` stream (one row per span). Mirrors the
+ * trace/session registries but maps to raw `spans` columns — `operation` +
+ * `tool_name` let a tool monitor scope to `execute_tool` spans of one tool.
+ * `duration_ns` is an alias column (read-time `end − start`). `gtePercentile` is
+ * intentionally unsupported here (no per-span distribution resolution); it fails
+ * loud (the builder throws) rather than silently mis-filtering.
+ */
+const SPAN_FIELD_REGISTRY: ChFieldRegistry = {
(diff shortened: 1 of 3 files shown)


geclos:
@claude review

PR #3558, opened 2026-06-14
feat(monitors): unified alert-kind vocabulary (event.*/metric.*), legacy marked

geclos:
> **Stacked on #3557 → #3556 → #3555.** Merge those first; base retargets as they land.

## What

Introduces the **target-agnostic alert kinds** for the unified monitor model — `event.matched`, `metric.threshold`, `metric.escalating` — alongside the existing kinds, which are now **clearly marked legacy** in code (per request):

- `savedSearch.*` (target on the alert's `source`) → superseded by the named unified kind; kept until existing saved-search monitors are migrated off them (removal is a later, **breaking-SDK** change).
- `issue.*` (system, event-driven) → folds in with the signals migration.

Unified kinds carry their target on the **monitor** (`{stream, filterSet, metric}`), so they have no `(sourceType, sourceId)`:
- `ALERT_INCIDENT_KIND_SOURCE_TYPE` is now `Partial` (absent for unified kinds); the source-event resolver returns `[]` for them.
- New `metric.*` conditions get a **float-capable** absolute threshold (error-rate `0.1`, p95 latency, cost — not integers), separate from the legacy count-only saved-search threshold.

## Why

This is the alert vocabulary the in-context tool/user monitors will use. It's the approved **additive** step toward the unified model: the only breaking part (removing `savedSearch.*` + SDK regen) is deferred to a later, deliberately-scheduled migration. Following the repo's existing pattern of clearly-marked "wired in a later milestone" scaffolding.

## Why it's safe

Strictly additive and behavior-preserving — **no existing kind/condition/severity/label value changed**, and nothing constructs a unified-kind monitor yet. The `Partial` relaxation is provably safe: the `undefined` branch is reachable only by unified kinds (never by legacy `issue.*`/`savedSearch.*`), so no legacy source-type check is disabled. All downstream formatters (human-readable alert, Slack/email templates, web incident markers) degrade to safe fallbacks for the new kinds.

## Verification

- Typecheck clean across `@domain/shared`, `@domain/monitors`, `@domain/alerts`, `@domain/notifications`, `@domain/integrations`, `@app/api`, `@app/web`.
- New shared tests (93 pass): kind-map completeness/consistency, source-type omission for unified kinds, float-absolute threshold accepted, aggregate-metric + expected + window accepted, sub-5-min window rejected.
- `knip` + `format` clean.
[...]

PR #3559, opened 2026-06-14
feat(monitors): target-on-monitor columns + entity + repository round-trip

geclos:
> **Stacked on #3558 → #3557 → #3556 → #3555.** Merge those first; base retargets as they land.

## What

Adds the unified query-time **target on the monitor row** — the home for tool/user/raw-stream monitors. Five nullable columns on `monitors` (`target_stream`, `target_filter_set`, `target_query`, `target_saved_search_id`, `metric`) and a `Monitor.target: MonitorTarget | null` entity field:

```
MonitorTarget = { stream, filterSet, query, savedSearchId, metric }
```

`savedSearchId` set ⇒ the predicate is resolved live from that saved search (inline `filterSet`/`query` left null); otherwise the inline predicate applies. All five columns null together ⇒ a legacy source-based alert or a system issue monitor (`target = null`).

## Why

This is the structural pivot from "target on the alert's `source`" to "target on the monitor" that the unified model needs. It's the persistence layer the evaluator (next PR) and the in-context tool/user UI (after) build on.

## Why it's safe

Strictly **additive and behavior-preserving**:
- Migration adds 5 **nullable** columns, no defaults/constraints/indexes changed — purely additive, reversible.
- Nothing constructs a non-null target yet; the evaluator/orchestrator and `listActiveSavedSearchAlerts` are **untouched** — firing still resolves from the saved-search source. `create-monitor` still creates only saved-search monitors (`target: null`).
- The repository mapper is symmetric and lossless; a corrupt row (`target_stream` set, `metric` null) surfaces loudly via `monitorSchema.parse` rather than fabricating a metric.

## Verification

- Typecheck clean: `@domain/monitors`, `@platform/db-postgres`, `@domain/admin`, `@app/api`, `@app/workers`, `@app/web`.
- `@domain/monitors`: 115 tests pass; `monitor-repository.test.ts`: 45 pass (incl. new round-trip cases — a spans/errorRate target round-trips, and a legacy saved-search monitor reads `target = null`).
- `knip` + `format` clean.
- Adversarial review: **Approve** — verified the migration is purely additive (varchar(24) correct for the saved-search CUID), mapper round-trips losslessly, all `Monitor` constructors set `target`, no firing-path change. Acted on its MINOR note (replaced the silent `metric ?? {kind:'count'}` fallback with loud `parse` failure).

## Follow-ups

- Evaluator reads `monitor.target` (prefers it; falls back to the saved-search source) + create path for unified-kind/target monitors.
- (Later) enforce the `savedSearchId` XOR inline-`filterSet` invariant with a `z.refine` once targets are constructed.

### DIFF
--- a/packages/domain/monitors/src/entities/monitor.ts
+++ b/packages/domain/monitors/src/entities/monitor.ts
@@ -4,12 +4,31 @@ import {
   alertIncidentSourceTypeSchema,
   alertSeveritySchema,
   cuidSchema,
+  filterSetSchema,
   monitorAlertIdSchema,
   monitorIdSchema,
+  monitorMetricSchema,
+  monitorStreamSchema,
   organizationIdSchema,
   projectIdSchema,
 } from "@domain/shared"
 import { z } from "zod"
+
+/**
+ * What a query-time monitor watches: a stream + predicate + the metric to measure
+ * (the unified `event.*`/`metric.*` model). `savedSearchId` set ⇒ the predicate
+ * resolves live from that saved search (inline `filterSet`/`query` left null);
+ * otherwise the inline `filterSet`/`query` apply. A monitor's `target` is `null`
+ * for legacy source-based alerts (target on the alert) and system issue monitors.
+ */
+export const monitorTargetSchema = z.object({
+  stream: monitorStreamSchema,
+  filterSet: filterSetSchema.nullable(),
+  query: z.string().nullable(),
+  savedSearchId: cuidSchema.nullable(),
+  metric: monitorMetricSchema,
+})
+export type MonitorTarget = z.infer<typeof monitorTargetSchema>
 
 /**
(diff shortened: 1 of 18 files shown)


geclos:
@claude review

geclos:
Triaged the review — verdict was Approve and I agree; all points are observations/nits, no code changes needed:
- **Index on `target_saved_search_id` / `savedSearchId`-set round-trip test** — premature here; no path queries by that column yet and `savedSearchId`-on-target isn't wired by any caller. Natural follow-ups when target-backed saved searches actually land.
- **`targetStream = ''` truthy edge** — can't occur (`monitorStreamSchema` rejects empty strings).
- **No FK / JSDoc duplication** — consistent with the no-FK convention and fine as-is.
Migration is additive/safe, mapper is symmetric, parse-on-read is the right pattern.

PR #3560, opened 2026-06-14
feat(monitors): make alert source nullable (sourceless unified-kind alerts)

geclos:
> **Stacked on #3559 → #3558 → #3557 → #3556 → #3555.** Merge those first; base retargets as they land.

## What

Makes a monitor alert's `source` **nullable**. Unified `event.*`/`metric.*` alerts watch the monitor's `target` (`{stream, filterSet, metric}`), not a `(sourceType, sourceId)` — so they carry no source.

- Migration: `monitor_alerts.source_type` → `DROP NOT NULL` (purely relaxing).
- Entity: `MonitorAlert.source: {type, id} | null`.
- Every source-based reader/serializer null-safed.

## Why

The alert-model prerequisite for the evaluator + create path that make tool/user monitors fire (next PR). Legacy alerts always have a source, so this is the last schema/entity relaxation before wiring firing.

## Why it's safe

Strictly additive and behavior-preserving:
- `DROP NOT NULL` changes no data and is reversible; only `monitor_alerts` touched (`alert_incidents.source_type` stays NOT NULL).
- Legacy `savedSearch.*`/`issue.*` alerts always have a source — the `run-*-alert` use-cases still `die` on a missing id (unreachable for them), and all source-based queries (live + fake repo) exclude null-source alerts exactly as a SQL `source_type = 'savedSearch'` predicate would.
- **No null-source alert can be created or serialized through any existing path** — `buildMonitorAlert` still rejects unified kinds and requires a source. `update-monitor-alert` validation can't loosen a legacy alert to null (its input `source.type` is non-nullable).

## Verification

- Typecheck clean: `@domain/monitors`, `@platform/db-postgres`, `@app/api`, `@app/web`, `@app/workers`, `@domain/alerts`, `@domain/notifications`, `@domain/admin`.
- `@domain/monitors`: 115 tests; `monitor-repository.test.ts`: 46 (incl. a real-DB round-trip of a `metric.threshold` alert with `source: null`).
- `knip` + `format` clean.
- Adversarial review: **Approve** — verified the migration is purely relaxing, every former `alert.source.X` access is null-safe and legacy-preserving, validation can't loosen legacy alerts, mappers are symmetric, and fake↔live repo parity holds. Acted on its MINOR: made the OpenAPI `source` field `.nullable()` so the published contract matches before unified alerts ship.

## Follow-ups

- Evaluator reads `monitor.target` (prefers it; else saved-search source) + unified-kind create path → **tool/user monitors fire end-to-end**.


geclos:
Adversarial-review note (non-blocking for this stacked PR): this PR makes `source` nullable in `apps/api/src/openapi/entities/monitor.ts` and `toMonitorAlertResponse`, but the generated `apps/api/openapi.json`, `mcp.json`, and the Fern TS/Python SDKs were not regenerated. The `API Manifests` workflow only runs on PRs whose base is `development`, so it won't fire here — but it **will** fail once this lands on `development` unless the manifests/SDKs are regenerated (`pnpm --filter @app/api openapi:emit` + `mcp:emit` + Fern generate). Consistent with the epic's deferred API/SDK surface; flagging so it isn't missed at merge time.

geclos:
@claude review

geclos:
Triaged the review — **Approve**, no changes needed. The three minor findings are all non-blocking and either resolved upstack or unreachable:

- **`event.matched` missing from `KINDS_WITHOUT_CONDITION`** — unreachable here (`event.matched` isn't user-creatable at this PR). Verified it's already handled at #3562, where the set is hoisted to `@domain/shared` and includes `event.matched` — i.e. fixed exactly where it becomes reachable.
- **`updateAlert` port doesn't carry `sourceType`** — latent. A legacy→unified kind change can't reach an inconsistent state: validation requires a unified kind to have no source, and `input.source ?? alert.source` can't null an existing source via update, so the transition is rejected rather than leaving a stale `source_type`. No action.
- **Fake repo `updateAlert` drops `sourceId` for null-source alerts** — correct behavior; the optional defensive assert isn't a convention here.

Also confirms the openapi/mcp/SDK regen is correctly deferred to merge-time (API Manifests only runs on `development`-based PRs).

PR #3561, opened 2026-06-14
feat(monitors): unified metric-alert evaluator (metric.threshold/escalating)

geclos:
> **Stacked on #3560 → #3559 → #3558 → #3557 → #3556 → #3555.** Merge those first; base retargets as they land.

## What

A pure evaluator for the unified alert kinds:
- `evaluateMetricAlert` (`metric.threshold`, point)
- `evaluateMetricEscalatingAlert` (`metric.escalating`, sustained)

Both compute the verdict over a monitor's resolved target (`{stream, filterSet, query, metric}`) via the `MetricSeriesReader`. The computational core of unified-monitor firing.

## How

Mirrors the saved-search evaluator's window / baseline / seasonal math (helpers now exported and shared — **legacy logic unchanged**), with the unified generalizations:
- the measured metric is the target's (not a hardcoded count);
- the absolute threshold is a **float `value`** (not an int count);
- a `multiplier` baseline is normalized to the window **only for accumulating metrics** (`count`/`sum`); intensive metrics (`avg`/`p95`/`errorRate`) compare directly;
- `metric.threshold` uses a fixed 5-min current window for all modes (a current-window rate is the right semantics for a float metric — the evaluator is pure, taking target+condition+now).

## Why it's safe

Pure and isolated — **not yet wired** to a state machine/orchestrator (firing + incidents are the next PR). The only change to the legacy evaluator is adding `export` to four math helpers (`WEEK_MS`/`mean`/`sampleStddev`/`baselineWindow`); no behavior change.

## Verification

- `@domain/monitors` typecheck clean; 120 tests pass (115 + 5 new).
- New unit tests (fake reader): absolute fire/no-fire, multiplier normalization split (same data → count fires at threshold ≈1, avg doesn't at threshold 12), per-bucket escalating scaling (10×1-min buckets, accumulating scaled vs intensive as-is).
- `knip` + `format` clean.
- Adversarial review: **Approve with nits** — verified legacy untouched, the accumulating-vs-intensive split is the correct generalization, escalating per-bucket math matches legacy, window choice deliberate. Acted on its Minor (gate `firstEventInWindow` on firing/activity, not the metric value — an intensive metric can be 0 with activity) and the test-comment nit.

## Follow-ups

- Wire the evaluator into state machines + orchestrator + **incident model** (sourceless incidents + notifications) → tool/user monitors fire end-to-end.
- Unified-kind create path; in-context tool/user UI + dashboard Target column.

### DIFF
--- a/packages/domain/monitors/src/index.ts
+++ b/packages/domain/monitors/src/index.ts
@@ -66,6 +66,14 @@ export { buildMonitorAlert } from "./use-cases/create-monitor-alert.ts"
 export type { DeleteMonitorError, DeleteMonitorInput } from "./use-cases/delete-monitor.ts"
 export { deleteMonitorUseCase } from "./use-cases/delete-monitor.ts"
 export type {
+  EvaluateMetricAlertError,
+  EvaluateMetricAlertInput,
+  EvaluateMetricEscalatingAlertInput,
+  MetricAlertEvaluation,
+  MetricEscalatingEvaluation,
+} from "./use-cases/evaluate-metric-alert.ts"
+export { evaluateMetricAlert, evaluateMetricEscalatingAlert } from "./use-cases/evaluate-metric-alert.ts"
+export type {
   EvaluateSavedSearchAlertError,
   EvaluateSavedSearchAlertInput,
   SavedSearchEscalatingEvaluation,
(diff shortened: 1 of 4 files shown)


geclos:
Adversarial-review note (non-blocking, follow-up): `mode: "expected"` is reachable for intensive metrics (errorRate/avg/p95) via the in-context form, but `seasonalAnomalyThreshold` floors σ at `max(observed, √expected, 1.0)` — count-calibrated. For `errorRate ∈ [0,1]` the threshold is ≥ ~1.0, so the alert can effectively never fire (same for sub-unit avg/p95 cost). The right fix is a product/statistical decision (drop the count floors for intensive metrics, or disallow `expected` mode for them) on a helper shared with issue detection, so it's intentionally left out of this triage pass rather than patched blind. Worth resolving before unified metric monitors are advertised.

geclos:
@claude review

PR #3562, opened 2026-06-14
feat(monitors): unified create path (target-on-monitor, sourceless alerts)

geclos:
> **Stacked on #3561 → … → #3555.** Merge those first; base retargets as they land.

## What

`createMonitorUseCase` can now create **target-on-monitor** monitors whose alerts are unified `event.*`/`metric.*` kinds (no source) — the domain capability behind tool/user monitors.

- `buildMonitorAlert` splits on kind: legacy `savedSearch.*` still require a saved-search source; unified kinds take **none** (and `event.matched` joins the no-condition set).
- `createMonitorUseCase` gains an optional `target` and enforces the **source-vs-target invariant**: a monitor is homogeneous (all-unified-with-target or all-legacy-with-source); mixing or missing/extra target is rejected.
- `USER_CREATABLE_ALERT_KINDS` gains the unified kinds; `KINDS_WITHOUT_CONDITION` is **hoisted to `@domain/shared`** (was duplicated + had drifted across the create/update alert use-cases).

## Why it's safe

Additive and behavior-preserving: legacy saved-search creation is unchanged, and the **public API/web create surface is untouched** (no `target` field, kinds still restricted there) — a unified-kind create via those entry points fails cleanly with a `ValidationError`, no broken monitor persists.

## Verification

- Typecheck clean: `@domain/shared`, `@domain/monitors`, `@app/api`, `@app/web`, `@app/workers`.
- `@domain/monitors`: 128 tests (incl. unified create round-trip, sourceless `buildMonitorAlert` paths, and the rejection cases).
- `knip` + `format` clean.
- Adversarial review: **Request changes → fixed.** It caught two MAJOR defects, both fixed + covered by tests:
  1. the source-vs-target check used `some()`, letting a **mixed legacy+unified alert array** persist an incoherent monitor → now rejects any mix (`field: "alerts"`);
  2. `update-monitor-alert`'s **stale local `KINDS_WITHOUT_CONDITION`** made a freshly-created `event.matched` alert un-updatable → fixed by hoisting the constant to `@domain/shared`.
  Also added the suggested direct `buildMonitorAlert` unit tests.

## Follow-ups

- The firing vertical: state machines + orchestrator + **sourceless incidents** (nullable `alert_incidents` + notification handling) → unified monitors fire end-to-end.
- In-context tool/user monitor UI + dashboard Target column.

### DIFF
--- a/packages/domain/monitors/src/use-cases/create-monitor-alert.test.ts
+++ b/packages/domain/monitors/src/use-cases/create-monitor-alert.test.ts
@@ -80,7 +80,11 @@ describe("buildMonitorAlert", () => {
       buildMonitorAlert(
         {
           kind: "metric.threshold",
-          condition: { kind: "metric.threshold", metric: { kind: "errorRate" }, threshold: { mode: "absolute", value: 0.1 } },
+          condition: {
+            kind: "metric.threshold",
+            metric: { kind: "errorRate" },
+            threshold: { mode: "absolute", value: 0.1 },
+          },
         },
         monitorId,
         at,
--- a/packages/domain/monitors/src/use-cases/create-monitor-alert.test.ts
+++ b/packages/domain/monitors/src/use-cases/create-monitor-alert.test.ts
@@ -68,4 +68,44 @@ describe("buildMonitorAlert", () => {
     )
     expect(error._tag).toBe("AlertConditionMismatchError")
   })
+
+  it("builds a sourceless unified alert with no condition (event.matched)", async () => {
+    const alert = await Effect.runPromise(buildMonitorAlert({ kind: "event.matched" }, monitorId, at))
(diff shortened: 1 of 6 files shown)


geclos:
@claude review

geclos:
Triaged the review. Fixed the one item that mattered; the rest don't warrant changes:
- **Misleading source error for unified kinds — fixed** (`d67d4a0e6`). The update path emitted `Source type must be "null" for event.matched`; now mirrors the create path (`Alerts of kind "X" take no source`). Also dropped the now-false "only saved-search kinds are user-owned" comment. (This was also flagged on #3558 — fixed here where unified kinds first become user-creatable/reachable.)
- **Stale comment / event.matched+target integration test** — comment removed above; the green path is covered at the unit level in `create-monitor-alert.test.ts`, so the integration-level case is optional. No change.

PR #3563, opened 2026-06-14
feat(monitors): firing vertical for unified metric monitors

geclos:
## What & why

PR9 of the unified-monitors epic. Wires the **unified (target-on-monitor) monitor family end to end** so tool/user/raw-stream monitors actually fire — alongside the existing saved-search family, on the same trigger/throttle. Stacked on `feat/monitors-unified-create`.

This is the **backend firing vertical only**. The in-context Tools/Users creation UI + dashboard Target column ship separately in PR10.

## Changes

- **Sourceless incidents.** `alert_incidents.source_type`/`source_id` become nullable — unified incidents carry no source (the target is recovered via `monitorAlertId`). The resulting cascade is handled everywhere a read path touched source:
  - `request-incident-notifications.ts`: a `SourcedIncident` type-guard **early-skips** sourceless incidents from the source-based notification path and narrows for all downstream helpers.
  - web serializers (`alerts.functions.ts`, `monitors.functions.ts`, `monitor-incidents-table.tsx`), the issue sweeper, `resolve-alert-incident`, the saved-search incident writer, and seeds — all null-guarded.
- **State machine** (`run-metric-monitor-alert.ts`): `event.matched` (point), `metric.threshold` (rearm open/close), `metric.escalating` (sustained gate, threshold frozen at open via `entrySignals`) — writing sourceless incidents.
- **Orchestrator** (`check-metric-monitors.ts`): lists active metric alerts, resolves each target (live saved-search ref or inline filterSet), dispatches with per-alert failure isolation; races a since-deleted saved search safely.
- **Repository** (`listActiveMetricMonitorAlerts`): only target-bearing monitors — **provably disjoint** from the saved-search lister, so no alert is double-fired.
- **Worker**: runs both families, **failure-isolated** so one family's list-query error can't starve the other on a tick.

## Deferred

Unified-incident notification *copy* — sourceless incidents open but do not notify yet (the source-based renderer doesn't cover them). Tracked for a follow-up.

## Testing

- `@domain/monitors` 131 tests green, incl. new `check-metric-monitors.test.ts` (threshold opens a sourceless incident; below-threshold opens none; no-target/legacy monitor ignored).
- `@domain/notifications` 56, `@domain/alerts` 10 green.
- typecheck/format/knip clean across affected packages.

## Review

Adversarial review run before opening: verdict **SHIP**, null-source cascade fully guarded, two firing paths provably disjoint (no double-fire). The one actionable MINOR — worker failure-isolation between the two families — is applied in this PR.

### DIFF
--- a/apps/web/src/domains/alerts/alerts.functions.ts
+++ b/apps/web/src/domains/alerts/alerts.functions.ts
@@ -38,7 +38,8 @@ export interface AlertIncidentRecord {
   readonly kind: AlertIncidentKind
   readonly severity: AlertSeverity
   readonly sourceType: AlertIncident["sourceType"]
-  readonly sourceId: string
+  /** `null` for unified (target-on-monitor) incidents — the target lives on the monitor. */
+  readonly sourceId: string | null
   readonly startedAt: string
   readonly endedAt: string | null
   /** Resolved name of the issue tied to the incident; `null` if not found (e.g., deleted). */
@@ -108,7 +109,7 @@ export const listProjectAlertIncidentsInRange = createServerFn({
         })
 
         const issueIds = Array.from(
-          new Set(incidents.filter((i) => i.sourceType === "issue").map((i) => i.sourceId)),
+          new Set(incidents.flatMap((i) => (i.sourceType === "issue" && i.sourceId !== null ? [i.sourceId] : []))),
         ).map(IssueId)
 
         const issues =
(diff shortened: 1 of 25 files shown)


geclos:
@claude review

geclos:
Triaged the review. Addressed the one substantive item; the rest are accurate observations that don't warrant changes here:

- **`metric.escalating` missing `lockAlertForUpdate` — fixed** (`346baee37`). Confirmed the asymmetry: `metric.threshold` locks before its read-then-insert, `metric.escalating` didn't. Added the same `FOR UPDATE` lock so concurrent ticks can't double-open an incident. Note: the saved-search escalating runner has the identical pre-existing unlocked pattern — left as a separate uniform follow-up rather than touching shipped behaviour in this PR.
- **`event.matched` / `metric.escalating` test coverage** — fair gap; deferring to a focused follow-up (orchestrator→state-machine→writer path), tracked with the epic.
- **`SavedSearchEntrySignals` reuse / `closeSavedSearchIncident` source guard** — structurally correct today; noted for if/when the types are reused more broadly. No change.

The review also confirmed the sweep change (`listProjectsWithActiveMonitorAlerts`) and the disjoint-firing/migration/worker-isolation analysis, which matches our own pass.

PR #3572, opened 2026-06-15
Add preset monitor creation flows

geclos:
## Summary

Adds recommended monitor creation flows for tool and user targets. Target monitor creation now opens a single modal with a Recommended/Advanced segmented control: Recommended shows opinionated aggregate monitor presets, while Advanced exposes the existing full monitor form in the same modal shell.

Tool presets cover failing tools, slow tools, usage spikes/drops, overuse for individual tools, and fleet-level all-tools monitors. User presets cover user errors, slow responses, activity spikes/drops, and cost spikes for both individual users and all-user traffic.

The metric monitor condition schema now supports an optional `above`/`below` direction for unified metric alerts, enabling usage-drop presets without changing saved-search monitor behavior.

## Validation

- `pnpm --filter @domain/monitors test -- src/use-cases/evaluate-metric-alert.test.ts`
- `pnpm --filter @domain/shared typecheck`
- `pnpm --filter @domain/monitors typecheck`
- pre-commit format + knip

Please attach screenshots or a short screen recording of the new monitor preset modal for reviewer context.

### DIFF
--- a/apps/web/src/routes/_authenticated/projects/$projectSlug/monitors/-components/advanced-monitor-create-fields.tsx
+++ b/apps/web/src/routes/_authenticated/projects/$projectSlug/monitors/-components/advanced-monitor-create-fields.tsx
@@ -1,0 +1,52 @@
+import { Input, Textarea } from "@repo/ui"
+import { AlertCardForm } from "./alert-card-form.tsx"
+import type { AlertDraft, AlertFieldErrors } from "./alert-form-helpers.ts"
+
+export interface AdvancedMonitorCreateValue {
+  readonly name: string
+  readonly description: string
+  readonly alert: AlertDraft
+  readonly nameError: string | undefined
+  readonly alertErrors: AlertFieldErrors
+}
+
+export function AdvancedMonitorCreateFields({
+  value,
+  onChange,
+  projectId,
+  projectSlug,
+}: {
+  readonly value: AdvancedMonitorCreateValue
+  readonly onChange: (next: AdvancedMonitorCreateValue) => void
+  readonly projectId: string
+  readonly projectSlug: string
+}) {
+  const set = (patch: Partial<AdvancedMonitorCreateValue>) => onChange({ ...value, ...patch })
+
+  return (
+    <>
+      <Input
+        required
+        label="Name"
+        placeholder="Tool error spikes"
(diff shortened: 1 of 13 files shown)


geclos:
@claude review

geclos:
Addressed the review feedback in `2a373de1a`:

- Restored the zero-baseline safety guard for above-direction metric alerts (`0 >= 0` no longer opens multiplier/expected incidents).
- Renamed the sigma helper parameters to match their actual `stddev` / `mean` inputs.
- Added coverage for below/multiplier, below/expected, zero-baseline above behavior, and below expected escalating thresholds.
- Extracted the duplicated Recommended/Advanced switch into `MonitorModeSwitch`.
- Aligned the metric-threshold preview wording with the UI phrasing.
- Removed the leftover `bodyPrefix` prop from the base create modal now that advanced mode switches within the same modal shell.

I left the preset `createPreset` toast behavior as-is since those flows do not expose editable fields; name/condition errors still surface as a destructive toast. I also left `targetWithFilter` as a shallow override because current user/all-user monitor targets never carry a status filter, and the error preset intentionally scopes the target to errored traces.

geclos:
Adversarial-review follow-up — fixed one real correctness issue in the directional-alert change (`2ceabe42a`):

**Absolute `below` alerts fired false incidents on empty windows.** `isMetricThresholdMet(0, N, "below")` is `true` (`0 <= N`), so an `absolute` + `below` monitor opened an incident on any quiet/no-traffic window — and the 5-min sweep evaluates metric-only projects, so it would spam. Added an activity gate (require `firstEventInWindow`) for **absolute below** in both the threshold and escalating paths, plus a regression test.

**`multiplier` and `expected` below are intentionally exempt** — they have a baseline (recent/seasonal) and the `threshold > 0` guard (from your "tighten directional metric alerts" commit) already suppresses the genuinely-no-traffic case while still firing when a busy target goes silent. Verified the existing `supports below-threshold multiplier alerts` test (empty current window vs busy baseline → fires) still holds.

The preset flows themselves are sound: they use `metric.escalating` + `expected` (seasonal) mode, where the seasonal threshold collapses to ~0 during legitimately quiet periods, so no false positives there. This fix only hardens the `absolute` path reachable via Advanced mode.

PR #3579, opened 2026-06-15
feat(web): per-turn index in the session drawer

csansoon:
Adds a **Turns** section to the Session tab of the session detail drawer: a per-turn index (one card per trace) for multi-trace sessions, sitting between the whole-session summary and the full Conversation transcript.

## what it adds

The Session tab previously collapsed a multi-turn session into single totals (one duration, one cost, etc.), so there was no way to see what each turn contained or which turn drove cost/latency without opening every trace. The Turns section lists each turn as a card with:

- input + output **preview** — what the turn was about and how it replied
- **duration · cost**, tool / error / annotation **chips**, and the **model** pinned right (shown only when it changes from the previous turn)
- a **health dot** flagging problem turns
- between-turn **gap rows** (clock + elapsed) surfacing think-time between turns, which the session duration breakdown deliberately excludes

Clicking a card opens that turn's trace. Only rendered for multi-trace sessions; single-trace sessions keep their existing Spans tab.

## health dot

Each turn is compared against the project-wide cohort baseline (`useTraceCohortSummary`), not its neighbors, so it's independent of the natural turn-over-turn cost ramp:

- red — the turn errored, or its duration/cost is in the project's top 1% (≥ p99)
- amber — top 5% (≥ p95)
- nothing otherwise

Percentiles are sample-gated (p95 needs ≥ 100 traces, p99 ≥ 1000), so on low-volume projects only errored turns light up — by design. Hovering the dot explains why.

## lazy previews

Message content isn't returned by list/collection queries, only by per-trace detail. Each card fetches its preview via `useTraceDetail` **only when it scrolls into view** (new `useInView` hook) and only while the section is open (`DetailSection` mounts its children on expand). Everything else is free: tool counts come from the already-loaded session spans, annotation counts from a batched hook, and the rest from `TraceRecord` aggregates.

The per-turn duration-composition tooltip reuses `computePerTraceBreakdowns`, extracted from `computeSessionDurationBreakdown` (no behavior change).

## decisions / follow-ups

- Turn-over-turn cost/latency deltas were prototyped and dropped: cost ramps every turn (each turn re-sends the accumulated context), so a per-turn delta just restates the ramp instead of signalling anything. The health dot, comparing against the project baseline, is the ramp-independent "is this turn off" signal instead.
- Moments per turn are deferred — that needs the message-index→trace map the Conversation tab builds.

## tests

Unit tests cover `computePerTraceBreakdowns`, `computeTurnHealth` (error / p95 / p99 / gated-null cases), per-turn tool-call stats, and input/output preview extraction. Typecheck and Biome clean.

### DIFF
--- a/apps/web/src/lib/hooks/useInView.ts
+++ b/apps/web/src/lib/hooks/useInView.ts
@@ -1,0 +1,36 @@
+import { useCallback, useRef, useState } from "react"
+
+/**
+ * Reports when an element first scrolls into view (sticky once seen). Uses a
+ * callback ref so the observer wires up the moment the node mounts. `rootMargin`
+ * expands the viewport to prefetch slightly ahead; intermediate scroll
+ * containers are clipped automatically, so a viewport root works inside nested
+ * `overflow` containers.
+ */
+export function useInView<T extends Element>({ rootMargin = "0px" }: { rootMargin?: string } = {}) {
+  const [inView, setInView] = useState(false)
+  const observerRef = useRef<IntersectionObserver | null>(null)
+
+  const ref = useCallback(
+    (node: T | null) => {
+      observerRef.current?.disconnect()
+      observerRef.current = null
+      if (!node) return
+      const observer = new IntersectionObserver(
+        (entries) => {
+          if (entries.some((entry) => entry.isIntersecting)) {
+            setInView(true)
+            observer.disconnect()
+            observerRef.current = null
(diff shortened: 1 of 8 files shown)


geclos:
@csansoon what do we do with this pr?

csansoon:
Nah let's close it. It was an exploration we ended up not liking

PR #3584, opened 2026-06-16
Improve onboarding with seeded sample projects

geclos:
## what changed

New organizations now get two projects during workspace provisioning:

- the user's default project, named from the organization (`<org name>'s project`)
- a separate `Sample project`, marked with `settings.isSample`

The sample project emits a `SampleProjectCreated` domain event. The domain-events worker publishes a `projects:seedDemo` task, and the projects worker starts `seedDemoProjectWorkflow` for that project. This kicks off the existing demo seeding path early so trace search embeddings and taxonomy data can be prepared while the user is still moving through onboarding.

Onboarding now lets the user rename their default project in the flaggers step. The telemetry step replaces “Skip for now” with an “Explore sample project” action when the sample project exists. Clicking it marks onboarding completed on the user project and navigates to the sample project.

The user project still shows the normal no-traces blank state after onboarding is completed via the sample-project path, instead of falling through to an empty sessions/traces table.

https://github.com/user-attachments/assets/15092dce-bf20-42e3-8c82-6a47cd8f3ac7

## validation

- `pnpm --filter @domain/shared typecheck`
- `pnpm --filter @domain/events typecheck`
- `pnpm --filter @domain/organizations typecheck`
- `pnpm --filter @domain/queue typecheck`
- `pnpm --filter @domain/organizations test -- provision-organization-workspace.test.ts`
- `pnpm --filter @app/workers test -- domain-events.test.ts`
- pre-commit `turbo format` + `knip`

## reviewer note

This includes onboarding UI changes; screenshots or a short recording of the flaggers and telemetry steps will help review.

### DIFF
--- a/apps/web/src/domains/organizations/organizations.functions.ts
+++ b/apps/web/src/domains/organizations/organizations.functions.ts
@@ -64,7 +64,7 @@ export const createOrganization = createServerFn({ method: "POST" })
         actorUserId: userId,
         name: data.name,
         slug,
-        defaultProjectName: "My project",
+        defaultProjectName: `${data.name.trim()}'s project`,
       }).pipe(
         withPostgres(
           Layer.mergeAll(ApiKeyRepositoryLive, ProjectRepositoryLive, OutboxEventWriterLive),
--- a/apps/web/src/domains/projects/projects.collection.ts
+++ b/apps/web/src/domains/projects/projects.collection.ts
@@ -74,6 +74,8 @@ export function createProjectMutation(name: string) {
       notifications: undefined,
       escalation: undefined,
       onboardingType: undefined,
+      onboardingCompleted: undefined,
+      isSample: undefined,
       sampling: undefined,
     },
     firstTraceAt: null,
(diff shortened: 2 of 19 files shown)


geclos:
@claude review

geclos:
Addressed the review items I thought were worth the tradeoff in `cbd4bb937`:

- Emit `OrganizationCreated` before `SampleProjectCreated` and updated the org provisioning test.
- Added a short note explaining why the sample project bypasses `createProjectUseCase`.
- Softened the onboarding tooltip so it does not over-promise sample data availability.
- Added `SampleProjectCreated` to the PostHog whitelist with test coverage.

I intentionally left a few lower-value nits alone: the project lookup is already org-scoped by `ProjectRepositoryLive.findById`, the legacy suggested-name fallback is harmless, the sample project null guard is acceptable defensive UI code, and the extra `completeProjectOnboarding` test did not seem worth the added surface area for this pass.

Validation:
- `pnpm --filter @domain/organizations exec vitest run src/use-cases/provision-organization-workspace.test.ts`
- `pnpm --filter @platform/analytics-posthog exec vitest run src/whitelist.test.ts`
- `pnpm --filter @domain/organizations typecheck`
- `pnpm --filter @platform/analytics-posthog typecheck`

Note: `pnpm --filter web typecheck` still fails on existing `@latitude-data/telemetry` module resolution, unrelated to this commit.

PR #3586, opened 2026-06-16
spec: revise signals — trackers model + phased build plan (LAT-664)

csansoon:
Revises `specs/signals.md` (the original LAT-664 unified-tracking spec) into a **tracker**-based model. Spec only — no code in this PR.

## the model

A **signal** is a tracked bucket of traces whose members are its **scores** — there is no separate occurrence ledger. A signal carries:

- an **origin** — `user` (hand-built) or `system` (auto-created);
- an optional **tracker** (a jsonb column, one per signal): the write-time detector, one of `semantic_similarity`, `script`, or `llm_as_judge`. A tracker-less signal is a **sink**;
- an optional **filters** `FilterSet` that pre-gates the tracker.

A tracker runs against each in-scope trace at ingest and writes a score; membership is the matched subset, where `passed = value ≥ threshold` is frozen at write time. Auto-discovery keeps working: flagger and human annotations route to (or create) tracker-less sink signals via the existing centroid + hybrid-search pipeline, and a sink can later be **promoted** to a judge tracker through today's "Monitor issue" / optimize-evaluation flow, with its accumulated annotations as the alignment ground truth.

## what changed from the prior spec

The original spec removed auto-discovery and added a separate `signal_occurrences` table. This revision reverses both:

- **Discovery stays**, modeled as tracker-less sink signals — the flagger and discovery machinery is reused unchanged.
- **Occurrences are scores.** `signal_occurrences` is dropped. Pure-tracker matches are immutable, so they write straight to the ClickHouse score-analytics table and skip the canonical Postgres row; judge and annotation scores keep the existing Postgres-canonical → ClickHouse path. That split is what lets a single ledger scale.
- **The tracker replaces the matcher-on-the-signal model**, and the `annotation_sink` tracker is gone — a sink is simply a signal with no tracker.

## data model

No new tables. `issues` → `signals` (adds `tracker`, `filters`, `origin`); `scores` gains `source_type` (with a new `tracker` value) and `signal_id`; `monitors` gains `target_signal_id`; the ClickHouse `scores` table gains `signal_id`. The unified `event.*` / `metric.*` alert kinds and `MonitorMetric` already exist — only `event.regressed` is new.

## build plan

The spec ends with a phased plan in which every phase is an independently shippable, non-breaking deploy. **MVP is phases 1–4:** rename issues→signals (behavior-identical), custom semantic signals, user-created judge signals (reusing the evaluation stack), and monitors on signals. Script trackers, unifying judge execution onto the tracker pipeline, moments→signals, and flagger consolidation are post-MVP.

Refs LAT-664.

### DIFF
--- a/dev-docs/signals.md
+++ b/dev-docs/signals.md
@@ -1,0 +1,273 @@
+# Signals
+
+A **Signal** is a tracked bucket of traces. Signals are the top-level observability entity of the reliability system: a signal defines *what* to watch, its **members** are the traces that exhibit it, and any signal can be watched over time by a **Monitor**.
+
+Signals unify two formerly separate systems — **Issues** (auto-created buckets of failed/annotated scores, monitored invisibly) and the user-configured alerting that ran over **Saved Searches** — into one model. A signal's members are its **Scores**; a signal is watched by a **Monitor**; a monitor owns **Alerts**; a fired alert opens an **Incident**, which is what notifies.
+
+```
+                tracker runs at ingest        monitors aggregate         alerts fire on        records the
+                (or annotation lands)         the score stream           conditions            firing
+  Trace ──────▶ SIGNAL ─────────────────────▶ MONITOR ─────────────────▶ ALERT ──────────────▶ INCIDENT ──▶ notifications
+                membership = its SCORES        (a metric over the
+                (write-time materialized)       signal's scores/traces)
+```
+
+Signals live in `@domain/signals` (the evolved `@domain/issues` package). The shared enums (`SignalOrigin`, `SignalTrackerType`, `ScoreSourceType`, alert kinds, conditions, severities) live in `@domain/shared` so the signals, monitors, scores, and notifications domains can reference them without a circular dependency.
+
+> **Spec**: `specs/signals.md` carries the design decisions, trade-offs, and the phased build plan. This doc describes the intended final system; consult the spec for phase status while the system is under construction.
+
+Related docs:
+
+- [scores.md](./scores.md) — the canonical scores model and the Postgres/ClickHouse split that signal membership rides on.
+- [monitors.md](./monitors.md) — the monitor / alert / incident machinery; signals are one monitor target type.
+- [evaluations.md](./evaluations.md) — the LLM-judge generation/alignment/execution stack that backs `llm_as_judge` trackers.
+- [conversation-intelligence.md](./conversation-intelligence.md) — the moment-label anchor matching that the semantic runner generalizes.
+- [reliability.md](./reliability.md) — the cross-domain reliability loop that produces scores.
+- [notifications.md](./notifications.md) — incidents drive notifications; the monitor mute gate is the only seam.
+- [domain-errors.md](./domain-errors.md) — `@domain/issues`/`@domain/signals` is the reference implementation for per-package error layout.
+
+## Why membership is materialized at write time
+
+Two structural decisions carry the whole model, and both follow from a single cost constraint: counting and alerting over *semantic* membership is only affordable if each trace's verdict is computed once, on arrival, and remembered.
+
+1. **A signal's occurrences ARE its scores.** There is no separate occurrence ledger. Every membership-bearing event — a tracker run, an annotation, a future custom push — writes a `scores` row carrying `signal_id`. A signal's membership is the subset of those rows that *matched* (`value >= threshold`). This collapses the former inconsistency where issues counted scores and saved searches counted traces into one counting unit, and reuses the entire scores pipeline.
+2. **Membership is materialized at write time.** A tracker is evaluated against each in-scope trace once, on arrival, and the verdict is frozen as a score (`passed`). Membership reads never re-apply a threshold at query time.
+
+The reasoning:
+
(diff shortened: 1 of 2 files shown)


geclos (dismissed):
Looks good

dev-docs/signals.md
```diff
@@ -0,0 +1,273 @@
...
+### Tracker
+
+The write-time detector embedded on a signal (`signals.tracker` jsonb, discriminated by `type`). Three types:
+
+```ts
+export const SIGNAL_TRACKER_TYPES = ["semantic_similarity", "script", "llm_as_judge"] as const
+export type SignalTrackerType = (typeof SIGNAL_TRACKER_TYPES)[number]
+
+export type SignalTracker =
+  | { type: "semantic_similarity"; semantic: SemanticAnchors; threshold: number }
```
geclos, line 75: missing code-rule based

specs/signals.md
```diff
@@ -37,75 +75,44 @@ Two structural decisions carry the whole spec and are stated up front:
 
 - an **origin**: `user` (built deliberately by a person) or `system` (auto-created by Latitude — see [Discovery](#discovery-sinks-and-promotion)). Origin is the differentiator between hand-built and discovered signals.
 - an optional **`filters`** (a `FilterSet`): a cheap, row-local pre-gate restricting which traces the tracker is even run against ("only `service = checkout`", "only traces above p90 latency"). Empty/absent = all traces. `filters` is only meaningful alongside a tracker — it gates tracker execution.
-- an optional **`tracker`** (a jsonb column, **one per signal**): the membership detector run at write time. `NULL` tracker = a **sink** (no write-time detection; membership comes only from annotations — this is how discovered signals work).
+- an optional **`tracker`** (a jsonb column, **one per signal**): the membership detector run at write time. A `NULL` tracker means a **sink** — no write-time detection, with membership coming only from annotations. This is how discovered signals work.
```
geclos, line 78: if the lack of trackes is the identifier for auto signals once a signal gets promoted it no longer becomes eligible for auto discovery assignment of scores?

specs/signals.md
```diff
@@ -37,75 +75,44 @@ Two structural decisions carry the whole spec and are stated up front:
...
 
 Constraints:
 
-- **Users cannot create a tracker-less signal.** A `user`-origin signal must have a tracker. Tracker-less (`NULL`) signals exist only as `system`-origin sinks. This deliberately keeps "plain filter slices" out of signals — those stay **saved searches** (a broad `filters`-only signal would write a score per trace; banning user-created tracker-less signals removes that footgun entirely).
-- **One tracker per signal.** A bucket has at most one detector. (A concept that needs two detectors — e.g. semantic *and* judge for "frustration" — is one signal that gets *promoted* from one tracker to another, or two signals; it is not one signal with two trackers.)
+- **Users cannot create a tracker-less signal.** A `user`-origin signal must have a tracker. Tracker-less (`NULL`) signals exist only as `system`-origin sinks. This deliberately keeps "plain filter slices" out of signals — those stay **saved searches**. (A broad `filters`-only signal would write a score per trace; banning user-created tracker-less signals removes that footgun entirely.)
```
geclos, line 84: i think we should explicitely split betwen auto created signals and user signals rather than rely on the lack of tracker as the identification mechanism

specs/signals.md
```diff
@@ -37,75 +75,44 @@ Two structural decisions carry the whole spec and are stated up front:
...
+- **One tracker per signal.** A bucket has at most one detector. A concept that needs two detectors — e.g. semantic *and* judge for "frustration" — is one signal that gets *promoted* from one tracker to another, or two signals. It is never one signal with two trackers.
 
 ### Tracker
 
-**The write-time detector embedded on a signal** (`signals.tracker` jsonb, discriminated by `type`). Three types:
+**The write-time detector embedded on a signal** — `signals.tracker` (jsonb, discriminated by `type`), at most one per signal. Three types: `semantic_similarity`, `script`, and `llm_as_judge`. Their shapes are defined once in [Data model → Shared contracts](#shared-contracts-domainshared-domainscores).
```
geclos, line 89: still missing code based rules tracker, i'd rather do that than script type

dev-docs/signals.md
csansoon: It's the `script` type:

```ts
export type SignalTracker =
 | { type: "semantic_similarity"; semantic: SemanticAnchors; threshold: number }
 | { type: "script";              source: string;            threshold: number }   // sandbox JS
 | { type: "llm_as_judge";        threshold: number }   // detector lives in evaluations.signal_id (1:1)

```

specs/signals.md
csansoon: Yes, should we keep annotating once a "better" tracker has been defined?

specs/signals.md
csansoon: okay i'll add a specific attribute for it

specs/signals.md
csansoon: Okay I'll rename it to `code_based_rules` then (?)
how are they defined tho. Are they not a script? they're like a series of specific rules that can be configured? Or what

PR #3587, opened 2026-06-16
Seed demo projects from derived snapshot

geclos:
## Summary

Adds a disk-backed derived-data snapshot for runtime demo project seeding. New demo projects still run the shared Postgres and ClickHouse base seeders, then import precomputed trace-search, conversation-intelligence, and taxonomy data instead of rerunning the expensive embedding/analyze/gardening pipeline for every onboarded org.

The snapshot is generated from the canonical Acme `default-project` corpus and includes:

- trace-search documents, message embeddings, and trace message occurrences
- session analyses, semantic moments, and moment labels
- taxonomy observations, clusters, lineage, and the completed taxonomy run

The importer rewrites organization/project scope, shifts timestamps by the target seed day, and deterministically remaps globally unique taxonomy IDs through the target `SeedScope`. Trace IDs, session IDs, content hashes, and vectors are copied because the base deterministic ClickHouse seed recreates the same corpus for each demo project.

## Snapshot maintenance

Adds `pnpm --filter @app/workers demo-project-snapshot:export` to regenerate `apps/workflows/src/seed-snapshots/demo-project-derived-v1/` from a local canonical project. The large message-embedding payload is split into multiple compressed parts so every tracked file stays below GitHub's per-file size limit.

`apps/workflows` build now copies `src/seed-snapshots` into `dist/seed-snapshots` so the runtime importer can read the snapshot after bundling.

## Temporal compatibility

`seedDemoProjectWorkflow` keeps the previous trace-search + taxonomy child workflow path behind the existing derived-data patch and only switches to the snapshot path behind a new `seed-demo-project-derived-snapshot-v1` patch, so in-flight workflow histories can replay.

## Validation

- `pnpm --filter @app/workflows test -- src/workflows/seed-demo-project-workflow.test.ts`
- `pnpm --filter @app/workflows build`
- Manual import of the generated snapshot into a temporary project succeeded
- `pnpm --filter @app/workflows typecheck` still fails on pre-existing unrelated `@domain/integrations` notification template errors


geclos:
@claude review

geclos:
@claude review

geclos:
Addressed review feedback in `8e62a3b4d` and `6816521b6`:

- changed the ClickHouse import batch drain to `batch.splice(0)`
- added `exportedAt` to `DemoProjectSnapshotManifest`
- added sync notes for the duplicated ClickHouse table lists
- added replay coverage for workflows that predate the derived-data patch (`postgres` → `clickhouse` only)

Validation run locally:
- `pnpm --filter @app/workflows test -- src/workflows/seed-demo-project-workflow.test.ts`
- `pnpm --filter @app/workflows typecheck`
- `pnpm --filter @app/workers check`

Leaving the non-atomic reset/import note as the documented v1 tradeoff from the PR description.

andresgutgon (approved):

PR #3593, opened 2026-06-16
perf(clickhouse): add session_id to sessions PRIMARY KEY

geclos:
## Why

Loading the session-detail drawer in production takes 1–3+ seconds, worst when opened **through the behaviour-detail drawer**.

Root cause: the `sessions` ClickHouse table is read by `(organization_id, project_id, session_id)`, but its `PRIMARY KEY` stopped at `(organization_id, project_id)` — `session_id` lived only in the `ORDER BY`. So the sparse primary index could only narrow to the project's range and then **scanned every granule in it** to locate the session, decompressing the large ZSTD message-payload columns (`input_messages` / `output_messages` / `system_instructions`) along the way. In the behaviour path this also competes with the heavy taxonomy queries running concurrently.

## What

Promote `session_id` into the `PRIMARY KEY`:

```
-PRIMARY KEY (organization_id, project_id)
+PRIMARY KEY (organization_id, project_id, session_id)
```

This turns the point read into an index seek. Verified on the dev dataset — a point lookup now prunes **granules 1/14 via binary search** instead of scanning all of them. It also aligns `sessions` with every other session-keyed table (`session_analyses` / `taxonomy_observations`, migration `00021`); `sessions` was the only outlier.

No application code changes — the key is physical and the query planner uses it automatically.

## How (migration `00031`)

ClickHouse can't change a populated table's `PRIMARY KEY` in place, so this rebuilds via the established create-new + `RENAME`-swap pattern (see `00013`):

1. Create the new-shape table (identical columns/index/partition/TTL, only the PK line changes).
2. Drop the MV, `RENAME` the live table out and the new one in, recreate the MV against the new table.
3. Backfill historical aggregate states from the frozen legacy table with `INSERT INTO sessions SELECT *` — copies the `AggregateFunction` states directly, **no span re-aggregation** (same single-shard + Replicated propagation pattern as `00027`).

Partitioning (`toYYYYMM(min_start_time)`), the `minmax` index, and the plan-aware + sandbox-hard-retention TTLs are carried forward verbatim. Both `unclustered/` and `clustered/` variants included. Reversible `Down` restores the original key.

### Caveat

The MV is dropped and recreated during the swap, so spans inserted in that sub-second DDL window are materialized by neither path (a negligible, bounded under-count for sessions active during the deploy). Everything before the swap is copied from the frozen legacy table.

## Verification

- Local `up` → `down` → `up` cycle preserves all sessions and span totals (3159 sessions / 81538 spans, byte-identical before/after).
- `EXPLAIN` confirms PK-driven granule pruning (1/14, binary search) on a point lookup.
- MV materializes new span inserts into the rebuilt table.
- chdb test schema (`testkit/schema.sql`) updated to the new key.
- Full `@platform/db-clickhouse` suite green (301 tests).

### DIFF
--- a/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/tool-pills.tsx
+++ b/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/tool-pills.tsx
@@ -2,7 +2,6 @@ import { cn, Tooltip } from "@repo/ui"
 import { formatCount } from "@repo/utils"
 import { Link, useParams } from "@tanstack/react-router"
 import type { ReactNode } from "react"
-import { useHasFeatureFlag } from "../../../../../domains/feature-flags/feature-flags.collection.ts"
 import type { SpanRecord } from "../../../../../domains/spans/spans.functions.ts"
 
 export interface ToolPillStat {
@@ -55,8 +54,7 @@ export function ToolPill({
   readonly tooltip?: ReactNode
 }) {
   const { projectSlug } = useParams({ strict: false })
-  const toolsEnabled = useHasFeatureFlag("tools")
-  const linked = toolsEnabled && typeof projectSlug === "string"
+  const linked = typeof projectSlug === "string"
 
   const pillClass = cn(
     "inline-flex h-7 max-w-full items-center gap-1.5 rounded-full border px-2.5 transition-colors",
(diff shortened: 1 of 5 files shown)


andresgutgon (dismissed):
Looks a heavy migration

geclos:
checked against production db size, the migration will be actually fast (table is small for ch standards)

PR #3595, opened 2026-06-16
feat(telemetry): Google ADK support + join tool calls and results for OpenInference traces

neoxelox:
Adds Google ADK to the supported instrumentations in the Python Telemetry SDK, and fixes a tool call/result pairing bug surfaced while testing ADK traces end to end against a live instance.

## Google ADK telemetry

Registers the OpenInference `openinference-instrumentation-google-adk` (==0.1.15) instrumentor as the `google_adk` instrumentation, bumps the Python SDK to 3.2.0, and adds a runnable weather-agent example (`examples/test_google_adk.py`), a docs page (`docs/telemetry/frameworks/google-adk.mdx`), and the nav entry.

## The pairing bug

In an ADK agent run that calls a tool, the tool call and its result did not join in the conversation view, and the result couldn't navigate to the tool's execution span.

Root cause is upstream in OpenInference: it emits **no message-side ids**. The assistant message's `tool_call` has no `id`, and the tool-result message has no `tool_call_id`. The only real id (`adk-…`) lives on the separate `execute_tool` span (`gen_ai.tool.call.id`). Our ingest minted a synthetic id on the assistant tool_call (`call_2_0`) but left the result's id null, so the two halves carried unrelated ids and couldn't be paired or navigated. rosetta-ai is left untouched — it faithfully translates what it's given — and the canonicalization stays out of the shared `safeStringifyJson`, whose callers depend on pass-through strings and insertion-order keys.

The fix has three parts:

**Ingest-time pairing** (`openinference.ts`). The content parser keeps a document-order queue of pending tool_calls. Each `role:"tool"` message gets its `tool_call_id` from an explicit `message.tool_call_id`, or by pairing to the matching pending tool_call (by name, else FIFO) — minting one shared id for call and result. `ToolCallData` and `ReassembledMessage` gained the id fields this needs.

**Read-time cross-span navigation** (`map-conversation-to-spans.ts`). A conversation tool_call whose id doesn't already resolve to an `execute_tool` span is linked to that span by name + canonicalized args + order, guarded so traces that already carry real ids are untouched. `SpanMessagesData` — and both the trace and session message queries — now select `tool_name` and `tool_input` to support the match. Argument canonicalization uses a new shared `stableStringify` helper (wraps `fast-json-stable-stringify`, the dep `hash` already uses).

**Session UI** (`session-slot.tsx`, session `conversation-tab.tsx`). The session drawer reused the trace `ConversationTab` but never passed `navigateToSpan`, which gates both the message- and tool-call-level "view execution span" affordances. Wired it through for single-trace sessions (where the Spans tab exists), so the session conversation matches the trace drawer.

## Validation

Verified end to end against a local instance: re-ingested an ADK weather-agent run and confirmed via the API that the assistant `tool_call.id` and the tool-result `tool_call_response.id` are now equal and non-null (were `call_2_0` vs `null`). Confirmed in the UI that the tool result joins the call and "view execution span" navigates to the `execute_tool` span, from both the trace and session views. Unit tests cover the parser pairing and the cross-span fallback (`@domain/spans` 653 pass, `@platform/db-clickhouse` span-repository 11 pass); `@repo/utils` + `@domain/spans` + `@app/web` typecheck clean.

### DIFF
--- a/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/session-detail-drawer/conversation-tab.tsx
+++ b/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/session-detail-drawer/conversation-tab.tsx
@@ -275,6 +275,7 @@ export function ConversationTab({
   searchQuery,
   focusMomentKind,
   focusMomentId,
+  navigateToSpan,
 }: {
   readonly projectId: string
   readonly session: SessionDetailRecord
@@ -285,6 +286,8 @@ export function ConversationTab({
   readonly focusMomentKind?: MomentKind | undefined
   /** Scrolls to this semantic moment when no label kind is focused. */
   readonly focusMomentId?: string | undefined
+  /** Navigates to a span in the session's single-trace Spans tab; enables the conversation's span-link affordances. */
+  readonly navigateToSpan?: ((spanId: string) => void) | undefined
 }) {
   const sessionId = session.sessionId
   // Annotations are an LLM-feedback feature — off under a sandbox scope.
@@ -385,6 +388,7 @@ export function ConversationTab({
       scrollContainerRef={scrollContainerRef}
       textSelectionPopoverControlsRef={textSelectionPopoverControlsRef}
       timeline={timeline}
+      {...(navigateToSpan ? { navigateToSpan } : {})}
       {...(labelsByMessageIndex.size > 0 ? { messageTrailingSlot } : {})}
       {...(searchQuery ? { searchQuery } : {})}
     />
--- a/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/session-detail-drawer/session-slot.tsx
+++ b/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/session-detail-drawer/session-slot.tsx
@@ -119,6 +119,12 @@ export function SessionSlot({
     onActiveTabChange("spans")
   }
 
+  function navigateToSpan(spanId: string) {
+    if (!singleTrace) return
+    setSelectedSpanId(spanId)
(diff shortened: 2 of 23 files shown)


apps/web/src/routes/_authenticated/projects/$projectSlug/-components/session-detail-drawer/conversation-tab.tsx
```diff
@@ -385,6 +388,7 @@ export function ConversationTab({
       scrollContainerRef={scrollContainerRef}
       textSelectionPopoverControlsRef={textSelectionPopoverControlsRef}
       timeline={timeline}
+      {...(navigateToSpan ? { navigateToSpan } : {})}
```
geclos, line 391: unrelated change?

geclos (approved):
looks good i think it's including some unrelated stuff changes but otherwise palantir

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/session-detail-drawer/conversation-tab.tsx
neoxelox: its needed to check if the "navigate to tool span" fix works

neoxelox: before we didnt show the action button on sessions because we forgot to add it...

neoxelox:
> looks good i think it's including some unrelated stuff changes but otherwise palantir

the other things are fixes, tool results weren't correctly joined to tool calls

PR #3600, opened 2026-06-17
docs(signals): model signal detectors as evaluations, drop the Tracker concept

csansoon:
Reworks `specs/signals.md` (with coordinated edits to `specs/sandbox-runtime.md`) so a signal's membership detector is an **evaluation**, not a separate "Tracker". This supersedes the interim trackers revision; the spec still describes the same Signals / Monitors / Alerts / Incidents consolidation — only the detector model changes.

## why

The trackers model wrapped three detector "types" (`semantic_similarity`, `script`, `llm_as_judge`) in a `signals.tracker` jsonb union — but a judge already *is* an evaluation, and scripts already run on the QuickJS sandbox runtime (`specs/sandbox-runtime.md`, Phases 0–1 shipped). The wrapper was redundant. Folding detection into the existing `evaluations` table removes a concept and a storage split, and keeps the judge/generation/alignment stack exactly as it ships today.

## the model

A signal's detector is an `evaluations` row, linked 1:1 via `evaluations.signal_id`; a signal with no evaluation is a **sink** (annotation-fed — the unchanged auto-discovery path). An evaluation is **always a script** that runs in the one sandbox engine:

- A judge is just a script that calls `llm()`. An optional `settings` object compiles deterministically to the `script`; a raw script is the advanced path. `script` stays `NOT NULL`.
- There is no `type` or `capability` column. Whether a run is `pure` (deterministic → ClickHouse-only) or `llm` (non-deterministic, costed → Postgres-canonical + ClickHouse) is detected from the script (does it call `llm()`?).
- Membership is the script's `Passed` / `Failed` verdict, stored as `scores.passed` — there is no host-side threshold. `value` is optional confidence for sort/UX.
- Occurrences are `scores WHERE signal_id = ? AND passed`, counted as distinct `trace_id` per signal.

## data model

No new tables. `issues` → `signals` (+ `origin`, `filters`, soft-delete; the detector is no longer a column on the signal). `evaluations` gains `settings`, relaxes `alignment` / `aligned_at` to nullable (judge-only), and gets a partial-unique index enforcing one active evaluation per signal — which needs a dedupe pre-migration for today's multi-evaluation-per-issue rows. `scores.source` → `source_type`, keeping `evaluation` and splitting `annotation` into `flagger` / `user`. ClickHouse `scores` gains `signal_id`.

## scope

Spec only — no code. MVP is Phases 1–4 (rename Issues → Signals; evaluation substrate + script signals; LLM-as-judge; monitors on signals). Semantic similarity is deferred to a future phase (Phase 7) — likely a `similarity()` / `embedding()` host function the script can call, with a possible native batch-runner optimization. The `evaluations.threshold` and per-detector-`type` ideas were considered and dropped during review.


geclos:
> There is no type or capability column. Whether a run is pure (deterministic → ClickHouse-only) or llm (non-deterministic, costed → Postgres-canonical + ClickHouse) is detected from the script (does it call llm()?).

without type you are gonna struggle to easily identify evaluation by its type and you need that to easily show create/edit forms specific for each type (code based evals need a simplified form even if they then compile to a script under the hood)

Also it'd like to be able to aggregate over type of evals (how many evals of type llm are in this project, how many of type code-based, etc..)

I know we then decouple the type from the source of truth (the script code), but it's a worthy tradeoff in this case.

specs/sandbox-runtime.md
```diff
@@ -15,46 +15,48 @@ This spec defines the real **sandbox runtime**: a portable, host-controlled, res
...
 
-One score-shaped contract for every detector. The script reports a **value**; the **host** derives membership from the signal's threshold. There are no dialects — what differs between judgment-bearing detectors and bare-membership detectors is only the persistence policy below.
+One score-shaped contract for every detector. The script reports a **verdict** — `Passed`/`Failed` — and that verdict IS membership; there is no host-side threshold. There are no dialects — what differs between judgment-bearing detectors and bare-membership detectors is only the persistence policy below.
 
-### Output: one score, host-derived membership
+### Output: one verdict, detector-decided membership
 
 ```ts
-// Script return type
-Score(value: number, feedback?: string)   // value ∈ [0, 1]
```
geclos, line 24: why not return the score as well? seems an arbitrary decision

specs/sandbox-runtime.md
```diff
@@ -15,46 +15,48 @@ This spec defines the real **sandbox runtime**: a portable, host-controlled, res
...
+Passed(value?: number, feedback?: string)   // matched
+Failed(value?: number, feedback?: string)   // not matched
+//   value ∈ [0,1] is OPTIONAL confidence for sort/UX (e.g. semantic similarity); it does NOT decide membership.
+//   Score(value, feedback?) is retained as a low-level alias for stored templates; new scripts return Passed/Failed.
 
 // Host-side result
 type RunResult = {
-  value: number        // signal-exhibition strength ∈ [0, 1]
+  passed: boolean      // the detector's verdict — this IS membership (no host-side threshold)
+  value?: number       // optional confidence ∈ [0, 1] for sort/UX
```
geclos: why is this optional

specs/sandbox-runtime.md
```diff
@@ -15,46 +15,48 @@ This spec defines the real **sandbox runtime**: a portable, host-controlled, res
...
 
 | Run | Persists |
 | --- | --- |
-| pure detector, `value ≥ threshold` | occurrence row only (with denormalized `value`) |
-| pure detector, below threshold | nothing |
-| `llm`-capability detector (evaluations included) | score row (value, feedback) via the canonical path, + occurrence when matched |
-| annotation linked to a signal | score row as today, + occurrence |
+| pure evaluation, matched (verdict `Passed`) | ClickHouse-only score row (`passed=true`, optional `value`) |
+| pure evaluation, not matched (verdict `Failed`) | ClickHouse-only score row (`passed=false`) — write-both default |
```
geclos: it's confusing that pure evals generate only clickhouse score and llm evals generate postgres scores. I'd write to the source of truth always (postgres), there should not be any difference in how llm evals treat scores vs other types of evals

specs/sandbox-runtime.md
```diff
@@ -63,20 +65,21 @@ Per the contract already documented in `dev-docs/evaluations.md`:
...
 | `llm(prompt, { schema })` | `llm` | Structured generation through the host (`@domain/ai`); the schema is required — schema-less calls throw in-sandbox; model/provider stay host-managed (`EVALUATION_SCRIPT_RUNTIME_MODEL`); remaining options are host-approved only |
-| `Score(value, feedback?)` | — | The single return type |
-| `Passed(value?, feedback)` / `Failed(value?, feedback)` | — | Sugar over `Score` (`value ?? 1` / `value ?? 0`); keeps stored templates valid |
+| `similarity(...)` / `embedding(...)` | `embeddings` | **FUTURE (`specs/signals.md` → Phase 7)** — semantic similarity as something a script can call. A host bridge mirroring `llm()`; needs an embeddings-ready execution lane (chunk embeddings exist only on the later `trace_search_embeddings` hop, not at trace-end). See the semantic-similarity future below. |
+| `Passed(value?, feedback?)` / `Failed(value?, feedback?)` | — | The return verdict — matched / not matched. This IS membership; `value` is optional confidence; `feedback` is judge-grade only |
+| `Score(value, feedback?)` | — | Low-level alias retained so stored templates stay valid; new scripts return `Passed`/`Failed` |
```
geclos: confuzio are we returning score or not in the script

neoxelox:
i would change the verb of "promoting" a signal to "tracking" a signal i guess

csansoon:
> @geclos 
> without type you are gonna struggle to easily identify evaluation by its type and you need that to easily show create/edit forms specific for each type (code based evals need a simplified form even if they then compile to a script under the hood)

That will be inside the `settings`. Evaluations can be of multiple "types", that just depends on what the script does. Auto-generated signals will have its evaluation created with a raw script, created by GEPA. But users will NOT, as for now, define the script directly. Instead, they will use a UI to define the evaluation settings, which will have types and stuff, and then we use that to create the actual script.

geclos:
> That will be inside the settings

is settings a jsonb? will i be able to query at scale e.g count of evaluations of type "semantic" vs evaluations of type "llmasjudge"?

specs/sandbox-runtime.md
csansoon: Good call — changed. The script now returns all three: a normalized `value` (0–1 score), `passed` (the membership verdict), and optional `feedback`. It's no longer a bare verdict. Fixed in 237f38bd7.

specs/sandbox-runtime.md
csansoon: Made it required. `value` (the normalized 0–1 score) is always returned now, alongside `passed` and optional `feedback`. 237f38bd7.

specs/sandbox-runtime.md
csansoon: Clarified — yes, always. The contract is `{ value (0–1 score), passed, feedback? }`, all returned. 237f38bd7.

specs/sandbox-runtime.md
csansoon: Agreed — dropped the split. All scores now write to both Postgres (canonical) and ClickHouse (analytics), uniformly; the only nuance is the existing draft lifecycle (a drafted annotation stays Postgres-only until published, then syncs). No per-type/per-capability difference. The CH-only path is kept only as a documented future scale lever for high-volume deterministic scripts (post-MVP), never a rule about eval kinds. 237f38bd7.

csansoon:
@geclos following up on the type point — the spec now spells out the distinction. A single script can combine `llm()` + semantic + deterministic checks, so an arbitrary script has no one "type". Type is a property of the **settings template** (the form a user picks; it drives the create/edit UI) and lives in `settings`; raw/GEPA scripts are custom. Your "how many llm vs code-based evals" is a separate, **multi-valued capability** question — answered by inspecting what each script actually does, which also covers raw/GEPA scripts that a single `type` column couldn't classify. Both axes are now in the spec. (237f38bd7)

csansoon:
@neoxelox renamed — "promote" → "track" a signal with an evaluation (`trackSignalUseCase`, `POST /{signalSlug}/track`, "Start tracking" in the UI, and the section is now "Discovery and tracking"). (237f38bd7)

geclos (approved):

PR #3601, opened 2026-06-17
feat(destinations): per-source config, status & delivery preview

andresgutgon:
## What

Builds on the merged multi-source engine (#3594) to make each source a first-class, configurable unit and to show users **what actually gets sent**.

- **`destination_sources` is now the full per-source record** — not just a cursor. It carries per-source **config** (`excludePayloads`, `maxRecordsPerRun`), an enable/disable **status**, and the sync cursor. Payload settings moved off the destination config into per-source config (`maxSpansPerRun` → `maxRecordsPerRun`); the destination config keeps `host` + `intervalMs`.
- **Per-kind capability** — `DESTINATION_KIND_META[kind].supportedSources` (`posthog → spans`) gates which sources a kind can export (semantic: PostHog is span-granular). The mapper is now **generic over the source record type** and resolved per `(kind, source)`.
- **Delivery preview** — a read-only "what gets sent" surface: sample the source's latest record → map with the candidate per-source config → return the events. Surfaced in the settings form as a **collapsible per-source preview** that re-fetches as you toggle exclude-payloads / change host (or shows "No data yet").
- **Atomic save** — per-source config saves in the same transaction as the destination (`updateDestination` takes `sourceConfigs`), so the edit form never half-applies.
- **Naming fix** — the misnamed "redaction" is renamed to **payload exclusion** (field exclusion, not PII scrubbing) across code, spec, and UI.
- Exposes `CollapsibleBlock` from `@repo/ui` (reusing the existing collapsible pattern).

## Migration

Additive (rename detected by drizzle): `destination_source_cursors → destination_sources` (+ index + RLS policy) and adds the `status`/`config` columns. The original production migration is untouched.

## UI

The PostHog form gains a **Sources** section — currently one card (spans) with its exclude-payloads toggle + the collapsible preview. Per-source enable/disable switches arrive once a kind supports ≥2 sources (the data model already supports it via `status`).

## Verification

- `pnpm typecheck` 82/82, `knip` clean, Biome clean
- Tests: domain destinations 80, spans 649, platform destinations 21 (against the real migration chain), data-destinations 60, clickhouse span 11

Refs LAT-684.


andresgutgon:
@claude review

andresgutgon:
On the 🔴 migration finding (`ADD COLUMN "config" jsonb NOT NULL` with no default): **not an issue here — the table has no rows in any deployed environment.** `destination_sources` (renamed from `destination_source_cursors`) only ever gets rows when a destination is created, and the **destinations feature is flag-gated and has never been enabled in staging or production**, so the table is empty. `ADD COLUMN … NOT NULL` succeeds on an empty table.

The general advice is correct for a populated table (`ADD … DEFAULT … ` then `DROP DEFAULT`, or nullable → backfill → `SET NOT NULL`) and we'll apply it when a table actually has rows. Here it's moot, so leaving the migration as-is.

andresgutgon:
Worked through the review (commit `42e02c2f8`). Summary of what changed and what didn't:

### Addressed

- **🟠 Per-source config replaced, not merged** — fixed. Added `destinationSourceConfigPatchSchema` (source-discriminated, no defaults) and `updateDestination` now merges each patch onto the stored source config inside the transaction, exactly like the destination-level config merge. So `maxRecordsPerRun` (and any future non-UI per-source field) survives a save instead of being reset to the Zod default. Create still seeds full config.
- **🟠 Unsafe `sourceConfig as SpansSourceConfig` cast** — fixed. The PostHog mapper now narrows on `sourceConfig.source` and `Effect.die`s on a non-spans config, so a future miswiring is a loud defect rather than a silent mismap. No cast.
- **🟠 N+1 in `listDestinations`** — added a comment documenting the per-destination source-state read and the `listByProjectId` batch as the escape hatch if counts grow. Left as N+1 (a handful of destinations per project today).
- **🟡 Missing `sourceConfigs` test** — added a test asserting a `{ source, excludePayloads }` patch merges onto the stored config with `maxRecordsPerRun` preserved (pairs with the existing destination-config preservation test).
- **🟡 Preview swallowed the error** — now surfaces the cause (`401`/schema/transport) in dev via `import.meta.env.DEV`, generic message in prod.
- **🟡 `excludePayloads` under `config.*`** — added an inline comment explaining it's a per-source setting kept there for form ergonomics and routed via `buildSourceConfigs`.

### Not addressed (with reason)

- **🔴 Migration `ADD COLUMN config NOT NULL`** — intentionally left as-is. The table has **no rows in any deployed environment** (destinations is flag-gated and never enabled in staging/prod), so `ADD COLUMN … NOT NULL` succeeds. The `DEFAULT … / DROP DEFAULT` pattern is correct for a populated table and we'll use it when one has rows. (Detail in the earlier comment.)

Verification: workspace typecheck clean, knip clean, domain destinations 82 tests + web typecheck green.

PR #3602, opened 2026-06-17
feat(web): group project sidebar into Observe / Understand / Refine

csansoon:
## what this does

Groups the project sidebar's navigation items into three labeled sections instead of one flat list:

```
OBSERVE        UNDERSTAND      REFINE
  Sessions       Issues          Monitors
  Users          Behaviors       Datasets
  Tools
```

The grouping follows what each item *is* relative to the user's data: **Observe** is the raw data you sent us (sessions, users, tools), **Understand** is what Latitude infers from it (issues, behaviors), and **Refine** is the tooling you set up on top to check and improve quality (monitors, datasets). Read top-to-bottom it's a funnel: your data → what we found in it → what you build on top.

## why

The previous order interleaved raw data, inferences, and user-created objects with no visible logic, so the list read as undifferentiated. The labeled groups make the three-part mental model explicit.

## how it works

`project-sections.ts` gains a `SectionGroupKey` (`observe` | `understand` | `refine`), each section is tagged with its `group`, and `PROJECT_SECTION_GROUPS` defines the group order and labels. A new `useVisibleProjectSectionGroups()` hook buckets the flag-filtered sections into their groups and drops any group left empty.

`useVisibleProjectSections()` stays a flat list (the command palette still consumes it, now in the new order), so this adds the grouped view without changing the existing one.

In the sidebar (`$projectSlug.tsx`), expanded mode renders a small uppercase muted header per group; collapsed (icon-only) mode replaces headers with a thin divider between groups so the grouping still reads without text.

## notes

Feature flags can shrink a group — with all flags off, **Refine** shows only Datasets under its own header. It never goes fully empty (Observe always has Sessions+Users, Understand always has Issues), but a one-item group is possible. The label names (Observe/Understand/Refine) went through a few iterations; open to swaps.


andresgutgon (dismissed):
Yes, before gerard notice

andresgutgon (approved):

andresgutgon:
Upps, se me ha resbalado el dedo sorry

PR #3607, opened 2026-06-17
feat(destinations): backfill — engine, worker chain, server-fn, UI, re-enable gap (LAT-681)

andresgutgon:
Adds **user-initiated historical backfill** for Data destinations (LAT-681/682): import a project's past traces into a destination (PostHog) without re-sending what live sync already owns.

## What it does
- **From the settings UI** — "Import history" on create and an on-demand **Backfill** button, with live progress and Cancel.
- **Coverage-bounded, paid once** — each source tracks how far back it's covered; a backfill fills `[start, coverage_start)` and stops, so it never overlaps live sync and re-triggers are no-ops. Reach is the org's plan retention (Free 30d / Pro 90d / Ent 365d), resolved and clamped on the backend — the UI never sees a retention number.
- **Paced & safe** — a self-advancing chain (one window at a time) on its **own queue** at low concurrency, so a big backfill can never starve live sync. Idempotent via deterministic event UUIDs.
- **Engine stays generic** — vendor mechanics (PostHog `historical_migration`, host rules) live in the adapter; billing lives behind a port.

## Notable behavior
- A backfill **never quarantines** the destination; a terminal failure writes a `failed` run row (visible in history) instead.
- A **credentials/host change re-opens** the backfill range (prior deliveries may have gone elsewhere).
- Backfilled events carry the span's original timestamp, so they land at their real date in PostHog — not "today".

## Where to review
- `packages/domain/destinations/src/use-cases/backfill-destination.ts` — window state machine + coverage bounding.
- `apps/workers/src/workers/destinations.ts` — the backfill lane.
- `apps/web/src/domains/destinations/destinations.functions.ts` + the destination card — server fns + UI.

Full design + open questions: `specs/data-destinations.md`.


andresgutgon:
@claude review please take into consideration spec documentation to balance the decisions

PR #3611, opened 2026-06-18
perf(clickhouse): promote trace_id into the traces PRIMARY KEY

geclos:
## Why

Session/trace **conversation loading is slow in production**. The dominant cost is `findByTraceId` — a point read on the `traces` rollup by `(organization_id, project_id, trace_id)` that fires on **every** trace/session drawer open and again when the conversation tab builds its span maps.

The rollup's `PRIMARY KEY` stopped at `(organization_id, project_id)` while `trace_id` lived only in the `ORDER BY`. So the sparse primary index could only narrow to the project's range, then **scanned every granule in it** — decompressing the large `ZSTD` message-payload columns along the way — to find a single trace. Seconds, for a project with history.

## What

Promote `trace_id` into the primary key: `PRIMARY KEY (organization_id, project_id, trace_id)` (the `ORDER BY` is unchanged). The index then seeks straight to the trace's granule(s).

ClickHouse can't change a populated table's `PRIMARY KEY` in place, so this rebuilds via the **create-new + `RENAME`-swap** pattern, mirroring **`00034`** which did exactly this for `sessions` (`session_id`). The full current schema (through `00032`), the `minmax` index, partitioning, the plan-aware + sandbox-hard-retention TTLs, and the materialized view are all carried forward **verbatim** — only the `PRIMARY KEY` line changes. Historical aggregate states are copied with `INSERT … SELECT *` (no span re-aggregation).

Adds `00037_traces_trace_id_primary_key.sql` in both `unclustered/` and `clustered/`.

## Risk / review notes

- **Production rollup rebuild.** It copies the whole `traces` table; run it in staging first and watch the swap window (spans inserted during the sub-second `RENAME` are captured by neither path — everything before is copied from the frozen legacy table; same trade-off as `00034`).
- The `INSERT … SELECT *` relies on `traces_v2` being byte-identical in column order/types to the live table. The column set was reconstructed from the migration chain and **cross-checked against the `00032` MV write-order**; full CI runs migrations against a real ClickHouse, so any mismatch fails loudly. Worth also diffing against `SHOW CREATE TABLE traces` on a migrated instance.
- Draft pending a staging migration run.

## Related

Complemented by the app-side cleanup in the conversation span-maps PR (removes the duplicate `findByTraceId`), which is independent of this one.

### DIFF
--- a/packages/platform/db-clickhouse/clickhouse/.migration-lock
+++ b/packages/platform/db-clickhouse/clickhouse/.migration-lock
@@ -4,6 +4,6 @@
 # create migrations in parallel, preventing incompatible
 # migrations from being merged without conflict resolution.
 #
-# Generated: 2026-06-17T14:18:41.302Z
-# Hash: 7b6b8e91-055f-4a4e-978e-3eb3caeaf316
+# Generated: 2026-06-18T09:02:47.853Z
+# Hash: a98b6156-4fdc-40df-8fa5-120ffd2a0140
 # Do not manually edit this file.
--- a/packages/platform/db-clickhouse/clickhouse/migrations/clustered/00037_traces_trace_id_primary_key.sql
+++ b/packages/platform/db-clickhouse/clickhouse/migrations/clustered/00037_traces_trace_id_primary_key.sql
@@ -1,0 +1,246 @@
+-- +goose NO TRANSACTION
+-- +goose Up
+
+-- Promote `trace_id` into the PRIMARY KEY of `traces` — clustered variant.
+-- See unclustered/00037_traces_trace_id_primary_key.sql for the rationale.
+--
+-- The trace/session drawer reads by (organization_id, project_id, trace_id) via
+-- findByTraceId, but the PK stopped at (organization_id, project_id) while
+-- trace_id lived only in the ORDER BY, so a point read scanned every granule in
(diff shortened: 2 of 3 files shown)


geclos:
### Schema parity verified against live `traces` (Latitude V2 staging)

Diffed the `traces_v2` definition in this migration against the live table on staging (`system.columns` / `system.tables`):

- **All 37 columns match** — 35 stored + the 2 `ALIAS` columns (`time_to_first_token_ns`, `duration_ns`) — on name, position, type, codec, and default.
- `sorting_key` (`organization_id, project_id, trace_id`) and `partition_key` (`toYYYYMM(min_start_time)`) are unchanged; the only delta is the intended `primary_key` → `(organization_id, project_id, trace_id)`.
- So `INSERT INTO traces SELECT * FROM traces_legacy` carries the aggregate states verbatim (the 2 ALIAS columns are excluded by `*` on both sides).

Incidental: the live engine reports `SharedAggregatingMergeTree` (Cloud's managed substitution for `Replicated*`/`MergeTree`), so both the `clustered` (`ReplicatedAggregatingMergeTree`) and `unclustered` (`AggregatingMergeTree`) variants resolve to the same Shared engine the current table uses.

Still recommend running the migration itself on staging to exercise the DDL / `RENAME` swap / MV recreation before promoting.

PR #3612, opened 2026-06-18
perf(web): build conversation span maps client-side, drop redundant trace fetch

geclos:
## Why

The conversation/session span-map server functions (`mapConversationToSpans`, `mapSessionConversationToSpans`) re-fetched the **full trace detail** via `findByTraceId` — decompressing the large `ZSTD` message columns — purely to obtain `allMessages`. But the client already holds `allMessages` from its own `useTraceDetail`. So every conversation view did the trace point lookup twice.

There was also a **latent correctness gap**: `messageSpanMap` is keyed by index into `allMessages`, yet the map was built on the server against a *second, independent* trace fetch. If that diverged from the copy the client renders (e.g. late-arriving spans changing the rollup aggregation between the two calls), messages would link to the wrong spans.

## What

- The server functions now return just the **message-bearing spans** (`SpanMessagesRecord[]`) and no longer call `findByTraceId`. Renamed to `listConversationMessageSpans` / `listSessionConversationMessageSpans` to match. The trace variant takes `startTime` from the client to bound the span scan; the session variant already received its window from the client.
- The hooks (`useConversationSpanMaps` / `useSessionConversationSpanMaps`) build the maps **client-side** via `buildConversationSpanMaps(allMessages, spans)`, keeping the same `{ messageSpanMap, toolCallSpanMap }` return shape — so downstream consumers are unchanged apart from passing `allMessages` (+ `startTime`).

Maps are now built against the **exact `allMessages` the client renders**, removing the duplicate point lookup *and* closing the index-misalignment gap.

## Notes

- `buildConversationSpanMaps` and `SpanMessagesData` are exported from `@domain/spans`'s **browser** entry, so importing them client-side is safe (no server/Effect code pulled into the bundle).
- Independent of the `traces` PRIMARY KEY migration PR, though they target the same slow path; the migration makes the remaining single `findByTraceId` (in `useTraceDetail`) cheap, and this removes the duplicate.
- Draft pending CI typecheck.

### DIFF
--- a/apps/web/src/domains/spans/spans.collection.ts
+++ b/apps/web/src/domains/spans/spans.collection.ts
@@ -1,3 +1,4 @@
+import { SpanId } from "@domain/shared"
 import type { SpanMessagesData } from "@domain/spans"
 import { buildConversationSpanMaps } from "@domain/spans"
 import { queryCollectionOptions } from "@tanstack/query-db-collection"
@@ -168,7 +169,12 @@ export const useSpanDetail = ({
 }
 
 const asMessageSpans = (records: readonly SpanMessagesRecord[]): readonly SpanMessagesData[] =>
-  records as unknown as readonly SpanMessagesData[]
+  records.map((record) => ({
+    ...record,
+    spanId: SpanId(record.spanId),
+    inputMessages: record.inputMessages as readonly GenAIMessage[],
+    outputMessages: record.outputMessages as readonly GenAIMessage[],
+  }))
 
 export function useSessionConversationSpanMaps({
   projectId,
@@ -210,7 +216,8 @@ export function useSessionConversationSpanMaps({
       })
       return buildConversationSpanMaps(allMessages ?? [], asMessageSpans(spans))
     },
-    enabled: enabled && projectId.length > 0 && sessionId.length > 0 && allMessages !== undefined,
+    enabled:
+      enabled && projectId.length > 0 && sessionId.length > 0 && latestTraceId.length > 0 && allMessages !== undefined,
   })
 }
 
--- a/apps/web/src/domains/spans/spans.collection.ts
+++ b/apps/web/src/domains/spans/spans.collection.ts
@@ -210,8 +210,7 @@ export function useSessionConversationSpanMaps({
       })
       return buildConversationSpanMaps(allMessages ?? [], asMessageSpans(spans))
     },
-    enabled:
-      enabled && projectId.length > 0 && sessionId.length > 0 && latestTraceId.length > 0 && allMessages !== undefined,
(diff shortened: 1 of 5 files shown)


apps/web/src/domains/spans/spans.collection.ts
```diff
@@ -189,35 +198,51 @@ export function useSessionConversationSpanMaps({
...
           sessionStartTime,
           sessionEndTime,
         },
-      }),
-    enabled: enabled && projectId.length > 0 && sessionId.length > 0 && latestTraceId.length > 0,
+      })
+      return buildConversationSpanMaps(allMessages ?? [], asMessageSpans(spans))
+    },
+    enabled:
+      enabled && projectId.length > 0 && sessionId.length > 0 && latestTraceId.length > 0 && allMessages !== undefined,
```
geclos: is latestTraceId check needed anymore since we are not using it?

geclos: Good catch. Dropped the `latestTraceId.length > 0` check from the `enabled` guard — `allMessages !== undefined` already implies the latest trace's detail loaded, so it was redundant. Kept `latestTraceId` in the queryKey though, since `allMessages` is derived from that trace and the cache needs to vary by it. Fixed in 0e0fb351d.

apps/web/src/domains/spans/spans.collection.ts
```diff
@@ -163,19 +167,24 @@ export const useSpanDetail = ({
   })
 }
 
+const asMessageSpans = (records: readonly SpanMessagesRecord[]): readonly SpanMessagesData[] =>
+  records as unknown as readonly SpanMessagesData[]
```
geclos: can we avoid escaping typecheck with `as unknown`?

apps/web/src/domains/spans/spans.collection.ts
geclos: Restored `latestTraceId.length > 0` in the enabled guard for symmetry with the queryKey — makes it explicit that the trace id is a real input. d5948a6a8

apps/web/src/domains/spans/spans.collection.ts
geclos: Done in d5948a6a8. The blanket `as unknown` was bridging two real gaps: `spanId` is a plain `string` on the wire record vs the branded `SpanId` in the domain type, and the message arrays are typed `object[]` on the record (they can't be `GenAIMessage[]` there — that index signature breaks TanStack's serialized return type). Replaced it with a `.map` that brands `spanId` properly and narrows the arrays with targeted casts (`object[]` → `GenAIMessage[]` is a legal assertion since GenAIMessage is a subtype of object, no `unknown` hop). So the only remaining casts are the unavoidable serialized↔domain narrowing, scoped to the two fields that genuinely differ.

PR #3618, opened 2026-06-18
feat(destinations): rate-limit backoff, idle auto-pause, and backfill record cap

andresgutgon:
Three related reliability changes to the data-destinations engine.

### 1. Back off on rate limits instead of quarantining (LAT-685)

A PostHog 429 means "slow down," not "broken." Delivery errors now carry a generic, destination-agnostic reason category, and a `rate_limited` failure backs off and retries (cursor untouched) without counting toward quarantine. Quarantine stays for genuine config faults (bad key, dead host).

### 2. Auto-pause idle destinations

After ~7 days of continuous empty runs, a destination is auto-paused so the sweep stops probing an abandoned project forever. No data is lost — resume is manual and its gap-backfill catches up within retention.

### 3. Cap a backfill to the most recent 1M records

Connecting a destination to a months-old project, or resuming one paused for months, could enqueue an unbounded historical import. Each backfill is now capped at **1,000,000 records, keeping the most recent ones** — the initiator raises the backfill's lower bound to the cap floor, so the existing oldest-first chain runs over just the newest-1M slice. Older history isn't auto-imported.

This is a deliberate **operational/product bound** (backfill duration + ClickHouse read volume), **not** a PostHog limit: PostHog documents no rate limit on `/batch/` capture (only a 20 MB body cap), so rate-safety is already handled by chunking + the 429 backoff. 1M ≈ 20 live-sync windows.


andresgutgon:
@claude review please

PR #3621, opened 2026-06-19
feat(signals): Phase 2 PR1 — evaluation engine cutover (scores mark behavior present; unified signals:match worker)

csansoon:
Phase 2 makes Signals something users configure themselves: you define a behavior you want to track, and an evaluation script decides, per trace, whether that behavior occurred. This is **PR1 of 3**, and it's internal plumbing only — it changes how the system records "did this behavior happen in this trace" and which worker runs evaluations. There's no user-facing way to create a Signal yet (PR2 adds the API/MCP/backfill; PR3 adds the builder UI). Full design in `specs/signals.md`.

## background: what a score's `passed` means

Every evaluation run writes a **score** row that carries a `passed` boolean. Until now the system was a *problem detector*: a score recorded a problem, and `passed = false` meant "this trace exhibits the problem." This PR re-frames scores around **occurrences** — a trace where the tracked behavior actually happened — and flips what `passed` means:

> **`passed = true` now means "the behavior is present in this trace" (an occurrence).**

That is the opposite of before, and it's now consistent for every kind of score: LLM-as-judge evaluations, automatic flaggers, and human annotations. A Signal's occurrences are just its scores with `signal_id = <that signal>` and `passed = true`, counted as distinct traces.

## why this ships as one deploy, not behind a feature flag

This doesn't only add code — it **changes the meaning of the `passed` value on every score already in the database**, and a migration rewrites those existing rows to match. A feature flag would require old code (which reads `passed = false` as "present") and new code (which reads `passed = true` as "present") to both run correctly against the same rows at the same time, which isn't possible once the meaning is flipped for everyone at once. So it ships as a single coordinated deploy with brief, accepted downtime — the same trade-off the Phase 1 Issues→Signals rename made.

## handling judges written before the cutover

An LLM-as-judge is a stored prompt. A judge written before this PR still prompts the model with the old "detect the problem" framing, so its raw output is in the old meaning of `passed`. Rather than regenerate every existing judge, each evaluation now has an `evaluations.legacy_polarity` flag, set to `true` for all pre-existing rows. When a flagged judge runs, its verdict is inverted at execution time so the score it writes lands in the new meaning. New and re-generated judges are written in the new framing with the flag `false`, so the flag clears itself out as judges get re-optimized over time. (There's a TODO to delete the flag once none remain.)

## the script decides `passed` (no host-side threshold)

Previously the host compared a numeric score against a threshold (`isScoreMatch`) to decide pass/fail. That's removed: the sandbox evaluation script returns the verdict directly via `Passed()` / `Failed()`, and `passed` is now a required field on the script's return value.

## one worker runs every evaluation: `signals:match`

A new worker, `signals:match`, becomes the single place evaluations are scheduled. It runs when traces finish (off `TracesIngested`, skipped for sandbox traces, debounced like the existing trace-end step), picks which evaluations apply to a trace (the existing sampling / turn / filter rules), and hands them to the existing `live-evaluations:execute` queue — the execution path itself is unchanged.

`trace-end` used to do this evaluation fan-out inline; it no longer does. It keeps everything else it owned: live-queue materialization, deterministic flaggers, saved-search monitors, trace-search refresh, and conversation intelligence. Evaluation scores are now written with their `signal_id` set whether or not the behavior was found; readers that want only occurrences filter on `passed = true`.

## renamed the score `source` column to `source_type` (Postgres only)

`scores.source` is renamed to `source_type` in Postgres and in the domain `Score` type (an in-place `RENAME COLUMN`). Two places intentionally keep the name `source`:
- the **ClickHouse** column — it's part of the table's sort key, so renaming it would force a full table rebuild;
- the **public `/scores` API field** — renaming it would break the SDK, so it's mapped to `source_type` internally instead.

The values (`evaluation` / `annotation` / `custom`) don't change.

## data migrations (run once, before the new code deploys)

- set `evaluations.legacy_polarity = true` on all existing evaluations;
- flip `passed` on all historical `evaluation` and `annotation` scores (Postgres `UPDATE`, ClickHouse `ALTER … UPDATE`), skipping errored rows;
- rebuild the hourly-occurrence rollup (the `scores_hourly_buckets` materialized view) so it counts only `passed = true` rows — the flip runs synchronously first so the rebuild sees the corrected data;
- update the discovery work-queue index to match `passed = true` rows.

Historical scores that weren't occurrences keep their `signal_id` empty (we don't back-fill it), so per-Signal totals from before the cutover are slightly under-counted. Accepted.

## follow-on fix: readers that want only occurrences

Because every evaluation score now carries a `signal_id` (not just the matching ones), code that lists "a Signal's scores" suddenly also saw the non-occurrences. Added an optional `passed` filter to `ScoreRepository.listBySignalId` and applied `passed = true` at the four readers that want occurrences only — Signal name/description generation, the two incident-email sample excerpts, and the web occurrences carousel — plus `listFlaggerSlugsBySignalId`.

Also repaired ~36 unit-test fixtures that the rename and the `passed` flip left stale (they still set the old `source` field or the old `passed` value). Only fixtures that build a domain `Score` were touched; ClickHouse-row, API-wire, and Signal-entity `source` fields were left as-is.

## validation

`@domain/signals` (191 tests), `@domain/notifications` (57), `@domain/scores` (31), `@domain/annotations` (68), and `@platform/db-postgres` (355) pass and typecheck clean. The ClickHouse-backed tests and the data migrations were validated by mirroring the known-good prior migrations and running a local `db:reset` (the ClickHouse `chdb` addon isn't built locally).

## deliberately out of scope (follow-ups)

- `evaluations.script_hash` → PR2 (only needed once the `alignment` column becomes nullable).
- User-created Signals — CRUD API + MCP tools + the settings→script compiler + historical backfill → PR2.
- The builder UI and dry-run preview → PR3.


csansoon:
@claude review this PR

csansoon:
Thanks for the review. Addressed the blocker + responses to the notes:

**🔴 Blocker — CH `00037` collision — fixed in `9245aa724`.** Confirmed: the `development` merge added `00037_traces_trace_id_primary_key.sql`, colliding with this PR’s `00037_signals_flip_historical_score_polarity.sql` in both `clustered/` and `unclustered/` (`create.sh`’s `fail_if_duplicate_versions` would abort the run). Renumbered the signals migrations preserving flip-before-rebuild order: flip → `00038`, rebuild → `00039`; updated the in-file cross-reference comments and regenerated `.migration-lock`. Duplicate-version check now passes in both dirs.

**🟡 Note 1 (PG `…103126` index churn)** — agreed, and an independent review flagged the same thing. It’s pre-deploy/transient so it’s safe; leaving the two-step as-is rather than rewriting migration history, but noted for future authors.

**🟡 Note 2 (`1 - value` inversion)** — acceptable for the temporary legacy bridge; `value` is display/sort-only under the new contract. Will fold a reminder into the `TODO(signals-legacy-polarity)` cleanup to revisit value-based display when the bridge is removed.

**🟡 Note 3 (TODO has no tracking issue)** — fair; will link a tracking issue to the `legacyPolarity` TODO.

**🟡 Note 4 (`signalId: null` in `ScoreCreated` → no-op discovery job)** — correct, not a bug; the dedupe + ownership short-circuit make it harmless. Noted as an efficiency item for when discovery volume matters.

apps/workers/src/workers/domain-events.ts
```diff
@@ -113,6 +113,31 @@ export const createDomainEventsWorker = ({
               },
             ),
           ),
+          // The signals matching pipeline runs every active evaluation against the trace. Gated on
+          // `!isSandbox` (sandboxes don't surface signals); debounced like trace-end so evaluations
+          // fire on the settled trace.
+          ...(isSandbox
```
andresgutgon: Don't do this. Events are sent unconditionally. Is the consumer that ignore when isSandbox is true. You're already sending `isSandbox`

apps/workers/src/workers/signals-match.test.ts
```diff
@@ -320,8 +320,8 @@ describe("runSignalsMatchJob", () => {
         sampling: 0,
       }),
     ])
-    const { source_type, ...scoreRow } = makeScoreRow({ id: "z".repeat(24), evaluationId: "f".repeat(24) })
-    await pg.db.insert(scores).values([{ ...scoreRow, sourceType: source_type }])
+    const { sourceType, ...scoreRow } = makeScoreRow({ id: "z".repeat(24), evaluationId: "f".repeat(24) })
+    await pg.db.insert(scores).values([{ ...scoreRow, sourceType: sourceType }])
```
csansoon, line 324: This is no longer needed, we can youst add `scoreRow` directly.

apps/workers/src/workers/domain-events.ts
csansoon: Good call — done in `7d9dc0e39`. Removed the publisher-side gate so `signals:match` is fanned out unconditionally with `isSandbox` in the payload (same as `trace-end`); the signals-match worker already early-returns on `isSandbox`. Also fixed the test that asserted the old gating.

andresgutgon (changes requested):
Merge drizzle migrations 🙏 this is crazy
image

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00038_signals_flip_historical_score_polarity.sql
```diff
@@ -0,0 +1,19 @@
...
+
+-- Signals cutover: a signal's occurrences are now its matching scores (passed = true = the behavior
+-- is present in the trace). Existing evaluation + annotation scores were written under the old
+-- problem-detector polarity (passed = false = exhibits), so flip them once to the new convention.
+-- Errored rows are left alone: errored is never an occurrence (passed stays false). Mirrors the
+-- Postgres flip in drizzle/20260618140517_signals-flip-historical-score-polarity. The CH column is
+-- still named `source` (the rename to source_type is Postgres-only; CH keeps `source` per the
+-- sort-key constraint). mutations_sync = 1 makes the flip complete before the next migration (00039)
+-- rebuilds scores_hourly_buckets, so that rebuild re-aggregates the already-flipped rows.
+ALTER TABLE scores UPDATE passed = NOT passed
```
geclos, line 12: is it worth all this migration/effort to change the meaning of passed? couldn't have we just gone with the meaning we already established before? What happens with existing evaluation monitors that are based in the previous meaning of passed?

geclos:
> Every evaluation run writes a score row that carries a passed boolean. Until now the system was a problem detector: a score recorded a problem, and passed = false meant "this trace exhibits the problem." This PR re-frames scores around occurrences — a trace where the tracked behavior actually happened — and flips what passed means:

> passed = true now means "the behavior is present in this trace" (an occurrence).

> That is the opposite of before, and it's now consistent for every kind of score: LLM-as-judge evaluations, automatic flaggers, and human annotations. A Signal's occurrences are just its scores with signal_id = <that signal> and passed = true, counted as distinct traces.


This seems like an arbitrary decision (just as arbitrary as deciding that evaluations detect problems rather than ocurrences), I'm unconvinced it's worth the pain of this PR just to switch the meaning of "passed"

packages/platform/db-clickhouse/clickhouse/migrations/unclustered/00038_signals_flip_historical_score_polarity.sql
csansoon: This is on purpose.

Until now, we were only detecting issues, so the evaluations FAILED when the issue was detected.
However, now evaluations are used to keep track of signals, and they can be both "good" or "bad", or even completely neutral. Evaluations are just checking whether a trace is an occurrance of the tracked signal. For tracking a positive signal (e.g.: "User Satisfaction"), it would be really weird that the script needs to return "Failed" when the signal has been found...
So, from now on, evaluations are specifically script that look for signals, and "Pass" when the signal has been found, even if that's something bad.

geclos:
@csansoon talked offline ok with this direction, please review table sizes in production and make sure the migration can be run.

andresgutgon (approved):

PR #3629, opened 2026-06-19
fix(destinations): late-materialize the window read to bound memory by page

andresgutgon:
## Problem

The backfill/sync window read (`listByIngestedAtWindow`) dedups the whole `(cursor, windowEnd]` window with `LIMIT 1 BY span_id`. `ingested_at` isn't in the `spans` sort key, so the cursor can't prune — **every read scans all of the org/project's rows**. The query selected the wide payload columns (input/output messages, tool defs/IO) *inside* that dedup, so the sort materialized the whole window's payloads.

Result on a payload-heavy project: **~1.04 GiB read, ~3.5 GiB peak memory** per window — *regardless of the page `LIMIT`*. So it kept tripping the per-query 4 GiB guard added in #3623 and stalling the chain ([Datadog issue `736c6ee4-6be6-11f1-b5f3-da7ad0900005`](https://app.datadoghq.eu/error-tracking/issues/736c6ee4-6be6-11f1-b5f3-da7ad0900005)), and lowering the page 5k→2k did nothing because **memory scaled with the window, not the page** (`query_log` showed `read_rows` 41k–70k no matter the page).

## Fix — late materialization

Two-phase within one query:
1. An inner `IN` subquery dedups the window reading **only `span_id` / `ingested_at`** → the sort stays in the low-MiB range even over the full window.
2. The outer query reads the wide payload columns for just the **≤`limit` winning spans**.

Same dedup/cursor/ordering semantics (latest `ingested_at` per span, keyset-resumable, `windowEnd`-scoped) — only the column-read volume changes.

## Verified against the failing project in prod (`query_log`)

| | Before | After (5k page, full window) |
|---|---|---|
| read bytes | 1.04 GiB | **69 MiB** |
| peak memory | ~3.5 GiB | **246 MiB** |
| duration | 1.5–3.3 s (then OOM) | 0.29 s |

Memory is now bounded by the **page**, not the window — so the existing **5,000** page is comfortable (the 4 GiB `max_memory_usage` guard stays as defense-in-depth).

## Notes

- Supersedes the page-size lever in #3628: with this fix the 5k page is safe, so #3628's 5k→2k change is no longer needed (its runs-table height fix is unrelated and can stand on its own).
- Tests: `@platform/db-clickhouse` 314 pass (incl. the `listByIngestedAtWindow` dedup / same-ms cursor-resume / windowEnd-scope / excludePayloads cases); typecheck + biome clean.
- Verification after deploy: occurrences of issue `736c6ee4…` drop to zero and the stalled backfill resumes.

### DIFF
--- a/packages/domain/destinations/src/constants.ts
+++ b/packages/domain/destinations/src/constants.ts
@@ -3,19 +3,17 @@ export const DESTINATION_INTERVAL_MS_MAX = 3_600_000
 export const DESTINATION_INTERVAL_MS_DEFAULT = 300_000
 
 /**
- * Hard ceiling on a single source read page, independent of (and clamping) the
- * configured `maxRecordsPerRun`. A window read serializes its whole page to
- * JSONEachRow in one ClickHouse response; span rows carry KB–MB payloads, so a
- * large page balloons `ParallelFormattingOutputFormat` — a 35k-row page cost
- * 4.2 GiB and 16s in production, tripping the server-wide OvercommitTracker and
- * resetting the client socket. Clamping the read page keeps every run bounded
- * regardless of a destination's stored config (legacy rows hold the old 50k
- * default). Cursor pagination already chains the rest, so smaller pages just
- * mean more, cheaper windows. Lowered to 2k after 5k still hit the per-query
- * 4 GiB `max_memory_usage` guard on destinations with large LLM payloads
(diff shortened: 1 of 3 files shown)


andresgutgon:
@claude review. Do we keep same behavior but fixing memory issue?

PR #3641, opened 2026-06-22
fix(web): self-heal stale server-fn calls instead of erroring on deploy skew

andresgutgon:
## Problem

A browser tab left open across a deploy calls a **server-function hash from the previous build**. The new server build's resolver (`getServerFnById`) no longer has that hash, so TanStack Start throws `Server function info not found for <hash>` (`@tanstack/start-server-core` → `handleServerAction`), which our `tracingRequestMiddleware` recorded as an exception and returned as a **500**.

This was the **#1 web error by volume — 189 occurrences / 30d** ([Datadog issue `b42b8026-5b49-11f1-aa94-da7ad0900005`](https://app.datadoghq.eu/error-tracking?query=&refresh_mode=sliding&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%22b42b8026-5b49-11f1-aa94-da7ad0900005%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D)), spiking around each deploy. It both polluted Error Tracking and silently broke the user's action (e.g. opening a session view).

A hash from the **current** build always resolves, so this is *exclusively* a client-staleness condition — never a fault in the running deploy. That makes it safe to special-case.

## Fix (two tiers)

**Tier 1 — server (`tracingRequestMiddleware`).** Classify the stale error as expected: skip `recordSpanExceptionForDatadog`, set the span `OK` + a `server_fn.stale` attribute, and re-throw. Stops it polluting Error Tracking; a genuine current-build resolver failure would have a different message and still be reported.

**Tier 2 — client (`staleServerFnReloadMiddleware`, new function middleware `.client()` phase).** When a stale call fails, `window.location.reload()` once so the tab pulls the new build and the user's action just works on retry instead of hitting the error boundary. Guarded against reload loops by a `sessionStorage` timestamp (`STALE_SERVER_FN_RELOAD_GUARD_MS`).

Shared predicate + reload decision live in `apps/web/src/lib/stale-server-fn.ts`, with `reload`/`now`/`storage` injected so the decision is pure and testable.

## Tests

`apps/web/src/lib/stale-server-fn.test.ts` (7 cases): predicate matches the TanStack message and rejects unrelated/non-Error values; reloads once on a stale error; ignores unrelated errors; suppresses a second reload inside the guard window (no loop); reloads again after the window; reloads when storage is unavailable. Full `@app/web` suite green (490 tests), `typecheck` + `biome` clean.

## Verification

- Server classification + reload decision are unit-covered.
- The end-to-end stale-tab flow (old tab → deploy → action → one reload → success) is best smoke-verified post-deploy; after this ships, occurrences on the issue above should drop toward zero and stale tabs should self-heal instead of 500ing.

Related but out of scope: the client-side `Failed to fetch dynamically imported module` (stale lazy chunk) is the same deploy-skew family surfaced through the React error boundary — can be addressed separately.

### DIFF
--- a/apps/web/src/lib/stale-server-fn.test.ts
+++ b/apps/web/src/lib/stale-server-fn.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, it, vi } from "vitest"
+import {
+  isStaleServerFnError,
+  maybeReloadForStaleServerFn,
+  STALE_SERVER_FN_RELOAD_GUARD_MS,
+} from "./stale-server-fn.ts"
+
+const staleError = () => new Error("Server function info not found for abc123def456")
+
+const fakeStorage = (initial: Record<string, string> = {}) => {
+  const map = new Map(Object.entries(initial))
+  return {
+    getItem: (key: string) => map.get(key) ?? null,
+    setItem: (key: string, value: string) => void map.set(key, value),
+    snapshot: () => Object.fromEntries(map),
+  }
+}
+
+describe("isStaleServerFnError", () => {
+  it("matches the TanStack stale server-fn error", () => {
+    expect(isStaleServerFnError(staleError())).toBe(true)
+  })
+
+  it("requires the full `…not found for <id>` shape, not just the phrase", () => {
+    // The phrase alone (without a function id) is not the resolver's error.
+    expect(isStaleServerFnError(new Error("Server function info not found"))).toBe(false)
+  })
+
+  it("ignores unrelated errors and non-Error values", () => {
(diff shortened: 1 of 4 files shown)


andresgutgon:
This fix only one part. This fix when a browser is stale and the backend deployed a new version. 
The problem now is we don't have blue/green deploy meaning users can refresh and still go to the old version and still hit the new version again ♻️ 💥 

BLOCKING THIS UNTIL `web` has 🟦 🟢

PR #3647, opened 2026-06-22
revert(signals): undo the passed-polarity inversion from Phase 2 PR1

csansoon:
Reverts the membership-polarity inversion that shipped in #3621 (Signals Phase 2 PR1). Membership is back to the original problem-detector convention: a score with `passed = false` means the signal's behavior is **present** (an occurrence); `passed = true` means absent. The host derives `passed` by thresholding the script's `value` again (`isScoreMatch`, default 0.5).

## Why

#3621 flipped `passed` so `passed = true` = behavior present across every score source, and ran a one-time migration flipping historical `evaluation` + `annotation` scores. That broke annotations: annotation scores overload `passed` as a **sentiment** field — the annotation UI writes thumbs-up = `passed = true` (positive), thumbs-down = `passed = false` (negative) — and that write path was never changed. So the migration inverted existing human annotations: negatives rendered as positive and vice-versa. We revert now and defer a polarity model that doesn't collide with annotation sentiment.

## Kept (not reverted)

Two PR1 changes are orthogonal to `passed` and stay:

- `scores.source` → `source_type` (Postgres column + domain field). The ClickHouse column, the `score.source` filter-DSL key, and the public `/scores` wire key stay `source`, so there is no SDK/wire change.
- The `signals:match` worker that moved evaluation scheduling off `TracesIngested`.

## Reverted

- **Sandbox contract** — the script returns `value` (+ optional `feedback`); the host derives membership via `isScoreMatch`. `Score`'s `passed` argument and the script-decides-membership model are removed.
- **Polarity** — baseline judge prompt, GEPA proposer, alignment/optimization scoring, the flagger default (`{ passed: false, value: 0 }`), discovery eligibility (`PassedScoreNotEligibleForDiscoveryError`), and alignment-example selection are back to `passed = false` = present.
- **`legacy_polarity`** — column, entity field, and the execution-boundary verdict inversion removed. Safe because no evaluation has been re-aligned since the cutover, so every stored script is pre-cutover (old polarity).
- **Always-stamp `signal_id`** — `run-live-evaluation` stamps `signal_id` only on matches again; occurrence reads drop the `passed = true` gates (membership = `signal_id` present).

## Data migrations

The #3621 data flip is live in production, so this adds new forward migrations (one Postgres; ClickHouse `00041`/`00042`) that restore the data. Three populations behave differently:

- **Evaluation + flagger (`source_id = 'SYSTEM'`) scores** — written uniformly in the new polarity; re-flipped in full (errored rows excluded).
- **Human annotations** — `passed` is sentiment and the write path never changed, so annotations created **after PR1's one-time migration ran** (`2026-06-18 22:00 UTC` — the migration-run time, not the later v0.3.11 app deploy) are already correct. Only the rows that existed when it ran are re-flipped.
- **`signal_id` cleanup** — the `passed` re-flip alone does not undo the always-stamp, so `signal_id` is cleared from non-occurrence evaluation rows (`passed = true` or `errored`), restoring the "`signal_id` only on matches" invariant. Scoped to evaluations — annotation `signal_id` comes from discovery/linking, not always-stamping.

The ClickHouse `scores_hourly_buckets` MV is rebuilt to its original no-passed-filter form, correct again once the always-stamp invariant is restored.

Sequencing: run the data migrations (re-flip + `signal_id` cleanup before the CH MV rebuild — enforced by `mutations_sync` + goose ordering) before or together with the reverted-code deploy. The `legacy_polarity` `DROP COLUMN` ships in the same PR as the code that stopped reading it; brief rollover downtime is acceptable (Phase-1 precedent).

## Verification

- Whole-workspace typecheck green (82/82).
- `@domain/{signals, annotations, scores, evaluations, flaggers, notifications}`, `op-gepa`, `sandbox-quickjs` green; the PR1-added occurrence-gating tests are deleted.
- `@platform/db-postgres` PGlite suite (356 tests) runs the Postgres revert migration end-to-end.
- The ClickHouse `00041`/`00042` migrations and the `chdb`-backed worker/API tests run in CI (no local `chdb` addon); they mirror the proven `00038`/`00039`.

## Follow-up

Thumbs-up annotations that were spuriously discovered during the ~3-day PR1 window (the flipped eligibility briefly let `passed = true` seed discovery) are a small, ambiguous set, not targeted here.

Refs #3621


csansoon:
@claude review this PR

csansoon:
Thanks for the thorough review. Addressed in `ed081e038`:

**Migration-deploy sequencing** — agreed, this is a hard requirement, not advisory. Added a `DEPLOY ORDER (hard requirement, not advisory)` note to the PG migration header spelling out the failure mode you described (PR1 new-polarity code still live after the re-flip writes fresh `passed=true` / always-stamped `signal_id` rows that this one-shot migration won't catch → mixed-polarity dataset). The reverted code and this migration ship in the same PR, so they deploy together.

**Annotation cutoff timestamp** — good catch on the conflation. The cutoff (`2026-06-18 22:00 UTC`) is when PR1's *migration ran* in production, not the v0.3.11 *app deploy* (~15h later); `470c8c355` had already corrected the timestamp but left the old "v0.3.11 deploy" wording. Clarified the comments in the PG migration and CH `00041` (clustered + unclustered), and fixed the PR description.

**MV rebuild window (`00042`)** — acknowledged: `scores_hourly_buckets` is empty between the DROP and the backfill INSERT, so escalation baselines read zero during that window. It's inherent to a synchronous MV rebuild and the migration runs once — we'll run it in a low-traffic window and note it in the deploy runbook (no code change).

The ✓ items (PG step ordering, `mutations_sync`, discovery index predicate, no-passed-filter MV, `legacy_polarity` drop timing) match my analysis — thanks for confirming. The window-discovered thumbs-up annotations remain a tracked follow-up.

geclos (approved):
LGTM

PR #3650, opened 2026-06-22
Improve monitor creation and session-based signal views

geclos:
## Summary

Adds source-first monitor creation and moves monitor/signal review surfaces toward sessions instead of traces.

Users can now create monitors from a selected source: saved searches, tools, users, or all sessions. Tool and user sources support either a specific tool/user or the whole source category. The alert form uses the selected target to offer metric options such as session volume, latency, cost, tokens, error rate, and match-style alert modes.

## Backend and analytics changes

Score analytics now reports signal impact in session terms where the UI needs session percentages and session lists. Signal detail sessions are fetched from ClickHouse and then hydrated through the session repository, instead of grouping traces in the browser.

Traced scores without a session id now fall back to `traceId` in the shared score write path. This keeps live evaluation and flagger scores queryable as session-keyed records when they run on trace rollups whose session id is empty.

Trace count metric semantics for trace streams now count distinct sessions, matching the product direction that monitor volume means session volume unless a tool-call stream explicitly counts calls.

## UI changes

- Monitor creation starts with source selection: saved search, tools, users, or sessions.
- Saved-search monitor creation uses the unified metric alert path.
- Monitor matching previews show sessions.
- Signal detail drawers list sessions from the backend.
- Tool detail recent activity shows sessions rather than individual tool calls.
- Signal list labels and percentages use affected sessions.

## Validation

- `pnpm --filter @platform/db-clickhouse typecheck`
- `pnpm --filter @domain/scores typecheck`
- `pnpm --filter @domain/evaluations typecheck`
- `pnpm --filter @domain/scores test -- write-score.test.ts`
- Commit hooks ran `turbo format` and `knip` for each commit.

Please attach screenshots or a short screen recording for the monitor creation modal, signal detail drawer, and tool detail activity table before review.


packages/platform/db-clickhouse/src/repositories/score-analytics-repository.ts
```diff
@@ -1754,6 +1769,81 @@ export const ScoreAnalyticsRepositoryLive = Layer.effect(
...
+
+          return yield* chSqlClient
+            .query(async (client) => {
+              const result = await client.query({
+                query: `SELECT
+                        normalized_session_id AS session_id,
+                        max(created_at) AS last_seen_at
+                      FROM (
+                        SELECT
+                          if(session_id = '', toString(trace_id), session_id) AS normalized_session_id,
```
andresgutgon: Find nicer using `coalesce`
```SQL
coalesce(nullIf(session_id, ''), toString(trace_id)) AS normalized_session_id
```

geclos:
@claude review

packages/platform/db-clickhouse/src/repositories/score-analytics-repository.ts
geclos: agree fixing

geclos:
Addressed the review feedback in 4ac354d01:

- Added explicit product-level monitor target kind (`tool`, `user`, `session`, `savedSearch`, `signal`) so `traces`/`spans` are treated as internal query streams rather than user-facing target categories.
- Added sessions support to the metric reader so session monitors no longer defect at evaluation time.
- Replaced the signal-session N+1 hydration with one batched ClickHouse query via `listBySessionIds`.
- Switched score session normalization to `coalesce(nullIf(session_id, ), toString(trace_id))`.
- Fixed initial monitor source selection for user targets and moved the saved-search target factory next to the other target helpers.

Verification: `pnpm typecheck`, `pnpm --filter @domain/monitors test -- create-monitor.test.ts check-metric-monitors.test.ts`, `pnpm openapi:emit`, `pnpm mcp:emit`, `pnpm generate:sdk`, `pnpm --filter @latitude-data/sdk typecheck`, `pnpm --filter @app/api typecheck`.

packages/platform/db-clickhouse/src/repositories/score-analytics-repository.ts
geclos: Addressed in 4ac354d01 — replaced the session-id normalization with the `coalesce(nullIf(...), toString(trace_id))` form.

andresgutgon (commented):
Missing CHANGELOG and bumping of SDKs

PR #3654, opened 2026-06-22
feat(monitors): add cacheHitRate metric kind for token analytics

andresgutgon:
## what this does

Adds a new `cacheHitRate` monitor metric kind so a monitor can alert when prompt-cache caching degrades. The metric is the token-weighted ratio over the evaluation window:

```
sum(cacheRead) / (sum(input) + sum(cacheRead) + sum(cacheCreate))
```

a `0..1` fraction, presented as a percentage. It joins the existing metric kinds (`count`, `errorRate`, `avg{field}`, `p95{field}`, `sum{field}`) on the `metric.threshold` / `metric.escalating` alert conditions.

## why

Customer ask: caching can silently degrade — cache-read tokens collapse while input tokens spike — and **total token usage barely moves**, so a `sum`/`avg` token monitor never fires even as the bill jumps several-fold. The metric schema's aggregating kinds each operate on a single token field and cannot express a quotient of two different token sums, so cache hit rate was not expressible. This adds the one kind that catches that failure mode before it shows up on an invoice.

The natural configuration is `direction: below` — open an incident when cache hit rate drops under a threshold (e.g. below 70%).

## no migration

The monitor's `metric` is a jsonb column (`packages/platform/db-postgres/.../monitors.ts`) and the schema is a Zod discriminated union, so the new kind is purely additive — no DB migration.

## evaluation

`metricAggregate` in the ClickHouse `MetricSeriesReader` builds the ratio for both the `traces` and `spans` streams. Divide-by-zero is guarded: a window with no input-side tokens has an undefined rate and reads `0` (not `nan`), mirroring how `errorRate` guards its empty case. Combined with the existing empty-window suppression in `evaluateMetricAlert`, a target with no traffic does not fire a `below` alert. On the spans stream the cache token columns are gated to usage operations (`chat`, `text_completion`, …) so wrapper/tool spans don't enter the denominator, consistent with the existing cost/tokens span columns.

cacheHitRate is intensive (a ratio), so it is **not** accumulating — `multiplier`/`expected` baselines compare directly, no window rescaling, same as `errorRate`/`avg`/`p95`.

## surfaces touched

- `monitorMetricSchema` (`@domain/shared`) — new no-field `cacheHitRate` member; `metricUnit` → `%`; `formatMetric` → "the cache hit rate".
- `MetricSeriesReader` (`@platform/db-clickhouse`) — ratio aggregate + cache token columns for both streams. Also renamed the local `BILLABLE_OPERATIONS_SQL` constant to `USAGE_OPERATIONS_SQL` to avoid overloading "billable" with the billing system.
- Web (`apps/web`) — "Cache hit rate" offered on the traces stream; threshold unit `%`; the existing data-driven form picks it up with no bespoke wiring.
- OpenAPI + MCP manifests and the TS/Python SDKs regenerated from the Zod source.

## tests

- `monitor-metric-units.test.ts` — unit/format/convert for `cacheHitRate` (`%`).
- monitors `helpers.test.ts` — human-readable sentence for a `below` cacheHitRate threshold.
- `metric-series-reader.test.ts` — token-weighted ratio over traces, the divide-by-zero guard (rows present, no input tokens ⇒ `0`), and the spans stream including usage-operation gating.

Typecheck + affected vitest suites pass locally (`@domain/shared`, `@domain/monitors` 141 tests, `@platform/db-clickhouse` reader suite).

Fixes LAT-691


andresgutgon:
@claude review

PR #3658, opened 2026-06-22
feat(telemetry): Hermes Agent integration (stream sessions to Latitude as OTLP traces)

### LINKED ISSUE #3400: [Feature]: Hermes Agent integration (stream sessions to Latitude as OTLP traces)
### What problem are you trying to solve?

Hermes Agent (https://github.com/nousresearch/hermes-agent) is Nous Research's
open-source agent harness — a fast-growing project (~27k stars) with a persistent
agent loop that runs code, browses, and connects to many messaging platforms.
Users running Hermes have no way to get full-fidelity observability of their agent
runs in Latitude today. Arize already ships observability for Hermes, so there's
both a proven integration path and competitive pressure.

### What would you like to happen?

Ship a `@latitude-data/hermes-telemetry` package following the same pattern as our
existing harness integrations (`@latitude-data/pi-telemetry`,
`@latitude-data/claude-code-telemetry`, `@latitude-data/openclaw-telemetry`):

- A Hermes plugin/hook/extension that streams each session to Latitude as OTLP
  traces — user prompts, model turns, tool calls, tool results, token usage, and
  the real system prompt that reached the model.
- A one-shot `npx -y @latitude-data/hermes-telemetry install` flow that prompts for
  API key + project slug and wires the hook into Hermes's config.
- Docs page under `docs/telemetry/frameworks/` (or a new "agent harnesses" section).

### Alternatives you've considered

Users can hand-roll the generic OTEL exporter (`docs/telemetry/otel-exporter.md`),
but that loses the turnkey install and the harness-specific span tree that makes
the other integrations valuable.

### What would you like to happen?

_No response_

### Alternatives you've considered

_No response_

GouravSingal-code:
> **Draft** — closes #3400. Opening early to confirm the packaging approach before polishing (see "Open question" below).

## What

A Hermes Agent integration that streams each session to Latitude as OTLP traces — user prompts, model turns, tool calls, tool results, token usage, and the real system prompt.

## Approach

The existing harness integrations (`pi`, `claude-code`, `openclaw`) wrap **JavaScript** tools, so a JS/npm plugin loads right into them. **Hermes is a Python harness** that loads in-process **Python** plugins (it discovers user plugins in `~/.hermes/plugins/` and via the `hermes_agent.plugins` pip entry-point — exactly how its bundled `observability/langfuse` plugin works). A JS module can't load inside a Python process, so the runtime connector here is a small **Python plugin**, and the npm package ships + installs it.

- **`plugin/`** — the Python Hermes plugin (`plugin.yaml` + `__init__.py`). Registers Hermes's lifecycle hooks (`pre/post_api_request`, `pre/post_llm_call`, `pre/post_tool_call`), assembles one trace per turn (`interaction` root → `llm_request` + `tool_execution` children, GenAI conventions), and ships OTLP to Latitude. Zero dependencies (stdlib HTTP, certifi when available), fail-open, content-gating via `--no-content`.
- **`src/`** — the TypeScript installer: `npx -y @latitude-data/hermes-telemetry install` prompts for API key + project, copies the plugin into `~/.hermes/plugins/observability/latitude/`, and writes `~/.hermes/.env`. Also includes an OTLP span-builder spec + tests mirroring the Python mapping.
- **`docs/telemetry/hermes.md`** + nav entry.

## ❓ Open question (the reason this is a draft)

The issue asks for an **npm** `@latitude-data/hermes-telemetry` like the other integrations — but those wrap JS tools, whereas Hermes is Python. Two options, both verified workable:

- **(A)** the npm package that installs the Python plugin (this PR) — matches the issue + the `npx install` UX.
- **(B)** a native **pip** package Hermes auto-discovers via its `hermes_agent.plugins` entry-point — arguably cleaner for a Python tool.

I built **(A)**; happy to switch to **(B)** (the Python plugin is identical — only the wrapper/publish path changes). **Which do you prefer?**

## Testing

Verified end-to-end against real services:
- Installed `hermes-agent` 0.17.0; Hermes's own `PluginManager.discover_and_load()` loads the plugin (runs `register(ctx)`), lists it as `latitude | enabled | user`, and registers all hooks (all are in Hermes's `VALID_HOOKS`).
- Drove a full turn through Hermes's own `invoke_hook(...)` → the plugin built the trace and shipped it → **`ingest HTTP 200`**, and the trace renders in Latitude (`interaction` → 2× `llm_request` → `get_weather` `tool_execution`).
- `cargo`-style unit/typecheck: package typechecks; installer functional test copies the plugin + writes `.env`; `knip` clean.

## Notes / TODO before "ready for review"

- Resolve the (A) vs (B) packaging question above.
- A live LLM-driven Hermes session (this PR drove Hermes's real hook dispatch with synthesized turn data; handlers are defensive `**kwargs`).
- Version/cost columns populate from real model latency + a priced model.

### DIFF
--- a/packages/telemetry/hermes/src/latitude_telemetry_hermes/__init__.py
+++ b/packages/telemetry/hermes/src/latitude_telemetry_hermes/__init__.py
@@ -1,0 +1,5 @@
+"""latitude — Hermes plugin that streams sessions to Latitude as OTLP traces."""
+
+from .hooks import register
+
+__all__ = ["register"]
--- a/packages/telemetry/hermes/src/latitude_telemetry_hermes/config.py
+++ b/packages/telemetry/hermes/src/latitude_telemetry_hermes/config.py
@@ -1,0 +1,102 @@
+"""latitude — Hermes plugin that streams sessions to Latitude as OTLP traces.
+
+Traces Hermes conversations, LLM calls, and tool usage to Latitude. One Hermes
+turn becomes one trace: an ``interaction`` root span with an ``llm_request``
+child per model call and a ``tool_execution`` child per tool call. Spans follow
+Latitude's GenAI semantic conventions (``gen_ai.*``) so they render natively in
+the Latitude trace viewer.
+
+Activation is handled by the Hermes plugin system — this plugin only loads when
+enabled via ``hermes plugins enable latitude``. At runtime it also
+requires credentials; if they're missing the hooks are inert (fail-open: a
+telemetry error never affects the agent).
+
+Required env vars (set in your environment or ``~/.hermes/.env``):
+  LATITUDE_API_KEY   - Latitude API key
+  LATITUDE_PROJECT   - Latitude project slug (or LATITUDE_PROJECT_SLUG)
+
+Optional env vars:
+  LATITUDE_BASE_URL  - ingest endpoint (default: https://ingest.latitude.so)
+  LATITUDE_NO_CONTENT - "true" to export structure/timing without prompts,
+                        responses, or tool I/O
+  LATITUDE_DEBUG     - "true" for verbose logging
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import ssl
+import threading
+from typing import Any, Dict, Optional
+
+
+def _ssl_context() -> ssl.SSLContext:
+    """Verified TLS context, preferring certifi's CA bundle.
+
+    Some Python installs (notably python.org builds on macOS) ship without a
+    usable system CA store, so a plain ``urlopen`` to https can raise
+    CERTIFICATE_VERIFY_FAILED. Hermes's HTTP stack already depends on certifi,
+    so use it when available; otherwise fall back to the system default.
+    """
+    try:
+        import certifi
+
+        return ssl.create_default_context(cafile=certifi.where())
+    except Exception:
+        return ssl.create_default_context()
+
+
+_SSL_CONTEXT = _ssl_context()
+
+logger = logging.getLogger(__name__)
+
+SCOPE_NAME = "@latitude-data/hermes-telemetry"
+PKG_VERSION = "0.1.0"
+
+# Bound on live trace state, so turns that never reach a clean finish
+# (interrupted / tool-only final step) can't leak forever.
+_MAX_RUNS = 256
+
+
+# ─────────────────────────── config ────────────────────────────────────────
+
+
+def _env(name: str, default: str = "") -> str:
+    return os.environ.get(name, default).strip()
+
+
+_CONFIG: Optional[Dict[str, Any]] = None
+_CONFIG_LOCK = threading.Lock()
+
+
+def _load_config() -> Dict[str, Any]:
+    api_key = _env("LATITUDE_API_KEY")
+    project = _env("LATITUDE_PROJECT") or _env("LATITUDE_PROJECT_SLUG")
+    base_url = _env("LATITUDE_BASE_URL") or "https://ingest.latitude.so"
+    enabled_flag = _env("LATITUDE_HERMES_TELEMETRY_ENABLED") or _env("LATITUDE_TELEMETRY_ENABLED")
+    enabled = enabled_flag.lower() not in {"0", "false", "no"} if enabled_flag else True
+    no_content = _env("LATITUDE_HERMES_NO_CONTENT") or _env("LATITUDE_NO_CONTENT")
+    allow_content = no_content.lower() not in {"1", "true", "yes"} if no_content else True
+    return {
+        "api_key": api_key,
+        "project": project,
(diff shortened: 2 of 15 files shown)


packages/telemetry/hermes/src/config.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+}
+
+export function resolveHermesDir(env: NodeJS.ProcessEnv = process.env): string {
+  if (env.HERMES_CONFIG_DIR) return expandHome(env.HERMES_CONFIG_DIR)
+  return join(userInfo().homedir, ".hermes")
+}
+
+export function resolveIdentity(env: NodeJS.ProcessEnv = process.env): UserIdentity {
+  const email = env.LATITUDE_HERMES_USER_EMAIL ?? gitConfig("user.email")
```
andresgutgon: Why do you need the user's email?

packages/telemetry/hermes/src/setup.ts
```diff
@@ -0,0 +1,208 @@
+import { copyFileSync, existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs"
+import { dirname, join } from "node:path"
+import { fileURLToPath } from "node:url"
+import * as p from "@clack/prompts"
+import pc from "picocolors"
```
andresgutgon: The fewer dependencies, the better. Please do the setup without colors; it would be better. No big deal tho but I prefer to be as lean as possible. Same for `@clack/prompts` and printing in the terminal

GouravSingal-code:
Hi @andresgutgon , I've converted the whole plugin into a pip package. Could you please check now ?

packages/telemetry/hermes/src/latitude_telemetry_hermes/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""latitude — Hermes plugin that streams sessions to Latitude as OTLP traces."""
+
+from .hooks import register
+
+__all__ = ["register"]
```
andresgutgon, line 5: You split into different files. Much better!

packages/telemetry/hermes/src/latitude_telemetry_hermes/py.typed
GouravSingal-code, line 1: py.typed is an empty file that tells type checkers (like mypy/pyright) "this package has real type hints, please use them."

Here's why it matters, with an example:

Our package has a function that returns a string:
latitude_telemetry_hermes/util.py
def trace_id() -> str:
   return os.urandom(16).hex()

Someone uses it wrong — they put the string into an int:
from latitude_telemetry_hermes.util import trace_id
count: int = trace_id()   # bug: trace_id() is a str, not an int

Without py.typed: the type checker ignores our hints and says nothing. The bug slips through.

With py.typed: the type checker reads our -> str hint and warns:
error: expression has type "str", variable has type "int"

andresgutgon (approved):

GouravSingal-code:
@andresgutgon , can you please merge this pr if everything looks good to you.

andresgutgon:
Hi @GouravSingal-code have you tried running latitude locally and seeing traces are ingested correctly?

GouravSingal-code:
Screenshot 2026-06-25 at 12 42 35 PM

GouravSingal-code:
Screenshot 2026-06-25 at 12 43 40 PM

andresgutgon:
Does it has the messages in the conversation tab? sorry for the interrogation xD

GouravSingal-code:
Screenshot 2026-06-25 at 12 48 29 PM

packages/telemetry/hermes/src/latitude_telemetry_hermes/config.py
```diff
@@ -0,0 +1,102 @@
...
+        return ssl.create_default_context(cafile=certifi.where())
+    except Exception:
+        return ssl.create_default_context()
+
+
+_SSL_CONTEXT = _ssl_context()
+
+logger = logging.getLogger(__name__)
+
+SCOPE_NAME = "@latitude-data/hermes-telemetry"
```
neoxelox, line 54: shouldn't it be `latitude-telemetry-hermes` ?

geclos (approved):

packages/telemetry/hermes/src/latitude_telemetry_hermes/config.py
andresgutgon: Can you this change @GouravSingal-code ? and I'll merge it

andresgutgon: On a second though I'll do the change later

andresgutgon:
Plugin published @GouravSingal-code 
https://pypi.org/project/latitude-telemetry-hermes/

Thanks for your contribution 💪

PR #3659, opened 2026-06-22
feat(monitors): classify listMonitorsForTarget by target kind

geclos:
## What

Adds an optional `targetKind` to `listMonitorsForTarget` so callers classify monitors by **product kind** (`user` / `tool` / `session` / `savedSearch`) instead of leaning on filter-set containment. Filter containment is left to do only identity narrowing (a specific `userId` / `toolName`).

This fixes the case where an **all-users** monitor — empty filter set, or one carrying an extra predicate like `status = error` — would *drop out* of the users dropdown because the old lookup used the filter set to classify.

## Changes

- **End-to-end plumbing** of `targetKind` through the domain port, use-case, both repositories (Postgres + fake), the API route, the web server-fn/collection, and the TS + Python SDKs.
- **Null-safe filter containment**: `coalesce(target_filter_set, '{}') @> :subset`, so no-filter targets match an empty subset.
- **Single source of truth for kind classification**: a canonical `monitorTargetKind` decoder in `@domain/monitors`, used by both the Postgres reads and the fake repo (removes the duplicated inference and any fake/Postgres divergence). The kind is recoverable from `stream`/`savedSearchId` (the create-time encoding), so nothing new is persisted — **no migration**.
- **`signal` excluded from the public request enum** until it has an encoding; it stays on the response `MonitorTarget` schema. Regenerated `openapi.json`, `mcp.json`, and both SDKs.
- **UI**: merged the monitors-list "Watching" column into the unified Target cell.

## Data model note

`kind` is the client-chosen product identity; `stream` / `filterSet` / `savedSearchId` / `metric` are compute-time invariants (how to compute the metric). `kind → stream` is a deterministic function, so `(stream, savedSearchId)` is an injective encoding of `kind` and decodes losslessly. A future `signal_id` column will give `signal` its own slot, parallel to `savedSearchId`.

## Tests

- New Postgres repo test covering the `coalesce(null)` path (no-filter target matches an empty subset).
- New test matching all-user monitors by kind without identity filters.
- `@platform/db-postgres` (358) and `@domain/monitors` (141) suites green; typecheck clean across domain, db-postgres, api, web, and the TS SDK.


geclos:
@claude review

PR #3661, opened 2026-06-23
feat(signals): present-verdict convention (passed = true = present)

csansoon:
Phase 2 PR2 of the Signals work. Establishes one convention: **a signal's behavior is "present" when `passed = true`** — the intuitive direction PR1 shipped inverted (`passed = false`) and #3647 reverted.

## what changed

Membership is `signal_id`; `passed` is the source-specific verdict (host-derived for evaluations by thresholding the script's `value`). The flip to `present ⇒ passed = true` lands at every site that reads or emits the verdict:

- **Writer** (`run-live-evaluation`) stamps `signal_id` when `passed === true`. Stored `value`/`passed` are untouched.
- **Discovery eligibility** (`check-eligibility`) treats an evaluation score as a discovery candidate only when present (`passed = true`); an absent run (`passed = false`, `signal_id = null`) is skipped so `resolveLinkedSignalId` can't assign it. The annotation branch is unchanged (negative annotations seed discovery).
- **Alignment scoring** (`evaluate-draft-against-examples`, `evaluate-optimization-candidate`) computes `predictedPositive = passed === true`.
- **Generation**: the baseline judge prompt sets `passed = true` when the behavior is present, and the GEPA proposer carries an explicit polarity rule so its prompt rewrites stay on the convention.

## no flag, no migration, no schema change

`passed` is host-derived per run, so the convention is purely how the runtime reads it and how judges are phrased — there is no per-evaluation column and no `scores` data migration. This is safe because no production evaluation uses the old convention. The discovery-work index keeps its existing predicate; the PR carries **zero schema change**.

## ClickHouse coverage for absent runs

Flipping the convention turned absent runs from `passed = true` to `passed = false`. `isImmutableScore` now treats every evaluation run as immutable on arrival regardless of verdict, so absent runs still sync to ClickHouse and remain denominators for pass-rate.

## testing

Whole-workspace typecheck (82/82). Green locally: `@domain/{scores,evaluations,signals}`, `@platform/op-gepa`, the touched `@app/workflows` alignment/optimize/refresh tests, and `@platform/db-postgres` evaluation-repository (`pg:generate` confirms no pending schema diff). ClickHouse-backed `@app/api`/`@app/workers` run in CI.

Spec: `specs/signals.md`, Phase 2 / PR2.


andresgutgon (dismissed):

csansoon:
@claude review this PR

csansoon:
Thanks for the review. Addressed the one consistency item; the rest are intentional and called out as non-blocking — confirming the reasoning:

- **`specs/sandbox-runtime.md` polarity** — good catch. That file still described generated judges in the old "problem-present ⇒ `passed = false`" convention, which contradicts the flip this PR makes. Fixed in `20e625261`: the present-verdict polarity is now described as per-evaluation (`membership_on_pass`), with the current convention being behavior-present ⇒ `Passed` (`passed = true`) and deprecated judges using the inverse. No example template changed — the wrapper (`result.passed ? Passed(1) : Failed(0)`) is polarity-neutral; the convention lives in the prompt text and the flag.
- **`check-eligibility` null-evaluation fallback** — the implicit invariant is intentional: `findById` does not filter `deletedAt`, so a soft-deleted evaluation still resolves and its real `membership_on_pass` is used. The `passed === false` fallback only applies to a genuinely orphaned `sourceId` (no row at all), which the writer never produces.
- **Discovery-work index broadening** — agreed, will monitor. The application-level gate keeps it correct; if old-convention passing scores make the candidate scan noisy, a narrower partial index is a cheap follow-up.
- **Optional `membershipOnPass` params** — left optional (default `false`) so the change is additive and behavior-preserving; every caller in this PR passes it explicitly. Tightening to required is a reasonable post-migration cleanup once no old-convention evaluations remain.

andresgutgon (changes requested):
Merge both migrations 🙏

csansoon:
Merged both migrations into one — `20260623122358_add-membership-on-pass-and-discovery-index` now carries the `membership_on_pass` column add plus the discovery-index predicate change. The two prior folders are removed, so the net diff has a single migration. Applies clean against PGlite (356/356). `89dcba905` 🙏

andresgutgon (dismissed):

csansoon:
Update: simplified the approach. Since no production evaluation uses the old convention, instead of the per-evaluation `membership_on_pass` flag I force one global convention — a signal's behavior is present when `passed = true`. That removes the column **and** the migration entirely, so there's nothing left to merge: the PR now carries zero schema change. `pg:generate` confirms no pending diff. (`67e6c4aad`, `1f5c67596`) 🙏

andresgutgon (approved):

PR #3662, opened 2026-06-23
fix(signals): avoid repeated list analytics

geclos:
Skips expensive signals list analytics after the first page so infinite scroll does not keep recomputing histogram and project-level totals. Continuation pages still hydrate the visible rows with evaluations, trends, tags, per-page occurrence aggregates, and sessions-based affected percentages.

Also removes the client-side next-page prefetch that was triggering those repeated analytics reads ahead of user scroll, and keeps bulk lifecycle enumeration on the lightweight path.

Validation: attempted `pnpm --filter @domain/signals test -- list-signals.test.ts`, but dependencies are not installed in this checkout (`vitest: command not found`).


packages/domain/signals/src/use-cases/list-signals.ts
```diff
@@ -529,12 +536,14 @@ export const listSignalsUseCase = (
     const evaluationRepository = yield* EvaluationRepository
     const sessionRepository = yield* SessionRepository
 
-    const windowMetricsEffect = scoreAnalyticsRepository.listSignalWindowMetrics({
-      organizationId: parsed.organizationId,
-      projectId: parsed.projectId,
-      ...(selectedTimeRange ? { timeRange: selectedTimeRange } : {}),
-      ...(parsed.signalIds ? { signalIds: parsed.signalIds } : {}),
-    })
+    const windowMetricsEffect = scoreAnalyticsRepository
```
andresgutgon, line 670: The weird thing is that we do the metrics query in the list query

PR #3665, opened 2026-06-23
Consolidate monitor incidents and rules

geclos:
## Summary
- Consolidates alert incidents into the incidents domain with source-keyed monitor and signal incident handling.
- Reworks monitors around a single target/rule model, updates mute gates and notification fan-out, and removes the old monitor alert API surface.
- Moves signal escalation through the shared incident flow and refreshes web, API, SDK, migration, seed, and docs surfaces for the final model.

## Verification
- `pnpm knip`
- Pre-commit hook: `turbo format`
- Verified during implementation: `pnpm --filter @app/api typecheck`, `pnpm --filter @app/workers typecheck`, `pnpm --filter @app/web typecheck`
- Verified during implementation: API, workers, web, Postgres, ClickHouse, monitors, notifications, signals, and incidents package tests
- Adversarial implementation/doc reviews completed during the phase work

## Review Notes
- Draft PR while the large monitor/signal/incidents consolidation gets a final reviewer pass.
- UI changed in monitor and signal flows; attach screenshots or a short recording before marking ready for review.


geclos:
@claude review

geclos:
@cursoragent review

packages/platform/db-postgres/src/create-better-auth.ts
```diff
@@ -114,7 +114,7 @@ export const createBetterAuth = (config: BetterAuthConfig) => {
   const stripeClient =
     stripeSecretKey && stripeWebhookSecret
       ? new Stripe(stripeSecretKey, {
-          apiVersion: "2026-05-27.dahlia",
+          apiVersion: "2026-04-22.dahlia",
```
andresgutgon: Why an older version?

geclos: Updated in `7326cc3c7` to `2026-05-27.dahlia`, which matches the latest API version exposed by the installed `stripe@22.2.0` types. I also aligned the auth CLI and billing provider Stripe clients in the same commit.

geclos:
Addressed the review feedback in three commits:

- `1fb9d95f7` implements the monitor sweep, removes the worker scaffolding/dead deps, wraps monitor escalation exits in the transaction with `IncidentClosed`, and emits `IncidentClosed` when deleting a monitor closes an open incident.
- `2b779b370` updates the existing consolidation migration to delete legacy multi-alert monitors before folding alerts, dedupe pre-existing duplicate open incidents, and enforce the partial unique open-incident index on `(organization_id, source_type, source_id) WHERE ended_at IS NULL`. I also reviewed the current seed/demo project paths: `packages/platform/db-postgres/src/seeds/monitors/index.ts` is a no-op under the new model, the demo snapshot only carries derived taxonomy/search data, and current monitor creation uses the single-rule model rather than `monitor_alerts`.
- `7326cc3c7` aligns Stripe clients to the current API version for the installed `stripe@22.2.0` package.

Verification run: package checks/typechecks for monitors, workers, API, web, db-postgres; `@domain/monitors` targeted tests; and `@platform/db-postgres pg:check`.

.github/workflows/api-manifests.yml
```diff
@@ -63,6 +63,9 @@ jobs:
       - name: Regenerate Fern-emitted TypeScript SDK
         run: ./fern/invoke.sh generate --group local-typescript --local --force
 
+      - name: Postprocess Fern-emitted TypeScript SDK
```
geclos: what is this and why it's in this pr

apps/api/src/routes/incidents.ts
```diff
@@ -53,28 +53,62 @@ const ListIncidentsQuerySchema = z.object({
...
+    .describe("Deprecated filter from the previous incidents model. This endpoint now filters by source and severity."),
   severities: z
     .preprocess((val) => (val === undefined || Array.isArray(val) ? val : [val]), z.array(z.enum(INCIDENT_SEVERITIES)))
     .optional()
     .describe("Restrict to incidents whose severity matches any value in this list."),
 })
 
+type ListIncidentsQuery = z.infer<typeof ListIncidentsQuerySchema>
+
+const parseLegacySourceType = (sourceType: string | undefined) => {
```
geclos: why legacy, how do legacy source types compare to the new ones

apps/api/src/routes/incidents.ts
```diff
@@ -53,28 +53,62 @@ const ListIncidentsQuerySchema = z.object({
...
+) => {
+  if (query.kinds !== undefined) {
+    throw new BadRequestError({ message: "`kinds` is no longer supported. Use `source_type` and `severities`." })
+  }
+
+  const legacySourceType = parseLegacySourceType(legacyQuery.sourceType)
+  const legacySourceId = parseLegacySourceId(legacyQuery.sourceId)
+
+  return {
+    sourceTypes: query.source_type
```
geclos: why does query accept source_type and sourceTypes?

.github/workflows/api-manifests.yml
geclos: This keeps the CI drift check aligned with local `pnpm generate:sdk`. Fern generates the TypeScript SDK first, then `scripts/postprocess-typescript-sdk.mjs` applies our checked-in postprocess step. Without this in `.github/workflows/api-manifests.yml`, CI can compare pre-postprocessed SDK output against the committed postprocessed SDK and report false drift. It is in this PR because the monitor/API contract changes regenerate the SDK and exposed that mismatch.

geclos: Correction after discussing this: the postprocess was the wrong direction. OpenAPI should remain the source of truth and Fern should be a translation layer, not something we patch after generation. I removed the postprocess script, package hook, and CI step in `da14c82d5`, then regenerated the TypeScript SDK directly from OpenAPI/Fern output.

PR #3668, opened 2026-06-23
feat(spans): ingest OpenClaw's native OpenTelemetry exporter

andresgutgon:
## Summary

OpenClaw now ships an official OpenTelemetry exporter (`@openclaw/diagnostics-otel`) that follows GenAI semantic conventions and is maintained by OpenClaw. This PR makes Latitude ingest its spans cleanly, **recommends it over our hand-rolled `@latitude-data/openclaw-telemetry` plugin**, and deprecates the old plugin.

All OpenClaw-specific handling is scoped to `scope_name = "openclaw"` and lives in the resolver layer — the generic `transform.ts` stays brand-agnostic.

## Ingest handling (`@domain/spans`)

- **Operation** (`resolvers/operation.ts`): classify `openclaw.run` → `invoke_agent` and `openclaw.tool.execution` → `execute_tool` (model calls already resolve to `chat` via `gen_ai.operation.name`). Fixes the trace structure + Tools facet / `tool_name` extraction.
- **Usage** (`resolvers/usage.ts`): OpenClaw never emits flat `gen_ai.usage.*` attrs — it buries per-call tokens **and the provider's real cost (USD)** inside the assistant-message JSON on `openclaw.content.output_messages`. We read it there (additive tokens, no inclusive/additive inference; real cost so `costIsEstimated = false`).
- **Status** (`resolvers/status.ts`): OpenClaw signals success out-of-band via `openclaw.outcome` and leaves the OTel status `unset` on success, so bars rendered gray. Derive `ok`/`error` from the outcome (OTel `error` and non-`completed` outcomes → `error`).
- **Drop** (`dropped-spans.ts`): the standalone `openclaw.model.usage` span is emitted in its own orphan trace and duplicates usage that already rides on the `model.call` spans — drop it at ingest so it doesn't surface as a phantom trace/session.

## Deprecation + docs

- README deprecation banners on `@latitude-data/openclaw-telemetry` + its CLI, pointing to the native exporter.
- Rewrote `docs/telemetry/openclaw.md` to recommend `@openclaw/diagnostics-otel` with a production-ingest config example (auth headers + `captureContent`), a "Migrating from the Latitude plugin" section, and a link to OpenClaw's OTel docs.

## Testing

- `pnpm --filter @domain/spans test` — 867 passing, incl. new unit tests for `resolveUsage` (embedded usage), `resolveStatusCode`, `isDroppedSpan`, and an end-to-end `transformOtlpToSpans` test for OpenClaw (drop + tokens + real cost + operation/status classification).
- `pnpm --filter @domain/spans typecheck` — clean.
- Validated end-to-end against real OpenClaw spans in local ClickHouse (per-call usage sums to the orphan span's totals; status/operation/tool resolve correctly).

### DIFF
--- a/packages/domain/spans/src/otlp/resolvers/usage.ts
+++ b/packages/domain/spans/src/otlp/resolvers/usage.ts
@@ -126,10 +126,11 @@ function extractEmbeddedUsage(outputMessagesJson: string): OpenclawUsage | undef
   try {
     const parsed: unknown = JSON.parse(outputMessagesJson)
     if (!Array.isArray(parsed)) return undefined
-    const withUsage = parsed.find(
-      (m): m is { usage: OpenclawUsage } =>
-        !!m && typeof m === "object" && typeof (m as { usage?: unknown }).usage === "object",
-    )
+    const withUsage = parsed.find((m): m is { usage: OpenclawUsage } => {
+      if (!m || typeof m !== "object") return false
+      const usage = (m as { usage?: unknown }).usage
+      return usage !== null && typeof usage === "object"
+    })
     return withUsage?.usage
   } catch {
     return undefined
--- a/packages/domain/spans/src/otlp/resolvers/usage.ts
+++ b/packages/domain/spans/src/otlp/resolvers/usage.ts
@@ -1,4 +1,5 @@
 import { computeTokenCost, getCostSpec } from "@domain/models"
+import { stringAttr } from "../attributes.ts"
 import type { OtlpKeyValue } from "../types.ts"
 import { resolveTokens } from "./usage/tokens.ts"
 import { first, fromFloat } from "./utils.ts"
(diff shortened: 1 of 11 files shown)


neoxelox (dismissed):

packages/domain/spans/src/otlp/resolvers/usage.ts
```diff
@@ -85,13 +86,115 @@ export interface ResolvedUsage {
...
 interface ResolveUsageInput {
   readonly attrs: readonly OtlpKeyValue[]
   readonly provider: string
   readonly model: string
 }
 
 export function resolveUsage({ attrs, provider, model }: ResolveUsageInput): ResolvedUsage {
+  // Some instrumentations carry usage in the message payload rather than as flat
+  // gen_ai.usage.* attrs (OpenClaw) — prefer that when present.
+  const embedded = resolveEmbeddedMessageUsage(attrs, provider, model)
```
neoxelox, line 196: make sure giving priority to this does not break other providers

andresgutgon: I think we're fine
Image

andresgutgon:
@claude review

andresgutgon:
Addressed the review findings in `c2f20544e`:

- (1) **(null type predicate in `extractEmbeddedUsage`)** — fixed: the predicate now requires `usage !== null` so a message with `usage: null` no longer matches (it was runtime-safe via the later `if (!usage)` guard, but the predicate was lying).
- (3) **(missing test)** — added a `resolveStatusCode` case asserting a failing `openclaw.outcome` (`abandoned`) overrides an OTel `ok` status → `error`. That's the intended contract: an explicit outcome failure wins.

No code change for the other two:
- (2) **(scope vs attribute gating)** — left as-is per @neoxelox's approval and the thread above: `openclaw.content.output_messages` is vendor-prefixed so collision with another provider is effectively impossible, and a non-OpenClaw span simply lacks the attribute and falls through to normal usage resolution. Happy to thread `scopeName` into `resolveUsage` as defense-in-depth if preferred.
- (4) **(`OPENCLAW_OK_OUTCOMES` lowercasing)** — intentional; no change.

`pnpm --filter @domain/spans test` → 868 passing, typecheck clean.

PR #3670, opened 2026-06-24
fix(core): stop enterprise setup from promoting every user to admin

andresgutgon:
## Problem

A self-hosted/enterprise customer reported that **all invited users were being converted to platform admins**.

`users.admin` is the platform/backoffice **super-admin** flag (gates `/backoffice`, admin procedures, cross-workspace search, billing, Latte debug) — not a workspace role.

## Root cause

The container startup command runs the setup script on **every boot**:

```
# packages/core/docker/Dockerfile:40
CMD sh -c "... && pnpm --filter @latitude-data/core scripts:setup"
```

In enterprise mode, `scripts/setup.ts` → `main()` called `migrateExistingUsersToAdmins()`, which did:

```sql
UPDATE users SET admin = true WHERE admin = false;
```

This indiscriminately promoted **every** user across all workspaces. So users invited correctly with `admin: false` (via `inviteUser` → `createUser`) were re-promoted to platform admin on every deploy/restart. The in-app invite flow itself was never the problem — this startup step overwrote it each boot.

## Fix

- Replace `migrateExistingUsersToAdmins` with **`ensureAtLeastOneAdmin`**: keeps the original intent (the instance always has at least one admin) but only promotes a **single** user (the oldest) when none exists, and is a no-op when an admin is already present.
- Harden `setupService` so that in enterprise mode only the **first** user of an empty instance becomes admin; subsequent signups/invited users default to non-admin.

## Backoffice admin toggle

Adds an **"Admin Access" toggle** to the Admin Tools section of the backoffice user detail page (`/backoffice/search/user/:email`) so an existing admin can grant/revoke platform admin access for any user.

- Gated by the `withAdmin` procedure (backoffice middleware).
- New core service `setUserAdmin` **rejects revoking the last remaining admin** (covers the last admin trying to demote themselves) — the instance can never be left without an admin.

## Notes / follow-up

- This stops *future* re-promotion. Instances that already ran the old script have everyone at `admin: true` already — demoting them is a separate, deliberate data migration (not included here).

## Verification

- `pnpm test src/services/users/setupService.test.ts` → 6/6 pass
- `pnpm test src/services/users/setAdmin.test.ts` → 4/4 pass
- `pnpm tc` (core + web) → clean

### DIFF
--- a/apps/web/src/actions/admin/users/setUserAdminAction.ts
+++ b/apps/web/src/actions/admin/users/setUserAdminAction.ts
@@ -1,0 +1,29 @@
+'use server'
+
+import { z } from 'zod'
+
+import { withAdmin } from '../../procedures'
+import { setUserAdmin } from '@latitude-data/core/services/users/setAdmin'
+import { unsafelyFindUserByEmail } from '@latitude-data/core/queries/users/findByEmail'
+import { NotFoundError } from '@latitude-data/constants/errors'
+
+export const setUserAdminAction = withAdmin
+  .inputSchema(
+    z.object({
+      userEmail: z.string(),
+      admin: z.boolean(),
+    }),
+  )
+  .action(async ({ parsedInput }) => {
+    const user = await unsafelyFindUserByEmail({
+      email: parsedInput.userEmail,
+    })
+
+    if (!user) {
+      throw new NotFoundError(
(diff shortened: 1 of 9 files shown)


andresgutgon:
@claude review

geclos (approved):

PR #3690, opened 2026-06-25
feat(signals): user-authored signals over the API and MCP

csansoon:
Signals have been discovery-only: the flagger and annotation pipeline clusters traces into `system` signals, and a user could "track" one with a generated judge. This PR makes signals user-authorable — a person creates, edits, and deletes a signal with its own evaluation over the public REST API and MCP. It is Phase 2 / PR3 of the signals work; discovery, monitors, and alerts are unchanged.

## origin: how a signal was created

New immutable `origin` column on `signals` (`user` | `system`), distinct from the existing `source` (`annotation` | `flagger` | `custom`, which records score provenance). `origin` gates behavior:

- `user` signals must have an evaluation and are never annotation-assignable.
- `system` signals are discovery-born and stay annotation-assignable even after a judge is generated.

`origin` is set at creation and never changes.

## evaluations: scripts beyond aligned judges

A signal's evaluation no longer has to be an annotation-aligned judge:

- `alignment` / `aligned_at` are nullable — a raw or deterministic script has no confusion matrix.
- `settings` (a declarative config; today `{ kind: "judge", criteria }`) compiles to the stored `script` on save via the sandbox, so capability detection and runtime parity match discovered judges.
- `script_hash` decouples each score's `metadata.evaluationHash` from `alignment` (backfilled from `alignment.evaluationHash` for existing rows).
- A partial-unique index enforces one active evaluation per signal.

## the QuickJS sandbox is the only execution path

The `evaluation-sandbox-runtime` feature flag and the legacy template-extraction executor are gone. Every evaluation script — live scoring, optimization, and alignment — runs through the QuickJS sandbox. Deterministic (non-judge) scripts now execute instead of being rejected by the old judge-template guard, which is what makes user-authored deterministic signals scorable.

## create / update / delete

New `POST`, `PATCH`, and `DELETE` routes under `/signals` (mirrored as MCP tools; TS and Python SDKs regenerated). The use-cases:

- **create** inserts an `origin: user` signal (no centroid) plus its evaluation, compiling `settings` or accepting a raw `script`; a bad script fails compilation with 422. Evaluation-less user signals are rejected. The evaluation scores forward only — new incoming traces from creation onward, with no historical backfill.
- **update** edits metadata/filters/priority and re-compiles + re-hashes an edited script; scoring is forward-only — existing rows are never re-scored.
- **delete** soft-deletes the signal and archives its active evaluation so it stops running on live traffic.

## UI

- The signals list shows every signal regardless of occurrence count, so a freshly created signal appears immediately (occurrence stats are synthesized as zero until scoring runs).
- The signal-detail evaluation section renders four states by origin: automatically evaluated (system, no eval), automatically aligned (system, eval — realign/remove available), custom script via API/MCP (user, raw script — read-only), and custom from settings (user, settings — settings editor deferred). Realign and remove apply only to system-origin evaluations.

## migration

One additive Postgres migration: the new `signals`/`evaluations` columns, the nullable relaxations, the active-detector index, a `script_hash` backfill, and a dedupe to one active evaluation per signal. No `scores` schema change and no ClickHouse change.

## scope and follow-ups

- The signal builder UI lands in PR4; the settings editor in the evaluation section is deferred with it.
- GEPA-from-builder and judge-builder are Phase 3; default-monitor provisioning is Phase 4.
- `origin` is not yet on the public API read response — the web reads the domain entity directly. It can be added when an external consumer needs to read it back.

## validation

Domain use-case tests (PGlite/chdb testkits) cover create rejecting evaluation-less user signals, the settings→script and raw-script paths, forward-only update, and soft-delete + evaluation archival. Full workspace typecheck, knip, and biome pass.


csansoon:
@claude review this PR

andresgutgon (commented):
Remember changelog and bump of version in SDKs

csansoon:
Thanks for the review — all of it is addressed across three commits on this branch:

- `50bc700ae` — enforce the UI's signal-origin rules in the API/MCP/domain
- `dc673560b` — backfill span annotation
- `19a6b1b9a` — SDK `6.2.0` bump + changelogs

## @claude review

**🔴 `monitorSignal` could realign a `user`-origin evaluation** — Fixed in `50bc700ae`. `monitorSignalUseCase` now rejects `origin: "user"` with a `BadRequestError` before either workflow path runs. The guard is threaded via a `signalOrigin` input so it covers the API route, MCP, *and* the web server fn (not just one call site), and there's a test asserting a user signal starts no `optimizeEvaluationWorkflow` and writes no outbox event. This matches the UI, which already hides realign for user signals — their evaluation is a self-managed script and must never be overwritten by the LLM alignment workflow.

**🟡 PATCH/DELETE unrestricted on system-origin signals** — Addressed in `50bc700ae`, per the product rule that the API mirrors the UI:
- `deleteSignal` now rejects `origin: "system"` ("ignore them instead") — discovery-born signals are dismissed through the lifecycle, not deleted.
- `updateSignal` only edits `name`/`description`/`filters` (the evaluation pre-gate, which the UI lets you change for both origins). There is **no manual eval-script update path** for any origin — a system signal's evaluation script is only ever changed by the realign workflow, and a user signal's is set at creation. So "can't manually edit a system eval's script" already holds with no extra guard needed.

**🟡 `updateSignal` doesn't propagate name/description to the evaluation** — Fixed in `50bc700ae`. Renaming a signal now syncs the active evaluation's `name` (it's created equal to the signal name), with a test.

**🟢 `signalId` carried but unused in `backfillSignalScoresUseCase`** — Done in `dc673560b`: it now `annotateCurrentSpan`s `projectId`/`signalId`/`evaluationId`, matching the create/delete use-cases.

**🟢 Unvalidated `JSON.parse` of the backfill cursor** — Acknowledged, left as-is. There's no existing `TraceListCursor` zod schema (it's a plain interface), and the cursor is written internally one line earlier, so adding+exporting a schema in `@domain/spans` isn't worth the cross-package churn for this low-risk path. Happy to add one if you'd prefer the belt-and-suspenders.

**🟢 No test for `monitorSignal` on a `user`-origin signal** — Added in `50bc700ae` (see the 🔴 item).

The non-blocking observations all check out and need no change — thanks for confirming the `aggregateType: "issue"` / `source: "custom"` / active-detector-index intent.

## @andresgutgon — changelog + SDK version bump

Done in `19a6b1b9a`. The regenerated `create`/`update`/`delete` signal endpoints are additive, so both SDKs go `6.1.0 → 6.2.0` (minor):
- `packages/sdk/typescript/package.json` and `packages/sdk/python/pyproject.toml`
- `## [6.2.0]` entries in both `CHANGELOG.md` files documenting `client.signals.create/update/delete` and the now-optional `Evaluation.alignment`/`alignedAt` fields.

(The version lives only in the hand-written package shell — Fern owns just `src/` — so the bump and changelogs survive the next `generate:sdk`.)

andresgutgon (dismissed):
LGTM

PR #3699, opened 2026-06-26
fix(web): restrict prompt editing when Prompt Management is off

neoxelox:
## Problem

When the **Prompt Management** product feature is turned off, the workspace is in observability-only mode — documents exist purely as auto-created containers for incoming traces (created/merged during span ingestion in `captureReferences.ts`, which is intentional). However, users could still **add/delete/rename prompts** in production, which was not intended.

Two leaks caused this:

1. **File-tree per-node menus were not gated by the feature flag.** The top toolbar buttons (`TreeToolbar`) correctly hid "New file/New folder" when the flag was off, but the per-node context menus (`DocumentHeader`/`FolderHeader`) gated rename/delete/new and the "set as main prompt" star only on `isMerged` — so they appeared on any draft commit.
2. **Draft (non-live) version URLs were directly reachable**, exposing the whole editing surface even with the flag off.

## Fix

- **Hide editing affordances on the file tree** — thread the existing `promptManagement` prop from `FilesTree` → `FileNode`/`NodeHeader`/temp nodes → `DocumentHeader` & `FolderHeader`. When the flag is off, the action lists return `[]` (so `NodeHeaderWrapper` renders no menu) and the "set as main prompt" star is hidden (`isMainDocument`/`setMainDocument` pass `undefined`, so `MainPromptIcon` returns `null`).
- **Redirect non-live versions to live** — in `CommitLayout`, when `promptManagement` is off and the requested commit is not the live `HEAD`, redirect to the live version's issues page (the canonical landing for observability-only mode). Targets the literal `HEAD_COMMIT` so there's no redirect loop, and only when a head commit exists (so projects with no merged version don't bounce to a 404).

When Prompt Management is **on**, behavior is completely unchanged.

## Scope / notes

- Trace-ingestion auto-creation of documents is intentionally left as-is — that's the intended observability behavior.
- This is a UI-level fix. The server-side write paths (web actions, gateway endpoints, core services) still don't enforce the flag; true server-side enforcement would be a separate, more invasive change and can follow up if desired.

## Testing

- `pnpm tc` (apps/web) passes.
- Manual: with a workspace's `promptManagerEnabled = false`, opening a draft commit URL bounces to `…/versions/live/issues`, and file/folder right-click menus no longer show rename/delete/new or the main-prompt star.

### DIFF
--- a/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/layout.tsx
+++ b/apps/web/src/app/(private)/projects/[projectId]/versions/[commitUuid]/layout.tsx
@@ -35,6 +35,7 @@ export default async function CommitLayout({
   let session: SessionData
   let project: Project
   let commit: Commit | undefined
+  let headCommit: Commit | undefined
   let isHead = false
   const { projectId, commitUuid } = await params
   try {
@@ -47,7 +48,7 @@ export default async function CommitLayout({
       workspaceId: workspace.id,
     })
 
-    const headCommit = await getHeadCommitCached({
+    headCommit = await getHeadCommitCached({
       workspace,
       projectId: project.id,
     })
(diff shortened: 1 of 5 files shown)


neoxelox:
@claude review

andresgutgon (approved):

PR #3708, opened 2026-06-26
fix: speed up signals list loading

geclos:
## What changed

Speeds up the Signals page server reads by avoiding the expensive "load every signal in the project" pass for table-page reads, and by letting the analytics request skip page item enrichment it does not render.

Datadog showed `listSignals` taking 4–18s and `getSignalsAnalytics` taking up to ~9s in production. The slow traces spent most of their time inside `issues.listSignals` before the instrumented ClickHouse spans; the ClickHouse child spans were sub-250ms. The uninstrumented work mapped to the default-listing path loading every signal from Postgres in 500-row pages before the table-row query or analytics response.

## Behavior preserved

The table data path still uses `signalRepository.listTableRows`, so filtering and ordering by Signals table columns remain server-side and unchanged. Per-row aggregate values are still loaded separately through `getSignalRowMetrics`, preserving occurrences, affected sessions %, and row trend behavior.

For the analytics-only request, the use case still computes the filtered candidate set, counts, priority counts, assignee counts, occurrence sum, and histogram, but it no longer runs page-only enrichments (`listBySignalIds`, trend, tags, occurrence detail) or returns page items.

## Validation

- `pnpm --filter @domain/signals test -- src/use-cases/list-signals.test.ts`
- `pnpm --filter @domain/signals typecheck`
- `pnpm --filter @domain/signals check`
- `pnpm --filter @app/web check` (passes with one pre-existing optional-chain warning in `alignment-stats-modal.tsx`)

### DIFF
--- a/apps/web/src/domains/signals/signals.functions.ts
+++ b/apps/web/src/domains/signals/signals.functions.ts
@@ -331,7 +331,7 @@ type ListSignalsRequest = z.infer<typeof listSignalsInputSchema>
 
 const runSignalsList = async (
   data: ListSignalsRequest,
-  options: { readonly includeAnalytics: boolean },
+  options: { readonly includeAnalytics: boolean; readonly includeItems?: boolean },
 ): Promise<SignalsListResultRecord> => {
   const { organizationId, userId } = await requireSession()
   const orgId = OrganizationId(organizationId)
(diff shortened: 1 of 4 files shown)


geclos:
@claude review

PR #3710, opened 2026-06-26
feat(api): implement signup via mcp

### LINKED ISSUE #3385: [Feature]: Implement signup via MCP
### What problem are you trying to solve?

Add support for user signup via MCP.

### What would you like to happen?

_No response_

### Alternatives you've considered

_No response_

rohankshah:
## What does this PR do?
Added a tool `createAccount` which exposes magic link signup flow via MCP
It replicates the behaviour of `betterAuth.api.signInMagicLink` to ensure the produced rows and the side effects are identical

Also added relevant test case

Closes #3385

## How was this tested?
Tested this using Gemini CLI to make the tool calls.

## Checklist

- [x] I have signed the CLA


packages/domain/events/src/testing/fake-outbox-event-writer.ts
```diff
@@ -0,0 +1,15 @@
+import { Effect } from "effect"
+import type { OutboxEventWriterShape, OutboxWriteEvent } from "@domain/events"
+
+const writtenEvents: OutboxWriteEvent[] = []
+
+export const createFakeOutboxEventWriter = () => {
```
andresgutgon, line 4: This goes in `@domain/events`

apps/api/src/routes/account.ts
```diff
@@ -57,6 +66,11 @@ const accountFernGroup = {
   "x-fern-sdk-method-name": "get",
 } as const
 
+const createAccountFernGroup = {
+  "x-fern-sdk-group-name": "account",
```
andresgutgon, line 78: Ask the llm to generate fern SDKs. After that ask to bump the versions of those SDKs and also to generate the entry in their `CHANGELOG.md`

packages/domain/users/src/use-cases/create-account.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const realBaseURL = new URL(input.webUrl)
+
+  const pathname = realBaseURL.pathname === "/" ? "" : realBaseURL.pathname
+  const basePath = "/api/auth"
+  const url = new URL(`${pathname}${basePath}/magic-link/verify`, realBaseURL.origin)
+  url.searchParams.set("token", verificationToken)
+  url.searchParams.set("callbackURL", "/")
+
+  yield* oAuthKeyRepo.createVerificationValue({
```
andresgutgon, line 81: What happen if the email already exists in Latitude? If this case is not handle. Manage it and throw an error that the api can return so the agent using the mcp can understand

packages/domain/events/src/testing/fake-outbox-event-writer.ts
rohankshah: Moved this to `@domain/events`

apps/api/src/routes/account.ts
rohankshah: This has been done

packages/domain/users/src/use-cases/create-account.ts
rohankshah: Handled case when email already exists. API returns a message 'Email already exists'. Also added a test case for this

apps/api/src/routes/account.ts
```diff
@@ -91,6 +113,47 @@ const getAccount = accountEndpoint({
...
+        withPostgres(
+          Layer.mergeAll(UserRepositoryLive, OAuthKeyRepositoryLive, OutboxEventWriterLive),
+          c.var.postgresClient,
+          c.var.organization.id,
+        ),
+        withTracing,
+      ),
+    )
+
+    console.log("response", response)
```
andresgutgon: 🔥

rohankshah: Let me remove this. 
Btw, I generated the sdks before rebasing, hence the merge conflicts.
I was thinking of going back to the previous commit => rebase with development => remove console.log and then generate the sdks again. 
Do you think this is a good idea?

andresgutgon: What happen if you run fern again. Does it solves the conflicts?

andresgutgon: Wait no. Those conflicts has to be solved manually. Keep the CHANGELOG from the `development` branch and also keep yours. This is not related with fern is a normal git conflict. Same with package json version. Check what version is now in `development` and bump it

rohankshah: Got it. Let me do that

andresgutgon (dismissed):

andresgutgon (approved):

neoxelox:
Thanks for the PR @rohankshah, and for getting deep into the MCP surface — really appreciate the effort here. 🙏

Before any line-level details, I think we should step back on the overall design, because it runs into how our MCP auth actually works:

**Every MCP tool already requires an authenticated caller.** Our MCP server is OAuth2 + Dynamic Client Registration, and auth is enforced at the transport level — an unauthenticated call to `/v1/mcp` is rejected with `401` before any tool executes. A valid token is also always bound to an existing user **and** a selected organization. So to invoke `createAccount`, the caller must already have an account and an org — which means it can't act as a "sign up to Latitude" entry point (you'd need an account to create one).

Since it's always invoked from inside an authenticated org, what this is really shaped like is our existing **`inviteMember`** tool — bringing a person into your organization by email. And `inviteMember` already does that correctly: it links the new person to the caller's org with a role and sends the email through our invitation flow.

`createAccount` as written doesn't do that part, though — the new user is never tied to the caller's organization (the verification only carries the email, and the event is emitted under `organizationId: "system"`), so the recipient just gets a generic magic-link signup and lands in onboarding disconnected from whoever "invited" them. So it ends up both redundant with `inviteMember` and incomplete as an invitation.

My suggestion would be to lean on the existing member-invitation path rather than adding a new tool. If you had a different scenario in mind that `inviteMember` doesn't cover, let's talk it through — happy to help shape it! 

(There are a few correctness/security details too — e.g. it hand-rolls better-auth's magic-link instead of going through the existing flow — but those are moot until we settle the direction.)

Overall I think doing a sign up fully automated by an agent through MCP (or other way) its actually a great idea but critical at the same time. We think this is core and we have to step back and think on how to design this ourselves.

Thanks again for your contribution!

PR #3734, opened 2026-06-29
feat(evaluations): run evaluations against a session runtime context

csansoon:
## summary

Evaluation scripts now run against a single `session` object instead of the lossy `conversation` context. It lets later PRs add deterministic conditions over metrics, tools, and status, and gives LLM judges the full session transcript — while keeping the existing per-trace TraceEnd pipeline and membership model unchanged.

This is **PR4a** of the signals Phase 2 builder work (`specs/signals.md`). PR4b (settings-defined conditions + codegen) and PR4c (builder UI) follow.

## the `session` object

A script's sole runtime global is now `session` (the old `conversation` / `issue` / `signal` globals are gone), assembled at evaluation time from the triggering trace's session. Base units: ns (`duration` / `timeToFirstToken`), microcents (`cost`), token counts. Schema:

```json
{
  "id": "string",
  "traceCount": "number",
  "spanCount": "number",
  "errorCount": "number",
  "duration": "number",
  "timeToFirstToken": "number",
  "cost": { "input": "number", "output": "number", "total": "number" },
  "tokens": {
    "input": "number",
    "output": "number",
    "total": "number",
    "cacheRead": "number",
    "cacheCreate": "number",
    "reasoning": "number"
  },
  "startTime": "string",
  "endTime": "string",
  "userId": "string",
  "tags": ["string"],
  "metadata": { "[key]": "string" },
  "conversation": [{ "role": "string", "content": "string" }],
  "traces": [
    {
      "id": "string",
      "name": "string",
      "status": "string",
      "errorCount": "number",
      "spanCount": "number",
      "duration": "number",
      "timeToFirstToken": "number",
      "cost": { "input": "number", "output": "number", "total": "number" },
      "tokens": {
        "input": "number",
        "output": "number",
        "total": "number",
        "cacheRead": "number",
        "cacheCreate": "number",
        "reasoning": "number"
      },
      "models": ["string"],
      "providers": ["string"],
      "finishReasons": ["string"],
      "tools": [
        {
          "name": "string",
          "input": "string",
          "output": "string",
          "error": "boolean",
          "duration": "number"
        }
      ]
    }
  ]
}
```

`session.conversation` is the deduped, session-wide transcript (its `toString()` renders `[role] content` lines) — the single message store judges interpolate as `${session.conversation}`. There is no raw per-span array: spans are rolled up per trace and tool spans are projected, so the object stays small even for long sessions.

## assembly

`loadScriptSessionContext` (`@domain/evaluations`) builds the object:

- conversation + aggregates from `SessionRepository.findBySessionId` (the deduped system + last-input + outputs reconstruction, same one conversation-intelligence uses). Orphan sessions with no rollup row fall back to the triggering trace.
- per-trace rollups from `SpanRepository.listBySessionId`.
- tool I/O from a new focused `SpanRepository.listToolSpansBySessionId` read — the light `Span` projection omits `tool_input`/`tool_output`, so this returns `execute_tool` spans with their I/O, error, and duration.

`run-live-evaluation` assembles the session and passes it to the executor; the alignment and optimization paths wrap a single example's conversation via `minimalScriptSession` (judges read only `session.conversation`).

## judge generation

The baseline and settings judge templates and the GEPA proposer now target `${session.conversation}`, so generated and optimized judges read the new global.

## scope and decisions

- **Triggering is unchanged.** Evaluations still run per trace at TraceEnd and membership stays keyed to the trace; the session is loaded at eval time from the trace's session. Moving to true session-settle triggering is a deferred follow-up.
- **No script migration.** This assumes no production evaluations on the old `conversation` global — existing stored scripts that reference it would break, and new generation uses `session.conversation`.
- Includes a spec refresh (`specs/signals.md`): Phase 1 / PR2 / PR3 marked complete, and the remaining builder work split into PR4a / PR4b / PR4c.

## testing

- Full workspace typecheck passes.
- Unit tests cover the sandbox binding (`session.conversation` view, legacy globals gone), the loader assembly (conversation fallback, per-trace rollups, tool projection + truncation), and the updated live / alignment execution paths.
- The new `listToolSpansBySessionId` ClickHouse query is exercised in CI (chdb).

### DIFF
--- a/apps/workers/src/workers/live-evaluations.ts
+++ b/apps/workers/src/workers/live-evaluations.ts
@@ -11,6 +11,8 @@ import {
 import {
   type ClickHouseClient,
   ScoreAnalyticsRepositoryLive,
+  SessionRepositoryLive,
+  SpanRepositoryLive,
   TraceRepositoryLive,
   withClickHouse,
 } from "@platform/db-clickhouse"
@@ -161,7 +163,7 @@ export const createLiveEvaluationsWorker = ({
       Effect.provide(RedisBillingSpendReservationLive(rdClient)),
       Effect.provide(RedisDetectorHealthTrackerLive(rdClient)),
       withClickHouse(
-        Layer.mergeAll(ScoreAnalyticsRepositoryLive, TraceRepositoryLive),
+        Layer.mergeAll(ScoreAnalyticsRepositoryLive, TraceRepositoryLive, SessionRepositoryLive, SpanRepositoryLive),
         chClient,
         OrganizationId(payload.organizationId),
       ),
--- a/packages/domain/evaluations/src/codegen/compile-settings-to-script.test.ts
+++ b/packages/domain/evaluations/src/codegen/compile-settings-to-script.test.ts
@@ -25,4 +25,11 @@ describe("compileSettingsToScript", () => {
     const interpolations = script.match(/\$\{[^}]+\}/g) ?? []
     expect(interpolations).toEqual([EVALUATION_CONVERSATION_PLACEHOLDER])
   })
+
+  it("interpolates session.conversation, not the legacy bare conversation global", () => {
+    expect(EVALUATION_CONVERSATION_PLACEHOLDER).toBe("${session.conversation}")
+    const script = compileSettingsToScript({ kind: "judge", criteria: "x" })
+    expect(script).toContain("${session.conversation}")
+    expect(script).not.toContain("${conversation}")
+  })
 })
--- a/packages/domain/evaluations/src/index.ts
+++ b/packages/domain/evaluations/src/index.ts
@@ -174,12 +174,10 @@ export {
 export {
   type ExecuteLiveEvaluationError,
   executeLiveEvaluationUseCase,
-  type LiveEvaluationConversationInput,
   type LiveEvaluationExecutionInput,
   type LiveEvaluationExecutionResult,
   type LiveEvaluationResultPayload,
   type LiveEvaluationSignalContext,
-  liveEvaluationConversationInputSchema,
   liveEvaluationExecutionInputSchema,
   liveEvaluationExecutionResultSchema,
   liveEvaluationResultPayloadSchema,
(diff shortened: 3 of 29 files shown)


csansoon:
@claude review this PR

csansoon:
@cursoragent review this pr

csansoon:
Thanks @cursoragent and Claude — addressed the actionable items.

**Fixed**
- Per-trace `tokens.total` now sums all components (`input + output + cacheRead + cacheCreate + reasoning`) so the breakdown is self-consistent with the session rollup, not just input + output — `ad6783c09`
- The three session reads (`findBySessionId` / `listBySessionId` / `listToolSpansBySessionId`) now run concurrently via `Effect.all` — `ad6783c09`
- Named the `MS_TO_NS` constant — `ad6783c09`
- Deep-freeze the whole `session` payload (traces, their arrays, tools), not just `conversation`; a runtime test asserts traces/tools are frozen — `d9ccefc02`
- Added the session-rollup loader path test (conversation + aggregates come from `SessionDetail`, not the trace) and a root-span fallback case — `a2ad1b425`
- Added a chdb integration test for `listToolSpansBySessionId` (execute_tool filter, span-id dedupe, start-time ordering, I/O + error mapping, orphan-session membership) — `c1fe59a65`. This also makes the description's "exercised in CI (chdb)" accurate.

**Deferred / acknowledged**
- *Breaking change (old globals removed):* assumes no production evaluations reference `conversation` / `issue` / `signal`. I'll confirm with an `evaluations.script LIKE` query before merge; new generation only emits `${session.conversation}`.
- *Signal metadata on `session`:* judges still get signal context baked into the generated prompt text; exposing `session.signal` for deterministic conditions is a PR4b concern (noted there).
- *App-side span aggregation / full-column fetch:* acceptable for PR4a; tracked as a follow-up perf item alongside the now-concurrent reads.
- *"lossy" JSDoc wording:* kept — it refers to the part-flattening in `formatGenAIMessage`, which is accurate.

Full-workspace typecheck green; evaluations 100 / sandbox-quickjs 36 unit tests pass (the chdb test runs in CI).

geclos:
@csansoon make sure evaluation scripts run after the session has been semantically embedded so that any semantic evaluations will run with embeddings for that session already computed

csansoon:
> @csansoon make sure evaluation scripts run after the session has been semantically embedded so that any semantic evaluations will run with embeddings for that session already computed

will do on a future PR, when adding support for semantic similarity. It will be its own PR

andresgutgon (approved):

PR #3740, opened 2026-06-29
fix(web): stop signals list priority sections from fragmenting

csansoon:
The signals list groups signals into priority sections (Urgent → High → Medium → Low → No priority). Users saw the same section repeated many times (`NO PRIORITY 128` over and over), higher-priority sections rendered *below* lower-priority ones, headers with no rows under them, and stacks of empty headers after scrolling down to load more and back up.

https://github.com/user-attachments/assets/88c27cc1-5f27-495d-8ac4-a8b45754dbfd

Two independent defects stacked.

## Rendering — the impossible-looking visuals

`InfiniteTable` grouped rows by *run-length*: it emitted a section header on every group *transition*. That assumes the data is already perfectly contiguous by group. The moment it isn't, you get a fresh header every time the group flips back — and because the model always pushes a data row after a header, two adjacent headers can't exist in the row model at all, which is the tell that the stacked/empty headers were a paint artifact, not the real order.

The artifact came from the virtualizer: it had no `getItemKey`, so `measureElement` cached row heights by array index. Every page load grows the display-row list and every reorder shifts index→row mapping, so cached heights got mis-attributed and rows/headers painted at stale offsets.

- Grouping now buckets rows by group in first-appearance order — exactly one header per group, all its rows contiguous — regardless of input order. Extracted as a pure `buildDisplayRows()` (`packages/ui/src/components/infinite-table/display-rows.ts`) with unit tests.
- The virtualizer gets a stable `getItemKey` (`group-…` / `row-…` / `skeleton-…`) so measured heights survive page growth and reorders.

Both grouping consumers (the signals list and the design-system demo) feed contiguous data and don't rely on a group key appearing in multiple places, so bucketing is strictly safer for both.

## Data — the aggravator

Each page is an offset slice of a re-sorted, mutating candidate set, fetched at different times and `flatMap`-concatenated. That returns the same signal in adjacent pages (duplicate React keys, repeated rows) and leaves priorities non-contiguous across page boundaries, which is what tripped the run-length grouper.

`useSignals` now dedupes loaded pages by signal id (earlier page wins, preserving top-of-list order). This is the lighter of the two options considered — it fixes the duplicate rows/keys without a backend change. The deeper fix (freezing an `asOf` ordering snapshot threaded through every page request) was deliberately deferred: it changes behavior (brand-new signals wouldn't appear until the view re-snapshots) and isn't needed to resolve the reported symptoms. The list stays live, so rows can still subtly reshuffle within a section while scrolling — acceptable for now.

The `No priority 128` count is unchanged: it's the global per-priority total from the analytics query and reads correctly once headers are no longer duplicated.

## Testing

- New `display-rows.test.ts` covers the bucketing, including the exact reported fragmentation (`none → medium → none → high → none` collapsing to three ordered sections). `pnpm --filter @repo/ui test` passes.
- `@repo/ui` and `@app/web` typecheck clean; Biome clean. Backend (`list-signals.ts`) untouched.

### DIFF
--- a/apps/web/src/components/signals/signal-priority-meta.tsx
+++ b/apps/web/src/components/signals/signal-priority-meta.tsx
@@ -33,3 +33,11 @@ export const SIGNAL_PRIORITY_META: Record<SignalPriorityGroupId, SignalPriorityM
   low: { label: "Low", icon: SignalLowIcon, iconColor: "foregroundMuted" },
   none: { label: "No priority", icon: MinusIcon, iconColor: "foregroundMuted" },
 }
+
+/**
+ * Fixed section order for the grouped issues list: most urgent first, "no
+ * priority" last. Section order is independent of the selected sort (which
+ * applies within each section). Mirrors `SIGNAL_PRIORITY_GROUPS` in
+ * `@domain/signals`.
+ */
+export const SIGNAL_PRIORITY_GROUP_ORDER: readonly SignalPriorityGroupId[] = ["urgent", "high", "medium", "low", "none"]
--- a/apps/web/src/domains/signals/signals.collection.ts
+++ b/apps/web/src/domains/signals/signals.collection.ts
@@ -255,7 +255,21 @@ export function useSignals(input: {
     }),
   })
 
-  const data = useMemo(() => pages.flatMap((page) => page.items), [pages])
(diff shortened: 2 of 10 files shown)


csansoon:
@claude, @cursoragent: review this pr

andresgutgon (approved):

PR #3745, opened 2026-06-29
fix: retry transient connection resets in ClickHouse client

geclos:
## What does this PR do?

Adds automatic retry logic for transient TCP connection resets when querying ClickHouse. When a pooled keep-alive socket is reused in the narrow window after the server (or load balancer) has closed it, the request fails with `ECONNRESET`, `EPIPE`, or "socket hang up" before reaching the server — making it safe to retry.

The fix:
1. **Detects connection reset errors** via a new `causesIncludeConnectionReset()` utility that walks error cause chains for `ECONNRESET`, `EPIPE`, or "socket hang up" messages
2. **Configures socket TTL** in the ClickHouse client to discard idle sockets before the server's keep-alive timeout (2.5s vs ClickHouse's 3s default), reducing the race window
3. **Retries up to 2 times** with a 50ms delay between attempts in `ChSqlClientLive`, allowing stale sockets to be evicted without masking real outages

Closes #

## How was this tested?

- Added comprehensive unit tests in `ch-sql-client.test.ts` covering:
  - Successful retry after transient resets
  - Exhaustion of retries on persistent failures
  - Non-reset errors are not retried
  - Errors are wrapped as `RepositoryError`
- Added unit tests for `causesIncludeConnectionReset()` in `errors.test.ts` covering:
  - Detection of `ECONNRESET` and `EPIPE` codes
  - Detection of "socket hang up" by message
  - Walking nested cause chains (e.g., wrapped `RepositoryError`)
  - Circular cause chain termination
  - False negatives for unrelated errors

## Checklist

- [x] I have signed the CLA

https://claude.ai/code/session_01MQF7jm5UEWmTRgAe4rFgn4

### DIFF
--- a/packages/domain/shared/src/errors.test.ts
+++ b/packages/domain/shared/src/errors.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, it } from "vitest"
 import {
   BadRequestError,
   ConflictError,
+  causesIncludeConnectionReset,
   NotFoundError,
   PermissionError,
   RepositoryError,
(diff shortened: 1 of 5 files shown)


geclos:
@cursoragent review this PR. Ensure the root cause for the bug is real and that the fix is sound.

PR #3773, opened 2026-07-01
feat(signals): builder UI (R1) — create, edit & preview signals from the web

csansoon:
R1 is the last MVP piece of Signals: a UI to author a signal without the API. Until now the signal detail page said "Editing its settings is coming soon" and the list had no way to create one. This adds a single create/edit builder with an on-demand live preview, the backend use-cases it needs, and the entry points into it.

https://github.com/user-attachments/assets/f6eb0386-a093-4b27-9ec6-28bb18c07825

## the builder

A stepped modal (`signal-builder-modal.tsx`) with a clickable step indicator and a fixed-height body:

- **Scope** — which traces the signal evaluates. Filter dimensions (tags / services / models / providers / metadata) are added from labeled chips and revealed progressively, so the default reads as one line ("evaluates every trace") instead of a wall of empty pickers. A sampling slider sets what share of matching traces are evaluated.
- **Evaluation** — the detector, as a declarative `settings` form across three tabs:
  - **Rules** — deterministic conditions over the `session` object, edited as a list. Adding one opens a nested sub-step: pick one of the ten condition types, then fill its settings (known-value comboboxes for tool names / finish reasons, friendly units for metric thresholds, inline regex validation).
  - **LLM** — a `criteria` textarea compiled to an `llm()` judge script; a trace joins the signal when the described behavior is present.
  - **Advanced** — raw-script authoring is a disabled "Coming soon" tab (wired in R2).
- **Test** — runs the draft evaluation over recent sessions and shows the hit rate, matches first, each as a card (prompt + duration/cost/tokens + verdict reason) that expands in place to full metrics and the conversation. No navigation away from the modal.
- **Details** (create only) — name + description.

Edit mode reuses the same modal for `origin=user`, settings-defined signals; name/description are edited from a separate rename modal on the detail header.

## live preview, worker-backed

The preview never persists a score, and it runs in a worker because `apps/web` can't run LLM generation. `previewEvaluationUseCase` (new, `@domain/evaluations`) compiles the draft `settings`/`script`, lists the latest ~10 sessions matching the draft filters, runs the sandboxed evaluation against each, and returns per-session verdicts plus a summary (first user message, duration, cost, tokens, trace/error counts). Web enqueues a `signals-preview` BullMQ job and polls Redis for the result (`SignalPreviewResult`, 5-minute TTL) — the same poll-a-job shape as alignment state. One path covers rule, judge, and raw script: the same compile → sample → sandbox body the live pipeline runs, minus the write.

## backend use-cases

- `previewEvaluationUseCase` — above.
- `updateSignalEvaluationUseCase` (new, `@domain/signals`) — recompiles a user signal's active evaluation from a new `settings` form. Rejects `origin=system` and raw-script (settings-null) evaluations. Edits are forward-only; existing scores keep their frozen membership.
- `createSignalUseCase` / `createEvaluationUseCase` gain an optional `sampling`.

## two decisions that diverge from the spec text

1. **The live pipeline gates on `signal.filters`, not `evaluation.trigger.filter`.** The spec described propagating a signal's filters into `trigger.filter`; instead `signals:match` now reads `signal.filters` directly (canonical) when selecting evaluations, and `trigger.filter` is retired. A signal with no filters matches everything. Any prod evaluation carrying a non-`{}` `trigger.filter` becomes inert — only the now-removed trigger-filter UI ever wrote one, so the risk is low.
2. **Settings edits recompile in place (same evaluation id), not archive + create.** Archiving the active evaluation and minting a new same-named row collided with the `(org, project, name)` unique index. Rather than invent a naming scheme, this drops that index (the `name` column stays) — one-active-per-signal is already enforced by the active-detector index. In-place recompile also matches the shipped realign path. Migration: `20260630090000_drop-evaluations-unique-name-index`.

## entry points

- **New signal** on the list and empty state.
- **Edit** and **Delete** on the signal detail page, for `origin=user` settings-defined signals (replacing the "coming soon" note). Sampling shows read-only there for user signals, since they edit it in the builder; system signals keep the inline click-to-change control.
- **Create signal from a trace search** — a command-palette action and a card in the save-search modal, shown only when filters are set. Neither auto-creates: both navigate to the Signals page with the builder pre-opened and the search's filters seeded into Scope (via `?newSignal=1&newSignalFilters=…`, cleared after one read). The semantic query isn't seeded yet (R3).

## api cleanup

Dropped the dead `QueuePublisher` layer on `createSignal` and fixed the stale "backfilled over recent history" prose in `apps/api/src/routes/signals.ts` (collection is forward-only). `openapi.json` / `mcp.json` regenerated; no new endpoints — the UI leads here.

## tested

- Domain unit tests: `preview-evaluation.test.ts`, `update-signal-evaluation.test.ts`, `build-trace-end-evaluation-selection.test.ts`.
- `typecheck` green across `@app/web`, `@domain/evaluations`, `@domain/signals`, `@app/workers`.
- Manual UI verification.

## not in this PR

- **Docs sync** — `dev-docs/signals.md` and the `specs/signals.md` R1 section still describe the pre-decision behavior (trigger.filter propagation, archive + create). Follow-up.
- **Advanced tab (R2)** — raw-script authoring and "describe your eval" generation; the tab is present but disabled.
- **Judge alignment** — user-created judges are never aligned (alignment is only for auto-discovered system signals), so the Judge tab compiles criteria to a forward-only `llm()` script with no alignment UI.


csansoon:
@claude @cursoragent review this PR

apps/web/src/routes/_authenticated/projects/$projectSlug/signals/-components/builder/signal-builder-modal.tsx
```diff
@@ -0,0 +1,365 @@
...
+    }
+    setIsSaving(true)
+    try {
+      const result = await createSignal.mutateAsync({
+        name: trimmedName,
+        description: description.trim(),
+        filters: filterSetOrNull(filters),
+        sampling,
+        evaluation,
+      })
```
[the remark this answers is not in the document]
csansoon: Fixed in 5b2e76c08 — `handleCreate` now guards a non-empty description (field error on the Details step) to match `createSignal`'s `z.string().min(1)`, instead of letting an empty string reach the server.

apps/web/src/routes/_authenticated/projects/$projectSlug/signals/-components/signal-rename-modal.tsx
```diff
@@ -0,0 +1,93 @@
...
+      async (value) => update.mutateAsync({ name: value.name.trim(), description: value.description.trim() }),
+      {
+        onSuccess: () => {
+          toast({ description: "Signal updated." })
+          onClose()
+        },
+        onError: (error) => toast({ variant: "destructive", description: toUserMessage(error) }),
+      },
+    ),
+  })
```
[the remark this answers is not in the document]
csansoon: Fixed in 50d9d0f67 — the rename form now has `onSubmit` validators for both name and description, so clearing a field shows a field error instead of submitting an empty string that fails server-side.

packages/domain/signals/src/use-cases/update-signal-evaluation.ts
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Recompiles a user signal's active evaluation from a new `settings` form, **in place** (same
+ * evaluation id). Updating in place — rather than archiving and minting a new row — is required by
+ * the `(org, project, name)` unique index (two non-deleted same-named rows collide) and matches the
+ * shipped realign path (`persistAlignmentResult`). Only `origin = 'user'` signals whose active
+ * evaluation is settings-defined are editable here: raw-script and system signals are rejected.
+ * Edits apply forward-only; existing scores keep their frozen membership. The signal's `filters`
+ * (the live pre-gate) are untouched and continue to gate the same evaluation.
+ */
```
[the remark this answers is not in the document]
csansoon: Fixed in 3469513cd — dropped the stale `(org, project, name)` unique-index rationale (that index is removed in this PR) and kept the reasons that still hold: parity with the realign path and avoiding a lineage/naming scheme.

csansoon:
Thanks for the review — addressed the three bugs and the actionable nits.

**Bugs**
- **1 — empty description on create** → `5b2e76c08`. `handleCreate` now requires a non-empty description (field error on the Details step), matching `createSignal`'s `z.string().min(1)`.
- **2 — empty description on rename** → `50d9d0f67`. Added `onSubmit` field validators for name + description in the rename modal.
- **3 — stale JSDoc** → `3469513cd`. Dropped the `(org, project, name)` unique-index rationale (that index is removed in this PR); kept the realign-path / no-naming-scheme reasons.

**Nits**
- **4 — preview poll not aborted on close** → `5b2e76c08`. `runSignalPreview` now takes an `AbortSignal`; the modal aborts on re-run and on close. This also fixes a stale re-run result overwriting the newer one.
- **5 — `traceIds[0]` looks like a skip** → `a7ec63be4` (comment only). It is not a skip: the first trace only anchors the load — `loadScriptSessionContext` resolves the `sessionId` from it and evaluates the whole session (all traces). Added a comment to that effect.
- **7 — spurious `updateSignal` on edit** → `5b2e76c08`. `handleSaveEdit` now calls `updateSignal` only when the filters actually changed.

**Declined**
- **6 — `describeError` duplicated** — leaving as-is. The two copies live in `@domain/evaluations` and `@app/workers`; sharing a 4-line helper would add a dependency edge between them that is not worth the coupling.

apps/web/src/routes/_authenticated/projects/$projectSlug/signals/$signalId/index.tsx
```diff
@@ -139,6 +159,15 @@ function SignalDetailPage() {
           }
           append={<SignalRelated projectId={project.id} projectSlug={projectSlug} signalId={signalId} />}
         />
+        {renameOpen && issue ? (
```
andresgutgon: `issue`?

apps/web/src/routes/_authenticated/projects/$projectSlug/signals/-components/builder/condition-meta.tsx
```diff
@@ -0,0 +1,535 @@
...
+      return c.toolName.trim().length > 0
+    case "finish_reason":
+      return c.value.trim().length > 0
+    case "empty_output":
+    case "json_output":
+    case "tool_failed":
+    case "error":
+      return true
+  }
+}
```
andresgutgon, line 535: This file is huge. Take a look carefully and look for opportunities to reduce ai slop if you see any. Like not doing DRY at all and repeating things.

apps/web/src/routes/_authenticated/projects/$projectSlug/signals/-components/builder/signal-builder-modal.tsx
```diff
@@ -0,0 +1,397 @@
...
+        await updateSignalEvaluation.mutateAsync({ settings: evaluation.settings, sampling })
+      }
+      await invalidateSignalQueries(projectId, initial.signalId)
+      toast({ description: "Signal updated." })
+      onClose()
+    } catch (error) {
+      toast({ variant: "destructive", description: toUserMessage(error) })
+      setIsSaving(false)
+    }
+  }
```
andresgutgon, line 237: Yeah this file is huge and not using `useForm` so you can use form field validations.

apps/workers/src/workers/billing.test.ts
```diff
@@ -26,12 +26,24 @@ import { outboxEvents } from "@platform/db-postgres/schema/outbox-events"
...
+// partitions around the real wall-clock month (`now()`), which the fake timers below can't move.
+// These tests freeze the clock to fixed 2026 dates and seed usage periods to match, so provision
+// the 2026 month partitions up front — otherwise an insert into an unprovisioned month fails with
+// "no partition of relation billing_usage_events found for row".
+beforeAll(async () => {
+  for (let month = 1; month <= 12; month++) {
+    const partitionStart = `2026-${String(month).padStart(2, "0")}-01`
+    await pg.client.exec(`SELECT latitude.create_billing_usage_events_partition('${partitionStart}'::date)`)
+  }
+})
```
andresgutgon: You dont need this. Is already fixed in development 🔥

packages/platform/db-postgres/src/repositories/alert-incident-repository.ts
```diff
@@ -194,28 +194,28 @@ const makeIncidentRepository = (): IncidentRepositoryShape =>
...
     listByMonitorId: ({ monitorId, limit, cursor }) =>
       Effect.gen(function* () {
         const sqlClient = (yield* SqlClient) as SqlClientShape<Operator>
         const where = and(
-          eq(alertIncidents.organizationId, sqlClient.organizationId),
-          eq(alertIncidents.sourceType, "monitor"),
-          eq(alertIncidents.sourceId, monitorId),
+          eq(incidents.organizationId, sqlClient.organizationId),
+          eq(incidents.sourceType, "monitor"),
+          eq(incidents.sourceId, monitorId),
```
andresgutgon: Why this rename? The file is called alert-incident

packages/platform/db-postgres/src/schema/alert-incidents.ts
```diff
@@ -29,5 +29,3 @@ export const incidents = latitudeSchema.table(
       .where(sql`ended_at IS NULL`),
   ],
 )
-
-export const alertIncidents = incidents
```
andresgutgon, line 33: Not sure why we need to do this refactor

apps/web/src/routes/_authenticated/projects/$projectSlug/signals/$signalId/index.tsx
csansoon: Fixed in e2bc0dbb5 — renamed the legacy `issue` variable to `signal` here, and cleaned up the stale copy in the same file (`Back to issues`, `Copy issue slug`, `This issue could not be loaded`).

apps/web/src/routes/_authenticated/projects/$projectSlug/signals/-components/builder/condition-meta.tsx
csansoon: Refactored in 6b3834479 — extracted the repeated "operator select + numeric value" block (`output_length`, `tool_call_count`, `metric`) into one shared `ComparisonRow`, and dropped the redundant single-`Select` wrapper `div`s. The rest of the file is the declarative per-type config (10 condition types), which I kept as-is. Happy to split the combobox widgets (`SuggestiveCombobox`/`ToolNameField`) into a sibling file if you want the file physically smaller — say the word.

apps/web/src/routes/_authenticated/projects/$projectSlug/signals/-components/builder/signal-builder-modal.tsx
csansoon: The two scalar fields (name/description) now validate inline — create in 2814026e8, rename in 620bc47f2. I held off converting the whole modal to `useForm` though: most of its state isn't form fields (filters, rule conditions, sampling, step index, the condition sub-editor), so `useForm` only really fits those two Details fields, which already show field errors now. I can extract the Details step into a small `useForm` subcomponent, or split this file up, if you'd prefer — what's your call?

apps/workers/src/workers/billing.test.ts
csansoon: Good catch 🙏 — rebased onto `development` (which carries the fix) and dropped my partition-provisioning in 5977e8ae6.

packages/platform/db-postgres/src/repositories/alert-incident-repository.ts
csansoon: Reverted in 01490ef67 — back to `alertIncidents`. The `drizzle-kit generate` "duplicate table" error was a local node-version quirk, not a real issue: `development` keeps the alias, and this PR's migration is hand-authored so it never runs `generate`.

packages/platform/db-postgres/src/schema/alert-incidents.ts
csansoon: Reverted in 01490ef67 — kept the `export const alertIncidents = incidents` alias. Same reasoning as the repository comment: the removal was an unnecessary workaround for a local drizzle-kit/node issue, and `development` keeps it.

csansoon:
Addressed the review + **rebased onto `development`** (force-pushed). Per-thread replies have the details; summary:

- **`issue` → `signal`** on the detail page (+ stale copy) — `e2bc0dbb5`
- **condition-meta DRY** — shared `ComparisonRow`, dropped redundant wrapper `div`s — `6b3834479`
- **billing partition fix dropped** — the rebase brought development's fix — `5977e8ae6`
- **`alertIncidents` alias/refactor reverted** — it was a local node-version drizzle-kit quirk — `01490ef67`

The rebase resolved cleanly; `openapi.json` / `mcp.json` regenerate with no drift, the SDK diff is just the createSignal prose, and `@app/web` / `@app/workers` / `@domain/*` / `@platform/db-postgres` typecheck green.

Left **one thread open** (`signal-builder-modal.tsx` → `useForm`): the two scalar fields already validate inline, but the modal's state is mostly non-form (filters, rule conditions, sampling, steps). I don't think a full `useForm` conversion fits — asked there whether you'd want the Details step extracted into a small `useForm` subcomponent instead. Your call.

andresgutgon (approved):

PR #3784, opened 2026-07-01
feat(sdk): Fern-generated latitude CLI, Latitude* SDK rename, API-key auth (LAT-703)

neoxelox:
Phase 1 of the agentic onboarding experience (`specs/agentic-experience.md`): auto-generate a `latitude` CLI from `apps/api/openapi.json`, the same way Fern already generates the TS and Python SDKs. Along the way it lands the SDK codegen cleanup the v7 Fern toolchain enables — renaming the SDK exports to `Latitude*` and modeling API auth as an org-scoped API key.

Most of the diff is regenerated SDK/CLI output. The reviewable surface is the Fern config, the API-key security scheme, the CLI publish workflow, and the root scripts.

## latitude CLI

A new Fern generator group (`local-cli` in `fern/generators.yml`) emits a single Rust binary into `packages/cli` from the OpenAPI spec — every API resource becomes a typed subcommand (`latitude projects list`, `latitude traces get`, `latitude analytics query`, …), with `--help`, a machine-readable `--schema`, and `latitude auth login` for the OS keyring.

- The generator is `fernapi/fern-cli-generator` (pinned `0.21.0`), with `ir-version: v67` (it isn't in the OSS generator registry, so the IR version has to be pinned) and `customCommands: false` (lean single crate). `packages/cli` is fully generator-owned; only `CHANGELOG.md` and `.fernignore` are hand-maintained.
- Root scripts: `generate:cli`, `generate:all`, `cli:build` (regenerate + compile a local binary), `cli:run` (invoke it). `sdk:check` is renamed `fern:check` (it validates every generator group).
- `publish-cli.yml` (fanned out from `publish-packages.yml` on push to `development`) builds five cross-platform binaries and cuts a GitHub Release. The CLI version comes from `packages/cli/CHANGELOG.md` and is injected at build time (the crate ships `0.0.0`); it's versioned independently of the SDKs, starting at `0.1.0`. Linux binaries are gnu so `auth login`/D-Bus keyring works, with rustls TLS; macOS/Windows use native-tls. `api-manifests.yml` regenerates the CLI and fails on drift.

## SDK changes (breaking — v7)

Both SDKs move to `7.0.0`:

- Exports drop the `Api` infix: `LatitudeApiClient` → `LatitudeClient`, `LatitudeApiError` → `LatitudeError`, etc. (and `import * as LatitudeApi` → `Latitude`).
- The auth argument is renamed `token` → `apiKey` (`api_key` in Python) and now falls back to the `LATITUDE_API_KEY` env var. It's driven from a single `API_SECURITY_SCHEME` constant (`apps/api/src/constants.ts`) carrying `x-fern-bearer`, shared by the live spec (`server.ts`) and the emitter — so the wire format stays `Authorization: Bearer <key>` across HTTP, both SDKs, and the CLI.
- The SDKs are regenerated on the v7 Fern toolchain. The HTTP API surface is unchanged (openapi.json deltas are limited to `info` metadata and the security scheme). Internal usage (`@platform/latitude-api`), examples, and tests are updated to `apiKey`.

## Not in this PR

Phases 2–5 of the spec — the unauthenticated bootstrap endpoint, the web claim flow, the agent skills, and the landing CTA — are follow-ups. The per-generator audience split (ship a future `bootstrap` command in the CLI but exclude it from the SDKs) is deferred to Phase 2, since there's no such endpoint to tag yet.

## Validation

`pnpm typecheck` (82/82), both SDK builds, the Python SDK tests, and `knip` pass. `publish-cli.yml` was exercised end-to-end with `act` (version detection → build → package → artifact upload), and the built CLI was run against production. Manifests and the generated SDKs/CLI are drift-clean against a fresh regenerate.

Refs LAT-703.


csansoon:
@claude review this pr

PR #3786, opened 2026-07-01
fix(signals): return BadRequestError instead of crashing on limit&gt;100

geclos:
## What does this PR do?

Fixes a production 500 in `listSignalsUseCase` (`packages/domain/signals/src/use-cases/list-signals.ts`).

**Datadog issue:** [`d9eced62-744f-11f1-a042-da7ad0900005`](https://app.datadoghq.eu/error-tracking/issue/d9eced62-744f-11f1-a042-da7ad0900005) — `ZodError` on `GET /v1/projects/:projectSlug/signals`, service `api`, 8 occurrences over the last day, still recurring.

**Root cause:** the route's shared `PaginatedQueryParamsSchema` (`apps/api/src/openapi/pagination.ts`) accepts `limit` up to **200** — that's the documented public API contract. `listSignalsUseCase`'s own input schema caps `limit` at **100** and validated it with a raw `listSignalsInputSchema.parse(input)` call. A client sending `limit` between 101 and 200 — valid per the route's own OpenAPI schema — passed the HTTP-layer validation, then hit the stricter internal cap and threw an unhandled `ZodError` straight out of the `Effect.gen` body, surfacing as a raw 500 instead of a clean 400.

Every sibling "list" use-case with the same narrower-than-the-route limit cap (`list-annotations.ts`, `list-scores.ts`) already wraps its `.parse()` in a small `parseOrBadRequest` helper that converts a `ZodError` into a `BadRequestError` (part of the use-case's declared error channel, mapped to HTTP 400 by the route). `list-signals.ts` was the one outlier missing this — this PR brings it in line with the established pattern rather than changing the actual limit semantics.

Closes #

## How was this tested?

Added a regression test in `list-signals.test.ts` that calls `listSignalsUseCase` with `limit: 150` (in-range for the route, out-of-range for the use-case) and asserts the effect fails with a `BadRequestError`.

- Confirmed the new test **fails** without the fix, reproducing the exact production stack trace (`ZodError` thrown at `list-signals.ts:505`, propagating as a defect).
- Confirmed the new test **passes** with the fix, along with the rest of the suite: `pnpm --filter @domain/signals test` → 87/87 passing.
- `pnpm exec biome check` clean on both changed files.
- `pnpm --filter @domain/signals typecheck` hits a pre-existing, unrelated environment issue (`@latitude-data/telemetry` type declarations unresolved in a fresh worktree) reproduced identically on `development` before this change — not introduced by this PR.

## Checklist

- [x] Lint, type-checking, and tests pass locally (see typecheck caveat above — pre-existing, unrelated to this change)
- [ ] I have signed the CLA

---
_Generated by [Claude Code](https://claude.ai/code/session_013LEsGEMnuMH7KSh6mnX51p)_

### DIFF
--- a/packages/domain/signals/src/use-cases/list-signals.ts
+++ b/packages/domain/signals/src/use-cases/list-signals.ts
@@ -9,7 +9,6 @@ import {
   type SignalWindowMetric,
 } from "@domain/scores"
 import {
-  BadRequestError,
   type ChSqlClient,
   cuidSchema,
   type FilterCondition,
@@ -91,18 +90,7 @@ const listSignalsInputSchema = z.object({
 })
 
 export type ListSignalsInput = z.input<typeof listSignalsInputSchema>
-export type ListSignalsError = RepositoryError | BadRequestError
-
-const formatValidationError = (error: z.ZodError): string => error.issues.map((issue) => issue.message).join(", ")
-
-const parseOrBadRequest = <T>(schema: z.ZodType<T>, input: unknown) =>
-  Effect.try({
-    try: () => schema.parse(input),
-    catch: (error) =>
-      new BadRequestError({
-        message: error instanceof z.ZodError ? formatValidationError(error) : "Invalid list signals input",
-      }),
-  })
+export type ListSignalsError = RepositoryError
 
 export interface SignalListAnalyticsCounts {
   readonly newSignals: number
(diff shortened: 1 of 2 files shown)


geclos:
@claude is openapi/pagination used for all paginated endpoints? if it's specific of signals we should instead align the api page limit with the actual use case real limit, rather than just returning a nicer error

geclos:
@claude if signals is the only outlier let's aligned it with the openapi page limit of 200 rather than arbitrarily limit it to 100

packages/domain/signals/src/use-cases/list-signals.ts
```diff
@@ -502,7 +514,7 @@ export const listSignalsUseCase = (
   ChSqlClient | EvaluationRepository | SignalRepository | ScoreAnalyticsRepository | SessionRepository | SqlClient
 > =>
   Effect.gen(function* () {
-    const parsed = listSignalsInputSchema.parse(input)
+    const parsed = yield* parseOrBadRequest(listSignalsInputSchema, input)
```
andresgutgon: Let it fail no?

geclos: yes cc @claude

geclos:
@cursoragent track ci jobs, fix all failing, do not stop until ci is fully green

PR #3815, opened 2026-07-02
fix(core): batch workspace deletion so it survives the statement timeout

csansoon:
Backoffice workspace deletion still never deletes the workspace. #3296 fixed the foreign key ordering and is deployed, but the deletion now dies further along: for any workspace with real traffic, the single-statement `DELETE FROM provider_logs WHERE workspace_id = $1` exceeds the 30-second `statement_timeout` our own pg pool sets (`POOL_CONFIG` in `packages/core/src/client/index.ts`), Postgres kills it, and the transaction rolls back. The failure is invisible because of the follow-up #3296 called out and deferred: the job returns the service `Result` without unwrapping it, so BullMQ records a completed job — no retry, no error tracking.

This is not a guess: the Datadog trace for today's deletion attempts of workspace 14242 shows `evaluation_results` deleted in 158ms, `spans` in 8.1s, then `provider_logs` cancelled at exactly 30,003ms with "canceling statement due to statement timeout", followed by a rollback and a job span with status ok.

## the fix

`destroyWorkspace` now runs in two phases and is no longer atomic — it is idempotent and designed to be retried until it completes:

- **Phase 1** drains the unbounded-growth tables (`evaluation_results`, `evaluation_results_v2`, `spans`, `provider_logs`, `document_logs`, `dataset_rows`, `events`) in batches of 5,000 rows, each batch its own implicit transaction, so no single statement approaches the 30s limit. Same approach as `cleanupWorkspaceOldLogsJob`. All batch subqueries run on indexed columns.
- **Phase 2** is the previous transactional delete, unchanged in order, now operating on near-empty big tables. It re-sweeps the phase 1 tables to catch rows written concurrently during phase 1, and raises its `statement_timeout` to 5 minutes with `SET LOCAL` because the final `DELETE FROM workspaces` cascade is one statement that can legitimately outlive the 30s default even after the big tables are drained.

`destroyWorkspaceJob` now unwraps the result and throws on failure, closing the #3296 follow-up. BullMQ retries it (3 attempts with backoff were already configured on enqueue, but never triggered because the job never threw), keeps the failed job for debugging, and the error becomes visible in APM instead of a green "completed".

## legacy rows without workspace_id

`provider_logs.workspace_id` and `document_logs.workspace_id` are nullable and old rows were never backfilled. Those rows are invisible to a workspace-scoped delete but still block the deletion: provider logs through `provider_api_keys`/`api_keys` (`onDelete: 'restrict'`) and document logs through the `commits` cascade (`onDelete: 'restrict'`). Phase 1 purges them through those references (provider logs via the workspace's provider keys, document logs via the workspace's commits). One shape is deliberately not pre-swept: provider logs with NULL workspace **and** NULL provider but an `api_key_id` set — `provider_logs.api_key_id` has no index, so a batched sweep on it would seq-scan per batch. If such rows exist, the deletion now fails loudly and we'll see it.

## how this was verified

The delete order was re-checked against the live FK graph, not the Drizzle files: all 279 migrations applied to a scratch database, then `pg_constraint` queried for every `RESTRICT`/`NO ACTION` edge into tables reachable from the workspace cascade and into the manually-deleted tables. Everything found is handled.

Tests cover an empty workspace, the #3296 scenario (spans, provider logs, events, legacy evaluation chain), data volumes larger than one batch (run with `batchSize: 2` to exercise the loop), and legacy NULL-`workspace_id` logs. Each new sweep was confirmed load-bearing by removing it and watching its test fail.

## out of scope

Workspace deletion only touches Postgres; spans backfilled to ClickHouse are never cleaned up. That predates this change and stays a follow-up.

### DIFF
--- a/packages/core/src/jobs/job-definitions/maintenance/destroyWorkspaceJob.ts
+++ b/packages/core/src/jobs/job-definitions/maintenance/destroyWorkspaceJob.ts
@@ -1,7 +1,6 @@
 import { Job } from 'bullmq'
 
 import { unsafelyFindWorkspace } from '../../../data-access/workspaces'
-import { Result } from '../../../lib/Result'
 import { destroyWorkspace } from '../../../services/workspaces/destroy'
 import { queues } from '../../queues'
 
@@ -32,6 +31,11 @@ export async function enqueueDestroyWorkspaceJob(workspaceId: number) {
 /**
  * Background job that permanently destroys a workspace and all associated data.
  * Delegates to the destroyWorkspace service for the actual deletion logic.
+ *
+ * Throws when the deletion fails so BullMQ retries the job and the failure is
+ * visible, instead of silently reporting a completed job with a swallowed
+ * `Result.error`. The deletion is idempotent, so retries resume where the
+ * previous attempt left off.
  */
 export const destroyWorkspaceJob = async (
   job: Job<DestroyWorkspaceJobData>,
@@ -39,9 +43,7 @@ export const destroyWorkspaceJob = async (
   const { workspaceId } = job.data
 
   const workspace = await unsafelyFindWorkspace(workspaceId)
-  if (!workspace) {
-    return Result.nil()
-  }
+  if (!workspace) return
 
-  return destroyWorkspace(workspace)
+  await destroyWorkspace(workspace).then((r) => r.unwrap())
 }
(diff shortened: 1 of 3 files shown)


andresgutgon (approved):
Did you tried?

csansoon:
> @andresgutgon 
> Did you tried?

Yes, although I don't have any workspace that would fail locally anyways, that's why I didn't catch this issue in the first place

PR #3816, opened 2026-07-02
fix(spans): recover Vercel AI SDK v6 output when the GenAI compat layer shadows the Vercel parser

andresgutgon:
## Problem

Traces in some projects show a fully populated envelope — model badges, tokens, cost, duration, full span tree — but the **input/output message content isn't visible**, even though the data is correct upstream and renders in other tools (e.g. LangSmith).

Diagnosed by inspecting the raw stored span attributes in production ClickHouse. On the affected LLM spans:

| attribute | present? |
| --- | --- |
| `gen_ai.input.messages` | ✅ |
| `gen_ai.output.messages` | ❌ **absent** |
| `ai.response.text` | ✅ |
| `ai.prompt` / `ai.prompt.messages` | ✅ |

`input_messages` was populated but `output_messages` was **empty** on every LLM span — even though `ai.response.text` was present.

## Root cause

Recent **Vercel AI SDK v6** builds emit a GenAI SemConv compat layer *alongside* the native `ai.*` attributes. The asymmetry is the bug: input is emitted as `gen_ai.input.messages`, but the model's reply stays **only** in `ai.response.text` — no `gen_ai.output.messages` is produced.

Content-parser dispatch (`content/index.ts`) is **first-match-wins**, and the first parser claims any span carrying `gen_ai.input.messages`:

```ts
canHandle: (attrs) => hasKey(attrs, "gen_ai.input.messages") || hasKey(attrs, "gen_ai.output.messages"),
parse: parseGenAICurrent,
```

So these spans route to `parseGenAICurrent`, which parses input fine but reads output from `gen_ai.output.messages` (absent) with a fallback only to the deprecated `gen_ai.completion.*` (also absent). It never looks at `ai.response.text`. `parseVercel` — the one parser that *does* read `ai.response.text` — is shadowed and never runs.

The envelope (tokens/cost/model) resolves from a **separate** enrichment path (`resolveAttributes`), which is why the trace looks complete while output is blank.

## Upstream (Vercel)

This is an incomplete-semconv emission on Vercel's side, part of their ongoing GenAI-semconv alignment: **[vercel/ai#2590 — Make Vercel AI SDK tracing fully compatible with OTEL GenAI semantic conventions](https://github.com/vercel/ai/issues/2590)**.

Confirmed version boundary from testing:
- **`ai@6.0.197`** — has the bug (emits `gen_ai.input.messages`, output only in `ai.response.text`).
- **`ai@6.0.218`+** — fixed upstream: emits `gen_ai.output.messages`, which `parseGenAICurrent`'s primary path already handles.

We still want this ingestion fix because:
1. It **retroactively surfaces traces already ingested** from the `≤6.0.197` era — an SDK upgrade only helps new traces.
2. It makes us **robust to any customer** pinned to the affected v6 range, without requiring them to upgrade first.

Customers should also be advised to move `ai` to `≥6.0.218` for clean output at the source.

## Fix

`parseGenAICurrent` now falls back to the Vercel output extractor when `gen_ai.output.messages` yields no content — mirroring the existing deprecated split-attr fallback. A content-bearing `gen_ai.output.messages` still takes precedence.

- `content/vercel.ts` — export `parseVercelOutput` (was a private `parseOutput`).
- `content/genai.ts` — after the deprecated fallback, recover output via `parseVercelOutput` when output is still empty.

## Tests

- **`content/genai.test.ts`** — unit coverage for the hybrid dispatch: recovers `ai.response.text` and `ai.response.toolCalls` output when `gen_ai.output.messages` is absent, and confirms a real `gen_ai.output.messages` still wins over the Vercel fallback.
- **`tests/vercel.test.ts`** — end-to-end ingestion example (an `ai`-scope span with `gen_ai.input.messages` + `gen_ai.system_instructions` + `ai.response.text`, no `gen_ai.output.messages`): asserts input, system, usage, model **and** output all resolve through `transformOtlpToSpans`.

Full `@domain/spans` suite: **884 passing**. Typecheck clean.

### DIFF
--- a/packages/domain/spans/src/otlp/content/genai.test.ts
+++ b/packages/domain/spans/src/otlp/content/genai.test.ts
@@ -547,4 +547,61 @@ describe("parseContent (GenAI current — gen_ai.{input,output}.messages)", () =
       expect((parts.find((p) => p.type === "text") as { content: string }).content).toBe("It is sunny.")
     })
   })
+
+  // Vercel AI SDK v6 emits gen_ai.input.messages (so this parser wins dispatch) but keeps the
+  // model's turn only in ai.response.* — no gen_ai.output.messages. Regression for output going
+  // blank while input still rendered.
+  describe("Vercel AI SDK v6 hybrid — gen_ai input + ai.response output", () => {
+    it("recovers output from ai.response.text when gen_ai.output.messages is absent", () => {
+      const result = parseContent([
+        str(
+          "gen_ai.input.messages",
+          JSON.stringify([{ role: "user", parts: [{ type: "text", content: "Summarize this lead." }] }]),
+        ),
+        str("ai.prompt.messages", JSON.stringify([{ role: "user", content: "Summarize this lead." }])),
+        str("ai.response.text", "Here is the summary."),
+      ])
+
+      expect(result.inputMessages).toEqual([
+        { role: "user", parts: [{ type: "text", content: "Summarize this lead." }] },
+      ])
+      expect(result.outputMessages).toEqual([
+        { role: "assistant", parts: [{ type: "text", content: "Here is the summary." }] },
+      ])
+    })
+
+    it("recovers tool-call output from ai.response.toolCalls when gen_ai.output.messages is absent", () => {
+      const result = parseContent([
+        str(
+          "gen_ai.input.messages",
+          JSON.stringify([{ role: "user", parts: [{ type: "text", content: "weather in SF?" }] }]),
+        ),
+        str(
+          "ai.response.toolCalls",
+          JSON.stringify([{ toolCallId: "call_1", toolName: "get_weather", input: { city: "SF" } }]),
+        ),
+      ])
+
+      const assistant = result.outputMessages.find((m) => m.role === "assistant")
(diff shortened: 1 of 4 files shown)


andresgutgon:
### Recommended customer action: upgrade `ai` to ≥ 6.0.218

The root cause is upstream in the Vercel AI SDK's telemetry emission — tracked in **[vercel/ai#2590 — Make Vercel AI SDK tracing fully compatible with OTEL GenAI semantic conventions](https://github.com/vercel/ai/issues/2590)**.

Confirmed version boundary:
- **`ai@6.0.197`** — buggy: emits `gen_ai.input.messages` but keeps the model's reply only in `ai.response.text` (no `gen_ai.output.messages`) → output renders blank in Latitude.
- **`ai@6.0.218`+** — fixed: emits `gen_ai.output.messages`, which our parser already handles cleanly.

**Advise affected customers to bump `ai` to `≥ 6.0.218`** for correct output at the source going forward. This ingestion fix stays valuable as a backstop: it retroactively surfaces traces already ingested from the `≤ 6.0.197` era, and covers any customer still pinned to the affected range.

geclos (approved):

PR #3826, opened 2026-07-03
feat(signals): describe-first agentic signal creation (R4)

csansoon:
Ships R4 from `specs/signals.md`: the signal builder now opens on a single textarea — describe what you want to track, and Latitude generates and creates the complete signal (name, description, filters, sampling, and a rule/judge/script evaluation), then navigates to its detail page. "Configure manually" drops into the existing wizard, which is unchanged. The method-cards intro screen is gone.

## how generation works

`createSignalFromPromptUseCase` (`@domain/signals`) runs in workers behind the new `signals-generate-signal` queue topic; the web enqueues and polls a Redis result at `org:{orgId}:signalGeneration:{generationId}` whose pending value carries a `step` line that feeds the intro's status text.

The use case is one bounded loop, not an agent with tools. It prefetches grounding — distinct tags/services/models/providers/tool names (`TraceRepository.distinctFilterValues`), sessions-per-day traffic, and one sample session — and injects it into the prompt so the model reconciles the user's wording against real values ("the ticket cancellation tool" → observed `cancel_ticket`; unobserved values inferred from naming patterns; user-quoted literals kept verbatim). Each draft is mapped through the shared `evaluationSettingsSchema`/`filterSetSchema`, compile-checked in the sandbox, and previewed against recent sessions via `previewEvaluationUseCase`; failures feed a repair turn, and one verdict-review turn lets the model confirm or revise. Hard cap of 4 `ai.generate` calls, then `createSignalUseCase` persists. Projects with no sessions skip the preview and still create.

Product decisions carried from the spec: the signal is created immediately (no review step — closing the modal only stops polling, and a Redis claim key makes BullMQ stall-redeliveries no-ops so a run never double-creates); the user's prompt is not persisted; kind preference is rule > judge > script so results stay form-editable; filters are generated only when they discard sessions that cannot match — the evaluation alone must express the ask.

## why the generation schema looks the way it does

Bedrock structured output compiles the schema into a constrained-decoding grammar with limits it reveals one error at a time: it rejects `oneOf`, `maxItems`, `maxLength`, and `minimum`/`maximum` (which zod's `.int()` emits implicitly), refuses grammars with many alternating object shapes, and caps union-typed parameters (every `.nullable()` counts) at 16 per schema. The generation schema is therefore deliberately dumb: rule conditions are one flat object (`type` enum + nullable per-type fields) instead of the 10-member union, kind payloads use empty-string/array sentinels instead of nulls, and no numeric or size constraints appear anywhere. Validation strength lives host-side: `mapGeneratedSignalDraft` reassembles real conditions, clamps name/sampling, rounds integer fields, and re-parses through the shared schemas — violations come back to the model as repair-turn feedback instead of failed requests. The full constraint list is in the commit trail for the next schema author.

## custom script tab becomes the compiled view

The Advanced tab now always shows the current evaluation as the exact script that runs: settings compile client-side (`compileSettingsToScript` is exposed through the `@domain/evaluations` browser entry after splitting the pure codegen from the sandbox-backed validators), rendered read-only with an explicit "Edit as custom script" detach that re-authors the evaluation as a raw script and clears the settings forms. Viewing the compiled code never changes what preview or save persist. This is slightly stricter than the spec's "editing re-authors" phrasing — the explicit detach prevents a stray keystroke from silently destroying a multi-condition form.

## retired

The R2 describe→script path is fully deleted: `createScriptFromPromptUseCase`, the `signals-generate-script` topic + worker, its web functions, and the in-tab generation UI. Its sandbox-contract system prompt survives embedded in the new signal-generation prompt. Any script-generation job in flight at deploy is dropped (5-minute-TTL interactive flow).

## ops notes

- Model resolves under the new `SIGNAL_GENERATOR` feature: `LAT_AI_EVALUATION_SCRIPT_GENERATOR_*` overrides stop working; use `LAT_AI_SIGNAL_GENERATOR_*`.
- Telemetry routes to a new `latitude-signal-generation` dogfood project (seeded in dev); it must exist in the prod dogfood org before rollout.

## testing

12 use-case tests cover the loop (happy paths per kind, filter mapping, repair turns for invalid drafts / compile errors / all-errored previews, review revision, zero-session path, budget exhaustion, credential errors, step reporting). Validated manually end-to-end against Bedrock.


csansoon:
@claude review this PR

csansoon:
Addressed the review — thanks. Point by point:

**1. Rate limiting** — fixed in 8c2ecc256. New per-org+project Redis counter (`enforceSignalGenerationRateLimit`, 20/hour in production) on `generateSignal`, same pattern as the export rate limit.

**2. Grounding prompt injection** — fixed in 524c3d8de. Observed values are whitespace-collapsed and capped at 100 chars, the grounding block is wrapped in `<observed_project_data>` delimiters, and the system prompt instructs the model to treat grounding and preview content as data, never instructions.

**3. Claim TTL** — fixed in 0cbec75e1. The claim key gets a dedicated 30-minute TTL, decoupled from the 10-minute result TTL.

**4. Order-dependent destructure** — fixed in b586d45c0. The grounding fetch is now a column-keyed `Effect.all` struct; also switched to unbounded concurrency for the six ClickHouse reads (your observation below).

**5. Prompt max length** — fixed in 8c2ecc256 (same commit as the rate limit; both gate the same endpoint's input). `max(2000)` on the schema plus `maxLength` on the intro textarea, sharing one constant.

On the observations: the quadruple-newline spacing is fixed in 524c3d8de. Review-turn revision going straight to create is intended (budget cap; the revision is still validated + previewed). Post-close creation without a toast is the locked product decision from the spec — the list shows the signal on next refetch; happy to revisit as a follow-up. Keeping the step strings exact in tests is deliberate for now: they double as the product copy contract, so a copy change failing the test is a feature until i18n forces the issue.

PR #3829, opened 2026-07-03
feat(web): central write-gate middleware + client scope stamp (G4)

andresgutgon:
## Task G4 — central write-gate + client scope stamp

Part of the **Showcase Demo Project** effort. Umbrella (spec) PR: **#3821** — see spec task **G4** and [Read-only enforcement](https://github.com/latitude-dev/latitude-llm/pull/3821) (layer 2).

Phase-1 groundwork: touches existing code but is **behavior-neutral for current users** — nothing user-visible changes until the showcase is wired up later.

### What this does
Read-only enforcement **layer 2** for the shared read-only showcase project. One global `functionMiddleware` entry in `apps/web/src/start.ts` with two halves:

- **Client (`.client()`)** — a global client-side middleware that stamps the current `ProjectScope` onto every outgoing server-fn request via `sendContext` (new pattern for this repo, framework-supported).
- **Server (`.server()`)** — reads that scope from `context` and throws a typed `ReadOnlyProjectError` (403) when the request is in a **read-only scope** (showcase/public) **and** the method is a write.

Method discipline makes the gate reliable and per-collection-code-free:
- **All `GET`s pass** (no write is a GET).
- **`POST`s are writes**, minus a 3-item POST-read allowlist: `previewEvaluation`, `listLinearTeamsForApiKey`, `listCursorRepositoriesForApiKey`.
- **Fails closed** on an unidentifiable POST (missing server-fn name).

### Why behavior-neutral
The gate only fires under a read-only scope, and **nothing sets a non-`live` scope yet**. `getCurrentProjectScope()` is a minimal live-default source, so the gate is inert. Verified: full `@app/web` test suite (516 tests) + `@domain/shared` and `@app/web` typechecks pass.

### Coordination notes
- **`ProjectScope`** (`apps/web/src/domains/project-scope/project-scope.ts`) is intentionally minimal — the coordination seam for **task G3**, which owns the full scope context + scope-aware read plumbing and will replace the live-default getter with a source driven by the resolved route scope.
- **`ReadOnlyProjectError`** lives in **`@domain/shared`** (client-safe) so it's importable by the **task G1** central `handleMutationError` (which branches on `_tag === "ReadOnlyProjectError"` to open the read-only modal). It satisfies `isHttpError`, so `start.ts`'s existing error shaping classifies it as an expected 4xx (logged at `warn`, kept out of Datadog).

### Where review should focus
1. **Middleware wiring** in `start.ts` — one entry does both client stamp + server gate; the client `sendContext.projectScope` is read as `context.projectScope` server-side. Ordering: after `tracingFnMiddleware` (outer) so a thrown `ReadOnlyProjectError` is traced as a 4xx.
2. **Allowlist correctness** — confirmed all three are `createServerFn({ method: "POST" })` genuine reads; `serverFnMeta.name` is the exported variable name.
3. **Error home** — `@domain/shared` vs `@domain/projects` (chose shared for client-safety + cross-cutting nature).
4. **Client-scope seam** — a live-default getter (no mutable holder / setter yet, to satisfy `knip`); G3/S3 add the dynamic source.


andresgutgon:
@claude review

PR #3834, opened 2026-07-03
Public design system

ntizish:
## Summary

Extracts the `@repo/ui` component showcase out of `apps/web`'s unauthenticated `/design-system` route and into a new standalone app, `apps/design-system`, deployed independently via Vercel (`design.latitude.so`). The old route already had no auth guard, so this doesn't change who can see the design system — it changes where it lives and how it ships.

## What changed

- New app `apps/design-system`: a Vite + TanStack Router SPA styled with Tailwind v4. It has its own `vite.config.ts`, `vercel.json` (SPA rewrite to `index.html`), and linked Vercel project, and needs no `turbo.json` or CI wiring — the existing `apps/*` workspace glob picks it up.
- Removed `apps/web/src/routes/design-system/*` (`index.tsx`, `button.tsx`, `colors.tsx`, `route.tsx` — ~1,280 lines) and regenerated `apps/web`'s `routeTree.gen.ts` to drop those routes. `chat.tsx` moved over via rename instead of being rewritten.
- Pages: About, Brand, Colors, Typography, Spacing, Shadows, Icons, a dynamic per-component demo route (`$component`), and dedicated pages for Button, Chat, and Infinite table. A shared shell/sidebar/theme toggle and a central `nav-config.ts` (General / Product / Components) drive navigation.
- Each demo page now pairs its live render with a copyable usage snippet (new `UsageSection`/`UsageCode` components on top of `@repo/ui`'s `CodeBlock`). `packages/ui`'s read-only CodeMirror block gained a `language` prop and JS/TS/JSX/TSX highlighting to support this — it previously only highlighted JSON.
- Added the Latitude logo SVGs to `apps/web/public` for the new app to reference, and excluded `.vercel` in the root `.gitignore`.

## Why

Latitude design practices with code usage examples are more accessible to the users, which is important as an Open Source app.


packages/ui/src/components/code-block/codemirror-readonly.tsx
```diff
@@ -24,10 +26,32 @@ function isJson(value: string): boolean {
   }
 }
 
+function languageSupport(language: string | undefined, isJsonContent: boolean): Extension | null {
```
andresgutgon, line 29: Why this change?

apps/design-system/src/routes/brand.tsx
```diff
@@ -0,0 +1,82 @@
...
+      title="Brand"
+      description="Latitude logo assets for product UI, marketing, and partner materials."
+    >
+      <UsageSection description="Import the logo mark from @repo/ui. Use the PNG or SVG lockup from /public when the wordmark is required.">
+        <UsageCode
+          lines={[
+            'import { LatitudeLogo } from "@repo/ui"',
+            "",
+            '<LatitudeLogo className="h-5 w-5" />',
+            '<img src="/latitude-logo.png" alt="Latitude" className="h-8 w-auto" />',
```
andresgutgon, line 47: So this is pointing to `latitude.so/latitude-logo.png`? Is not `.svg`? Also could not this app has their own assets instead of depending on our app

apps/design-system/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "format": "biome format --write src && biome check --fix src",
+    "typecheck": "tsgo -p tsconfig.json --noEmit"
+  },
+  "dependencies": {
+    "@repo/ui": "workspace:*",
+    "@tanstack/react-router": "1.168.10",
+    "lucide-react": "1.9.0",
+    "react": "catalog:",
+    "react-dom": "catalog:",
+    "rosetta-ai": "catalog:"
```
andresgutgon: Why we need `rosseta-ai` here?

apps/design-system/src/routes/brand.tsx
ntizish: idk, what's better? in terms of components, the entire concept is about being dependent on the main app, so that when we update components, the changes are pushed to the DS right away

packages/ui/src/components/code-block/codemirror-readonly.tsx
ntizish: To enable syntax highlight in code preview
Screenshot 2026-07-06 at 11 41 52

apps/design-system/package.json
ntizish: AI did me dirty, fixing

apps/design-system/src/routes/brand.tsx
ntizish: ok, this is bad, fixing

geclos:
@claude review

geclos:
I'd fix the code mirror bug, and review if we need the orphaned public assets Claude mentions

andresgutgon (approved):

PR #3849, opened 2026-07-05
feat: import historical traces from Langfuse, LangSmith and Braintrust

geclos:
Customers evaluating or leaving another observability platform have no way to bring their history with them. Forward OTLP ingestion already works; this adds a self-serve **Imports** wizard in project settings that backfills traces and spans from Langfuse, LangSmith or Braintrust into one Latitude project.

Implements `specs/observability-migration.md` (LAT-721) through Phase 6.

## the engine

One engine, three adapters. The engine owns scheduling, cursors, limits, retries, idempotency, stats and ClickHouse writes; an adapter only knows how to call its vendor and normalize rows into `SpanDetail`. `@domain/imports` holds the entities, ports and use-cases, `@platform/import-sources` the three adapters.

A job is one row in `import_jobs`: encrypted credentials, non-secret config snapshotted at confirmation, cursor, stats, a bounded run history, and the terminal error. One queued or running import per org, enforced by a partial unique index rather than application logic.

Progress runs as a BullMQ page chain — each job reads one bounded page, persists progress, and re-enqueues itself — which matches the data-destinations backfill model and needs no Temporal workflow. Span ids are derived deterministically from vendor ids, so re-running a range dedupes through ClickHouse's `ReplacingMergeTree` and a retry resumes mid-range instead of re-reading everything.

Spans are written through `SpanRepository.insert`, not OTLP HTTP. Going back out through the wire protocol would mean re-encoding rows we have already normalized, and losing fidelity in both directions.

## imported traces are processed and billed like ingested ones

Writing spans to ClickHouse and stopping there would produce history that none of the product's analysis runs on: no `TraceEnded`, so no conversation intelligence, no embeddings, no semantic search, no flaggers. Each page therefore publishes the same `TracesIngested` event `processIngestedSpansUseCase` publishes, carrying the org's plan snapshot and its `isSandbox` bit. Every existing consumer follows with no second code path, and sandbox orgs keep skipping both the billing and the LLM fan-out.

That processing costs what live traffic costs, so imported traces are billed at one credit each. Billing already dedupes on `trace:{org}:{project}:{trace}`, so a trace whose spans arrive across two pages is charged once.

Reviewers will want to look at two consequences of that decision:

**The budget counts traces, by root span.** Counting distinct trace ids per page would over-count by up to 10x when a trace's spans scatter across pages, which would stop imports far short of what the user paid for. A trace has exactly one root span however its spans are spread, so counting roots is exact, order-independent, and agrees with what billing charges for. It undercounts only when a trace's root is absent from the source data.

**Two ceilings, both ending in `capped` rather than a failure.** `config.maxTraces` is the ceiling the user accepted, defaulted to whatever their plan currently affords. Separately, the org's live remaining usage is re-read before every page, which is what catches usage spent mid-import by live ingestion or another project — a snapshot cannot. Which ceiling stopped a job is derived from its own counters (`importCapReason`), not stored: the trace ceiling is the only one a job can *reach*, so a capped job that fell short of it ran out of plan usage.

## newest first, without trusting vendor sort parameters

If an import can only afford part of a range, the part it keeps should be the most recent. That has to hold for all three sources, and **Langfuse's observations endpoint has no sort parameter in either API version**, so a per-adapter sort hint could not deliver it.

Ordering is therefore the engine's job. It reads the range as day-wide windows walking backwards from `rangeTo`, with each adapter paging inside a window using its own cursor. Newest-first becomes a property of the engine, and an adapter that ignored a sort hint would degrade to day-granular ordering rather than to arbitrary ordering. Windows whose first page is empty widen 4x, bounded, so a sparse year costs a handful of requests rather than 365. LangSmith and Braintrust are additionally read descending inside the window, so for those two the truncation is exact.

## limits come from the plan, and are shown before confirming

`resolveImportLimits` is the single derivation, used by both the wizard and job creation, so what a user is shown is what the backend accepts.

- **Trace ceiling**: defaults to the org's remaining plan usage; hard ceiling 1,000,000.
- **Range**: defaults to the last `min(90, planRetentionDays)` days and can never reach further back than retention. ClickHouse drops spans past `retention_days` measured from the span's own `start_time`, so importing older history would bill for rows that are then deleted. A Pro org cannot import more than 90 days however high the product cap is.
- An org with no usage left is refused up front rather than handed a job that caps on its first page.

The wizard shows the plan, the traces left this period, the reset date and the effective lookback before the user confirms, and the import card explains a cap in terms of which ceiling was hit.

## adapters

Each source's contract came from its own published reference rather than guesswork:

- **Langfuse** reads the **v2** observations API. v2 returns only the field groups named in `fields` and omits the rest, so the adapter names every group it maps and a test asserts that list — getting it wrong would import spans with no content, metadata, model, usage or tags and still look successful. Pages follow `meta.cursor`, clamped to v2's 1000-row maximum. The host is resolved and pinned to a public address, so a self-hosted URL cannot be pointed at a private one.
- **LangSmith** asks for `order: "desc"` and pages with `cursors.next`. Its `session_id` is a project id rather than a conversation id, so sessions come from a configurable run metadata key, and the preview reports whether that key actually appears in the sample.
- **Braintrust** queries BTQL, which has no parameter binding, so the project id and cursor are allow-listed rather than escaped before being composed into the query.

All three fail loudly when a full page returns no continuation cursor, so a broken pagination contract surfaces as a visible error rather than an import that quietly drops the rest of a window.

## operational behaviour

Per-source request pacing, with `Retry-After` honoured exactly and bounded so a permanent throttle still terminates. A page is bounded by a timeout with the socket aborted, so a hung source releases its worker slot. Credentials are scrubbed on every terminal transition and collected again on retry. Cancellation is cooperative between pages; spans already imported stay. `ProjectDeleted` deletes the project's imports, which also frees the org's single import slot. `ImportStarted` and `ImportFinished` reach PostHog through the outbox, both emitted from one helper so no terminal transition can skip them.

## data model note

The job row is the source of truth, so the last 25 processed pages live in a bounded jsonb ring buffer on `import_jobs` rather than a second table. This deliberately differs from `destination_sync_runs`, where the runs table *is* the source of truth because a destination has no cumulative progress row. A second table here would have been a projection of data the job already had, needing a pruning cron and a join; the array is safe precisely because pages are sequential — one active import per org, each page publishing exactly one successor, so there is never more than one writer.

## testing

166 domain tests, 133 adapter tests, and repository round-trip coverage for the jsonb config, cursor, stats and ring buffer. The engine tests cover the window walk, root-span trace counting, both cap paths, `TracesIngested` publication, cancellation, the retry and rate-limit paths, and idempotency across a re-import. Adapter tests pin each vendor's pagination and range contract, including the fail-loud guard. Workspace typecheck, lint and knip are clean, and `drizzle-kit check` passes.

## what still needs a live vendor account

The three pagination contracts come from published API definitions and the official LangSmith SDK, not from a live run. Nothing here has been exercised against a real Langfuse, LangSmith or Braintrust project, so a staging run per source and the throughput measurement are still open, and are recorded as such in the spec. The fail-loud cursor guards exist so that a wrong contract shows up as an error rather than as silent data loss.

Also out of scope, per the spec: blob/Parquet imports for enterprise scale, scores and annotations, datasets, prompt-registry enrichment beyond span metadata, and continuous sync.

Refs `specs/observability-migration.md`

## Summary by CodeRabbit

* **New Features**
  * Added historical trace imports from Langfuse, LangSmith, and Braintrust.
  * Added a project Settings → Imports page with connection testing, project selection, preview, progress tracking, cancellation, and retry support.
  * Imports run in the background with newest-first processing, plan-based limits, redaction, rate-limit handling, and detailed run history.
* **Documentation**
  * Added setup guides and provider-specific field mapping documentation.
* **Bug Fixes**
  * Improved billing credit availability calculations and project cleanup behavior.


geclos:
@cursoragent triage comments and babysit pr

geclos:
@cursoragent ensure all ci is green

geclos:
@claude review

andresgutgon:
This is HUGE

PR #3850, opened 2026-07-05
fix(spans): bound session span-list reads to stop ClickHouse OOMs

geclos:
Fixes [Datadog issue 92bd265a](https://app.datadoghq.eu/error-tracking/issue/92bd265a-77d8-11f1-8550-da7ad0900005): `RepositoryError: Repository listBySessionId failed: (total) memory limit exceeded: would use 16.63 GiB … (while reading column attr_string)` — opening a session detail drawer killed the query at the server-wide OvercommitTracker, endangering other tenants' queries on the same replica.

## root cause

The session was an 11-hour agent run with ~11,700 spans, instrumented with the Vercel AI SDK, which puts the **entire conversation history** into the `ai.prompt` / `ai.prompt.messages` string attributes of every generation span. Ingest stores attributes verbatim (by design), so this one session carries **~30 GiB of uncompressed `attr_string`** (avg 2.6 MiB per span, measured on production).

`listBySessionId` selected `LIST_COLUMNS` — which included `attr_string`/`attr_int`/`attr_float`/`attr_bool`/`resource_string` — for every span in the session, with no limit and no per-query memory cap. With ~935 rows per read granule × 2.6 MiB, a single decompressed block of `attr_string` is multi-GiB, and the multi-threaded PrefetchedReadPool reads several concurrently → instant OOM at the 14.4 GiB replica cap.

No session-level consumer renders these maps: the session drawer computes duration breakdowns and tool pills from timing/tool columns, and the only UI that shows attributes is the single-span detail view, which reads them through `findBySpanId`.

## changes

**`listBySessionId` no longer reads the attribute maps.** The inner scan selects a lean column set; the outer select returns the maps as empty typed literals (`CAST(map(), …)`), so ClickHouse never touches those columns on storage. The row shape and the `Span` domain type are unchanged; the port JSDoc documents that the maps come back empty and points attribute consumers at `findBySpanId`. Verified against the exact production session that OOMed: the rewritten query returns all 11,745 spans in **1.1 s reading 1.4 MiB** (previously: killed after allocating 16+ GiB).

**Session membership predicate is now sargable.** `coalesce(nullIf(session_id, ''), toString(trace_id)) = {sessionId}` wraps both columns in functions, defeating the `idx_session_id`/`idx_trace_id` bloom-filter skip indexes — every granule of the org/project was scanned (matters most for the evaluation-runtime callers, which pass no time bounds; the table PK `(org, project, start_time)` can't prune those). It's split into bare column equalities: `session_id = {sessionId}`, plus an `OR (session_id = '' AND trace_id = {sessionId})` arm only when the id is 32 chars (orphan session ids are trace ids; any other length can't match a `FixedString(32)`). Production check on the same project without time bounds: 693,700 rows read with the old predicate vs 76,696 with the new one, identical counts. Applied to the three sharers of the predicate: `listBySessionId`, `listToolSpansBySessionId`, `findMessagesForSession`.

**Bounded-read settings on all multi-span reads.** The `max_memory_usage: 4 GB` + `output_format_parallel_formatting: 0` defense that `listByIngestedAtWindow` already used is extracted into a shared constant and applied to `listByTraceId`, `listByProjectId`, `listBySessionId`, `listToolSpansBySessionId`, `findMessagesForTrace`, `findMessagesForSession`, and `listRecentDetailsByProjectId` — a pathological query now fails its own request with a retryable `RepositoryError` instead of tripping the server-wide OvercommitTracker.

Ingest is deliberately untouched: attributes keep being stored verbatim as they came, with message extraction into the denormalized payload columns as today.

## tests

New `listBySessionId` suite (chdb, real schema): attribute maps come back empty while `findBySpanId` still returns them; orphan single-trace sessions keyed by trace id still match; a conversation session whose id happens to be 32 chars still matches; an empty session id matches nothing (the naive predicate split would have matched every orphan span); optional time-window scoping. Full `@platform/db-clickhouse` suite passes (388 tests), typecheck (`tsgo`) and biome clean.

## verification after deploy

Occurrences of issue `92bd265a-77d8-11f1-8550-da7ad0900005` should stop; opening the session drawer for session `tsk_r2S9n6loNm9ebb` (project `gu0lbj2fu843hghax4rdi1ie`) should render instead of erroring.

[Slack Thread](https://latitudedata.slack.com/archives/C03BZFX9KCK/p1783190893830399?thread_ts=1783190893.830399&cid=C03BZFX9KCK)

<div><a href="https://cursor.com/agents/bc-e3d6da95-a843-5c2c-9db9-0e7257f84ae6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e3d6da95-a843-5c2c-9db9-0e7257f84ae6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/domain/spans/src/ports/span-repository.ts
+++ b/packages/domain/spans/src/ports/span-repository.ts
@@ -77,7 +77,11 @@ export interface SpanRepositoryShape {
    * Every span in a session. Membership mirrors the `sessions_mv` grouping key
    * (`coalesce(nullIf(session_id, ''), toString(trace_id))`), so it covers both
    * conversation-id sessions and orphan single-trace sessions (whose spans carry
-   * no `session_id` and are keyed on their `trace_id`).
+   * no `session_id` and are keyed on their `trace_id`). The dynamic attribute
+   * maps (`attrString`/`attrInt`/`attrFloat`/`attrBool`/`resourceString`) come
+   * back empty: a session's span count is unbounded and instrumentors can put
+   * whole conversations into attributes, so reading them here is a memory
+   * hazard — fetch a span's attributes via `findBySpanId`.
    */
   listBySessionId(input: {
     readonly organizationId: OrganizationId
--- a/packages/platform/db-clickhouse/src/repositories/span-repository.test.ts
+++ b/packages/platform/db-clickhouse/src/repositories/span-repository.test.ts
@@ -158,6 +158,140 @@ describe("SpanRepository", () => {
     })
   })
 
+  describe("listBySessionId", () => {
+    const SESSION_ID = SessionId("session-list")
+    const HEX32_SESSION_ID = SessionId("99999999999999999999999999999999")
+    const ORPHAN_TRACE = TraceId("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb")
+
+    const insertListFixture = () =>
+      runCh(
+        insertJsonEachRow(ch.client, "spans", [
+          makeSpanRow({
+            session_id: SESSION_ID,
+            span_id: "aaaa000000000001",
+            name: "older",
+            attr_string: { "ai.prompt": "huge-blob" },
+            attr_int: { retries: 2 },
+            resource_string: { host: "h1" },
+            ingested_at: "2026-01-01 00:00:00.000",
+          }),
+          makeSpanRow({
+            session_id: SESSION_ID,
+            span_id: "aaaa000000000001",
+            name: "newer",
+            attr_string: { "ai.prompt": "huge-blob" },
+            ingested_at: "2026-01-01 00:00:01.000",
+          }),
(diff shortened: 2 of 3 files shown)


geclos:
@claude review

andresgutgon:
I changed my mind. Artifacts are awesome

PR #3879, opened 2026-07-06
fix(web): show onboarding illustrations in the claim flow

neoxelox:
The temporary-organization **claim** onboarding reused the same step components as the normal project onboarding, but only showed the left-hand form on each step — none of the right-side illustrations. Everything else (traces intro gallery, signals feed, Slack queue mock) was missing.

## Cause

The step visuals aren't a prop on the step components. Each step module (`role`, `flaggers`, `slack`, `telemetry`) exports two separate components: `Left` (the form/content) and `Right` (the illustration). The normal `OnboardingFlow` renders a two-column layout where the right column is a carousel of `OnboardingGallery` + each step's `Right`. `ClaimOnboarding` rendered only `Step.Left` in a single centered column and never mounted a right pane, so reusing the step components got the left halves only.

## Change

Extract the normal flow's right-column carousel into a shared `OnboardingRightPane` and mount it in both flows.

- **`onboarding-right-pane.tsx`** (new) — takes the flow's ordered `steps` + `currentStep` and renders the carousel, sliding to the slide that matches the current step. One source of truth for the onboarding visuals.
- **`onboarding-flow.tsx`** — swaps its inline right column for `OnboardingRightPane`. Behavior is unchanged: `activeSteps.map(→slide)` reproduces the old visible-slides list and `steps.indexOf(currentStep)` reproduces the old active index.
- **`claim.$token.tsx`** — `ClaimOnboarding` now uses the same two-column shell (left column matches the normal flow's layout) and mounts `OnboardingRightPane`. Its `activeSteps` is `["role","flaggers","slack"]` when Slack is configured, else `["role","flaggers"]`, so the claim flow shows the intro/flaggers/slack visuals matching its steps.

## Validation

`pnpm --filter @app/web typecheck` and `biome check` pass. No behavior change to the normal onboarding flow; the claim flow now renders the illustrations on every step.

### DIFF
--- a/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
+++ b/apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding-flow.tsx
@@ -16,8 +16,7 @@ import { submitOnboarding } from "../../../../../domains/users/user.functions.ts
 import { getQueryClient } from "../../../../../lib/data/query-client.tsx"
 import { toUserMessage } from "../../../../../lib/errors.ts"
 import { createFormSubmitHandler } from "../../../../../lib/form-server-action.ts"
-import { CarouselSlide, CarouselTrack } from "./onboarding/carousel-track.tsx"
-import { OnboardingGallery } from "./onboarding/onboarding-gallery.tsx"
+import { OnboardingRightPane } from "./onboarding/onboarding-right-pane.tsx"
 import * as FlaggersStep from "./onboarding/steps/flaggers-step.tsx"
 import * as RoleStep from "./onboarding/steps/role-step.tsx"
 import * as SlackStep from "./onboarding/steps/slack-step.tsx"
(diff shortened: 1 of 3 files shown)


geclos (approved):
## Review — clean, correct refactor ✅

Extracts the onboarding right-pane carousel into a shared `OnboardingRightPane` and mounts it in both the main onboarding flow and the claim flow (which previously rendered no illustrations). Good DRY consolidation that fixes the reported gap.

### Correctness — no bugs found
- **Behavior preserved.** The new `activeRightSlideIndex = Math.max(0, steps.indexOf(currentStep))` is equivalent to the old `visibleRightSlides.indexOf(STEP_TO_RIGHT_SLIDE[step])`, since `visibleRightSlides` is a 1:1 map of `steps` (unique steps, bijective `STEP_TO_RIGHT_SLIDE`). The `Math.max(0, -1) → 0` out-of-range fallback is retained.
- **Types align.** `ClaimOnboardingStep` (`role|flaggers|slack`) is a subset of `OnboardingStep`, so the `activeSteps` arrays passed from the claim flow are assignable (`ReadonlyArray` covariance).
- **Imports/exports resolve.** `CarouselTrack`/`CarouselSlide`, `OnboardingGallery`, and each `Step.Right` exist with matching signatures; the `ComponentProps<typeof FlaggersStep.Right>[...]` prop reuse is correct.
- **No desynced panes.** In the claim flow `step` is never `"slack"` when `slackEnvConfigured` is false (that transition only fires when configured), so left/right panes stay in sync with `activeSteps`.

### Minor nit (non-blocking)
`onboarding-right-pane.tsx` carries a 2-line comment explaining *why* the pane is separate. The repo convention (CLAUDE.md) favors one-line comments with rationale in the commit/PR rather than the code. Defensible as it flags a real non-obvious constraint, but consider trimming to one line.

### Caveat
I could not run `pnpm --filter web typecheck` in the review environment (workspace deps not installed → `tsgo` absent), so the type analysis above is by inspection. CI's typecheck should confirm.

---

PR #3881, opened 2026-07-06
feat(showcase): showcase resolver chokepoint (S2)

andresgutgon:
## Task S2 — Showcase resolver

Part of the **Showcase Demo Project** effort — umbrella PR #3821. Builds on the Phase-1 groundwork + S1 (pointer table `latitude.showcase` + `@domain/showcase` repo, `wantsShowcase` org flag from G5) already on `development`.

Implements the **single read chokepoint** that resolves the pinned showcase project + org from the `latitude.showcase` pointer.

### What it does
- **`resolveShowcaseUseCase`** (`@domain/showcase`): the resolver core.
  - **Authorization:** "authenticated **AND** the requesting org's `wantsShowcase` flag is true" → else `NotFoundError` (404). A missing org and a false/absent flag are indistinguishable from "no showcase" — all 404, so a dismissed org leaks nothing.
  - **Pointer read is Redis-cached** (`showcase:current`, global key, no `org:` prefix since the pointer isn't org-scoped). Cache errors degrade to the DB read; only live resolutions are cached, never the null.
  - Returns **exactly one** `{ organizationId, currentProjectId }`. Callers must thread that same `organizationId` into **both** `withPostgres`/`withClickHouse` **and** the repo method arg — ClickHouse has no RLS backstop, so the layer org and the query-param org must be the same value.
- **`resolveShowcaseAccess`** (`apps/web/src/server`): thin session + Live-layer wiring, modeled on `resolveSandboxAccess` / `resolveOrgScope`. Reads run under the requesting org's own Postgres scope (both `organizations` and the pointer table have no RLS policy), no RLS bypass.

### Scope / non-goals
Per the task: **only** the resolver. No `$projectSlug` loader branch, no `useProjectsCollection` merge, no routes, no regeneration, no UI (those are S3/S4+).

**Behavior-neutral:** nothing calls the resolver yet (S3 wires it). `knip.json` ignores the dormant server file + `@domain/showcase` dep for `apps/web` until S3 removes them.

### Tests
`resolve-showcase.test.ts` — authz true (resolves), authz false → 404, no preference set → 404, no showcase → 404, showcase row with `currentProjectId = null` → 404. All 7 `@domain/showcase` tests pass.

### Verification
- `pnpm --filter @domain/showcase typecheck` ✅
- `pnpm --filter @domain/showcase test` ✅ (7 passed)
- `pnpm --filter @app/web typecheck` ✅
- `knip` ✅

### Review focus
- The authz predicate + the "one org id into all layers" invariant (the CH-no-RLS trap).
- Cache key/TTL choice and cache-error degradation.
- The temporary `knip.json` ignore entries (to be removed by S3).

### DIFF
--- a/packages/domain/showcase/src/use-cases/resolve-showcase.test.ts
+++ b/packages/domain/showcase/src/use-cases/resolve-showcase.test.ts
@@ -1,0 +1,144 @@
+import { createOrganization, type Organization, OrganizationRepository } from "@domain/organizations"
+import {
+  CacheStore,
+  type CacheStoreShape,
+  NotFoundError,
+  OrganizationId,
+  ProjectId,
+  SqlClient,
+  type SqlClientShape,
+} from "@domain/shared"
+import { Effect } from "effect"
+import { describe, expect, it } from "vitest"
+import { createShowcase } from "../entities/showcase.ts"
+import { ShowcaseRepository } from "../ports/showcase-repository.ts"
+import { createFakeShowcaseRepository } from "../testing/fake-showcase-repository.ts"
+import { resolveShowcaseUseCase } from "./resolve-showcase.ts"
+
+const REQUESTING_ORG_ID = OrganizationId("requestingorgggggggggg01")
+const SHOWCASE_ORG_ID = OrganizationId("showcaseorgggggggggggg01")
+const SHOWCASE_PROJECT_ID = ProjectId("showcaseprojectttttttt01")
+
+const sqlClient: SqlClientShape = {
+  organizationId: REQUESTING_ORG_ID,
+  transaction: (effect) => effect,
+  query: () => Effect.die(new Error("unexpected query")),
+}
+
+const noopCache: CacheStoreShape = {
+  get: () => Effect.succeed(null),
+  set: () => Effect.void,
+  delete: () => Effect.void,
+}
+
+const provideDeps = <A, E>(
+  effect: Effect.Effect<A, E, SqlClient | CacheStore | ShowcaseRepository | OrganizationRepository>,
+  {
+    requestingOrg,
+    showcaseRepository,
+  }: {
+    requestingOrg: Organization | null
+    showcaseRepository: ReturnType<typeof createFakeShowcaseRepository>["repository"]
+  },
+) =>
+  effect.pipe(
+    Effect.provideService(SqlClient, sqlClient),
+    Effect.provideService(CacheStore, noopCache),
+    Effect.provideService(ShowcaseRepository, showcaseRepository),
+    Effect.provideService(OrganizationRepository, {
+      findById: (id) =>
+        requestingOrg ? Effect.succeed(requestingOrg) : Effect.fail(new NotFoundError({ entity: "Organization", id })),
+      listByUserId: () => Effect.die(new Error("unused")),
+      save: () => Effect.die(new Error("unused")),
+      delete: () => Effect.die(new Error("unused")),
+      countBySlug: () => Effect.die(new Error("unused")),
+      listExpiredUnclaimed: () => Effect.die(new Error("unused")),
+    }),
+  )
+
+const orgWithFlag = (wantsShowcase: boolean | undefined) =>
+  createOrganization({
+    id: REQUESTING_ORG_ID,
+    name: "Requesting Org",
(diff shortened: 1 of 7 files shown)


packages/domain/showcase/src/use-cases/resolve-showcase.ts
```diff
@@ -0,0 +1,127 @@
...
+} from "@domain/shared"
+import { Effect } from "effect"
+import { z } from "zod"
+import { ShowcaseRepository } from "../ports/showcase-repository.ts"
+
+/**
+ * Redis key for the resolved showcase pointer. It names the one live showcase
+ * org+project, so it is global (not org-scoped) and carries no `org:` prefix;
+ * the `latitude:` namespace is applied by the Redis client. The pointer table
+ * is the source of truth and invalidates this key when it changes.
```
geclos, line 20: Docstring is ahead of the code: nothing currently invalidates `showcase:current`. `create-showcase.ts` never touches the cache, and there's no update/swap path yet, so *"the source of truth ... invalidates this key when it changes"* isn't true today.

It's harmless right now — `create` leaves `currentProjectId` null and null is never cached — but it's a latent landmine: whoever implements the current→next swap **must** `delete(SHOWCASE_POINTER_CACHE_KEY)`, or a stale pointer serves the wrong project for up to the 300s TTL. Consider softening this to "the swap path must invalidate this key" until that code lands, so the comment doesn't read as an already-honored invariant.

---

packages/domain/showcase/src/use-cases/resolve-showcase.test.ts
```diff
@@ -0,0 +1,144 @@
...
+const SHOWCASE_ORG_ID = OrganizationId("showcaseorgggggggggggg01")
+const SHOWCASE_PROJECT_ID = ProjectId("showcaseprojectttttttt01")
+
+const sqlClient: SqlClientShape = {
+  organizationId: REQUESTING_ORG_ID,
+  transaction: (effect) => effect,
+  query: () => Effect.die(new Error("unexpected query")),
+}
+
+const noopCache: CacheStoreShape = {
```
geclos, line 28: All five cases inject this no-op cache, so the caching path in `resolveShowcasePointer` — `parseCachedPointer`, the cache-hit early return, and the `set(..., { ttlSeconds })` on a miss — has no coverage. Since the Redis cache is the core of this chokepoint, worth adding: (1) a store returning a serialized pointer to exercise the hit branch, and (2) an assertion that `set` is called with the resolved value on a miss (and that a live resolution, not a null, is what gets cached).

---

geclos (commented):
## Review — showcase resolver chokepoint (S2)

**Verdict: solid, no blocking issues.** Two non-blocking notes inline. The domain/web layering, tenancy scoping, and error semantics are all correct.

### Verified correct
- **RLS reasoning holds.** The docstrings claim `showcase` and `organizations` have no RLS policy, so the runtime `latitude_app` user reads them directly under the requesting-org scope. Confirmed against the S1 schema/migration: `showcase` has no `organizationRLSPolicy`, the migration grants `SELECT`/`UPDATE` only, and `find()` reads the `id=1` singleton directly. Correctly uses `getPostgresClient()` rather than the admin client — unlike `resolveSandboxAccess`, which needs admin only for its system-scope fallback.
- **Effect requirements satisfied.** `SqlClient | CacheStore | ShowcaseRepository | OrganizationRepository` are all provided by `withPostgres(Layer.merge(...))` + `RedisCacheStoreLive` in `resolveShowcaseAccess`.
- **Uniform 404** — missing requesting org, `wantsShowcase` false/absent, no showcase, and null `currentProjectId` all collapse to the same `NotFoundError`, leaking nothing about whether a showcase exists. Well covered by the 5 tests.
- **Global (non-org-scoped) cache key** `showcase:current` is a deliberate, documented exception to the `org:`-prefix convention — legitimate since it names the single showcase, and the `latitude:` namespace is applied by the client.
- Cache errors correctly degrade to the DB read, and only live resolutions are cached (never null).

### Non-blocking notes (inline)
1. `resolve-showcase.ts` — docstring claims cache invalidation that isn't implemented yet (latent landmine for the future swap path).
2. `resolve-showcase.test.ts` — the cache read/hit/set path is untested (all cases use a no-op cache).

---

PR #3882, opened 2026-07-06
feat(ui): AgentTextarea shader input for agent generation

csansoon:
Adds `AgentTextarea` to `@repo/ui`: a textarea for fields whose content triggers an AI agent, framed by a WebGL shader that signals the input's nature while idle and takes the field over while the agent works. First consumer is the signal builder's describe-first step, replacing the plain textarea + spinner row.

https://github.com/user-attachments/assets/b5104036-a317-4774-932c-93172c2a159b

## the component

Idle, it looks like a regular textarea — the shader only draws an animated hairline border: traveling brand-color spots over a constant ring, thickening slightly on focus. Setting `status` (a `string | null`) puts it in the loading state: the border shrinks back into its hairline while a translucent tide rises from the bottom edge, the typed text fades out, and the status string crossfades in the center as it changes. Component size never changes — the textarea stays mounted at `opacity-0` so autosize height holds. Resolve reverses everything.

The fill is "speaking pools": a quiet slice of the brand ramp with five soft pools of light rising and falling along the bottom edge out of phase, like a slowed-down voice level meter. It paints only where the composition departs from its anchor color (black in dark, near-white in light), over a `bg-background` backdrop layer — so the surface you see while loading is the textarea's own background in both themes, and loading reads as added light rather than a background swap. Everything is parameterized in box fractions, so the composition holds at any textarea size.

`useStagedStatus(stages, active)` ships alongside: it maps elapsed time to stage labels (`{ atSeconds, label }[]`) for fake-progress flows like the ~30s signal generation.

## the shader

One WebGL1 canvas per instance (Safari-compatible, zero dependencies), one fragment shader for both border and fill, driven by smoothed uniforms (`coverage`, `intensity`, `focus`, time scale). The border treatment is adapted from Paper Design's pulsing-border shader (Apache-2.0), reworked for textarea geometry: spots travel by perimeter arc-length with offset-contour corner radii, so speed and size stay uniform around corners of any aspect ratio. The fill ramp transcribes the exact 10 radial stops of the design system's marketing gradient preset A (`latitude-design-system/agent/visual/gradients/A.json`); marketing gradients are formally marketing-only, so this is a deliberate crossover scoped to the agent-working moment. Border and fill share one five-slot palette (blue, lavender, cyan-white, blue, yellow) with light-mode counterparts mixed by a `u_light` uniform fed from a `.dark`-class MutationObserver.

Tuning happened in a temporary live-settings playground; the final values are baked as GLSL and module constants, and the settings surface was removed — the public API is `status` plus the usual textarea props.

## lifecycle and performance

`prefers-reduced-motion`, WebGL unavailable, shader compile failure, and unrecovered context loss all fall back to a static CSS border/fill; the component stays a fully functional textarea. Cleanup never calls `loseContext()` — React Suspense/Activity reconnects re-run effects, and a context lost by our own cleanup is permanently unrestorable (its lost event fires after the listeners are gone); the backing store shrinks to 1×1 instead.

The render loop pauses when the tab is hidden or the canvas is offscreen, caps near 60fps so 120Hz displays don't double GPU cost, and caps DPR at 2. The fragment shader early-exits outside the border band and skips the pool composition wherever the fill is invisible — interior pixels cost almost nothing while idle. The textarea sits on its own compositor layer (`transform-gpu`) so keystroke repaints don't contend with the animating canvas.

## surfaces

- Signal builder, describe step: `status` driven by the generation worker's step events.
- Design-system index: a showcase section with a simulated generation run (the earlier standalone playground page is removed).

Stage selection logic is unit-tested; the shader states were verified manually in light and dark, in the builder modal and the showcase.


packages/ui/src/components/agent-textarea/shader-surface.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const canvasRef = useRef<HTMLCanvasElement | null>(null)
+  const [fallback, setFallback] = useState(false)
+
+  // fragmentSource is read once here; remount (key) to switch shaders.
+  useMountEffect(() => {
+    const canvas = canvasRef.current
+    if (canvas === null) return
+    const disable = (reason: string) => {
+      console.warn(`AgentTextarea shader disabled: ${reason}`)
+      setFallback(true)
```
geclos, line 125: **Fallback path bypasses effect cleanup → observers leak for the component's lifetime.** `useMountEffect` is `useEffect(fn, [])`, so the cleanup only runs on unmount — never when `fallback` flips to `true`. When `disable()` is reached at runtime (shader compile failure, or context lost past the 1.5s grace window), the fallback div renders and the canvas unmounts, but the `MutationObserver` on `document.documentElement`, the `ResizeObserver`, the `IntersectionObserver`, and the `visibilitychange` listener stay attached. The theme observer in particular keeps invoking `readLightTarget()` on every dark/light toggle indefinitely.

Consider tearing down the observers/listeners inside `disable()` (extract a `teardown()` the cleanup and `disable()` both call), so the fallback state stops doing work.

---

packages/ui/src/components/agent-textarea/shader-surface.tsx
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    let restoreTimer: ReturnType<typeof setTimeout> | null = null
+    const onContextLost = (event: Event) => {
+      event.preventDefault()
+      contextLost = true
+      syncRunning()
+      restoreTimer = setTimeout(() => disable("WebGL context lost"), CONTEXT_RESTORE_GRACE_MS)
+    }
+    const onContextRestored = () => {
```
geclos, line 248: **Runaway rAF after a late context restore.** Because the `webglcontextrestored` listener is never removed on the fallback path (same root cause as the observer leak above), this sequence burns GPU/CPU indefinitely: context is lost → the 1.5s grace timer fires `disable()` (fallback shown, canvas detached) → the driver later dispatches `webglcontextrestored` → `onContextRestored` → `initialize()` → `syncRunning()` restarts the `requestAnimationFrame` loop, which then calls `gl.drawArrays` every frame on the detached 1×1 canvas until `AgentTextarea` unmounts. A `teardown()` in `disable()` that removes these listeners (and guards `initialize()` behind `!fallback`) would close this.

---

packages/ui/src/components/agent-textarea/use-staged-status.ts
```diff
@@ -0,0 +1,32 @@
...
+  return stages[Math.max(reached - 1, 0)]?.label ?? null
+}
+
+export function useStagedStatus(stages: ReadonlyArray<StagedStatusStage>, active: boolean): string | null {
+  const [, setTick] = useState(0)
+  const activeRef = useRef(active)
+  const startedAtRef = useRef<number | null>(null)
+  activeRef.current = active
+  if (active && startedAtRef.current === null) startedAtRef.current = Date.now()
+  if (!active) startedAtRef.current = null
```
geclos: **A transient `active` false→true flip regresses the stage back to the first.** `if (!active) startedAtRef.current = null` clears the start timestamp on *any* inactive render, and the next active render re-captures `Date.now()`. If a consumer's `active`/loading derivation flickers `false` for a single render after, say, 10s of loading (a debounce, a re-derived boolean), elapsed jumps back to ~0 and the label regresses from "Writing" to "Reading", restarting the whole multi-second progression even though loading never actually stopped. Since this is an exported reusable hook, it's worth only resetting when loading has genuinely ended (e.g. keep the start time and reset on a true→false→true edge you actually intend), rather than on every inactive render.

---

packages/ui/src/components/agent-textarea/shader-surface.tsx
```diff
@@ -0,0 +1,263 @@
...
+      contextLost = false
+      syncRunning()
+    }
+
+    let restoreTimer: ReturnType<typeof setTimeout> | null = null
+    const onContextLost = (event: Event) => {
+      event.preventDefault()
+      contextLost = true
+      syncRunning()
+      restoreTimer = setTimeout(() => disable("WebGL context lost"), CONTEXT_RESTORE_GRACE_MS)
```
geclos, line 246: **Orphaned `restoreTimer` can force fallback despite a healthy context.** `restoreTimer` is assigned unconditionally here and in the revive branch (line 226), without clearing any pending timer first. On a double context-loss (lost → briefly restored → lost again), this second assignment orphans the first timer reference; `onContextRestored` later clears only the current reference, so the orphaned timer still fires `disable("WebGL context lost")` after 1.5s and drops to the static fallback even though the context was successfully restored. Clear the existing timer before (re)scheduling: `if (restoreTimer !== null) clearTimeout(restoreTimer)`.

---

packages/ui/src/components/agent-textarea/shader-surface.tsx
```diff
@@ -0,0 +1,263 @@
...
+  if (vertex === null || fragment === null) return null
+  const program = gl.createProgram()
+  if (program === null) return null
+  gl.attachShader(program, vertex)
+  gl.attachShader(program, fragment)
+  gl.linkProgram(program)
+  if (!gl.getProgramParameter(program, gl.LINK_STATUS)) return null
+
+  gl.useProgram(program)
+  gl.bindBuffer(gl.ARRAY_BUFFER, gl.createBuffer())
```
geclos: **GL program, shaders, and vertex buffer are never deleted → leak across re-initialize.** `setupProgram` creates a `WebGLProgram`, two `WebGLShader`s, and a fresh buffer here, but nothing is ever `gl.deleteProgram` / `gl.deleteShader` / `gl.deleteBuffer`'d. On a Suspense/Activity hide→show cycle the effect re-runs and — per the file's own comment — `getContext` returns the same non-lost context, so `initialize()` calls `setupProgram` again and orphans the previous set on that context. Repeated hide/show or context-lost/restored cycles accumulate GPU objects. Delete the old program/shaders/buffer before recompiling (or reuse the existing program instead of rebuilding).

---

packages/ui/src/components/agent-textarea/agent-textarea.tsx
```diff
@@ -0,0 +1,130 @@
...
+  className,
+  minRows = 2,
+  maxRows,
+  disabled,
+  onFocus,
+  onBlur,
+  ...props
+}: AgentTextareaProps) {
+  const [focused, setFocused] = useState(false)
+  const loading = status !== null
```
geclos: **Empty-string `status` silently activates the loading state.** `loading = status !== null` means `status=""` yields `loading === true`: the textarea gets disabled and faded to `opacity-0`, `showDestructiveBorder` is forced off (suppressing error styling), and the overlay renders an empty `Text.H5` — an inert, invisible, error-suppressed widget with no label explaining why. Consider treating empty/whitespace as "not loading" (e.g. `const loading = status != null && status.length > 0`), so only a meaningful status string enters the loading state.

---

packages/ui/src/components/agent-textarea/use-staged-status.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { useMountEffect } from "../../hooks/use-mount-effect.ts"
+
+interface StagedStatusStage {
+  readonly atSeconds: number
+  readonly label: string
+}
+
+export function selectStage(stages: ReadonlyArray<StagedStatusStage>, elapsedSeconds: number): string | null {
+  const reached = stages.filter((stage) => elapsedSeconds >= stage.atSeconds).length
```
geclos: **`selectStage` assumes `stages` is sorted ascending by `atSeconds`, undocumented.** It counts reached stages and indexes `stages[reached - 1]`, which is only correct for ascending input. e.g. `selectStage([{0,"Reading"},{14,"Refining"},{3,"Writing"}], 5)` → `reached = 2` → returns `"Refining"` when the correct current stage is `"Writing"`. For an exported hook, either sort defensively or document the precondition (and ideally add a test covering unsorted input).

---

packages/ui/src/components/agent-textarea/agent-textarea.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const showDestructiveBorder = errors !== undefined && errors.length > 0 && !loading
+
+  const [statusPair, setStatusPair] = useState<{ current: string | null; previous: string | null }>({
+    current: status,
+    previous: null,
+  })
```
geclos, line 49: **Minor/cosmetic:** on `status` A→null, this setState-during-render commits `{current: null, previous: A}`, but that same render has `loading === false`, so the whole overlay block unmounts before the `animate-out`/`fade-out` can play. The last status label pops out instantly rather than crossfading; only the textarea's opacity transitions. Intermediate transitions (A→B) animate correctly — it's just the final label on exit. If the exit crossfade matters, the overlay would need to stay mounted for one transition duration after loading ends.

---

geclos (commented):
Automated review of the `AgentTextarea` component. Really nicely built — the shader lifecycle handling (reduced-motion, context loss, visibility/intersection pausing, the Suspense/Activity `loseContext` reasoning) is thorough, and prop forwarding, SSR/hydration, the crossfade state machine, and `selectStage` boundaries all hold up. No crashing bugs found; everything below is a leak, a latent robustness edge, or cosmetic.

**The one theme worth acting on first** is that the runtime fallback path bypasses the effect's cleanup. `useMountEffect` is `useEffect(fn, [])`, so cleanup runs only on unmount — when `disable()` sets `fallback` at runtime, the canvas unmounts but the observers, listeners, and (on a late context-restore) the rAF loop keep running for the component's lifetime. Extracting a `teardown()` that both the cleanup and `disable()` call would resolve the two highest findings together.

Findings, most-severe first:
1. Fallback path leaks observers/listeners (shader-surface).
2. Late `webglcontextrestored` after grace-disable restarts a rAF loop drawing to a detached canvas (shader-surface).
3. `useStagedStatus` regresses to stage 1 on a transient `active` flip.
4. `onContextLost` orphans the previous `restoreTimer`, forcing fallback despite a healthy restored context.
5. GL program/shaders/buffer never deleted → leak on re-initialize.
6. `status=""` silently disables and hides the textarea.
7. `selectStage` depends on `stages` being sorted (undocumented).
8. Last status label skips its exit crossfade (cosmetic).

Non-blocking heads-up unrelated to the diff: GitHub reports this PR as `mergeable_state: dirty` (merge conflicts with `development`) — worth a rebase.

---

packages/ui/src/components/agent-textarea/shader-surface.tsx
csansoon: Fixed in eab750b27: `disable()` now runs the same idempotent `teardown()` the unmount cleanup uses, so the theme/resize/intersection observers and the visibilitychange + context listeners are all removed when the fallback engages at runtime.

packages/ui/src/components/agent-textarea/shader-surface.tsx
csansoon: Fixed by the same teardown in eab750b27 — `webglcontextrestored` is removed when the grace timer disables, so a late restore can no longer reach `initialize()` or restart the rAF loop.

packages/ui/src/components/agent-textarea/use-staged-status.ts
csansoon: Fixed in 684c6adba: the start-time reset moved from render to the 1s interval tick, so only an inactive state that persists to the next tick clears it. Trade-off: a genuine stop+restart within the same second continues the previous timeline — the right bias for a progress label.

packages/ui/src/components/agent-textarea/shader-surface.tsx
csansoon: Fixed in 89fa65b9a — the pending grace timer is cleared before scheduling a new one on each context loss.

packages/ui/src/components/agent-textarea/shader-surface.tsx
csansoon: Fixed in 940eadc64: program, shaders, and buffer are tracked as one handle bundle, released before every recompile and on unmount; the compile/link failure paths also delete whatever they had created.

packages/ui/src/components/agent-textarea/agent-textarea.tsx
csansoon: Fixed in c724fa286: blank (empty or whitespace-only) status no longer enters the loading state.

packages/ui/src/components/agent-textarea/use-staged-status.ts
csansoon: Fixed in cf72082f2: `selectStage` now picks the latest reached stage regardless of input order, falling back to the earliest stage before any is reached, with an unsorted-input test.

packages/ui/src/components/agent-textarea/agent-textarea.tsx
csansoon: Leaving this one as is: the exit pop is masked by the fill draining over the same ~500ms (the label sits on a surface that is already dissolving), and keeping the overlay mounted past `loading` would add a delayed-unmount timer for a label that is disappearing anyway. Will revisit if the exit reads poorly in practice.

csansoon:
Review addressed: 7 findings fixed in eab750b27, 684c6adba, 89fa65b9a, 940eadc64, c724fa286, cf72082f2 (one commit per issue; the two fallback-path leaks share the teardown fix). The cosmetic exit-crossfade finding is declined with rationale in its thread. Also merged `origin/development` — the design-system showcase moved to the new `apps/design-system` app as an `agent-textarea` registry entry, and the PR is mergeable again.

PR #3883, opened 2026-07-06
fix(design-system): reuse the main app's favicon

ntizish:
## Summary

`apps/design-system` had no favicon at all. This copies `apps/web/public/favicon.svg` into `apps/design-system/public/` and links it in `index.html`, so the design system tab icon matches the main app instead of showing the browser default.


andresgutgon (approved):

geclos (approved):
Reviewed — looks good, safe to merge. ✅

**What it does:** Adds `apps/design-system/public/favicon.svg` (a byte-for-byte copy of `apps/web/public/favicon.svg`) and a `<link rel="icon" type="image/svg+xml" href="/favicon.svg" />` in `index.html`.

**Verified:**
- The copied SVG is identical to the source in `apps/web`. ✔
- The `<link>` matches how the main app declares its icon (`apps/web/src/routes/__root.tsx` → `links: [{ rel: "icon", type: "image/svg+xml", href: "/favicon.svg" }]`), so the tab icon will match. ✔
- `apps/design-system` is a plain Vite app; Vite serves `public/` at the root, so `/favicon.svg` resolves correctly. Vercel preview deploy is green. ✔

**Two non-blocking notes (neither introduced by this PR):**
1. The SVG uses JSX-style `fillRule`/`clipRule` attributes rather than the standard SVG `fill-rule`/`clip-rule`. In a standalone `.svg` served as a static asset, browsers ignore the camelCase forms and fall back to `fill-rule: nonzero`, which can change how the ring-shaped paths render. Since this is a verbatim copy of the already-shipping web favicon it renders identically to the main app, so it's fine to merge as-is — but worth fixing at the source (`apps/web/public/favicon.svg`) in a follow-up if the icon ever looks off.
2. The asset is now duplicated across two apps' `public/` dirs. Given they're separate Vite builds with no shared static-asset pipeline, copying is the pragmatic choice; just a maintenance heads-up if the brand mark ever changes.

---

PR #3887, opened 2026-07-06
feat(showcase): regeneration + swap + daily cron (S4)

andresgutgon:
## What & why

**Task S4** of the Showcase Demo Project (umbrella **#3821**, spec on `fix/one-demo-project-for-all`). Implements blue/green **regeneration + the atomic swap + a daily cron** for the shared read-only Showcase, building on the merged Phase-1 groundwork + S1 (pointer table, `@domain/showcase`, guarded create).

Regeneration builds a fresh `next` project via `seedDemoProjectWorkflow` (re-anchored to "now"), quality-gates it, then **auto-swaps** the pointer. **S4 owns the atomic swap.**

## How it works

`regenerateShowcaseWorkflow` (Temporal) drives the state machine:

```
child seed build → quality gate (≥N traces landed) → mark next ready → atomic swap
```

- **Begin** (daily cron worker, `apps/workers/src/workers/showcase.ts`): resolves the showcase org from the pointer, provisions a fresh `next` project (`settings.isShowcase`), an api key, and queue assignees, flips the pointer to `building`, and starts the workflow. No-ops when no showcase exists / a build is already in flight / the showcase org has no members yet.
- **Build**: `executeChild(seedDemoProjectWorkflow, …)` — the same seed the per-signup demo used, re-anchored via `timelineAnchorIso`.
- **Quality gate**: `assertShowcaseNextQualityActivity` asserts the built project actually has telemetry (distinct traces ≥ 100); a silently empty/partial seed fails here (non-retryable) so the swap never runs.
- **Swap** (`swapShowcaseUseCase`, S4-owned): an atomic, **row-locked** (`SELECT … FOR UPDATE`) transaction that asserts `next_state='ready'`, sets `current ← next`, clears `next`, resets to idle — serializing a manual backoffice swap against the scheduled one — then invalidates the Redis cache (`latitude:showcase:current`).

**Failure is not a modeled state**: a failed/thin build throws before the swap, so `current` is untouched (→ Datadog) and the stale `next` is reclaimed later (S5). No `failed` enum value.

**Cron**: daily off-peak `0 4 * * *` UTC on a new `showcase` queue, registered idempotently on worker boot.

**Exclusions**: showcase projects carry `settings.isShowcase` and are excluded from taxonomy gardening (alongside `isSample`). Retention: showcase telemetry already bypasses per-org plan retention via the seed's fixed 30-day retention + daily re-anchoring (reused seed, unchanged) — no live-ingest path touches it.

## Behavior-neutral

The showcase isn't surfaced to users yet (S3/S6/C1), so regenerating/swapping has **no user-facing effect**.

## Tests

- `swap-showcase.test.ts` (use-case, fake repo): swap asserts `next_state='ready'` first; happy-path flip + cache invalidation; serialization (second swap sees idle → `ShowcaseNotReadyError`).
- `showcase-repository.test.ts` (live PGlite adapter): `beginNextBuild` / `markNextReady` / `swap` transitions; not-ready build leaves `current` intact; no double-flip.

## Review focus

- The swap transaction / row-lock semantics in `showcase-repository.ts`.
- The workflow ordering as the failure gate (`regenerate-showcase-workflow.ts`) + the explicit quality-gate threshold.
- The begin-step guards in the cron worker (skip when no showcase / build in flight / no members) and the choice to place project/api-key provisioning in the worker (avoids new domain dep edges).

## Out of scope (later tasks)

Retirement/cleanup of the OLD project + stale-`building` GC (S5); UI/loader/collection-merge (S3/S6); backoffice trigger (S7).


andresgutgon:
@claude review

packages/platform/db-postgres/src/repositories/showcase-repository.ts
```diff
@@ -46,6 +54,68 @@ export const ShowcaseRepositoryLive = Layer.effect(
...
+              if (!locked) return yield* Effect.fail(new ShowcaseNotFoundError())
+              if (locked.nextState !== "ready" || !locked.nextProjectId) {
+                return yield* Effect.fail(new ShowcaseNotReadyError({ nextState: locked.nextState }))
+              }
+
+              const [swapped] = yield* sqlClient.query((db) =>
+                db
+                  .update(showcase)
+                  .set({
+                    currentProjectId: locked.nextProjectId,
```
geclos, line 115: Nit (non-blocking): the swap promotes `locked.nextProjectId` from the pointer, not the specific project the workflow gated (`input.projectId`). In the normal single-flight path these are identical, so this is safe today — it relies on BullMQ's per-job locking plus the worker's `nextState !== null` guard keeping regenerations strictly single-flight. Worth a one-line comment noting that dependency, since the swap trusts the pointer rather than re-checking it against the just-gated project. If two regenerations ever overlapped, the gate would run on project A while the swap promoted project B.

---

packages/platform/db-postgres/src/repositories/showcase-repository.ts
```diff
@@ -46,6 +54,68 @@ export const ShowcaseRepositoryLive = Layer.effect(
...
+        Effect.gen(function* () {
+          const sqlClient = (yield* SqlClient) as SqlClientShape<Operator>
+          const [row] = yield* sqlClient.query((db) =>
+            db
+              .update(showcase)
+              .set({ nextState: "ready", updatedAt: new Date() })
+              .where(eq(showcase.id, SHOWCASE_SINGLETON_ID))
+              .returning(),
+          )
+          if (!row || !row.nextProjectId) return yield* Effect.fail(new ShowcaseNotFoundError())
```
geclos: Minor robustness (non-blocking): `markNextReady` runs the `UPDATE … set nextState: "ready"` first and only *then* checks `!row.nextProjectId`. If it were ever called with a null `next`, it would have already written an inconsistent `ready` + null-`nextProjectId` row before failing. Not reachable in the current flow (always preceded by `beginNextBuild`, which sets `nextProjectId`), but asserting `nextProjectId IS NOT NULL` in the `WHERE` — or reading before writing — would make the port method self-defensive.

---

geclos (commented):
## Review — showcase regeneration + swap + daily cron (S4)

**Verdict: solid, well-tested, no blocking issues.** Verified locally against head `3be3bd7`:

- `@domain/showcase` tests: 5 passed (swap ready-path, still-building rejection, concurrent-swap serialization)
- `@platform/db-postgres` PGlite repo tests: 5 passed (beginNextBuild / markNextReady / swap + FOR-UPDATE serialization + no-double-flip)
- Typecheck clean: `@domain/showcase`, `@domain/shared`, `@domain/queue`, `@platform/db-postgres`, `apps/workers`, `apps/workflows`

### Strengths
- **Atomic swap is correct** — `SELECT … FOR UPDATE` on the singleton, assert `next_state = 'ready'`, flip, reset, all in one transaction; the fake repo faithfully mirrors the assert-then-flip so both test layers cover the manual-vs-scheduled race guard.
- **Clean failure model** — a thin/failed build throws before the swap (`ShowcaseQualityGateError` is non-retryable, so the deterministic count doesn't burn the retry budget), leaving `current` untouched; post-commit crash on swap self-heals via the `ShowcaseNotReadyError` → "already swapped" catch.
- **Correct Temporal hygiene** — `timelineAnchorIso`/`queueAssigneeUserIds` captured in the worker before start (no `new Date()`/`Math.random()` in workflow code), fixed regen `workflowId` + `WorkflowAlreadyStartedError` dedup, per-project child seed id.
- Best-effort cache invalidation can't roll back a committed swap and self-heals via TTL; `isShowcase` gardening/retention exclusion mirrors the existing `isSample` pattern.

Two low-severity, non-blocking notes inline (swap trusting the pointer vs. the gated project id; `markNextReady` validating after writing). Neither affects the happy path or the tested behaviors.

---

apps/workers/src/workers/showcase.ts
```diff
@@ -0,0 +1,170 @@
...
+        )
+
+        if (prepared === null) {
+          logger.info("Showcase organization has no members yet — skipping regeneration", { organizationId })
+          return
+        }
+
+        const workflowStarter = yield* Effect.promise(() => getWorkflowStarter())
+        yield* workflowStarter
+          .start("regenerateShowcaseWorkflow", prepared, { workflowId: REGENERATE_WORKFLOW_ID })
```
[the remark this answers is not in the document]
andresgutgon: Fixed in `83167bd`. The cron handler no longer skips when a build is in flight — it now **resumes** the pointer's existing `next` via an idempotent workflow start (`prepareRegeneration`). So a transient failure between `beginNextBuild` and a durable workflow start no longer wedges the pointer in `building`: the next run re-drives the same `next` project (a genuinely in-flight run still dedups via `workflowId`), and no new project is provisioned while `next` is set, so nothing is orphaned.

packages/platform/db-postgres/src/repositories/showcase-repository.ts
```diff
@@ -46,6 +54,68 @@ export const ShowcaseRepositoryLive = Layer.effect(
...
+          return toDomainShowcase(row)
+        }),
```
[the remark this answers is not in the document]
andresgutgon: Fixed in `83167bd`. `markNextReady` now guards `next_project_id IS NOT NULL` in the `WHERE`, so an idle pointer matches no row and the `UPDATE` has no side effect — no more inconsistent `ready` + null-`next`. Added a live-repo regression test (`markNextReady on an idle pointer fails without writing an inconsistent ready state`).

apps/workflows/src/activities/showcase-regeneration-activities.ts
```diff
@@ -0,0 +1,130 @@
...
+            currentProjectId: showcase.currentProjectId,
+          }),
+        ),
+      ),
+      Effect.asVoid,
+      Effect.catchTag("ShowcaseNotReadyError", (error) =>
+        Effect.sync(() =>
+          logger.warn("Showcase swap skipped — pointer not ready (already swapped?)", { nextState: error.nextState }),
+        ),
+      ),
```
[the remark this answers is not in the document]
andresgutgon: Fixed in `83167bd`. The activity now only swallows `ShowcaseNotReadyError` when `nextState === null` (idle / already-swapped). A still-`building` (or otherwise non-null) state re-fails, so the workflow reports the inconsistency instead of silently stopping without a swap.

packages/platform/db-postgres/src/repositories/showcase-repository.ts
andresgutgon: Added the clarifying comment in `83167bd` — the swap promotes whatever `next` the pointer names, trusting strictly single-flight regeneration (BullMQ per-job lock + the worker resuming an existing `next` rather than forking a second one).

packages/platform/db-postgres/src/repositories/showcase-repository.ts
andresgutgon: Fixed in `83167bd` — same change as the `markNextReady` thread above: guarded `next_project_id IS NOT NULL` in the `WHERE` so a null-`next` call is a no-op `UPDATE` instead of writing `ready` and then failing. Added a regression test.

geclos:
can we move the cleanup of annot queues to another pr please

PR #3888, opened 2026-07-06
Add design.latitude.so Vercel CNAME to production DNS

geclos:
## Summary

Adds a production Route53 CNAME for `design.latitude.so` pointing to Vercel (`6ac0d6b5a212d2a5.vercel-dns-016.com.`), matching the pattern used for other Vercel-hosted subdomains like `one-pager.latitude.so` and `41st.latitude.so`.

The record was already applied to production via `pulumi up --stack production` so this PR keeps infrastructure code in sync with live DNS.

## Test plan

- [x] `pulumi preview --stack production` showed only `latitude-production-design-cname` create
- [x] `pulumi up --stack production` applied the record
- [x] `dig +short design.latitude.so CNAME` returns `6ac0d6b5a212d2a5.vercel-dns-016.com.`

### DIFF
--- a/infra/lib/dns.ts
+++ b/infra/lib/dns.ts
@@ -300,6 +300,16 @@ export function createDnsRecords(
       allowOverwrite: true,
     })
 
+    // CNAME for design.latitude.so -> Vercel
+    records.designCname = new aws.route53.Record(`${name}-design-cname`, {
+      zoneId: hostedZoneId,
+      name: "design.latitude.so",
+      type: "CNAME",
+      records: ["6ac0d6b5a212d2a5.vercel-dns-016.com."],
+      ttl: 300,
+      allowOverwrite: true,
+    })
+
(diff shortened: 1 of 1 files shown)


geclos (commented):
Reviewed — looks clean and low-risk. 👍

- The new `design.latitude.so` CNAME is structurally identical to its siblings (`www`, `41st`, `one-pager`): same `zoneId`, `type: "CNAME"`, `ttl: 300`, `allowOverwrite: true`, and the same `vercel-dns-016` target family `41st.latitude.so` already uses.
- `records` is typed `Record<string, Route53Record>`, so the new `designCname` key needs no type change.
- Comment style matches the existing per-record convention.

Non-blocking note: the description mentions the record was already applied to prod via `pulumi up` before this PR. That's an accepted IaC-sync pattern here and it's transparently disclosed — just flagging that trunk briefly reflected prod rather than the reverse.

Nothing needs changing.

---

PR #3890, opened 2026-07-06
refactor(api): extract operation definitions into transport-neutral @repo/operations

csansoon:
Moves every public API endpoint definition out of `apps/api` into a new workspace package, `@repo/operations` (`packages/operations`), and makes the definitions transport-neutral. This is the groundwork for agentic signal creation: an internal agent needs to call the same read operations the MCP exposes, in-process from a worker, without HTTP round-trips or minting credentials. It also tightens the existing "one declaration fans out everywhere" pipeline — the same definition now produces the HTTP route, OpenAPI operation, MCP tool, TS/Python SDK methods, CLI command, and an in-process agent tool.

External surface is provably unchanged: `openapi.json` and `mcp.json` re-emit byte-identical, and a full `pnpm generate:all` produces zero drift in the TS SDK, Python SDK, and CLI.

## the new layer

`@repo/operations` sits between apps and domain. It owns the operation contracts — route config, public Zod schemas, and the logic that validates input, orchestrates `@domain/*` use-cases, and maps to response shapes. `apps/api` keeps only transport: middleware (auth, org context, rate limiting), the MCP HTTP transport, the non-operation routes (health, well-known, bootstrap), manifest emit scripts, and mounting.

The factory (renamed `defineOperation`) accepts two forms:

- **`execute(input, ctx)`** — transport-neutral: an Effect receiving pre-validated `{ params, query, body }` and an `OperationContext` (organization, auth, platform clients — everything `c.var` used to provide). The HTTP handler is generated; failures in the error channel reach `honoErrorHandler` exactly as before. The `tools` module (8 read-only analytics operations) is converted as the pilot; its pre-existing HTTP tests pass unchanged.
- **`handler`** — the existing Hono handler, unchanged. The other 16 modules moved as-is and convert opportunistically.

Two route-config fields become declarative: `group`/`sdkMethod` replace the per-module fern helpers (renamed in place to the `x-fern-*` extensions with an order-preserving transform, since emitted key order follows declaration order and the manifests are diffed byte-for-byte in CI), and `rateLimitTier` replaces the imperative `mountHttp(app, createTierRateLimiter(...))` calls. The per-module `createXxxRoutes` factories collapse into `OperationModule` manifests mounted by `mountOperationModules`, which throws on a missing tier — the emit scripts boot the same assembly, so an undeclared tier cannot ship.

## agent toolsets

`defineToolset({ groups, exclude })` selects operations by `group` and shapes them as in-process tools. Selection is asserted at definition time: an unmatched group, a stale exclude, or a selected operation that is non-tool, not `readOnlyHint: true`, or handler-form throws — an agent can never be handed a mutation, by construction. `invokeOperation` runs a tool from the flat MCP-style input: validate against the flattened route schema (same coercions/defaults as the HTTP validators), split back into `{ params, query, body }`, call `execute` with a caller-built context. Tenancy is baked into the context; org identity is never part of the model-visible input.

Adding an operation to a selected group adds it to the toolset automatically, but not silently: each toolset has a checked-in manifest snapshot (`toolsets/__snapshots__/*.manifest.json`, serialized like `mcp.json`), so the diff shows exactly what an agent's tool surface gained. Ships `toolsAnalyticsToolset` (the 8 tools operations) as the first consumer.

## decisions worth knowing

- `typedResponses` is a literal-keyed twin of `openApiResponses` with identical runtime output, used only by execute-form operations. Retyping `openApiResponses` itself would strict-check ~100 legacy handlers at once. Its response keys stay literal because deriving them from a generic `extraErrors` type parameter collapses `OperationOutput` inference to a number index under tsgo — `extraErrors` statuses land in the spec but aren't typed as returnable variants yet.
- `rateLimitTier` is enforced by the mount-time throw rather than being type-required, which keeps machinery tests from carrying dead tiers; `group`/`sdkMethod` are type-required.
- Toolset definitions live in `packages/operations/src/toolsets/`, not `@domain/*` — a domain-owned toolset would create a package cycle since `@repo/operations` imports domain packages.
- `account.ts` split: `getAccount` moved; the unauthenticated bootstrap route stayed in `apps/api` (`routes/bootstrap.ts`) because it needs the RLS-bypassing admin client and global rate limiter.
- HTTP route tests stayed in `apps/api/src/routes/*.test.ts` deliberately — they test through `registerRoutes` + `app.fetch()` so the full middleware chain runs, and they double as the behavior-parity harness for handler→execute conversions.

## validation

Byte-identical manifest emission was gated after every stage. Fern regeneration (TS SDK, Python SDK, CLI) shows zero drift. Package tests (74) cover the factory, execute wrapper, mounting, invocation, and toolset assertions; route suites (signals, tools, api-keys, datasets) and the MCP transport tests pass. `members.test.ts` and `account.test.ts` fail locally on pre-existing env-dependent cases — verified identical failures on the parent commit.

Docs are synced: the api-endpoints skill is rewritten for the new authoring flow, architecture-boundaries and AGENTS.md describe the layer, and dev-docs path references are updated.

Follow-up track (separate PRs): the `AIAgent` tool-loop capability in `@domain/ai`, the signal-creation agent consuming a toolset from the worker, and opportunistic handler→execute conversions.


geclos:
This is critical make sure to manually test it connecting an mcp to your local instance and running some ops

csansoon:
@claude review this pr
@cursoragent review this pr

csansoon:
Thanks for the review — addressed the actionable findings. All three fixes keep `openapi.json` / `mcp.json` / the toolset snapshot **byte-identical** (verified), so nothing on the wire moved.

**#1 — registry double-registration (Medium): fixed in `84e8df3ff`.** Went with *fail-loud* rather than the silent dedupe guard: an operation `name` is both the OpenAPI `operationId` and the MCP tool name, so it must be globally unique — a collision is always a bug (or a re-run that forgot `resetOperationRegistry`). `registerOperation` now throws on a duplicate name. That matches the "assert at boot" ethos and turns a would-be flaky tool count into a hard failure; a silent dedupe could instead mask a real double-mount. Added a test.

**#2 — `flattenRouteInputSchema` per invoke (Low): fixed in `5a588c0d6`.** `toToolsetTool` now flattens the static route schema once at build time and reuses it via a new `invokeOperationWithFlat`; `invokeOperation` keeps the flatten-per-call convenience wrapper for one-off use. Agent hot loops no longer rebuild `z.object(shape)` per call.

**#3 — `typedResponses` swallowing `extraErrors` (Low): fixed in `a5dc13071`.** Removed `extraErrors` from the signature entirely — that's the pit-of-success version of your suggestion. Execute-form operations that genuinely need extra error statuses use `openApiResponses` (and forgo the typed output union) until the helper grows typed support. No caller passed `extraErrors`, so the manifests are unchanged.

**#4 — `RoleSchema` description (Minor): deferring, on purpose.** Valid nit, but it's pre-existing (the description came over verbatim when `account.ts` moved), and moving it changes the emitted `openapi.json` / `mcp.json` / SDK — which would break this PR's byte-identical guarantee, the whole validation story here. I'll fix it in the follow-up track that already owns spec/description changes rather than muddy the "move without drift" claim.

**#5 — comment density in `define-operation.ts` (Style):** intentional. This is the public API surface of a new package, which is the CLAUDE.md "contract docs where the file documents every sibling" exception, and most of the long blocks are documenting genuine traps (the currying, the order-preserving x-fern rename that byte-identity depends on). Happy to trim any that are pure rationale-narration if reviewers point at specific ones.

**Minor observations** — all by-design and documented (full `OperationContext` mirroring the middleware chain, mount-time `rateLimitTier` enforcement, handler-form ops staying MCP-visible via HTTP dispatch). No change.

@geclos — agreed that connecting a real MCP client to a local instance and running some ops is the key remaining validation before merge; the byte-identical emission proves the *shape* didn't drift but not that dispatch still works end-to-end. That manual pass is on the list before this merges.

geclos (dismissed):

geclos:
@cursoragent review

geclos:
@claude generate a publish an artifact to explain the diff of this pr, thanks

PR #3901, opened 2026-07-07
Fix modal searchable selects

geclos:
Fixes the searchable repository selector inside the Cursor connection modal. The dropdown now renders its popover content inside the modal DOM, so the dialog focus trap no longer treats the search input/options as outside interactions. Other popovers keep the existing portaled behavior by default.

Also updates the signal builder Custom script tab based on Agentation feedback:
- adds CodeMirror placeholder support in `RichTextEditor`
- shows a placeholder generated from `compileSettingsToScript` so the blank script editor displays a real evaluation script shape without seeding editable content
- removes the explanatory text under the detector tabs

Validated with browser QA:
- connected Cursor locally with the provided API key
- opened the repository selector, focused and typed in search, scrolled results, and selected `latitude-dev/latitude-llm`
- opened the signal builder Custom script tab and confirmed the generated placeholder is visible while Next remains disabled until real content is entered

Checks run:
- `pnpm --filter @repo/ui typecheck`
- `pnpm --filter @repo/ui check`
- commit hook workspace `format` + `knip`

UI changes are visible in the Cursor integration modal and the signal builder Custom script tab; screenshots or a short recording would help reviewers.

### DIFF
--- a/packages/ui/src/components/popover/primitives.tsx
+++ b/packages/ui/src/components/popover/primitives.tsx
@@ -14,16 +14,10 @@ const PopoverAnchor = PopoverPrimitive.Anchor
 const PopoverClose = PopoverPrimitive.Close
 
 type PopoverContentProps = ComponentPropsWithoutRef<typeof PopoverPrimitive.Content> & {
-  portaled?: boolean
+  container?: ComponentPropsWithoutRef<typeof PopoverPrimitive.Portal>["container"]
 }
 
-function PopoverContent({
-  className,
-  align = "center",
-  sideOffset = 4,
-  portaled = true,
-  ...props
-}: PopoverContentProps) {
+function PopoverContent({ className, align = "center", sideOffset = 4, container, ...props }: PopoverContentProps) {
   const content = (
     <DismissableLayerBranch>
       <PopoverPrimitive.Content
@@ -45,9 +39,7 @@ function PopoverContent({
     </DismissableLayerBranch>
   )
 
-  if (!portaled) return content
-
-  return <PopoverPrimitive.Portal>{content}</PopoverPrimitive.Portal>
+  return <PopoverPrimitive.Portal container={container}>{content}</PopoverPrimitive.Portal>
 }
 PopoverContent.displayName = PopoverPrimitive.Content.displayName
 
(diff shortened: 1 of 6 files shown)


geclos:
@claude review

packages/ui/src/components/popover/primitives.tsx
```diff
@@ -13,34 +13,41 @@ const PopoverAnchor = PopoverPrimitive.Anchor
...
+          "data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2",
+          zIndex.popover,
+          className,
+        )}
+        {...props}
+      />
+    </DismissableLayerBranch>
   )
+
+  if (!portaled) return content
```
andresgutgon: You don't need this refactor. Just pass the ref where it mounts 

Image

geclos: You were right. Radix `Popover.Portal` already supports `container`, so we do not need the `portaled={false}` escape hatch.

Changed in `21ab3ffd4`: `PopoverContent` now forwards an optional `container` to `PopoverPrimitive.Portal`, and searchable `Select` passes its wrapper ref as that container. That keeps the content portaled, but mounts it inside the dialog subtree instead of `document.body`.

QA with the Cursor modal after this change:
- connected locally with the Cursor API key
- opened the repository selector
- confirmed the search input focuses and accepts `latitude-llm`
- confirmed filtering and selecting `latitude-dev/latitude-llm` works
- confirmed the inner options container scrolls (`scrollTop = 800`)

Targeted checks passed locally: `@repo/ui` typecheck/check and `@app/web` typecheck/check. Commit hook also ran workspace format + knip.

packages/ui/src/components/popover/primitives.tsx
```diff
@@ -13,34 +13,33 @@ const PopoverAnchor = PopoverPrimitive.Anchor
...
-          {...props}
-        />
-      </DismissableLayerBranch>
-    </PopoverPrimitive.Portal>
+type PopoverContentProps = ComponentPropsWithoutRef<typeof PopoverPrimitive.Content> & {
+  container?: ComponentPropsWithoutRef<typeof PopoverPrimitive.Portal>["container"]
+}
+
+function PopoverContent({ className, align = "center", sideOffset = 4, container, ...props }: PopoverContentProps) {
+  const content = (
```
andresgutgon: Redundant variable assignation. Revert

geclos: Addressed in `7f0c4bad4`: removed the redundant `content` variable and returned the portal JSX directly, keeping the `container` behavior intact. Targeted `@repo/ui` check/typecheck pass locally.

andresgutgon (approved):

PR #3915, opened 2026-07-07
feat(billing): meter AI work per operation at cost-based credit prices

geclos:
## What does this PR do?

Changes how we bill AI work: instead of flat prices per feature scan, we now charge each LLM call and semantic query at what it actually costs us, plus a margin. Signal scans keep a flat baseline credit, with the AI charges stacked on top.

### How we bill now

| Action | Price |
|---|---|
| `trace` | 1 credit (unchanged) |
| `eval-scan` | 1 credit baseline for **every** live evaluation scan — LLM, semantic, or rules-only. |
| `llm-call` | estimated provider cost × 1.3, converted to credits ($0.002/credit), rounded up, minimum 1. Flat 30 credits when we can&#39;t estimate (model not in the pricing registry, or the provider didn&#39;t report usage). |
| `semantic-query` | estimated embed cost × 2, same conversion. Flat 15 credits when the adapter doesn&#39;t report tokens. |

So a rules-only scan bills 1 credit; a judge scan bills 1 credit + its LLM calls at cost. Cost comes from the provider-reported token usage, priced through the same model registry we already use to price customer spans (Voyage embeds use a hardcoded rate since they&#39;re not in the registry).

Billing happens once, at the AI layer: every `generate` and every query-time `embed` charges automatically when the calling flow provides a metering scope. Cache hits are never charged. Retries don&#39;t double-charge.

Flows that bill after this PR: live evaluations, flagger classification and annotation, eval alignment, GEPA optimization — plus five flows a codebase sweep found running completely unbilled: **session analysis** (one LLM pass per completed session), **signal discovery** (name/description generation), **signal detail refresh**, **taxonomy cluster naming**, and **annotation publication enrichment**.

### Revenue impact vs the old model

The old flat prices (`flagger-scan` 30, `live-eval-scan` 30, `eval-generation` 1,000) had margins that were random:

- A typical judge live-eval costs us ~$0.005 but billed 30 credits ($0.06). Overcharged 12×. It now bills 1 + ~4 credits, and only long sessions or expensive models pay more.
- A GEPA optimization run can burn ~$30–120 in tokens but billed 1,000 credits ($2). We were eating up to $118 per run. It now bills what it consumed, plus 30%.
- Flagger classifier calls, session analysis, signal discovery/refresh, taxonomy naming, and annotation enrichment were all completely free. They&#39;re now billed per call — session analysis alone runs on every completed session.

Net effect: every AI operation carries a guaranteed 30% margin (100% on semantic queries), every scan contributes a baseline credit, typical users pay less, and heavy users stop being subsidized.

### Not billed yet (documented in dev-docs/billing.md)

- **Web/API semantic search** (`listTraces`/`listSessions` `query` param and the web search boxes): the query embed costs us money but the endpoints don&#39;t provide a metering scope yet. Blocker is UX: deciding what an out-of-credits org sees (the planner already falls back to lexical search when the embedder is unavailable, so that&#39;s the likely answer).
- **Signal-generation agent**: uses the separate `AIAgent` service, which bypasses the metering layer entirely — needs metering on the agent service itself.
- **Eval/signal previews**: run `llm()` free, treated as an authoring/testing surface by design for now.

N/A

## How was this tested?

- Unit tests for the metering layer: cost-based pricing against real registry prices, flat fallbacks, no billing without a scope, cache/document embeds never charged, plus direct tests of both credit formulas and credits-override validation.
- All affected suites pass after merging `development`: `@domain/billing`, `@domain/evaluations`, `@domain/flaggers`, `@platform/ai`, workers billing suite, workflows alignment suite. Full workspace typecheck passes.

## Checklist

- [x] I have signed the CLA

https://claude.ai/code/session_01Pv3tfechUodwQspjfLv8Qk


geclos:
Triage of the Codex and CodeRabbit findings, addressed in 95c3006d:

**Fixed**
- `withAIMetering` now runs `tapError` before `tap`, so a failed cost-based record can no longer also commit the flat fallback record (Codex, CodeRabbit).
- Live evaluations re-fail when the metering record write fails instead of persisting a false errored score and losing the usage event — the worker retries and the idempotency keys dedupe the replay (Codex).
- The flagger draft use-case authorizes `llm-call` only on the annotator-fallback branch; saving classifier-provided feedback makes no AI call and now succeeds even out of credits (Codex, CodeRabbit).
- `withActivityAIMetering` fails blocked activities with `BadRequestError`, which the workflow retry policy marks non-retryable — out-of-credits orgs fail fast instead of burning Temporal retries. This also covers the empty-incremental-refresh concern: the refresh no longer retries when blocked (Codex ×2).
- `recordUsageEvent` dies on credits overrides that aren't positive integers, with boundary tests (CodeRabbit).

**Intentional tradeoffs, not changing in this PR**
- *Cap overshoot (Codex P1, CodeRabbit "reserve upper-bound cost")*: flows authorize one flat-estimate `llm-call` before AI work; per-call metering then records exact credits, which can exceed the estimate. A flow can therefore overshoot a cap by the calls in one flow — the same intentional coarseness as the trace ingest gate, documented in dev-docs/billing.md. Reserving a true per-request maximum requires knowing max output tokens per model upfront and a reconcile step; if overshoot proves material in production we'll revisit.
- *GEPA parallel sequence keys (Codex, CodeRabbit)*: correct observation — GEPA's parallel judge fan-out has scheduler-dependent sequence keys, so retry dedup relies on the 24h AI cache rather than key stability. Documented limitation; threading stable per-example identities through the optimizer is a follow-up.
- *Fallback-model pricing (Codex)*: when the Bedrock adapter falls back to a different model, metering prices the configured model. The registry rates for the configured and fallback models are close, and the 1.3× margin absorbs the difference; surfacing the executed model through `GenerateResult` is a follow-up.
- *Workflow-level eval-generation authorization (Codex)*: kept deliberately — it's the cheap boundary gate that stops an out-of-credits org before the Temporal fan-out starts. The 30-credit reservation without a matching record only affects the in-memory cap counter, which re-syncs from Postgres.
- *`semanticSimilarity()` embeds not billed as semantic-query (Codex)*: intentional. Those embeds are content-addressed documents — a fixed query string embeds once ever per project — and since 95c3006d every scan bills a baseline 1-credit `eval-scan` that covers this orchestration.

Also in 95c3006d per product direction: every live-eval scan now bills the flat 1-credit `eval-scan` (renamed from `deterministic-eval-scan`) as a baseline, with cost-based LLM/semantic charges stacking on top.

---

andresgutgon (dismissed):

PR #3917, opened 2026-07-07
Fix OAuth token validation to enforce live organization membership

geclos:
## Summary
Previously, an OAuth access token issued to a user who was subsequently removed from their organization would continue to pass validation until the Redis cache entry expired (up to 5 minutes). This PR closes that window with two complementary changes:

1. **INNER JOIN members in lookupByToken** - on initial DB lookup, any token whose user is no longer a member of the bound organization returns no row and is immediately rejected, caching a null for 5 seconds.
2. **Live membership re-check on every cache hit** - after a valid result is retrieved from cache, hasLiveMembership is called against Postgres before returning it. If the user has been removed since the cache was populated, the cache entry is invalidated, a null is written in its place, and the request is rejected. This limits the effective revocation window to a single DB query rather than the full ~300 s TTL.

### DIFF
--- a/packages/platform/oauth-token-auth/src/validate-oauth-token.test.ts
+++ b/packages/platform/oauth-token-auth/src/validate-oauth-token.test.ts
@@ -225,6 +225,25 @@ describe.skipIf(!nodeSupportsUint8Hex)("validateOAuthAccessToken (integration, N
     expect(result).toBeNull()
   })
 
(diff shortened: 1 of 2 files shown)


geclos:
@claude review

geclos:
@claude babysit this pr, address comments and review

geclos:
@claude update pr title and description to properly describe what is implemented

PR #3920, opened 2026-07-08
feat(tracing): persist customIdentifier as a queryable column on spans

andresgutgon:
## Problem

Enterprise customers want **cost/usage per tenant**. Their plan: pass the tenant id as `customIdentifier` to the run API, then query the Latitude Postgres DB grouped by tenant.

Today that doesn't work:
- `document_logs` (which had a `custom_identifier` column) is **deprecated and no longer written** — the run path and even the `createDocumentLog` endpoint now write the `spans` table. So the customer's old query returns only pre-migration data.
- The `spans` table had **no** column for the tenant id.
- The value *was* already reaching spans: `runDocumentAtCommit` sets it in baggage, the telemetry SDK normalizes the key to `latitude.custom_identifier`, and `BaggageSpanProcessor` copies it onto **every** span as an attribute — but **nothing extracted it at ingestion**, so it was never persisted.

## Fix

Persist `customIdentifier` as a first-class, indexed column on `spans`, extracted in `extractLatitudeReferences` so it lands on **every span type — including the completion spans that carry cost**. That makes per-tenant cost a simple `GROUP BY custom_identifier, SUM(cost)` with **no self-join**.

- **constants**: add `ATTRIBUTES.LATITUDE.customIdentifier = 'latitude.custom_identifier'` and `customIdentifier` on `SpanReferenceMetadata`.
- **ingestion**: extract it in `extractLatitudeReferences` (shared by all span specs).
- **Postgres**: `custom_identifier` column + `(workspace_id, custom_identifier)` index; extract in postgres `bulkCreate`; Drizzle migration `0276`.
- **ClickHouse**: parity column + extraction; clustered + unclustered migrations `0009`.
- **backfill jobs**: carry `custom_identifier` when re-inserting spans to ClickHouse.
- **tests**: ingestion persists `custom_identifier` on completion spans from the attribute, and leaves it null when absent.

No telemetry/propagation change is needed — the attribute already reaches all spans; this only adds extraction + storage.

## Customer query after this ships

```sql
SELECT custom_identifier, SUM(cost) AS total_cost, SUM(tokens_completion) AS out_tokens
FROM latitude.spans
WHERE workspace_id = :workspaceId AND type = 'completion' AND custom_identifier IS NOT NULL
GROUP BY custom_identifier;
```

## Notes / deploy considerations

- Additive, backwards-compatible (nullable column) → single PR (not a destructive migration).
- **Only populated going forward.** Spans ingested before this deploy have `custom_identifier = NULL`. The `latitude.custom_identifier` attribute isn't retained on old rows, so historical backfill isn't possible from the span data alone.
- The `CREATE INDEX` on `spans` is non-concurrent; on a large table consider building it `CONCURRENTLY` out-of-band.
- Base branch is `latitude-v1` (the enterprise customer branch), matching where this is needed.

## Testing

- `pnpm tc` — clean across all packages.
- `packages/core` tracing/spans suite — **189 passed**, including the two new ingestion assertions.
- ClickHouse migrations were **not** applied locally (local CH is diverged from the repo's migration set) — the SQL mirrors the existing `spans` ALTER conventions in both variants.

### DIFF
--- a/packages/constants/src/tracing/span.ts
+++ b/packages/constants/src/tracing/span.ts
@@ -153,6 +153,12 @@ export type SpanReferenceMetadata = {
   projectId?: number
   testDeploymentId?: number
   previousTraceId?: string
+  /**
+   * Caller-provided tenant/customer identifier, passed as `customIdentifier`
+   * to the run API. Propagated to every span in the trace and persisted as a
+   * queryable column so cost/usage can be aggregated per tenant.
+   */
+  customIdentifier?: string
 }
 
 export type ToolSpanMetadata = BaseSpanMetadata<SpanType.Tool> &
--- a/packages/constants/src/tracing/span.ts
+++ b/packages/constants/src/tracing/span.ts
@@ -291,6 +291,7 @@ export type Span<T extends SpanType = SpanType> = {
   previousTraceId?: string
 
   source?: LogSources
+  customIdentifier?: string
 
   tokensPrompt?: number
   tokensCached?: number
--- a/packages/constants/src/tracing/attributes.ts
+++ b/packages/constants/src/tracing/attributes.ts
@@ -51,6 +51,7 @@ export const ATTRIBUTES = {
     experimentUuid: 'latitude.experiment_uuid',
     source: 'latitude.source',
     externalId: 'latitude.external_id',
+    customIdentifier: 'latitude.custom_identifier',
     testDeploymentId: 'latitude.test_deployment_id',
     previousTraceId: 'latitude.previous_trace_id',
 
--- a/packages/core/clickhouse/migrations/clustered/0009_add_custom_identifier_to_spans.down.sql
+++ b/packages/core/clickhouse/migrations/clustered/0009_add_custom_identifier_to_spans.down.sql
@@ -1,0 +1,2 @@
+ALTER TABLE spans ON CLUSTER default
+  DROP COLUMN IF EXISTS custom_identifier;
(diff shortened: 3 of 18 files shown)


andresgutgon:
@claude review

andresgutgon:
Thanks @chatgpt-codex-connector — both addressed in `13e287369`:

- **P1 (concurrent index):** the `spans_workspace_custom_identifier_idx` migration now uses `CREATE INDEX CONCURRENTLY` (with the `COMMIT;`-then-concurrent pattern used by the repo's other spans-index migrations, e.g. `0262`), so it no longer holds a write-blocking lock during deploy.
- **P2 (identifier length):** the column is now unbounded `text` instead of `varchar(256)`, matching the run API's unconstrained `customIdentifier` and the legacy `document_logs.custom_identifier` type — long identifiers won't fail the span insert.

Also exposed `customIdentifier` on the `Span` type. Migration re-verified against Postgres (column `text`, index present) and the ingestion test passes.

csansoon (approved):

packages/constants/src/tracing/span.ts
```diff
@@ -153,6 +153,12 @@ export type SpanReferenceMetadata = {
   projectId?: number
   testDeploymentId?: number
   previousTraceId?: string
+  /**
+   * Caller-provided tenant/customer identifier, passed as `customIdentifier`
+   * to the run API. Propagated to every span in the trace and persisted as a
+   * queryable column so cost/usage can be aggregated per tenant.
+   */
```
csansoon, line 160: yapping

PR #3926, opened 2026-07-08
perf(flaggers): normalize instruction-extractor cache key and raise verbatim threshold

geclos:
## What does this PR do?

Cuts the cost of the flagger instruction extractor (the LLM that summarizes long inspected-agent system prompts before classification).

The extractor result was cached in Redis keyed by an exact `sha256` of the **raw** system prompt. Customer system prompts embed volatile per-call content (timestamps, UUIDs, user ids, emails), so every trace produced a new hash, every lookup missed, and the extractor re-read the full prompt on every trace — despite a 30-day TTL meant to amortize it.

Changes:

- **Normalize the prompt before hashing** — `normalizeSystemPromptForCacheKey` replaces UUIDs, ISO datetimes, times, emails, long hex runs, and digit runs with stable placeholders and collapses whitespace, so per-call drift maps to the same cache key. The extractor LLM still receives the raw, un-normalized prompt; normalization only affects key derivation.
- **Bump the cache key version to v2** since the key semantics changed.
- **Raise the verbatim threshold from 1,200 to 6,000 chars** (`FLAGGER_INSPECTED_AGENT_VERBATIM_MAX_CHARS`, moved to `constants.ts`) — mid-size prompts now go verbatim to the classifier, skipping the extractor roundtrip entirely; the extractor only runs for genuinely large prompts.

A follow-up PR stacks similarity-based (SimHash) cache reuse on top, for prompts whose prose itself drifts.

Closes #

## How was this tested?

- New unit tests for `normalizeSystemPromptForCacheKey` (volatile-content equivalence, materially different prompts staying distinct, whitespace collapsing, hex vs digit placeholders).
- New integration test proving two traces whose system prompts differ only by timestamp/UUID share a single extractor call (cache hit on the second).
- Existing cache tests updated to the v2 key format; long-prompt fixtures grown past the new 6,000-char threshold so they still exercise the extractor path.
- `pnpm --filter @domain/flaggers test` → 241 passed, 2 skipped (pre-existing gated regression suites); `pnpm --filter @domain/flaggers typecheck` clean.

## Checklist

- [x] I have signed the CLA

https://claude.ai/code/session_019Mgw9ns6S1mWgyde7s5PpP

---
_Generated by [Claude Code](https://claude.ai/code/session_019Mgw9ns6S1mWgyde7s5PpP)_

### DIFF
--- a/packages/domain/flaggers/src/constants.ts
+++ b/packages/domain/flaggers/src/constants.ts
@@ -34,6 +34,8 @@ export const FLAGGER_DEFAULT_ANNOTATOR_MODEL = {
   maxTokens: 2048,
 } as const
 
+export const FLAGGER_INSPECTED_AGENT_VERBATIM_MAX_CHARS = 6_000
+
 export const FLAGGER_DRAFT_DEFAULTS = {
   passed: false,
   value: 0,
--- a/packages/domain/flaggers/src/use-cases/run-flagger.test.ts
+++ b/packages/domain/flaggers/src/use-cases/run-flagger.test.ts
@@ -21,7 +21,11 @@ import { createFakeTraceRepository } from "@domain/spans/testing"
 import { Cause, Effect, Layer } from "effect"
 import { describe, expect, it } from "vitest"
 import { z } from "zod"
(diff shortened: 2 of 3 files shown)


geclos:
@cursoragent review

PR #3928, opened 2026-07-08
perf(flaggers): reuse instruction extractions across similar system prompts

geclos:
## What does this PR do?

Stacked on #3926. That PR's volatile-token normalization collapses templated drift (dates, UUIDs, ids), but when users free-form edit the prompt prose the exact-hash cache still misses and the instruction extractor re-runs — even though the extracted summary (a coarse "what the agent is and does" description) would be identical. This PR adds similarity-based reuse:

- **SimHash util in `@repo/utils`** — `simhash64` (64-bit SimHash over lowercased word 3-shingles hashed with FNV-1a, pure BigInt) and `hammingDistance64`.
- **Per-project similarity index in Redis** — `org:{orgId}:flaggers:inspected-agent-context:v2:index:{projectId}`, a zod-validated bounded list (16 entries, 30-day TTL) of `{sketch, contentKey}` pairs for cached `understood=true` extractions. Read-modify-write races are lossy by design (worst case one extra extraction).
- **New lookup flow** — exact content-key hit → similarity hit (Hamming ≤ 6 against the index; reuses the matched extraction and backfills the exact key so the next identical prompt fast-paths) → run the extractor. `understood=false` results are never indexed nor reused via similarity (prompt-specific, short TTL).
- **Telemetry** — `flagger.inspectedAgentContextSource` span annotation (`verbatim` / `exact-hit` / `similar-hit` / `extracted` / `fallback` / `missing`) so the hit mix and extractor call rate are graphable.

The Hamming threshold of 6 was validated empirically: a ~17.8k-char prompt with a few words changed scored distance 4; an unrelated prompt of similar shape scored 32.

Closes #

## How was this tested?

- New unit tests for `simhash64`/`hammingDistance64` (identical → 0, near-duplicates → small distance, unrelated → large distance, empty/one-word inputs don't throw).
- New flagger tests: rephrased-but-similar long prompts share one extractor call with exact-key backfill; dissimilar prompts trigger a second extraction; `understood=false` cached results are not reused via similarity; malformed index JSON behaves as a miss without throwing.
- `pnpm --filter @domain/flaggers test` → 245 passed, 2 skipped (pre-existing gated regression suites); `@repo/utils` simhash tests all pass (4 pre-existing `base64.test.ts` failures on Node 22 are unrelated); typecheck clean on both packages.

## Checklist

- [x] I have signed the CLA

https://claude.ai/code/session_019Mgw9ns6S1mWgyde7s5PpP

---
_Generated by [Claude Code](https://claude.ai/code/session_019Mgw9ns6S1mWgyde7s5PpP)_


geclos:
@cursoragent review

packages/utils/src/simhash.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export function hammingDistance64(a: bigint, b: bigint): number {
+  let diff = (a ^ b) & MASK_64
+  let distance = 0
+  while (diff !== 0n) {
+    diff &= diff - 1n
+    distance++
+  }
+  return distance
+}
```
geclos, line 54: is there an established npm dependency or package we can use instead of implementing our own diff algo?

geclos: I checked the npm ecosystem before hand-rolling this, and there's no well-maintained option:

- `simhash-js` (MIT) — closest match, but last published 4+ years ago, 32-bit hash only (we want 64-bit for a usable Hamming range), no TypeScript types, ~9k downloads/mo
- `simhash` — v0.1.0, effectively abandoned, no license declared in the manifest, ~200 downloads/mo
- `minhash` (MIT) — different algorithm (set similarity, no compact sketch to store per entry); would need more storage per index entry and its own maintenance story

Given the whole implementation is ~50 lines of pure arithmetic (FNV-1a + bit voting) with no I/O and full unit coverage, a tiny unmaintained dependency seemed like a worse trade than owning it — supply-chain surface and the repo's license-audit requirement included. Happy to swap if you'd rather take `simhash-js` and accept the 32-bit sketch.

---

packages/utils/src/simhash.ts
```diff
@@ -0,0 +1,54 @@
+const FNV_OFFSET_BASIS_64 = 0xcbf29ce484222325n
+const FNV_PRIME_64 = 0x100000001b3n
+const MASK_64 = 0xffffffffffffffffn
+const SIMHASH_BITS = 64
+
+function fnv1a64(text: string): bigint {
+  let digest = FNV_OFFSET_BASIS_64
+  for (let i = 0; i < text.length; i++) {
+    digest ^= BigInt(text.charCodeAt(i))
+    digest = (digest * FNV_PRIME_64) & MASK_64
```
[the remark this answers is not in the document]
geclos: The suggested fix doesn't apply: `MASK_64` is already a `bigint` literal (`0xffffffffffffffffn`), so `BigInt(MASK_64)` is an identity call, and the expressions on lines 10 and 47 are bigint-only — TypeScript would reject mixing `number`/`undefined` into them, so no implicit undefined→number conversion is possible there (this applies to the line-47 finding in this file too).

The only expression in the file where an operand is *typed* as possibly `undefined` is the indexed `bitVotes[bit] += …` accumulation (in-bounds by construction, but `number | undefined` to the checker). I've made that explicit in f268e74 (`bitVotes[bit] = (bitVotes[bit] ?? 0) + …`), which should clear the alert on re-scan.

---

PR #3975, opened 2026-07-10
feat(taxonomy): Custom Behaviors Phase 1a — contracts & schema slice (LAT-747)

andresgutgon:
## What & why

Phase 1a of Custom Behaviors — the **contracts + schema fork point** (Linear **LAT-747**). This thin slice decouples Phase 2 (workflow) from Phase 3 (web): once it lands on `development`, both phases run in parallel against these types. **Interfaces, entities, and migrations only — no use-case bodies, no workflow, no UI** (those are LAT-748/749/750).

Resolved decisions from **LAT-746** applied here: **Q2 → fixed 7d** (no `lookbackDays` column, a named constant instead) and **Q3 → reuse the shared FilterSet, reject `topics`, keep `moments`** (no separate registry).

## Changes

**Postgres**
- New `custom_behaviors` table (`schema/custom-behaviors.ts`): id, org, project, name, slug (unique per project), `filter_set` jsonb, status, timestamps, RLS — cloned from the `taxonomy-clusters` / `taxonomy-runs` conventions. **No `lookbackDays` column.**
- Additive nullable `custom_behavior_id` on `taxonomy_clusters` and `taxonomy_runs` (`NULL` = global). Additive `ALTER … ADD COLUMN` only.
- Migration `20260710131953_custom-behaviors` generated with `pg:generate`.

**ClickHouse**
- New `custom_behavior_assignments` table created via `ch:create` (both unclustered + clustered `00046_*`), mirroring the `taxonomy_observations` layout with `assigned_cluster_id`, keyed by `custom_behavior_id` so the global `taxonomy_observations.assigned_cluster_id` is **never** mutated. Phase 2 writes it; Phase 3 reads it.

### Why a separate table (and what it deliberately does *not* duplicate)

A custom behavior produces a **different** clustering of the same observations, and one observation can match several behaviors — so scoped assignments are a fan-out `(observation, behavior) → scoped cluster`. That can't be a column on the one-row-per-observation `taxonomy_observations` (its `assigned_cluster_id` is single-valued and must stay the global assignment). Storing scoped assignments as extra *rows* in `taxonomy_observations` would instead duplicate the heavy `embedding Array(Float32)` + `projection_metadata` once per behavior — the real waste. A dedicated table avoids that.

The table follows ClickHouse's write-once/read-many grain: **denormalize the cheap read-side scalars, never the expensive projection payload.**
- **Kept (denormalized on purpose):** `session_id` for join-free "sessions in this scoped cluster" reads (same reason `taxonomy_observations` denormalizes it); `start_time` on the event-time axis so `PARTITION BY toYYYYMM(start_time)` + TTL stay aligned with `taxonomy_observations` (a behavior's assignments expire in step with the observations they reference); bloom indexes on `session_id` / `observation_id` / `assigned_cluster_id` for point lookups.
- **Not copied:** `embedding`, `projection_metadata`, `analysis_hash`, `moment_id`, `projection_hash`, `projection_method` — the heavy immutable blob that already lives once in `taxonomy_observations`. Reads that need it join back on `(org, project, observation_id)`, which is that table's own sort prefix.

**`@domain/taxonomy`**
- `CustomBehavior` + `CustomBehaviorAssignment` entities.
- `customBehaviorFilterSetSchema` — the shared `filterSetSchema` plus a refinement that rejects the `topics` field (a behavior scoped on behaviors is circular). The exclusion lives in the shared domain contract so both the web save path and the Phase 2 workflow enforce it.
- `CustomBehaviorRepository` (PG CRUD) and `CustomBehaviorAssignmentRepository` (CH) ports.
- `CUSTOM_BEHAVIOR_LOOKBACK_DAYS = 7` named constant — Phase 2 consumes it as a sampling-activity **parameter**, so a future selectable window is an additive column swap.

**`@domain/shared`**
- `CustomBehaviorId` branded id + `customBehaviorIdSchema`.

## Validation

- `pnpm --filter @platform/db-postgres typecheck` ✅ and full workspace typecheck ✅ (incl. `apps/workflows` + `apps/web` — new types/ports import cleanly).
- Drizzle migration generates cleanly (additive: 1 new table + RLS + 2 ADD COLUMN).
- CH migration created via `ch:create` (not hand-written); not applied here (CI/Phase 2 applies).
- Unit test: a `FilterSet` with `topics` fails validation; one with `moments` passes (`custom-behavior.test.ts`).

## Review focus

- **`custom_behavior_assignments` shape** — see "Why a separate table" above; the denormalization of `session_id` / `start_time` is a deliberate ClickHouse read-locality call. Worth a sanity check against Phase 2/3 access patterns.
- **`status` enum** — `pending | generating | ready | failed` (default `pending`). The issue lists a `status` column without enumerating values; this is a forward-compatible generation-lifecycle guess Phase 2 can extend additively.
- No `deletedAt` on `custom_behaviors` (matches the explicit column list + the taxonomy tables it clones); the CRUD port exposes a hard `delete`.

### DIFF
--- a/packages/domain/shared/src/id.ts
+++ b/packages/domain/shared/src/id.ts
@@ -62,6 +62,7 @@ export type WrappedReportId = Id<"WrappedReportId">
 export type TaxonomyClusterId = Id<"TaxonomyClusterId">
 export type TaxonomyRunId = Id<"TaxonomyRunId">
 export type TaxonomyLineageId = Id<"TaxonomyLineageId">
+export type CustomBehaviorId = Id<"CustomBehaviorId">
 
 // Integration-related IDs
 export type SlackIntegrationId = Id<"SlackIntegrationId">
@@ -105,6 +106,7 @@ export const WrappedReportId = (value: string): WrappedReportId => value as Wrap
 export const TaxonomyClusterId = (value: string): TaxonomyClusterId => value as TaxonomyClusterId
 export const TaxonomyRunId = (value: string): TaxonomyRunId => value as TaxonomyRunId
 export const TaxonomyLineageId = (value: string): TaxonomyLineageId => value as TaxonomyLineageId
+export const CustomBehaviorId = (value: string): CustomBehaviorId => value as CustomBehaviorId
 export const SlackIntegrationId = (value: string): SlackIntegrationId => value as SlackIntegrationId
 export const SlackDeliveryId = (value: string): SlackDeliveryId => value as SlackDeliveryId
 export const DestinationId = (value: string): DestinationId => value as DestinationId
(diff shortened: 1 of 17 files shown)


andresgutgon:
@claude review please

PR #3978, opened 2026-07-10
feat(wrapped): add skills breakdown to Claude Code Wrapped (report V3)

csansoon:
Adds a **Skills** breakdown to Claude Code Wrapped: which skills a project's agents leaned on during the week, alongside the existing tokens/files/tools stats.

## What counts as a skill use

A skill use is a span that either:
- **reads a `SKILL.md`** file (`Read` / `NotebookRead` whose `file_path` ends in `SKILL.md`), or
- **invokes the `Skill` tool** (`tool_name = 'Skill'`).

Skills are identified by **name only**, so the same skill merges across every path it was loaded from. The name comes from the `SKILL.md` parent directory for reads, and from the `Skill` tool's `skill` argument otherwise (plugin-namespaced slugs like `plugin:name` are kept verbatim). Usage is counted per span — one session using a skill three times counts three.

## What's shown

- **Public viewers:** distinct skills used + total skill uses. No skill names.
- **Members:** the two public counts, plus the top-3 skills (name + count) and a per-workspace skill breakdown in each workspace deep-dive.

Per-workspace attribution keys off the span-level `metadata['workspace.name']` rather than a file path, so `Skill`-tool calls — which carry no `file_path` — still land in the right workspace.

## Report V3

This is a data-shape change, so it follows the existing freeze-and-version convention: V1 and V2 stay frozen, and V3 is added alongside on both the web and email `(type, reportVersion)` dispatch tables. `reportV3Schema` extends V2 with a `skills` object (`distinctUsed`, `totalUses`, `top[]`) and per-workspace `skills`. Old V1/V2 blobs keep rendering through their original renderers untouched.

`redactForPublic` strips `skills.top` for non-member viewers (the distinct/total counts stay public). The Wrapped email mirrors the public numbers only — distinct + total, no names.

## Data path

- `@platform/db-clickhouse`: a `getSkillUsage` query (distinct + total + top-3 in one round-trip) and a `skills` CTE folded into the workspace deep-dive. Both share two SQL constants — `SKILL_MATCH_PREDICATE` and `SKILL_NAME_EXPR` — for the match and name-derivation logic.
- `@domain/spans`: the V3 schema, the `getSkillUsage` port method, and the build-report assembly wiring.
- `@app/web` + `@domain/email`: the V3 renderer/template and dispatch entries.

## Testing

- A chdb-backed `getSkillUsage` test covers name derivation from both sources, name-merge across paths, distinct/total counts, and top-3 ranking.
- An `assembleReport` unit test covers the `skills` and per-workspace `skills` fields flowing through to V3.
- Existing `build-report`, `run-wrapped`, and `wrapped-fan-out` suites pass; typecheck is clean across all touched packages.

One coverage note: the per-workspace deep-dive isn't exercised in the chdb test because its pre-existing command-classifier CTE uses `lowerUTF8`, which the local chdb build lacks. That path runs against real ClickHouse in production and is covered at the assembly layer.


csansoon:
@claude review this pr

csansoon:
Thanks for the review — addressed the two actionable cleanups, notes on the rest below.

**Fixed**
- **`SkillsSection` duplicated `formatCompact`** — now imports it from `WrappedReportV1` instead of redefining. `1ae70ae2a`
- **Email dispatcher version chain** — dropped the `resolvedVersion === 3 / === 2` JSX arms; `renderWrappedHtml` now renders the registry-resolved `Template` directly (all versions share the same runtime props), so new versions need only a `TEMPLATE_BY_TYPE_VERSION` entry. Also refreshed the stale dispatch doc comment. `19c52196a`

**Kept as-is**
- **`SkillUsageCHRow.top_counts` `| string` branch** — intentional. The `num()` cast and `number | string` row types are the established defensive pattern across this reader (ClickHouse JSON can surface UInt64 as either), so I kept it consistent with the sibling rows rather than special-casing this one.

**Deferred (pre-existing, out of scope here)**
- **`WorkspaceDeepDiveSection` in `v1/` carrying V3 logic** — agree a neutral `claude-code/`-level location is cleaner, but moving the shared primitive touches V1/V2/V3 imports; better as a focused follow-up than mixed into this PR.
- **No test for the `totalUses <= 0` guard** — there's no component-test harness for these wrapped renderers today; the guard is a trivial early-return. Left as a note rather than standing up render-test infra for one branch.
- **Per-workspace chdb coverage gap** — as noted in the PR description, the deep-dive path trips `lowerUTF8` (absent in the local chdb build) via the pre-existing command-classifier CTE; it runs in production and is covered at the assembly layer.

PR #3980, opened 2026-07-10
fix(clickhouse): reduce trace detail memory usage

geclos:
## What changed

- add a summary-only trace lookup for trace-end orchestration and signal matching
- resolve trace message fields from source `spans` rows instead of finalizing wide `argMinIf`/`argMaxIf` aggregate states from `traces`
- apply the spans-backed message path to point lookups, batched detail reads, metadata reads, and conversation chunks
- preserve the materialized view's operation filters when reconstructing input, output, and system messages

## Why

Production ClickHouse periodically reaches its replica memory ceiling while worker fan-out runs concurrent trace-detail lookups. A full lookup can consume roughly 690 MiB because ClickHouse must deserialize message aggregate states from every matching active part. Bursts reached 23 detail queries per second and caused OvercommitTracker to cancel reads at the 18 GiB server limit.

Reading message candidates from `spans` bounds aggregation by the spans in the requested trace. Callers that only need orchestration fields no longer read conversation content.

Datadog issue: https://app.datadoghq.eu/error-tracking/issue/c70cdfb2-74fa-11f1-b700-da7ad0900005

## Verification

- `pnpm --filter @domain/spans typecheck`
- Biome checks for all changed files
- `git diff --check`
- production query-log analysis confirmed the failing query shape, concurrency bursts, and per-query memory profile

The ClickHouse repository test suite is currently blocked during module loading by an unrelated missing `@domain/monitors` workspace resolution. The commit hook's formatting pass succeeded; its workspace `knip` step was blocked by the existing `@tanstack/react-start/plugin/vite` package-export mismatch.

### DIFF
--- a/packages/domain/spans/src/ports/trace-repository.ts
+++ b/packages/domain/spans/src/ports/trace-repository.ts
@@ -85,6 +85,12 @@ export interface TraceRepositoryShape {
     readonly traceId: TraceId
   }): Effect.Effect<TraceDetail, NotFoundError | RepositoryError, ChSqlClient>
 
+  findSummaryByTraceId(input: {
+    readonly organizationId: OrganizationId
+    readonly projectId: ProjectId
+    readonly traceId: TraceId
+  }): Effect.Effect<Trace, NotFoundError | RepositoryError, ChSqlClient>
+
   findMetadataByTraceId(input: {
     readonly organizationId: OrganizationId
     readonly projectId: ProjectId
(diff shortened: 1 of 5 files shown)


geclos:
@claude review

geclos:
@cursoragent review

geclos:
@cursor Thanks. Addressed the actionable minor notes in `b5b86a559`: metadata lookups now use a narrower spans aggregation, the detail/chunk queries share operation-filter fragments, and `findConversationChunk` has explicit message-assembly coverage.

geclos:
@claude Thanks. Addressed the duplicated query predicates and added chunk regression coverage in `b5b86a559`. I kept the parallel summary/message reads for full-detail lookups: missing traces are uncommon, while serializing the queries would add summary-query latency to every successful detail read. I also left error wrapping at the public repository-operation boundary to avoid nested `RepositoryError`s; the failing SQL remains available as the cause.

PR #4007, opened 2026-07-13
fix(clickhouse): strip timezone offsets from DateTime64 filter params

geclos:
## What does this PR do?

Fixes a production regression in ClickHouse trace/session filtering: date-range filters on `startTime` (and any other field bound as `DateTime64(9, 'UTC')`) failed with a `RepositoryError` when the filter value carried a timezone offset other than a bare `Z`.

`mapDateTime64UtcQueryParam` (`packages/platform/db-clickhouse/src/registries/helpers.ts`) only stripped a trailing `Z`. A date string with a `+00:00`/`-05:00` offset — e.g. as produced by Postgres timestamps — kept its offset suffix, and ClickHouse's `DateTime64(9, 'UTC')` binder rejected the parameter as only partially parsed:

```
Value 2026-06-09 22:41:43.054269+00:00 cannot be parsed as DateTime64(9, 'UTC') for query parameter 'f_0'
because it isn't parsed completely: only 26 of 32 bytes was parsed: 2026-06-09 22:41:43.054269.
```

This is the same class of bug as #3924 (which validated integer-typed filter params against fractional values) but for the `DateTime64` case — the earlier fix didn't cover it, and the issue regressed on 2026-07-09 with this variant. The fix generalizes the suffix-strip to any `Z` or numeric UTC offset (colon or no colon), fixing the shared helper at its source so every call site benefits: `startTime` on both the trace and session ClickHouse field registries, plus `session-intelligence-filters.ts`.

### Root cause

`mapDateTime64UtcQueryParam` used `value.endsWith("Z") ? value.slice(0, -1) : value`, which only handles the `toISOString()`-style suffix. Any other timezone suffix format passed through unmodified into the ClickHouse query parameter, which the binary protocol parser rejects since `DateTime64(9, 'UTC')` expects a bare `YYYY-MM-DD HH:MM:SS.ffffffffff` value (the `'UTC'` in the column type already fixes the timezone).

### Fix

Replace the `Z`-only check with a regex that strips any trailing `Z` or `[+-]HH:MM`/`[+-]HHMM` offset before converting the `T` separator to a space.

Datadog Error Tracking issue [ET-291](https://app.datadoghq.eu/error-tracking/issue/04a76546-7743-11f1-97b7-da7ad0900005) (`04a76546-7743-11f1-97b7-da7ad0900005`), `RepositoryError: ... cannot be parsed as DateTime64(9, 'UTC')`, service `api`, resource `POST /v1/projects/:projectSlug/traces/list`. Comment posted on the issue with this PR link.

## How was this tested?

Added test cases to `packages/platform/db-clickhouse/src/registries/helpers.test.ts` covering: a trailing `Z`, a `+00:00` offset (the reported regression), a non-UTC colon offset (`-05:00`), a no-colon offset (`+0000`), a bare value with no suffix, and non-string passthrough. Confirmed the new offset-suffix tests fail without the fix (reproducing the exact reported error) and pass with it.

- `pnpm --filter @platform/db-clickhouse test -- src/registries/helpers.test.ts` — 10/10 passed
- `pnpm --filter @platform/db-clickhouse test -- src/registries src/filter-builder.test.ts` — 49/49 passed (no regressions in sibling filter-building tests)
- `pnpm --filter @platform/db-clickhouse typecheck` (tsgo) — clean
- `pnpm exec biome check` on changed files — clean

## Checklist

- [ ] I have signed the CLA

---
_Generated by [Claude Code](https://claude.ai/code/session_01DiVSVda1Q2abAzu3La82zY)_

### DIFF
--- a/packages/platform/db-clickhouse/src/filter-builder.ts
+++ b/packages/platform/db-clickhouse/src/filter-builder.ts
@@ -12,6 +12,7 @@ export interface ScalarFieldMapping {
   readonly isArray?: boolean
   readonly arrayContains?: boolean
   readonly mapValue?: (value: FilterCondition["value"]) => FilterCondition["value"]
+  readonly valueExpression?: (paramName: string, options: { readonly array: boolean }) => string
 }
 
 export interface SyntheticFieldMapping {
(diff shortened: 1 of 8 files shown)


geclos:
@cursoragent review

PR #4009, opened 2026-07-13
Route signal notifications to assignees

geclos:
Signal discovery and escalation notifications now target the signal assignee when one is set. Unassigned signals keep the existing project-member fan-out.

Assignee-targeted notifications also skip shared Slack routes so they remain private to the assigned user.

Tests cover assigned and unassigned recipient selection for discovery and escalation notifications.

## Verification

- `pnpm --filter @domain/notifications test -- request-incident-notifications.test.ts request-signal-discovered-notifications.test.ts`
- `pnpm --filter @domain/notifications typecheck`
- commit hooks: workspace format and knip


geclos:
@claude review

PR #4017, opened 2026-07-14
feat(experiments): project-scoped variant comparison

neoxelox:
## What this adds

**Experiments** — a project-scoped container that compares two or more **Variants** against one **baseline**, across every analytic Latitude computes for sessions, users, tools, signals, and behaviours. A Variant is a saved population selector: a `FilterSet` + a search `query` + a `timeRange`. For each non-baseline Variant, every metric is shown as a value plus a signed percentage change versus the baseline, colored by whether the change is good or bad.

The entity is close to Monitors — project-scoped, org-RLS'd, slug-referable, soft-deleted, embedded JSON config, list + detail pages, full HTTP/MCP/SDK/CLI surface — with two differences: it owns an ordered array of embedded Variants (no child table) instead of one target + rule, and it computes comparison analytics **on read** rather than evaluating a rule on a schedule. No incidents, jobs, or domain events.

## Model

- **Variant** (embedded in the `experiments.variants` jsonb array, max 10): a stable cuid `id`, a name, a `baseline: boolean` flag, a `filterSet` (sessions mode), a nullable `query`, and a `timeRange` (relative / absolute / null → last 30 days).
- **Baseline** lives on the Variant (the flag), not on the experiment. Exactly one Variant is flagged when the array is non-empty; helpers keep that invariant across add / remove / set-baseline. Variant-name uniqueness is a write-time rule, so legacy rows with duplicate names still parse on read.
- **Population grain is sessions**: a Variant's population is the set of `session_id`s matching its filter + range + query, and child entities (tools/signals/behaviours) are scoped to that set.

## The new backend piece: `VariantMetricsReader`

No existing engine takes `(filters + range + query) → population aggregates` for tools/signals/behaviours or distinct users, so this PR adds `VariantMetricsReader` (port in `@domain/experiments`, ClickHouse adapter in `@platform/db-clickhouse`). It builds the population once from the sessions analytics stream, **materializes the matching `session_id`s into an array**, and passes them as a param to each metric query (`session_id IN {sessionIds:Array(String)}`). Materializing rather than nesting the aggregate-heavy population subquery is deliberate: the nested form is rejected by ClickHouse with `ILLEGAL_AGGREGATION`. Semantic queries are treated as best-effort and the affected Variant is flagged `approximate`.

`getExperimentComparisonUseCase` runs the reader for every Variant, attaches per-metric deltas vs the baseline, and flags Variants whose session/user population deviates from the baseline by more than 25%.

## Web

The detail view is a **table** (`comparison-table.tsx`): one supercolumn per Variant (baseline first), columns dividing the available width equally down to a floor. The baseline column is sticky-left with a soft stuck shadow while scrolled, and its definition rows are wrapped by the brand-gradient `AnimatedBorder` shader (extended here with per-corner radii + an intensity prop). Shared rows hold the variant header, the population editors (a full-width `DateRangePicker`, the segmented `SearchInput`, and a reusable `FilterBuilder`), the headline panels, then a collapsible section per metric entity with one aligned row per metric. A list page, create/rename/delete/import modals, a nav entry, and a command-palette search round it out.

Variant edits are a client-side actions hook that derives the next `variants` array and calls the shared `updateExperimentUseCase`, so the web and API enforce invariants through one seam.

## API / SDK / MCP / CLI

Five operations under `/projects/:projectSlug/experiments` — `list` / `create` / `get` / `update` (PUT) / `delete` — all routed through the domain use-cases. `get` returns the full comparison bundle and resolves top-list ids to display names; it fans out ClickHouse per Variant, so it carries the `high` rate-limit tier. Regenerated `openapi.json`, `mcp.json`, the TS + Python SDKs, and the CLI; bumped both SDKs to `9.1.0` and added a CLI `7.1.0` CHANGELOG entry.

## Testing

- `@domain/experiments`: 66 unit tests (entities, helpers, use-cases); the reader has chdb-testkit tests in `@platform/db-clickhouse`.
- HTTP integration (`apps/api/src/routes/experiments.test.ts`): full lifecycle plus duplicate-name rejection, exercising the ClickHouse metrics path end to end.
- An MCP `tools/list` case for the five new tools.

## Notes

- Docs: `dev-docs/experiments.md` (internal) and a public `docs/experiments/overview.mdx` in the Understand group. The planning spec (`specs/experiments.md`) is removed now that the feature is complete.
- The Postgres migration (`20260713120949_experiments`) is checked in but has not been run against a database in this branch.


packages/platform/db-clickhouse/src/repositories/variant-metrics-repository.ts
```diff
@@ -0,0 +1,357 @@
...
+ * Per-query safety caps for the population-scoped metric queries. A query that blows past these
+ * fails its own request (a retryable RepositoryError) instead of tripping the server-wide
+ * OvercommitTracker and killing unrelated tenants' queries. The execution cap sits under the 30s
+ * HTTP `request_timeout` so ClickHouse aborts the query and frees the shared socket before the
+ * client abandons it. Search-plan settings (from a semantic query) still layer on top, but these
+ * caps always win. Mirrors `BOUNDED_READ_SETTINGS` in the span repository.
+ */
+const BOUNDED_METRIC_SETTINGS = {
+  max_execution_time: 25,
+  max_memory_usage: "4000000000",
```
geclos: hard to parse what's this 4GBs?

neoxelox: yes

neoxelox: The max_memory_usage setting [defines the maximum amount of RAM (in bytes) a single query can allocate](https://kb.altinity.com/altinity-kb-setup-and-maintenance/altinity-kb-memory-configuration-settings/). If a query exceeds this limit, ClickHouse aborts it with an exception.

neoxelox: similar to BOUNDED_READ_SETTINGS on the span reader

geclos: 4GBs is absolutely bonkers 4 concurrent queries like these will OOM clickhouse

packages/platform/db-clickhouse/src/repositories/variant-metrics-repository.ts
```diff
@@ -0,0 +1,357 @@
...
+/**
+ * Per-query safety caps for the population-scoped metric queries. A query that blows past these
+ * fails its own request (a retryable RepositoryError) instead of tripping the server-wide
+ * OvercommitTracker and killing unrelated tenants' queries. The execution cap sits under the 30s
+ * HTTP `request_timeout` so ClickHouse aborts the query and frees the shared socket before the
+ * client abandons it. Search-plan settings (from a semantic query) still layer on top, but these
+ * caps always win. Mirrors `BOUNDED_READ_SETTINGS` in the span repository.
+ */
+const BOUNDED_METRIC_SETTINGS = {
+  max_execution_time: 25,
```
geclos: also 25 seconds is bonkers as well

packages/platform/db-clickhouse/src/repositories/variant-metrics-repository.ts
```diff
@@ -0,0 +1,357 @@
...
+      return yield* computeVariantMetrics(chSqlClient, input)
+    }),
+  computeSummaryMetrics: (input) =>
+    Effect.gen(function* () {
+      const chSqlClient = (yield* ChSqlClient) as ChSqlClientShape<ClickHouseClient>
+      return yield* computeSummaryMetrics(chSqlClient, input)
+    }),
+})
+
+export const VariantMetricsReaderLive = Layer.succeed(VariantMetricsReader, make())
```
geclos, line 357: Is this computing all variants in a single go? Would it mayhaps be better to compute each variant in parallel instead? Unsure about this

neoxelox: its done concurrently yes
```
const computed = yield* Effect.all(
      resolved.map(({ variant, range }) =>
        reader
          .computeVariantMetrics({
            organizationId: experiment.organizationId,
            projectId: experiment.projectId,
            filterSet: variant.filterSet,
            query: variant.query,
            range,
          })
          .pipe(Effect.map((metrics) => ({ variant, range, metrics }))),
      ),
      // Bound the per-comparison query fan-out onto the shared ClickHouse pool: each variant fires up
      // to METRIC_QUERY_CONCURRENCY queries, so this cap keeps total in-flight pool-friendly.
      { concurrency: VARIANT_METRIC_CONCURRENCY },
    )
```

andresgutgon (changes requested):
Monster PR. Split

PR #4021, opened 2026-07-14
feat(traces): validate trace filter fields at the API boundary and add endTime filter

geclos:
## What does this PR do?

`listTraces` (and every operation that takes a trace filter set) accepted an **open** `filters` record: the wire schema validated the condition *shape* but never the field *names*. The ClickHouse filter builder silently drops keys it doesn't recognize, so a caller who filtered by an unsupported field — e.g. `endTime`, which was never a registered filter field — got an **unfiltered, all-time result set with no error**. That's the trap a user hit (thousands of all-time traces when they expected a windowed count).

This PR does two things:

1. **Fail loud instead of silently ignoring.** A new `TraceFilterSetSchema` validates every filter key against the canonical allow-list and returns a `400` naming the unknown field, so `{ "endTimee": [...] }` is rejected instead of quietly returning everything. The allow-list also flows into `openapi.json` and `mcp.json` as a self-documenting field reference, so SDK / MCP / CLI callers can see exactly what is filterable.
2. **Add the `endTime` filter field.** A trace's `endTime` (max span end) is registered in the trace and session ClickHouse field registries, so it now compiles to a `HAVING` clause (like `duration`/`cost`) instead of being dropped. `startTime` already worked; `endTime` is its symmetric sibling.

### Design notes / scope

- The canonical set of valid trace filter field names lives in `@domain/shared` (telemetry fields + `startTime`/`endTime` + `score.*` + `metadata.*`) as the single source the boundary validates against. Drift-guard tests keep the ClickHouse registries and the domain list in sync.
- Validation is applied at the **operations boundary** (not inside `buildClickHouseWhere`, whose silent-skip is load-bearing for internal callers that peel off score/session-intelligence filters before the registry sees them). It covers `listTraces` and every other operation that takes a `TraceRef` / `TracesRef` filter set (annotations, scores, dataset import, export).
- `endTime` is exposed as an **API/MCP/SDK/CLI** filter field, matching how `startTime` is treated. It is intentionally **not** added to the generic web filter dropdown: that dropdown only renders status/text/multiSelect/numberRange fields and `startTime` isn't in it either — time windowing in the web UI is handled by the dedicated time picker. A web end-time control would be a separate product change.
- `gtePercentile` / `ltePercentile` are rejected on `startTime` / `endTime` (percentile resolution only rewrites duration/ttft/cost; leaving those ops through would 500 in `buildClause`).

### Generated surface

`openapi.json` / `mcp.json` were regenerated via the emit scripts, and the TS + Python SDKs and the `latitude` CLI were regenerated with Fern (`pnpm generate:all`). Package versions were bumped to ship the new surface: TypeScript + Python SDKs **9.2.0**, CLI changelog **7.2.0**, with changelog entries for TraceFilterSet / `endTime` filter validation.

### Follow-ups (not in this PR)

- Same allow-list validation could be extended to the other filter surfaces (`querySpans`, saved searches, signals) — each has its own registry, so it needs a per-surface allow-list.
- Unrelated sibling bug spotted while here: the API's `TRACE_SORT_FIELDS` (`endTime`, `durationNs`, `tokensTotal`, `costTotalMicrocents`) don't match the repo's `SORT_COLUMNS` keys (`duration`, `cost`, `spans`, …), so those `sortBy` values silently fall back to `startTime` ordering — same silent-drop class, worth a separate fix.

Closes #

## How was this tested?

- `@domain/shared`: unit tests for `isTraceFilterFieldName` / `unknownTraceFilterFields` (accepts telemetry/time/score/metadata; rejects typos, unknown `score.*`, bare `metadata.`, and session-only fields on the trace surface).
- `@platform/db-clickhouse`: drift-guard tests (`TRACE_FIELD_REGISTRY` keys == domain telemetry list; `SCORE_FILTER_KEYS` == domain score list; `endTime` present on both registries) + a filter-builder test proving `endTime` compiles to a parameterized `end_time` comparison.
- `@repo/operations`: `TraceFilterSetSchema` accepts the documented fields, rejects unknown ones with a path-tagged `400` message, and rejects `gtePercentile` on time fields.
- `apps/api`: route tests that `POST /list` returns `400` for an unknown filter field and `200` for `startTime`/`endTime` windows.
- `pnpm typecheck`, `biome check`, and `knip` pass for touched packages; `openapi.json` / `mcp.json` regenerated via the emit scripts; TS/Python SDKs + CLI regenerated via Fern so the `api-manifests` drift check is green.

## Checklist

- [ ] I have signed the CLA

<div><a href="https://cursor.com/agents/bc-884bcf02-2fef-4b9a-a409-5a2fd8562888"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-884bcf02-2fef-4b9a-a409-5a2fd8562888"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


geclos:
@claude review

geclos:
@claude review again

andresgutgon (changes requested):
Remember bump version and CHANGELOGs in SDKs

andresgutgon (approved):

PR #4022, opened 2026-07-14
feat(spans): detect Claude Code subagents by classifying interaction as invoke_agent

csansoon:
Part of the subagent-visualization effort. Extends automatic subagent detection to Claude Code, which previously showed up as a single flat "Main agent" with no subagents identified.

## What changes

Claude Code's `interaction` span (`span.type: interaction`, and the native `claude_code.interaction` name) now resolves to the `invoke_agent` operation instead of `prompt`. An `interaction` orchestrates one turn's generations and tool calls — and nests via `tool:Agent` into a child `interaction` — which is structurally the same thing every other framework's `invoke_agent` span represents. It also fits the operation's lossy-wrapper contract: no own tokens/cost, with the real usage on the `chat` leaves underneath.

With that one reclassification, `buildAgentGraph` handles Claude Code with no graph-logic change:

- the root `interaction` becomes the backed (non-virtual) main agent,
- plain tool calls with no generations stay plain tools,
- a `tool:Agent` that wraps a nested `interaction` is detected as a subagent, with the nested interaction collapsing in as its identity span.

## Agent naming

Claude Code Task subagents emit only `subagent.id` (`"<name>:<id>"`, e.g. `"Explore:ab63322…"`) — no `subagent.type` or agent-name attribute. Added a lowest-priority `agentName` candidate that takes the name half (split on `:`), so a subagent labels as "Explore" rather than falling back to the "Agent" tool name. Real name attributes still win when present.

## Label fix

Because the main now has a real backing span, its name is a framework token (`"interaction"`). The agent-graph label helper used that span name for the main node, which would surface "interaction" instead of "Main agent". The main now defaults to "Main agent" and is overridden by the resolved `agentName` when one exists — the same two-phase resolution already used for subagents. This also fixes the pre-existing Vercel case, where the main showed "ai.generateText".

## Notes

- Forward-only. `operation` is computed at ingest, so existing Claude Code spans keep `prompt` until reprocessed; no migration.
- No change to the usage/cost or conversation rollups — those already gate on the `chat` leaves and exclude `invoke_agent`.

## Testing

`@domain/spans` unit tests + typecheck + biome pass. Added coverage for the Claude Code trace shape, the generic main-label fallback, and `subagent.id` name extraction (plus its precedence under `subagent.type`). Affected web suites (agents breakdown, conversation timeline, duration composition) pass.

### DIFF
--- a/dev-docs/claude-code-telemetry.md
+++ b/dev-docs/claude-code-telemetry.md
@@ -36,7 +36,7 @@ Each turn produces three kinds of spans, all routed through the existing `apps/i
 
 | `span.type` | Maps to `Operation` | Carries |
 | --- | --- | --- |
-| `interaction` | `prompt` | Root of the turn. `user_prompt`, `session.id`, `interaction.duration_ms`. |
+| `interaction` | `invoke_agent` | Root of the turn — the agent boundary that orchestrates the turn's generations and tool calls. `user_prompt`, `session.id`, `interaction.duration_ms`. |
 | `llm_request` | `chat` | Child of interaction. `model`, token counts (input/output/cache_read/cache_creation), `gen_ai.input.messages`, `gen_ai.output.messages` (full conversation as JSON). |
 | `tool_execution` or `tool` | `execute_tool` | Child of llm_request, one per tool call. `tool.name`, `tool.id`, `tool.input`, `tool.output`. |
 
--- a/dev-docs/pi-telemetry.md
+++ b/dev-docs/pi-telemetry.md
@@ -79,7 +79,7 @@ The extension emits one Latitude trace per pi user prompt/agent run. All spans a
 ## Trace shape
 
 ```text
-interaction                         span.type=interaction → operation prompt
+interaction                         span.type=interaction → operation invoke_agent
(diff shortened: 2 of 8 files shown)


csansoon:
@claude review this pr

csansoon:
Thanks — addressed the stale docs in `22abed494`:

- `dev-docs/claude-code-telemetry.md` ingestion-shape table: `interaction` row now maps to `invoke_agent` (with a note that it's the turn's agent boundary).
- `dev-docs/pi-telemetry.md`: both the trace-shape diagram and the resolver-contract table updated to `invoke_agent`.

Checked the rest of `dev-docs` and the public `docs/` — no other references to the old `prompt` mapping (remaining `interaction` + `prompt` hits are about the user's prompt text, not the operation).

On the minor coverage note: leaving those as-is. The generation-level `agentName` fallback is framework-agnostic and already covered by "falls back to an owned generation's agentName…" in `build-agent-graph.test.ts`; the `subagentLabelFor` → `"Subagent"` branch is a pre-existing defensive fallback, not behavior this PR introduces.

PR #4026, opened 2026-07-14
fix(auth): require confirmation before magic-link verification

geclos:
## What changed

- Rewrite outgoing Better Auth magic links to a scanner-safe `/auth/verify` URL.
- Keep the one-time token and every generated verification parameter in the URL fragment, including nested callback search parameters.
- Add an explicit confirmation screen that reconstructs the Better Auth verification URL only after the user clicks Continue.
- Show a recovery state for missing or malformed links and update email template previews.

## Why

Email security scanners can preload links before the recipient opens the message. A direct GET to Better Auth consumes the one-time token and can create a session for the scanner. URL fragments are not sent with the initial HTTP request, so the confirmation page can render without consuming the link.

## Validation

- `pnpm --filter @app/web exec vitest run src/lib/auth/magic-link.test.ts`
- `pnpm --filter @domain/email check`
- `pnpm --filter @domain/email typecheck`
- `pnpm --filter @app/web build`
- Manual Mailpit QA for existing-user sign-in, new-user signup and onboarding, invalid links, and scanner-style GETs without confirmation

## Screenshot

Magic-link confirmation page

### DIFF
--- a/apps/web/src/routes/auth/verify.tsx
+++ b/apps/web/src/routes/auth/verify.tsx
@@ -1,0 +1,63 @@
+import { Button, Icon, Text, useMountEffect } from "@repo/ui"
+import { createFileRoute, Link } from "@tanstack/react-router"
+import { AlertCircle, ShieldCheck } from "lucide-react"
+import { useState } from "react"
+import { AuthScreen } from "../../components/auth-screen.tsx"
+import { hasMagicLinkVerificationToken, reconstructMagicLinkVerificationUrl } from "../../lib/auth/magic-link.ts"
+
+export const Route = createFileRoute("/auth/verify")({
+  component: VerifyMagicLinkPage,
+})
+
+function VerifyMagicLinkPage() {
+  const [hasVerificationToken, setHasVerificationToken] = useState<boolean | undefined>(undefined)
+
+  useMountEffect(() => {
+    setHasVerificationToken(hasMagicLinkVerificationToken(window.location.hash))
+  })
+
+  const confirm = () => {
+    const verificationUrl = reconstructMagicLinkVerificationUrl({
+      fragment: window.location.hash,
+      origin: window.location.origin,
+    })
+
+    if (verificationUrl) {
+      window.location.assign(verificationUrl)
+    } else {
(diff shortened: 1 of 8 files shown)


packages/domain/email/src/templates/signup-existing-account-magic-link/EmailTemplate.tsx
```diff
@@ -44,5 +44,5 @@ export function SignupExistingAccountMagicLinkEmail({
 
 SignupExistingAccountMagicLinkEmail.PreviewProps = {
   userName: "Alex",
-  magicLinkUrl: "https://console.latitude.so/auth/verify?token=signup-existing-preview",
+  magicLinkUrl: "https://console.latitude.so/auth/verify#token=signup-existing-preview",
```
geclos, line 47: why this change in these urls?

geclos: These are preview-only props. I changed them so React Email previews and link checks use the same fragment-based `/auth/verify#token=...` shape as production links instead of documenting the old query-string form. No runtime behavior depends on these values.

apps/web/src/routes/auth/verify.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <AuthScreen title="Confirm your sign in" description="Review your request before continuing to Latitude.">
+      <div className="flex flex-col gap-4 rounded-xl border border-border bg-muted/50 p-6">
+        {hasVerificationToken === false ? (
+          <>
+            <div role="alert" className="flex flex-col items-center gap-3 text-center">
+              <Icon icon={AlertCircle} size="lg" className="text-destructive" />
+              <Text.H5 color="foregroundMuted">
+                This confirmation link is missing or invalid. Request a new link and try again.
+              </Text.H5>
+            </div>
```
andresgutgon, line 42: We have alert design system component

geclos: wouldn't look good in this ui though, as it's already in a gray bounding box

andresgutgon (approved):

PR #4035, opened 2026-07-15
feat(spans): support GenAI memory operation spans (Phase 0, LAT-729)

csansoon:
## What does this PR do?

First step of Memory Observability (LAT-729): make OTEL GenAI **memory operation** spans first-class on the Spans tab, and start capturing the attributes those spans carry. No storage engine yet — this is the self-contained on-ramp, and it doubles as a probe for whether any current traffic already emits `gen_ai.memory.*`.

The PR also lands the full design doc it implements against: `specs/memory-observability.md` (the git-like snapshot/diff/blame model, the four planned features, and the phased plan). Only Phase 0 is code here.

https://github.com/user-attachments/assets/3839edc3-003c-4c66-a661-3db6ad15156d

### Memory spans, end to end

- **Operation enum.** `operationSchema` now names the 7 memory operations — `create_memory`, `update_memory`, `upsert_memory`, `delete_memory`, `search_memory`, `create_memory_store`, `delete_memory_store` — alongside `MEMORY_OPERATIONS` / `isMemoryOperation` exports. (The enum was already a `z.union([enum, z.string()])` pass-through, so these strings flowed through un-normalized before; now they're typed and classifiable.)
- **Structured attribute capture.** `resolveAnyValue` in the OTLP transform previously dropped `arrayValue`/`kvlistValue` attributes on the floor. It now flattens them to a JSON string so structured payloads — notably `gen_ai.memory.records` (the record content array) — survive into `attr_string`. This is a general fix: any previously-dropped structured attribute now persists.
- **Spans-tab rendering.** Memory spans get a `Brain`/`Database` icon in the tree, an amber (`bg-warning/60`) waterfall bar, a **"Memory" filter toggle** (wired through both the trace and session span views), and a detail panel showing store / record / count, the search query, and the parsed records content — with a "Content not captured" fallback when the payload is absent.
- **TS SDK.** New `memoryScope` capture option → stamped as `latitude.memory.scope` on both the processor and tracer paths. This is the scope key the later phases group memory history by (falls back to the span's `user.id`, then an empty bucket).

### Decisions worth flagging (deviations from the spec's original Phase 0 plan)

- **No ClickHouse migration.** Filtering rides the indexed `operation` column; scalar memory attrs and the flattened `records` JSON ride the existing `attr_string`/`attr_int` maps already returned by the span-detail read path. The dedicated `resolvers/memory.ts`, resolved columns, and ingest-side scope resolution move to Phase 1 with the ledger.
- **The array/kvlist flattening is intentionally general** and relies on `CODEC(ZSTD)` for the extra text. If a pathological emitter bloats `attr_string`, the follow-up is a length cap in `resolveAnyValue` (noted in the spec).
- **Web keeps its own 7-string `isMemoryOperation`** (`spans-tab/memory-operations.ts`, type-only domain import) rather than pulling the domain runtime helper into the client bundle — accepted duplication.
- **`OPERATION_ICON`'s `satisfies` is vacuous** (the `z.string()` catch-all collapses the `Exclude` to `never`), so the memory icon entries are a manual, non-compiler-enforced addition.

Deferred to Phase 1 (all recorded in the spec's implementation notes): the memory ledger + content-addressed blobs + current-state projection, the `resolvers/memory.ts` resolver, and the `"update_memory · <record.id>"` tree-row label.

Refs LAT-729 (Linear). Phase 0 of a multi-phase effort — does not close the issue.

## How was this tested?

- Typecheck clean: `@domain/spans`, `@latitude-data/telemetry`, `@app/web`.
- Unit tests pass (added transform, span-filter, and processor cases): domain 916, telemetry 140, web 611.
- `knip` and Biome clean.

Manual UI check (memory span → trace on the Spans tab): confirm the memory icon + amber bar, that the "Memory" filter narrows the tree, and that the detail panel shows store/record/count + parsed records. Cleanest producer is an OTLP `POST /v1/traces` span with `gen_ai.operation.name = "update_memory"` and `gen_ai.memory.records` sent as a structured array (exercises the transform fix).

## Checklist

- [x] I have signed the CLA

### DIFF
--- a/packages/domain/spans/src/otlp/transform.ts
+++ b/packages/domain/spans/src/otlp/transform.ts
@@ -30,6 +30,21 @@ function nanosToDate(nanos: string | undefined): Date {
   return new Date(ms)
 }
 
+function anyValueToPlain(value: OtlpAnyValue | undefined): unknown {
+  if (!value) return undefined
+  if (value.stringValue !== undefined) return value.stringValue
+  if (value.boolValue !== undefined) return value.boolValue
+  if (value.intValue !== undefined) return Number(value.intValue)
+  if (value.doubleValue !== undefined) return value.doubleValue
+  if (value.arrayValue !== undefined) return (value.arrayValue.values ?? []).map(anyValueToPlain)
+  if (value.kvlistValue !== undefined) {
+    const result: Record<string, unknown> = {}
+    for (const entry of value.kvlistValue.values ?? []) result[entry.key] = anyValueToPlain(entry.value)
+    return result
+  }
+  return undefined
+}
+
 function resolveAnyValue(
   value: OtlpAnyValue | undefined,
 ): { type: "string" | "int" | "float" | "bool"; value: string | number | boolean } | null {
@@ -38,6 +53,10 @@ function resolveAnyValue(
   if (value.boolValue !== undefined) return { type: "bool", value: value.boolValue }
   if (value.intValue !== undefined) return { type: "int", value: Number(value.intValue) }
   if (value.doubleValue !== undefined) return { type: "float", value: value.doubleValue }
+  // Structured OTLP values (e.g. gen_ai.memory.records) are flattened to a JSON string so they survive in attr_string.
+  if (value.arrayValue !== undefined || value.kvlistValue !== undefined) {
+    return { type: "string", value: JSON.stringify(anyValueToPlain(value)) }
+  }
   return null
 }
 
--- a/packages/domain/spans/src/otlp/transform.ts
+++ b/packages/domain/spans/src/otlp/transform.ts
@@ -1,5 +1,6 @@
 import { ExternalUserId, OrganizationId, ProjectId, SessionId, SimulationId, SpanId, TraceId } from "@domain/shared"
 import type { SpanDetail, SpanKind, SpanStatusCode } from "../entities/span.ts"
+import { anyValueToPlain } from "./any-value.ts"
 import { attrArray, stringAttr } from "./attributes.ts"
 import { parseContent } from "./content/index.ts"
 import { isDroppedSpan } from "./dropped-spans.ts"
@@ -28,21 +29,6 @@ function nanosToDate(nanos: string | undefined): Date {
   if (!nanos || nanos === "0") return new Date()
   const ms = Number(BigInt(nanos) / BigInt(1_000_000))
   return new Date(ms)
-}
-
-function anyValueToPlain(value: OtlpAnyValue | undefined): unknown {
-  if (!value) return undefined
-  if (value.stringValue !== undefined) return value.stringValue
-  if (value.boolValue !== undefined) return value.boolValue
-  if (value.intValue !== undefined) return Number(value.intValue)
-  if (value.doubleValue !== undefined) return value.doubleValue
-  if (value.arrayValue !== undefined) return (value.arrayValue.values ?? []).map(anyValueToPlain)
-  if (value.kvlistValue !== undefined) {
-    const result: Record<string, unknown> = {}
-    for (const entry of value.kvlistValue.values ?? []) result[entry.key] = anyValueToPlain(entry.value)
-    return result
-  }
-  return undefined
 }
 
 function resolveAnyValue(
--- a/packages/platform/db-clickhouse/src/repositories/span-repository.ts
+++ b/packages/platform/db-clickhouse/src/repositories/span-repository.ts
@@ -491,9 +491,12 @@ export const SpanRepositoryLive = Layer.effect(
               // matching parts and is expensive for traces with many spans or
               // large payload columns. The newest ingested row wins, matching
               // the table's ReplacingMergeTree(ingested_at) semantics.
-              query: `SELECT ${LIST_COLUMNS}
+              // Drop the attr maps (same OOM hazard as listBySessionId — a trace's
+              // spans can each carry whole conversations or memory records in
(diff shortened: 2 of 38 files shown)


csansoon:
@claude review this PR

andresgutgon (dismissed):

packages/domain/spans/src/otlp/transform.ts
```diff
@@ -38,6 +53,10 @@ function resolveAnyValue(
   if (value.boolValue !== undefined) return { type: "bool", value: value.boolValue }
   if (value.intValue !== undefined) return { type: "int", value: Number(value.intValue) }
   if (value.doubleValue !== undefined) return { type: "float", value: value.doubleValue }
+  // Structured OTLP values (e.g. gen_ai.memory.records) are flattened to a JSON string so they survive in attr_string.
+  if (value.arrayValue !== undefined || value.kvlistValue !== undefined) {
+    return { type: "string", value: JSON.stringify(anyValueToPlain(value)) }
```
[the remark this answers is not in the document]
csansoon: Good catch, and correct — this PR made it worse by flattening `gen_ai.memory.records` into `attr_string`, which `listByTraceId` was reading for every span in the trace.

Fixed in `95bbf6aa2`: `listByTraceId` now returns empty attr maps at the ClickHouse read (`LIST_COLUMNS_LEAN` + `EMPTY_ATTR_MAP_COLUMNS`), mirroring what `listByTraceIds` and `listBySessionId` already do for the same reason.

Verified safe for both consumers before dropping the maps:
- The web trace-list server function returns `SpanRecord`, which has no `attrString`/`attrInt` (those live only on `SpanDetailRecord` from `findBySpanId`), so the serializer was discarding them anyway.
- The only domain consumer (`resolveLastLlmCompletionSpanId` via `resolve-score-trace-context`) reads only `operation`/`endTime`/`spanId`.

Also documented the empty-maps contract on the `listByTraceId` port docstring, matching `listBySessionId`.

csansoon:
Thanks for the review. Addressed the two flattener findings:

**Finding 1 (reuse — three copies of the AnyValue flattener):** fixed in `5cf98a899`. Extracted a single `anyValueToPlain` into `packages/domain/spans/src/otlp/any-value.ts` and pointed all three call sites at it — `transform.ts` (`anyValueToPlain`), `resolvers/enrichment.ts` (`otlpValueToPlain`), and `content/genai.ts` (`anyValueToJs`) are gone. Canonicalized on the `value.arrayValue?.values` form the enricher/parser already used, so their behavior is unchanged.

**Finding 2 (`__proto__` key shadowing):** folded into the same commit — the shared helper now captures a `__proto__` key as an own property via `Object.defineProperty` instead of letting plain assignment hijack the object prototype and drop the value. The fast path (normal keys) stays a plain assignment, so no per-attribute `defineProperty` cost on ingest.

**Finding 3 (`OPERATION_ICON` `satisfies` is non-enforcing):** agreed, and thanks for confirming the analysis. Leaving it as-is for this PR — a literal-only `Operation` variant to restore exhaustiveness is a broader change I'd rather not fold into the Phase 0 slice.

Separately, addressed Codex's inline P2 about `listByTraceId` reading the attr maps (`95bbf6aa2`).

`@domain/spans` typecheck + 916 tests pass, biome + knip clean. The `@platform/db-clickhouse` repository test is chdb-backed and I couldn't run it locally, so I'm relying on CI for that one — the change mirrors the already-tested `listByTraceIds`/`listBySessionId` queries.

packages/platform/db-clickhouse/src/repositories/span-repository.ts
```diff
@@ -487,9 +487,12 @@ export const SpanRepositoryLive = Layer.effect(
...
               // large payload columns. The newest ingested row wins, matching
               // the table's ReplacingMergeTree(ingested_at) semantics.
-              query: `SELECT ${LIST_COLUMNS}
+              // Drop the attr maps (same OOM hazard as listBySessionId — a trace's
+              // spans can each carry whole conversations or memory records in
+              // `attr_string`; the span detail view reads attributes via findBySpanId).
+              query: `SELECT ${LIST_COLUMNS_LEAN}, ${EMPTY_ATTR_MAP_COLUMNS}
                     FROM (
-                      SELECT ${LIST_COLUMNS}
+                      SELECT ${LIST_COLUMNS_LEAN}
```
andresgutgon, line 499: Is this change part of this PR?

csansoon: Yes. It came out of Codex's P2 inline review on this PR (the now-resolved thread on `transform.ts`). The chain: this PR starts flattening large `gen_ai.memory.records` payloads into `attr_string` (the transform change), and `listByTraceId` was selecting the full attr maps for every span in the trace — so opening a trace with big memory records would read/materialize those blobs even though nothing on the trace-list path uses them.

The fix just makes `listByTraceId` return empty attr maps at the CH read (`LIST_COLUMNS_LEAN` + `EMPTY_ATTR_MAP_COLUMNS`), mirroring what `listByTraceIds` and `listBySessionId` already do for the same reason. Verified safe: the web trace-list serializer returns `SpanRecord` (no `attrString`/`attrInt` — those are `SpanDetailRecord`-only, via `findBySpanId`), and the only domain caller (`resolveLastLlmCompletionSpanId`) reads just `operation`/`endTime`/`spanId`.

Happy to split it into its own PR if you'd rather keep this one to the Phase 0 slice — but since this PR's flattening is what makes that read load-bearing, I kept them together.

PR #4037, opened 2026-07-15
feat(taxonomy): Custom Behaviors Phase 5 — scoped gardening + auto-garden (LAT-760)

andresgutgon:
## What & why

Phase 5 of Custom Behaviors (LAT-760). Makes a custom behavior a **filter-scoped *living* taxonomy** with the **same trend affordances as the global Behaviours page** — born/continue/die, novelty, new-this-week, spiking, a real time window, and segments — with **no structural schema change** (one additive nullable column). Also makes custom behaviors **fully automatic like the global tree**: creating one auto-starts gardening and a cron sweep keeps it living; there is **no manual trigger**.

The reframing: a gardened custom behavior is the same kind of thing as the global tree, filter-scoped — not a different model. Global derives trend from *latest snapshot + `start_time` windows + stable cluster ids* (Hungarian matcher), and `custom_behavior_assignments` already mirrors that shape, so this is algorithm + read + cron wiring, not schema.

## Changes

**1. Periodic scoped cron sweep** (`apps/workers`)
- New `taxonomy`/`gardenCustomBehaviorSweep` task + `CUSTOM_BEHAVIOR_GARDENING_CRON_*` constants, registered in `server.ts` on the same 6h cadence as the global `gardenSweep`.
- `runGardenCustomBehaviorSweepJob` fans out one `gardenCustomBehavior` (`reason: "cron"`) per eligible behavior, deduped on the behavior's workflow id.
- Eligibility via a cross-org **admin** query `listGardenableCustomBehaviors` (live project — same demo/showcase/soft-delete exclusions as `listGardenableProjectRefs`; past the cadence throttle). Admin/RLS-bypass because the sweep is a system job with no single "current org" — RLS can't express a cross-tenant list; each enqueued job re-scopes to its own org.
- One **additive nullable** column `custom_behaviors.last_gardened_at`, stamped at run start as the throttle anchor. **No enabled flag** — a behavior gardens for its whole lifetime; deleting it is the only off switch.

**2. Auto-garden, no manual trigger**
- `createCustomBehaviorUseCase` enqueues the first gardening run on create (best-effort; the sweep backstops).
- Removed the manual Generate/regenerate button (list + detail) and the `generateCustomBehaviorFn` server function + `useGenerateCustomBehavior` hook. The `generateCustomBehavior` **domain** use-case stays (reused by create and the dev/QA helper).

**3. Scoped trend read** (Part 4)
- New `CustomBehaviorAssignmentRepository.getClusterTrendCounts`, windowed over `custom_behavior_assignments.start_time` (scoped mirror of the global read), wired into the scoped branch of `listProjectBehavioursUseCase` (was hardcoded `trendCounts = []`).
- Scoped `getClusterAssignmentCounts` gains an optional `start_time` window so the time picker filters volume too.

**4. Re-enabled UI** (`apps/web`, Part 5)
- The scoped custom-behaviours page now passes `segment` / `onSegmentChange` / `timeFilter` / `timeRange` (via `useAnalyticsTimeWindow` + `TimeFilterDropdown`) to the reused `BehavioursView`.
- The pending/empty view is reframed as an **auto-garden waiting state**: under 15 matching observations is a normal steady state, not pre-Generate. It shows the live preview count vs the threshold and keeps waiting-for-data / generating / ready / failed visually distinct so it never reads as broken.

**5. Lineage + write semantics** (Parts 2 & 3 — already shipped in Phase 2, now tested + documented)
- Scoped lineage continuity already runs the Hungarian matcher against the behavior's own prior clusters (`listActiveByProject({ customBehaviorId })`), so cluster ids survive regeneration.
- Writes are insert-only (`upsertMany` into a `ReplacingMergeTree`); nothing truncates the scoped slice on re-run, so history accumulates until TTL. Added an explicit accumulate/no-truncate test.
- Documented the whole mechanism in `dev-docs/taxonomy.md`.

## Invariants held
- Global Behaviours page + Topics filter unchanged; global reads filter `custom_behavior_id IS NULL`, scoped reads never touch `taxonomy_observations.assigned_cluster_id`.
- Out of scope (as global lacks it too): per-run membership-migration history — the only thing that would force `reassignment_run_id` into the sort key.

## Validation
- `pnpm typecheck` — full workspace green (87/87).
- Unit: `@domain/taxonomy` (99) incl. new scoped-trend + accumulate/no-truncate tests; `@platform/db-postgres` (9) incl. new `listGardenableCustomBehaviors` + migration-applied repo tests; create-CRUD updated for auto-start.
- ClickHouse-backed worker tests (`taxonomy.test.ts`, incl. the new sweep-handler cases) run in **CI** — the `chdb` native binary isn't built in this worktree, so they're skipped locally (pre-existing env limitation, not a code issue).
- **Not** run: full running-app end-to-end (repeated sweeps showing real trends/segments). It needs the Temporal + workers + CH stack and a seeded behavior with ≥15 matching observations; local dev seed cohorts are under-threshold and `chdb` isn't available here. Best verified against the LAT-752 QA seed (which includes a ≥42-obs generatable cohort and an under-threshold cohort for the waiting state).

## Follow-up split, if this is too large
The cron/worker + schema (Part 1 + auto-garden) is separable from the read + UI (Parts 4–5) if a smaller PR is preferred; they meet only at `last_gardened_at` and the scoped read branch. Editing a behavior's filter does not yet expedite a re-garden (it lands on the next sweep) — a candidate follow-up.


andresgutgon:
@claude review

andresgutgon:
@claude review again please

PR #4051, opened 2026-07-16
fix(cost): stop double-counting cached and reasoning tokens

andresgutgon:
## Problem

A customer running a reasoning model (**azure/gpt-5.1**) reported that the run-prompt API response cost/usage and the Traces UI both show wrong numbers. Investigation confirmed a token/cost **accounting bug**, separate from the earlier accumulated-usage fix (#3914).

Providers report `cachedInputTokens` as a **subset** of `promptTokens`, and `reasoningTokens` as a **subset** of `completionTokens`. Our code treated all four as disjoint and summed them, so cached and reasoning tokens were counted/charged twice.

For one real run:
| | Tokens | Cost |
|---|---|---|
| API response | 34,603 ✅ | $0.178 ❌ (reasoning billed twice) |
| Traces UI | 43,043 ❌ (adds cached+reasoning) | $0.248 ❌ (worse — also re-adds in `enrichCost`) |
| **Correct** | **34,603** | **≈ $0.108** |

## Fix

1. **`estimateCost` / `estimateCostBreakdown`** (`ai/estimateCost/index.ts`) — price only the non-overlapping remainder in the base buckets: `input` on `prompt − cached`, `output` on `completion − reasoning`. `cached` and `reasoning` are still each priced once at their own rate. Reported token counts stay as the raw provider values. This corrects cost everywhere (this is the shared cost function).
2. **`enrichCost`** (`tracing/spans/specifications/completion.ts`) — pass the **raw** provider token counts instead of pre-adding `cached`→input and `reasoning`→output, which had made the stored span cost double-count even harder than the API response.
3. **Trace token totals** (`queries/clickhouse/spans/computeDocumentTracesAggregations.ts` + `tracing/spans/fetching/computeDocumentTracesAggregations.ts`) — total tokens = `tokens_prompt + tokens_completion` instead of summing all four fields.
4. **Tests** — corrected the breakdown expectations and added a `does not double-count cached/reasoning` case.

## Impact & notes

- **Non-cached, non-reasoning models are unaffected** (`cached = 0`, `reasoning = 0` ⇒ `prompt − 0`, `completion − 0`), so no regression for standard models — only the previously over-charged cases change.
- `estimateCost` is used platform-wide (quota, billing displays, per-span cost), so this touches all cost figures for models that report cached/reasoning tokens — in the correcting (lower) direction.
- **Historical spans** keep their previously-stored inflated `spans.cost`; only newly-ingested spans get the corrected value. A backfill would be a separate task if we want history corrected.

## Testing

Docker was down locally, so relying on CI for `tc` + tests. The updated + new unit tests in `estimateCost/index.test.ts` cover the non-double-count behavior; the math reconciles to the customer's exact figures (response $0.178 and UI $0.248 reproduced from the old formula; corrected ≈ $0.108).

### DIFF
--- a/apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/experiments/comparison/route.test.ts
+++ b/apps/web/src/app/api/projects/[projectId]/documents/[documentUuid]/experiments/comparison/route.test.ts
@@ -300,9 +300,9 @@ describe('GET /api/projects/[projectId]/documents/[documentUuid]/experiments/com
         expect(data[0].runMetadata.count).toBe(2)
         expect(data[0].runMetadata.totalDuration).toBe(3000)
         expect(data[0].runMetadata.totalCost).toBe(300)
-        expect(data[0].runMetadata.totalTokens).toBe(
-          50 + 25 + 100 + 50 + 10 + 5,
-        )
+        // prompt + completion only; cached (10) and reasoning (5) are subsets
+        // already included in prompt/completion, so they are not added again.
+        expect(data[0].runMetadata.totalTokens).toBe(50 + 25 + 100 + 50)
       })
 
       it('ignores spans that do not belong to the experiment', async () => {
--- a/apps/web/src/components/tracing/spans/Completion.tsx
+++ b/apps/web/src/components/tracing/spans/Completion.tsx
@@ -137,11 +137,11 @@ function DetailsPanel({ span }: DetailsPanelProps<SpanType.Completion>) {
           {!!span.metadata.tokens && (
             <MetadataItem
               label='Tokens'
+              // prompt already includes cached, and completion already includes
+              // reasoning, so the total is prompt + completion (not the sum of
+              // all four, which would double-count the cached/reasoning subsets).
               value={(
-                span.metadata.tokens.prompt +
-                span.metadata.tokens.cached +
-                span.metadata.tokens.reasoning +
-                span.metadata.tokens.completion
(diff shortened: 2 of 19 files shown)


geclos (approved):

andresgutgon:
Addressed in `80f49cacf`.

**P2 (breakdown token counts):** `estimateCostBreakdown` now stores **disjoint** counts per category — `prompt` = non-cached remainder, `completion` = non-reasoning remainder, with `cached`/`reasoning` as their own buckets. So each `CostRow` shows tokens that match its cost, and `totalCost`/`totalTokens`/`totalInputTokens`/`totalOutputTokens` in `costs.ts` sum correctly (prompt+cached = raw prompt, completion+reasoning = raw completion) with no double-count. The raw provider values still live in the API response `usage`.

While here I also swept the rest of the same copy-pasted "sum all four token fields" pattern so nothing else double-counts:
- ClickHouse: `getExperimentRunMetadata`, `weeklyEmailLogs`, `fetchConversation` (+ the two trace aggregations already in the first commit)
- Postgres/TS: `experimentsRepository`, `data-access/conversations/shared`, `data-access/weeklyEmail/logs`, span `export`
- web-ui: `Completion`/`AggregatedCompletionsDetails` token totals, and `aggregateCompletionSpans` breakdown builders (now disjoint to match `estimateCostBreakdown`)

`pnpm tc` is green. Note historical span rows keep their previously-stored inflated `cost`; only newly-ingested spans get the corrected value.

PR #4074, opened 2026-07-16
refactor(memories): make store_id the sole memory key, remove scope

csansoon:
Reorients memory observability so a **store** (`gen_ai.memory.store.id`) is the unit everything groups under, and deletes the derived `scope` concept. This is a pre-step before Phase 3 (the Memory page): it re-bases the engine so that page is built on the real OTEL identity rather than a concept we're removing.

## why

There is no memory scope on the wire — `scope` was a Latitude-derived string (`gen_ai.memory.scope → latitude.memory.scope → user.id → ""`). Its user-id fallback fragmented a store shared by multiple users: the same record split into one `memory_current` row per user, and last-writer-wins never reconciled across them. Record identity is really `(store_id, record_id)` (decision D3), so `scope` was an extra axis on top of the true one. Removing it makes a shared store one manifest and turns `user_id` into what it actually is — an access annotation, not an identity.

## what changes

- **ClickHouse** (`00053_memory_drop_scope`): `memory_events` and `memory_current` recreated without the `scope` column, re-keyed on `(store_id, record_id)`. `memory_blobs` is untouched. The tables hold no durable data yet, so this is a `DROP + CREATE`, not a data-preserving rebuild.
- **`@domain/memories`**: `scope` removed from the entities. The `scope` parameter on `readCurrentSnapshot` / `readManifestAt` / `readLatestStoreWipes` becomes `storeId`; `readRecordVersions` drops it. `MemorySnapshot`, `MemoryDiff`, and the per-record session-summary now carry `storeId`. The materializer no longer resolves a scope and seeds its live-record set per store.
- **`@domain/spans`**: `MemoryOperationSpan` drops `scopeAttr` / `latitudeScopeAttr`; the span adapter stops reading `gen_ai.memory.scope` / `latitude.memory.scope`.
- **TS SDK**: the `memoryScope` capture option and its `latitude.memory.scope` attribute are removed — with scope gone they did nothing.
- **Web**: the session/trace memory-summary breakdown groups records by store id.
- **Spec** (`specs/memory-observability.md`): Scope resolution → Store identity, D4 superseded, Features 2–4 restated store-centric, and Phase 3 re-tasked as store list → store detail (records + blame + the users who accessed it) → a memory section on the existing user page.

## behavior change to know

A store written by two different users is now a single manifest with last-writer-wins across them, where before each user got an isolated fragment. For isolated per-user memory the emitter sets `store.id` to the user id, so those become distinct stores. A materializer test locks this in: a second user's upsert of a shared record classifies as `update`.

## deploy

Single-step column drop, so during a rolling deploy in-flight old pods briefly write and read the dropped column and fail. Memory materialization is a retryable trace-end job and the summary read degrades to empty, so both self-heal after cutover. Accepted because the tables are testing-phase with no production data.

## scope

Engine and spec only. The Memory page UI (Phase 3) is not in this PR.

## testing

Typecheck across memories, spans, db-clickhouse, telemetry, web, workers, ingest, and api. Tests: `@domain/memories` use-cases (20), the chdb-backed adapter suites against the regenerated schema (8), and the telemetry SDK (140).

Refs LAT-729.


csansoon:
@claude review this pr

csansoon:
@codex good catch, and correct — this was a regression I introduced. The original `recordKey`/`scopedKey` used a NUL delimiter; my full-file rewrites of the summary and fake silently replaced the (invisible) NUL with a space, so opaque ids containing spaces could collide.

Fixed in a2d1c6a00: restored a collision-free delimiter across every record-identity key in the read path — the summary's `recordKey`, the fake repo (`recordKey`/`blobKey` plus the `readSessionMemoryEvents`/`readRecordVersions` dedup and membership sets), the adapter's `readRecordVersions` membership set, and normalized `compute-memory-diff.ts` to match. It's now written as an explicit escape sequence rather than a raw NUL byte, so it stays visible in source and can't be clobbered by a future edit. Domain (20) and chdb adapter (8) suites pass.

csansoon:
Re: the note on `MemoryRecordChange.storeId` being redundant now that diffs are per-store — agreed it's not a bug. Keeping it as-is: a `MemoryRecordChange` is more self-describing when it carries its own `(storeId, recordId)`, and Feature 4's changed-file view iterates changes directly, so having the store on each change avoids threading `MemoryDiff.storeId` through the render. Cheap to keep, and it's the natural shape if a multi-store diff ever lands. No change.

(Also addressed @codex's P2 record-key collision in a2d1c6a00 — see that thread.)

PR #4084, opened 2026-07-17
feat(taxonomy): pure relative divisive builder (LAT-771)

andresgutgon:
## What & why

Phase 2 of the [Taxonomy adaptive clustering](https://linear.app/latitude/project/taxonomy-adaptive-clustering-15ed0cc96307/overview) project (LAT-771). Absolute cosine similarity isn't comparable across projects, so the fixed root sibling-cosine gate (`0.85`) collapses narrow-domain corpora into a single cluster. This ships the **node-relative separation** builder (calibrated in Phase 1, #4079) alongside the existing static builder, and hardens the clustering worker's resource bounds.

**Production stays on the static builder.** This PR does not change what any org sees — the relative builder is dormant until the Phase-3 rollout gate (`off` → `shadow` + Datadog dashboard → per-org `adaptiveTaxonomyClustering` flag → all) chooses it. Nothing here is ungated.

## Two explicit builders (no default)

Both divisive builders share the same spherical-k-means core and differ only in the split-acceptance gate. There is no unprefixed builder — every caller picks one explicitly.

| | Function | Schedule | Used by |
| --- | --- | --- | --- |
| **Static** (absolute sibling-cosine) | `buildStaticHierarchicalClusters` | `TAXONOMY_TREE_STATIC_DEPTH_SCHEDULE` | **production** (gardening use-case + worker), calibration A/B baseline |
| **Relative** (node-relative separation) | `buildRelativeHierarchicalClusters` | `TAXONOMY_TREE_RELATIVE_DEPTH_SCHEDULE` | its own tests + the calibration harness (dormant in prod) |

## Changes

**Relative builder — `packages/domain/taxonomy/src/clustering.ts` + `constants.ts`**
- New `RelativeDepthSchedule` (`maxDominantChildFraction`, `minRelativeSeparation`, `withinDistanceQuantile`, `routingSimilarityQuantile`) and calibrated `TAXONOMY_TREE_RELATIVE_DEPTH_SCHEDULE` (root separation `0.45`, tightening to `0.55`/`0.65` by depth).
- Gate is **size + dominance + score + relative-separation**:
  - `memberDistance = 1 - cosine(x, assignedCentroid)`; `withinDistance = quantile(memberDistance, withinDistanceQuantile)`
  - `closestSiblingDistance = min(1 - cosine(childA, childB))`; `relativeSeparation = closestSiblingDistance / max(withinDistance, 1e-6)`
  - cosine distances clamped to `[0, 2]`; quantiles via linear interpolation at `(n-1)*q`.
- Per-split member-confidence routing threshold (lower-tail quantile of member-to-own-centroid similarity, min across children, floored by the global absolute threshold).
- Deterministic tie-breaking: lower K, then earlier restart, on exact Calinski–Harabasz ties.
- Returns `{ root, diagnostics }` with **bounded, embedding-free** `RelativeClusteringDiagnostics` (selected K by depth, accepted/rejected counts + reasons, separation/threshold distributions).

**Static builder** — the previous algorithm, unchanged, kept as `buildStaticHierarchicalClusters` (`StaticDepthSchedule` / `TAXONOMY_TREE_STATIC_DEPTH_SCHEDULE`). It is the production path today, the calibration-harness baseline, and the Phase-6 removal target.

Both builders stay pure functions of embeddings, schedule, project-derived seed, and k-means constants — no settings, run history, `Date`/time, external services, or `Math.random()`. Relative member distances are O(n) per candidate; sibling distances O(K²), K ≤ 10.

**Worker — `apps/workflows/src/activities/taxonomy-clustering-worker.ts`**
- A single wall-clock deadline (`TAXONOMY_CLUSTERING_WORKER_TIMEOUT_MS`), `worker.terminate()` + `clearTimeout` on every terminal path, and Node `resourceLimits.maxOldGenerationSizeMb` (`TAXONOMY_CLUSTERING_WORKER_MAX_OLD_GEN_MB` = 512). Lifecycle is managed with a `using` `DisposableStack` (needs `ESNext.Disposable` in the workflows tsconfig lib; Node ≥25 provides it natively). Builder-agnostic; the worker runs the static builder today.

**Tests** — new `clustering.test.ts` drives the relative builder on the shipped schedule (determinism, narrow-domain 3–5 children, unimodal suppression, finite + self-consistent bounded diagnostics, `quantile`) and asserts a partition-signature match against the Phase-1 calibrated candidate. Calibration harness/test unchanged (static baseline). Worker test covers the static bare-tree path.

## Phase-4 pointer: shadow-comparison shape symmetry

The static builder returns a bare `ClusteringTreeNode` and does **not** emit a diagnostics object — deliberately. The `off`/static path has no relative-separation, routing thresholds, or relative rejection reasons to report, and tree shape is a pure function of the node (one `treeShape(root)` walk, as the calibration harness already does).

When Phase 4 builds the `taxonomy.gardenTaxonomyWorkflow.shadowComparison` event, compute the `static.*` **and** `adaptive.*` shape fields (`rootChildCount`, `nodeCount`, `leafCount`, `maxDepth`) with **one shared `treeShape(root)` helper applied to both trees**, and use `RelativeClusteringDiagnostics` only for the relative-only metadata (rejection reasons, separation/threshold percentiles, selected-K). Note `RelativeClusteringDiagnostics` carries `nodeCount`/`leafCount`/`maxDepth` but not `rootChildCount` (read from `root.children.length`) — routing both sides through `treeShape` keeps the two columns symmetric and avoids drift. Do not add a diagnostics object to the static builder just for symmetry; it's slated for removal in Phase 6.

## Deferred to later phases
- Rollout mode gate `off/shadow/enforced` + `adaptiveTaxonomyClustering` per-org flag, and switching the gardening path to the relative builder — **Phase 3/4**.
- Removing `computeSplitLinkThreshold` and wiring member-confidence thresholds into the persisted plan; staging + atomic swap + full live-window reassignment — **Phase 3**.
- Bounded comparison telemetry + Datadog shadow dashboard — **Phase 4**.

## Test plan
- `pnpm --filter @domain/taxonomy typecheck` / `--filter workflows typecheck` ✅
- `pnpm --filter @domain/taxonomy exec vitest run src/clustering.test.ts src/calibration/calibration.test.ts src/use-cases/build-hierarchical-taxonomy.test.ts` — 65 passed ✅
- `pnpm --filter workflows exec vitest run src/activities/taxonomy-clustering-worker.test.ts` — 1 passed ✅
- Worker `using` lifecycle QA'd end-to-end (tsx dev path + compiled cjs under plain Node 25): correct result, error-path rejects, clean process exit ⇒ deadline timer cleared and worker terminated.

### DIFF
--- a/apps/workflows/src/activities/taxonomy-clustering-worker-entry.ts
+++ b/apps/workflows/src/activities/taxonomy-clustering-worker-entry.ts
@@ -3,7 +3,7 @@ import { type BuildHierarchicalClustersInput, buildHierarchicalClusters } from "
 
 interface WorkerSuccessMessage {
   readonly ok: true
-  readonly tree: ReturnType<typeof buildHierarchicalClusters>
+  readonly result: ReturnType<typeof buildHierarchicalClusters>
 }
 
 interface WorkerErrorMessage {
@@ -21,8 +21,8 @@ const errorMessage = (error: unknown): WorkerErrorMessage => {
 
 try {
   if (!parentPort) throw new Error("Taxonomy clustering worker started without a parent port")
-  const tree = buildHierarchicalClusters(workerData as BuildHierarchicalClustersInput)
-  parentPort.postMessage({ ok: true, tree } satisfies WorkerSuccessMessage)
+  const result = buildHierarchicalClusters(workerData as BuildHierarchicalClustersInput)
+  parentPort.postMessage({ ok: true, result } satisfies WorkerSuccessMessage)
 } catch (error) {
   parentPort?.postMessage(errorMessage(error))
 }
--- a/apps/workflows/src/activities/taxonomy-clustering-worker-entry.ts
+++ b/apps/workflows/src/activities/taxonomy-clustering-worker-entry.ts
@@ -1,9 +1,9 @@
 import { parentPort, workerData } from "node:worker_threads"
-import { type BuildHierarchicalClustersInput, buildHierarchicalClusters } from "@domain/taxonomy"
+import { type BuildStaticHierarchicalClustersInput, buildStaticHierarchicalClusters } from "@domain/taxonomy"
 
 interface WorkerSuccessMessage {
   readonly ok: true
-  readonly result: ReturnType<typeof buildHierarchicalClusters>
+  readonly tree: ReturnType<typeof buildStaticHierarchicalClusters>
 }
 
 interface WorkerErrorMessage {
@@ -21,8 +21,8 @@ const errorMessage = (error: unknown): WorkerErrorMessage => {
 
 try {
   if (!parentPort) throw new Error("Taxonomy clustering worker started without a parent port")
-  const result = buildHierarchicalClusters(workerData as BuildHierarchicalClustersInput)
-  parentPort.postMessage({ ok: true, result } satisfies WorkerSuccessMessage)
+  const tree = buildStaticHierarchicalClusters(workerData as BuildStaticHierarchicalClustersInput)
+  parentPort.postMessage({ ok: true, tree } satisfies WorkerSuccessMessage)
 } catch (error) {
   parentPort?.postMessage(errorMessage(error))
 }
--- a/apps/workflows/src/activities/taxonomy-clustering-worker.ts
+++ b/apps/workflows/src/activities/taxonomy-clustering-worker.ts
@@ -3,15 +3,15 @@ import { dirname, extname, resolve } from "node:path"
 import { fileURLToPath, pathToFileURL } from "node:url"
 import { Worker } from "node:worker_threads"
 import {
-  type BuildHierarchicalClustersInput,
-  type BuildHierarchicalClustersResult,
+  type BuildStaticHierarchicalClustersInput,
+  type ClusteringTreeNode,
   TAXONOMY_CLUSTERING_WORKER_MAX_OLD_GEN_MB,
   TAXONOMY_CLUSTERING_WORKER_TIMEOUT_MS,
 } from "@domain/taxonomy"
 
 interface WorkerSuccessMessage {
   readonly ok: true
-  readonly result: BuildHierarchicalClustersResult
+  readonly tree: ClusteringTreeNode
 }
 
 interface WorkerErrorMessage {
@@ -46,14 +46,14 @@ const workerEntryUrl = () => {
  *     process.
  *   - a single wall-clock deadline bounds the run and terminates a hung or
  *     looping worker. It covers the entire worker invocation — one build today,
- *     static plus adaptive shadow later — so shadow mode shares one budget.
+ *     static plus relative shadow later — so shadow mode shares one budget.
  *   - the timer and the thread are registered on a `using` DisposableStack, so
  *     `clearTimeout` + `worker.terminate()` run exactly once when this scope
  *     exits, on every terminal path (message, error, exit, timeout).
  */
 export const buildHierarchicalClustersInWorker = async (
-  input: BuildHierarchicalClustersInput,
-): Promise<BuildHierarchicalClustersResult> => {
+  input: BuildStaticHierarchicalClustersInput,
+): Promise<ClusteringTreeNode> => {
   const entry = workerEntryUrl()
   // The deadline timer and the worker thread are torn down (LIFO) when this
   // scope exits — i.e. once the awaited promise settles, on every terminal path
@@ -67,7 +67,7 @@ export const buildHierarchicalClustersInWorker = async (
   })
   stack.defer(() => void worker.terminate())
 
-  return await new Promise<BuildHierarchicalClustersResult>((resolvePromise, rejectPromise) => {
+  return await new Promise<ClusteringTreeNode>((resolvePromise, rejectPromise) => {
     const timeout = setTimeout(() => {
       rejectPromise(new Error(`Taxonomy clustering worker timed out after ${TAXONOMY_CLUSTERING_WORKER_TIMEOUT_MS}ms`))
     }, TAXONOMY_CLUSTERING_WORKER_TIMEOUT_MS)
@@ -75,7 +75,7 @@ export const buildHierarchicalClustersInWorker = async (
 
     worker.once("message", (message: WorkerMessage) => {
       if (message.ok) {
-        resolvePromise(message.result)
+        resolvePromise(message.tree)
         return
       }
       const error = new Error(message.error)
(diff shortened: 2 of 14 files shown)


apps/workflows/src/activities/taxonomy-clustering-worker.ts
```diff
@@ -31,35 +36,55 @@ const workerEntryUrl = () => {
...
+ */
+export const buildHierarchicalClustersInWorker = async (
+  input: BuildHierarchicalClustersInput,
+): Promise<BuildHierarchicalClustersResult> => {
+  const entry = workerEntryUrl()
+  // The deadline timer and the worker thread are torn down (LIFO) when this
+  // scope exits — i.e. once the awaited promise settles, on every terminal path
+  // (message, error, exit, timeout). Promise settlement is idempotent, so the
+  // first event wins and later ones are no-ops.
+  using stack = new DisposableStack()
```
andresgutgon, line 62: @geclos I'm using `using` here. I think this is the first time in our code. Just take a look.

apps/workflows/src/activities/taxonomy-clustering-worker.ts
```diff
@@ -31,35 +36,55 @@ const workerEntryUrl = () => {
...
+  const worker = new Worker(entry.url, {
+    workerData: input,
+    execArgv: entry.isTypeScriptSource ? ["--import", "tsx"] : [],
+    resourceLimits: { maxOldGenerationSizeMb: TAXONOMY_CLUSTERING_WORKER_MAX_OLD_GEN_MB },
```
andresgutgon, line 66: Adding resource limits to the worker 👀

andresgutgon:
@claude review

apps/workflows/src/activities/taxonomy-clustering-worker-entry.ts
```diff
@@ -21,8 +21,8 @@ const errorMessage = (error: unknown): WorkerErrorMessage => {
 
 try {
   if (!parentPort) throw new Error("Taxonomy clustering worker started without a parent port")
-  const tree = buildHierarchicalClusters(workerData as BuildHierarchicalClustersInput)
-  parentPort.postMessage({ ok: true, tree } satisfies WorkerSuccessMessage)
+  const result = buildHierarchicalClusters(workerData as BuildHierarchicalClustersInput)
```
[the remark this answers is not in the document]
andresgutgon: Fixed. This was flagged against the first revision; the PR was since reworked (commit 54fa42184) so the worker entry now calls `buildStaticHierarchicalClusters` and `planHierarchicalTaxonomyUseCase` builds with the static schedule. Production stays byte-identical on the static builder — the relative builder is shipped but has no runtime caller until the Phase-3 off/shadow/enforced gate selects it. Resolving.

packages/domain/taxonomy/src/clustering.ts
```diff
@@ -291,36 +358,298 @@ const calinskiHarabaszScore = (
...
+      return { memberIndices, centroid, children: [], depth }
+    }
+    acceptedSplits++
+    const depthKs = selectedKByDepth[depth] ?? []
+    depthKs.push(best.k)
+    selectedKByDepth[depth] = depthKs
+    acceptedRelativeSeparations.push(best.relativeSeparation)
+    routingThresholds.push(best.splitLinkThreshold)
+    const children = best.clusterMemberIndices.map((childIndices) => recurse(childIndices, depth + 1))
+    return { memberIndices, centroid, children, depth }
```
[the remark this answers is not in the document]
andresgutgon: Valid, and deferred to Phase 3 by design. In this PR the relative builder is not the production or persistence path — gardening still runs the static builder, for which `computeSplitLinkThreshold` (sibling-cosine) is the correct descent gate, so nothing strands here today. The spec's Phase 3 ("Store accepted routing thresholds" + "Remove computeSplitLinkThreshold") is where the member-confidence threshold is carried onto the persisted node and the sibling-cosine value is dropped. One implementation note for that phase: the relative diagnostics currently expose `routingThresholds` as a flat per-split array, not per-node, so Phase 3 will attach the selected threshold to each `ClusteringTreeNode` (which today carries no `splitLinkThreshold`) before `collectNodes` persists it. Tracking it there rather than here.

apps/workflows/src/activities/taxonomy-clustering-worker.ts
andresgutgon: Rationale: it replaces the manual `settled`/`settle`/`cleanup` bookkeeping. The timer and the worker thread are registered on the `DisposableStack`, so `clearTimeout` + `worker.terminate()` run exactly once (LIFO) when the async scope exits — on every terminal path (message, error, exit, timeout). Promise settlement is already idempotent, so the first event wins and the explicit guard is redundant.

It needs the function to be `async` and `await` the result (disposal fires at scope exit, not when the Promise executor returns synchronously) and `ESNext.Disposable` in the tsconfig `lib` — Node >=25 provides `DisposableStack`/`Symbol.dispose` natively, and tsdown (target node25) emits the `using` keyword natively, no downlevel helper. I scoped the lib addition to `apps/workflows/tsconfig.json` only, not repo-wide, so this doesn't silently enable `using` everywhere — easy to promote to the base config if we want it broadly.

QA'd end-to-end (tsx dev path + the compiled `.cjs` under plain Node 25): correct result, error path rejects, and the process exits cleanly — which is the proof disposal ran, since a leaked worker thread or the pending 5-min deadline timer would otherwise keep the event loop alive.

apps/workflows/src/activities/taxonomy-clustering-worker.ts
andresgutgon: Rationale: `maxOldGenerationSizeMb` caps that worker isolate's V8 heap. The divisive build is the one allocation-heavy step and runs on externally-shaped input (a tenant's ≤1,500 × 2,048 embedding sample). Without a cap, a pathological corpus or a future allocation regression would grow the shared workflows activity process and could OOM-kill it — taking down every other in-flight Temporal activity in that process. With the cap, a blow-up hits the worker's own OOM instead, surfaced to us as a rejection via the error/exit handlers and torn down by the same DisposableStack — contained to the disposable worker.

It pairs with the wall-clock deadline to bound both time and memory of the build, and it's one shared budget that will cover static plus the relative shadow build once Phase 4 runs both in the one worker.

512MB is the calibrated ceiling from Phase 1 (recorded in BASELINES.md); the max-sample build peaks well under it (~tens of MB of growth), so it's a deliberately generous backstop/tripwire that only fires on a gross regression, never on a healthy build.

PR #4101, opened 2026-07-19
fix(ingest): don't crash the whole span batch on a malformed traceId

geclos:
## What does this PR do?

Fixes a production `TypeError: Cannot read properties of undefined (reading 'replace')` in `transformSpan` (`packages/domain/spans/src/otlp/transform.ts:180`) — [Datadog issue `28550eee-82d3-11f1-b2ea-da7ad0900005`](https://app.datadoghq.eu/error-tracking/issue/28550eee-82d3-11f1-b2ea-da7ad0900005), `workers` service, 99 occurrences, first seen 2026-07-18 (new regression, first triage of this issue).

**Root cause:** `apps/ingest`/`workers` decode OTLP/JSON span-ingestion bodies with a raw `JSON.parse(...) as OtlpExportTraceServiceRequest` cast (`packages/domain/spans/src/use-cases/process-ingested-spans.ts:56`) — the payload is cast, not validated. `transform.ts` already guards against this for `attributes` (see the `attrArray` helper and its comment in `packages/domain/spans/src/otlp/attributes.ts`), but `transformSpan` still called `span.traceId.replace(/-/g, "")` unguarded. A span whose `traceId` field is missing or non-string (a malformed client payload) throws, and because `transformSpan` runs inside `transformOtlpToSpans`'s loop inside an `Effect.gen`, the exception propagates as a defect that fails the **entire ingestion batch** for that request — not just the one malformed span — causing legitimate telemetry in the same batch to be dropped/retried.

**Fix:** `transformOtlpToSpans` already has a well-established "skip and count, don't crash" mechanism for spans with an unresolvable `projectId` (`rejectedSpans`). Extended the same guard to a missing/non-string `traceId`: such spans are now rejected (counted in `rejectedSpans`) instead of throwing, and the rest of the batch is processed normally.

Note: on the ingest/worker path where this crash occurred, `transformOtlpToSpans`'s `rejectedSpans` count is not currently wired up to the OTLP `partialSuccess` response (that response is built earlier, from project-slug resolution only) — bad-`traceId` spans are silently dropped rather than reported back to the client. That's pre-existing behavior for this code path, not a regression from this change, and out of scope here; flagging in case it's worth a follow-up.

Datadog: https://app.datadoghq.eu/error-tracking/issue/28550eee-82d3-11f1-b2ea-da7ad0900005

Closes #

## How was this tested?

- Added tests in `packages/domain/spans/src/otlp/tests/project-scoping.test.ts`:
  - a span with a missing (`undefined`) `traceId` is rejected, not thrown
  - a span with a non-string `traceId` (object) is rejected, not thrown
  - a mixed batch keeps processing valid spans after skipping one with an empty `traceId`
- Verified the new tests **fail** against the pre-fix code with the exact same stack trace as the Datadog issue (`TypeError: Cannot read properties of undefined (reading 'replace')` at `transform.ts:180`), and **pass** with the fix applied.
- `pnpm --filter @domain/spans test` — all otlp/project-scoping tests pass (pre-existing unrelated failures in `ingest-spans.test.ts` are a sandbox Node-version issue — `Uint8Array.toBase64` requires Node 24+ — reproduced identically on `development` without this change).
- `pnpm exec biome check` on changed files — clean.

**Verification after deploy:** occurrences of Datadog issue `28550eee-82d3-11f1-b2ea-da7ad0900005` should drop to zero. (Malformed-`traceId` spans are dropped in the worker, not surfaced via `partialSuccess` — see note above.)

## Checklist

- [ ] I have signed the CLA

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed, missing, non-string, or empty trace IDs.
  * Invalid spans are now safely rejected without interrupting processing of other valid spans in the same batch.
  * Strengthened sampling key extraction and span transformation to avoid crashes from invalid trace ID values.
  * Added additional test coverage for robustness with malformed trace IDs.

### DIFF
--- a/packages/domain/spans/src/otlp/tests/project-scoping.test.ts
+++ b/packages/domain/spans/src/otlp/tests/project-scoping.test.ts
@@ -390,4 +390,107 @@ describe("transformOtlpToSpans trace ID normalization", () => {
     expect(spans).toHaveLength(1)
     expect(spans[0]?.traceId).toBe("0af7651916cd43dd8448eb211c80319c")
   })
+
+  it("rejects a span with a missing trace ID instead of crashing the batch", () => {
+    const { spans, rejectedSpans } = transformOtlpToSpans(
+      {
+        resourceSpans: [
+          {
+            resource: { attributes: [str("service.name", "test")] },
+            scopeSpans: [
+              {
+                scope: { name: "scope", version: "1" },
+                spans: [
+                  {
+                    traceId: undefined as unknown as string,
+                    spanId: "missing-trace",
+                    name: "missing-trace",
+                    startTimeUnixNano: "1710590400000000000",
+                    endTimeUnixNano: "1710590401000000000",
+                    attributes: [],
+                    status: { code: 1 },
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+      },
+      ctx,
+    )
+    expect(rejectedSpans).toBe(1)
+    expect(spans).toHaveLength(0)
+  })
+
+  it("rejects a span whose trace ID is a non-string value instead of crashing the batch", () => {
+    const { spans, rejectedSpans } = transformOtlpToSpans(
+      {
+        resourceSpans: [
+          {
+            resource: { attributes: [str("service.name", "test")] },
+            scopeSpans: [
+              {
+                scope: { name: "scope", version: "1" },
+                spans: [
+                  {
+                    traceId: { "0": "a" } as unknown as string,
+                    spanId: "object-trace",
+                    name: "object-trace",
+                    startTimeUnixNano: "1710590400000000000",
+                    endTimeUnixNano: "1710590401000000000",
+                    attributes: [],
+                    status: { code: 1 },
+                  },
+                ],
+              },
+            ],
+          },
+        ],
+      },
+      ctx,
+    )
+    expect(rejectedSpans).toBe(1)
+    expect(spans).toHaveLength(0)
+  })
+
+  it("keeps processing the rest of the batch when one span has an invalid trace ID", () => {
+    const { spans, rejectedSpans } = transformOtlpToSpans(
+      {
+        resourceSpans: [
+          {
+            resource: { attributes: [str("service.name", "test")] },
+            scopeSpans: [
+              {
+                scope: { name: "scope", version: "1" },
+                spans: [
+                  {
+                    traceId: "0af7651916cd43dd8448eb211c80319c",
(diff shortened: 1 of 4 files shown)


geclos:
@cursoragent review

geclos:
Thanks for the review — both nits addressed:

- Confirmed the `rejectedSpans` reporting gap: `ingestSpansUseCase`'s partial-success count is computed from project-slug resolution before the payload is queued, and `processIngestedSpans` (the worker path, where this crash happened) discards `transformOtlpToSpans`'s `rejectedSpans`. Corrected the PR description so the deploy-verification signal is "Datadog issue goes to zero," not client-visible `partialSuccess.rejectedSpans`. That gap is pre-existing, not introduced by this change, so left wiring it up as a follow-up rather than in scope here.
- Updated the stale `TransformResult.rejectedSpans` JSDoc to mention both rejection reasons (929d2b1).

Left the `nanosToDate`/`BigInt` follow-up as a note for a future issue if it shows up in Error Tracking — agreed it's out of scope for this crash fix.

---

geclos:
Both good catches — responses inline:

**"Move trace-id validation before sampling" (line 254):** confirmed and fixed in 8bb1717. `extractSamplingKey`'s `traceId` fallback handed a raw, non-string `traceId` straight to `deterministicSample`, which calls `crypto.createHash("sha256").update(key)` — that throws on a non-string `key`. This is reachable on the synchronous ingest HTTP path (not just the worker) whenever a project has fractional sampling enabled, so it's a second, independent crash site for the same "OTLP/JSON cast, not validated" root cause. Now returns `null` for a missing/non-string `traceId` (falls through to no sampling decision, same as when no session id is found), with regression tests for both cases.

**"Include trace-id rejects in the ingest response" (line 255):** this is the same gap Cursor's review already flagged — confirmed it's real (the OTLP `partialSuccess.rejectedSpans` is computed synchronously in `ingestSpansUseCase` from project-slug resolution *before* the payload is queued; the worker's `transformOtlpToSpans` rejection, including this PR's new traceId check, happens later and isn't wired back into that response). Already corrected the PR description to reflect this as the real verification signal (Datadog issue count, not client-visible `rejectedSpans`). Threading traceId validation into the synchronous `inspectPayload`/`ingestSpansUseCase` path so it's reported at request time would be a good follow-up, but it's a larger, cross-cutting change to the ingest counting/response pipeline — keeping this PR scoped to fixing the crash.

---

PR #4165, opened 2026-07-22
feat(web): drag-to-resize span detail panel in Spans tab

csansoon:
Replaces the Expand/Minimize tree toggle in the Spans tab with a drag-to-resize span detail panel, matching the pointer-event resize pattern already used elsewhere (the memory activity panel, the span tree/waterfall splitter).

## before

Selecting a span in the Spans tab showed its details in a bottom panel whose size was controlled by an "Expand tree"/"Minimize tree" button — a binary toggle between a 192px-capped tree (detail takes the rest) and a 50/50 split. There was no way to pick an in-between size.

## now

The detail panel carries an explicit height and exposes a `role="separator"` grab handle on its top border. Drag it (or focus it and use ↑/↓) to resize; the tree above takes the remaining space via flex. Dragging the panel nearly shut collapses it and deselects the span, same as the ✕ button. The panel opens at 50% of the available height, clamped so it never drops below a minimum or crowds out the tree, and a `ResizeObserver` keeps it in bounds when the drawer resizes.

The toggle button, its icons, and the `isMinimized`/`onToggleMinimized` state threaded through the tree components and both drawer wrappers are gone.

## shared surface

The Spans tab renders from one set of components (`GroupedSpanTree`, `SpanTree`, `SpanDetail`), so the change lands in both the Session drawer and the Trace drawer. The new resize logic lives in `span-detail/use-detail-resize.ts`, the vertical sibling of the tree's existing horizontal `useResizablePanel`.

## Summary by CodeRabbit

* **New Features**
  * Added resizable span detail panels with pointer and keyboard controls.
  * Panels can be resized within available space and closed by dragging below the minimum height.
  * Improved accessibility with resize-handle keyboard support and ARIA attributes.

* **Improvements**
  * Simplified span tree controls by removing minimize/maximize behavior.
  * Standardized resizing behavior across span details and record activity panels.

### DIFF
--- a/apps/web/src/lib/hooks/useResizablePanelHeight.ts
+++ b/apps/web/src/lib/hooks/useResizablePanelHeight.ts
@@ -1,20 +1,37 @@
 import { useMountEffect } from "@repo/ui"
 import { useCallback, useRef, useState } from "react"
 
-export const MIN_PANEL_HEIGHT = 140
-const DEFAULT_PANEL_HEIGHT = 320
-// Reserve this much above the detail panel so the filters bar and a few tree rows stay visible.
-export const MIN_CONTENT_ABOVE = 160
-// Drag the panel shorter than this and it closes instead of clamping to the minimum.
-const CLOSE_DRAG_THRESHOLD = 96
-const KEYBOARD_STEP = 24
+const DEFAULT_KEYBOARD_STEP = 24
 
-// Vertical sibling of the span-tree `useResizablePanel`: drags the span detail panel taller/shorter.
-export function useDetailResize(onClose: () => void) {
+interface Options {
+  readonly minHeight: number
+  readonly minContentAbove: number
+  readonly closeThreshold: number
+  readonly defaultHeight: number | "half"
+  readonly onClose: () => void
+  readonly keyboardStep?: number
+}
+
+/**
+ * Drag/keyboard resize for a panel pinned to the bottom of a flex container.
+ *
+ * Attach `panelRef` to the panel root (its `parentElement` is the sizing
+ * container) and apply `height` where the panel's resizable area lives.
+ * `minContentAbove` reserves space so the content above never collapses;
+ * dragging shorter than `closeThreshold` calls `onClose` instead of clamping.
+ */
+export function useResizablePanelHeight({
+  minHeight,
+  minContentAbove,
+  closeThreshold,
+  defaultHeight,
+  onClose,
+  keyboardStep = DEFAULT_KEYBOARD_STEP,
+}: Options) {
   const panelRef = useRef<HTMLDivElement>(null)
-  const [height, setHeight] = useState(DEFAULT_PANEL_HEIGHT)
+  const [height, setHeight] = useState(typeof defaultHeight === "number" ? defaultHeight : minHeight)
   const [isDragging, setIsDragging] = useState(false)
-  const drag = useRef<{ startY: number; startHeight: number; max: number } | null>(null)
+  const drag = useRef<{ startY: number; startHeight: number } | null>(null)
   const moveRef = useRef<((event: PointerEvent) => void) | null>(null)
   const upRef = useRef<(() => void) | null>(null)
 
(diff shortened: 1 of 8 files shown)


apps/web/src/routes/_authenticated/projects/$projectSlug/-components/trace-detail-drawer/tabs/spans-tab/span-detail/index.tsx
```diff
@@ -19,9 +20,39 @@ export function SpanDetail({
...
 
   return (
-    <div className="flex flex-col flex-1 overflow-hidden border-t border-border">
+    <div
+      ref={panelRef}
+      className="relative flex shrink-0 flex-col overflow-hidden border-t border-border"
+      style={{ height }}
+    >
+      {/* biome-ignore lint/a11y/useSemanticElements: resize handle requires div for drag events */}
+      <div
```
andresgutgon, line 45: We already have all resize machinery. Why coding again?

apps/web/src/lib/hooks/useResizablePanelHeight.ts
```diff
@@ -0,0 +1,100 @@
+import { useMountEffect } from "@repo/ui"
```
andresgutgon, line 1: Again. A lot is already done for resize panels

andresgutgon (changes requested):
Check existing resize "technology" before introducing other way 🙏

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/trace-detail-drawer/tabs/spans-tab/span-detail/use-detail-resize.ts
```diff
@@ -0,0 +1,100 @@
...
+// Drag the panel shorter than this and it closes instead of clamping to the minimum.
+const CLOSE_DRAG_THRESHOLD = 96
+const KEYBOARD_STEP = 24
+
+// Vertical sibling of the span-tree `useResizablePanel`: drags the span detail panel taller/shorter.
+export function useDetailResize(onClose: () => void) {
+  const panelRef = useRef<HTMLDivElement>(null)
+  const [height, setHeight] = useState(DEFAULT_PANEL_HEIGHT)
+  const [isDragging, setIsDragging] = useState(false)
+  const drag = useRef<{ startY: number; startHeight: number; max: number } | null>(null)
```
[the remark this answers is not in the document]
csansoon: Fixed in 533533555. This file is gone — the logic moved into the shared `useResizablePanelHeight` hook (`apps/web/src/lib/hooks/`), which now recomputes `maxHeight()` on every pointer move instead of snapshotting it at drag start, so a mid-drag drawer shrink can no longer restore an out-of-bounds height.

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/trace-detail-drawer/tabs/spans-tab/span-detail/index.tsx
csansoon: Good catch — there was no shared hook to reuse, only two local copies: the memory activity panel's `usePanelResize` (`record-content-view.tsx`) and the span-tree's horizontal `useResizablePanel`. Rather than add a third, I extracted `useResizablePanelHeight` into `apps/web/src/lib/hooks/` and pointed both vertical panels (span detail + memory activity) at it. Done in 533533555.

apps/web/src/lib/hooks/useResizablePanelHeight.ts
csansoon: Same fix as the other thread — this file is deleted in 533533555. The one shared `useResizablePanelHeight` hook now backs both the span detail panel and the memory activity panel.

csansoon:
Addressed the resize-machinery feedback in 533533555: extracted a single shared `useResizablePanelHeight` hook (`apps/web/src/lib/hooks/`) and migrated both vertical panels — the span detail panel and the memory activity panel — onto it, so there is now one implementation rather than three copies.

Also folded in the CodeRabbit notes: the stale-max bug is fixed (max recomputed each pointer move), and the implementation-narrating comment is gone. I kept the keyboard step at 24 (the value both vertical panels already used) rather than the span-tree's horizontal 8 — the two vertical panels are now consistent with each other.

Re-requesting review when you have a moment 🙏

PR #4176, opened 2026-07-22
fix(telemetry): resolve Claude Code memory dir across git worktrees

csansoon:
Follow-up to #4140, which added memory-observability spans for Claude Code auto memory. Those spans never emitted for a session running in a git worktree, so a share of real usage was silently dropped.

## the bug

The hook located the auto-memory directory as the transcript's sibling: `dirname(transcript)/memory`. That only holds when the session's working directory is the same directory Claude Code stores memory under. With git worktrees it isn't. Claude Code keeps one memory store per repository, under the **main worktree**, while a linked worktree's session writes its transcript under that worktree's own project directory. The two paths differ, so `classifyMemoryTool`'s containment check failed and every `Read`/`Write`/`Edit` against the memory store was skipped — no child memory span, nothing on the Memory page.

A session run directly in the repo root was unaffected, since there the two directories coincide — which is why it went unnoticed.

## the fix

Match the memory store structurally instead of assuming a fixed location. Any path shaped like `<projectsRoot>/<store>/memory/<record>` — for any `<store>` under the shared projects root (`dirname(dirname(transcript))`) — is a memory operation, with `gen_ai.memory.store.id` set to the owning `<store>` slug and `gen_ai.memory.record.id` to the path under that store's memory dir.

- A session in the repo root behaves exactly as before (the store slug equals the transcript's slug).
- A worktree session writing the repo's main-worktree store now resolves to that store.
- Subagents ride the same projects root.

`MemoryEmitOptions.dir` becomes `projectsRoot`, and `memoryDirFromTranscript` becomes `memoryProjectsRoot`. The consume side is untouched: classification and materialization read the `gen_ai.memory.*` attributes, and the store id is still a single project slug.

A fully relocated `autoMemoryDirectory` outside `~/.claude/projects` stays an unhandled case, unchanged from before.

## validation

- Unit tests for the worktree case (a store under a different slug than the session) in both `memory.test.ts` and `otlp.test.ts`, plus a non-memory-subdirectory boundary case. Typecheck, build, format, and knip are green.
- Checked end to end against prod: with this build installed, a memory write from a git worktree session emits the `upsert_memory`/`update_memory` spans with `store.id` set to the main-clone slug — where the previous build emitted plain tool spans with no memory child.

Bumps `@latitude-data/claude-code-telemetry` to 0.0.14.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Claude Code telemetry auto-memory spans being silently skipped during Git worktree sessions.
  - Memory operations now correctly attribute `store` and `record` identifiers to the owning repository’s memory store.
  - Kept behavior unchanged for sessions run from the repository root.

- **Tests**
  - Expanded coverage for worktree-style paths, invalid/non-memory locations, and memory content capture.

- **Chores**
  - Bumped the Claude Code telemetry package to version 0.0.14.

### DIFF
--- a/packages/telemetry/claude-code/CHANGELOG.md
+++ b/packages/telemetry/claude-code/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **Memory spans now emit from git worktrees (e.g. Conductor).** The hook derived the auto-memory directory as the transcript's sibling (`dirname(transcript)/memory`), but Claude Code keeps one memory store per repository under the **main worktree**, while a linked worktree's session transcript lives under that worktree's own project directory. The two paths differ, so every memory operation from a worktree session was silently skipped. The classifier now recognizes any `~/.claude/projects/<store>/memory/<record>` path under the shared projects root, setting `gen_ai.memory.store.id` to the owning `<store>` slug — so a worktree session writing the repo's main-worktree store is captured. Sessions run directly in the repo root are unaffected.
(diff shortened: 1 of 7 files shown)


csansoon:
@coderabbitai addressed in 735ab475f — collapsed both `memory.ts` comments to single-line notes documenting just the worktree constraint (dropped the multi-line rationale). Thanks!

PR #4204, opened 2026-07-23
feat(taxonomy): behaviors become a catalog, each with views (LAT-782)

andresgutgon:
Implements **LAT-782**, consolidated: it absorbs the preset-catalog / create-modal / custom-facet phases and the IA rework that came out of showing the old screen to the team. Builds on Phases 1–3 (already on `development`). No new migrations.

## The problem

Behaviors was a single tree with a scope dropdown on top of it, and that dropdown was the only clue a project could have more than one tree. Nobody read it that way. Shown the screen cold, the team saw a pile of unrelated trees and could not say what a "behavior" was, what the dropdown listed, or how the entries related to each other.

## The model, and the words for it

A **behavior** is a question your sessions get grouped by. Latitude ships one, Topics ("what was this about?"), and users add more: five curated presets (user goal, outcome, friction reason, assistant approach, capability gap) or one they author. A behavior of that kind is internally a **facet** — extracted once per session, then clustered in the answer's embedding space instead of the transcript's. A **view** is a saved filtered slice of a behavior.

The UI now uses only those words. **"Lens" is gone from the product vocabulary**; the schema keeps `facet` as the internal term, and `facet-selection.ts` / `resolve-facet-selection.ts` replace the old lens-named modules.

## Screens

`/behaviours` is a grid of behavior cards: name, what it groups by, a teaser of the tree three levels deep, session and view counts. `+ Behavior` opens the preset catalog, with an authoring form behind it. Hovering a card blurs its content and surfaces a "View behavior" label; the whole card is the anchor, so the label is a styled span rather than a nested button. A project with nothing gardened drops the header entirely and shows the blank slate alone, carrying both actions — a title introducing an empty grid introduces nothing.

`/behaviours/:slug` is one behavior: its tree (unchanged from what ships today), its views as a row of links under the title, and `+ View`. The topic behavior sits among the others under the reserved `lat-topics` slug and renders exactly what production renders now. `+ View` only appears on a behavior, never on a view — a view is already a slice.

Creating a behavior analyzes every sampled session through it once. That cold start streams the extracted answers, a health read on them (clear rate, distinct answers, unclear rate) and refine/stop, so a question that does not discriminate is caught before its tree lands rather than after.

## Routes

- `/behaviours/:slug/views/:view` — views nest under their behavior. A flat `/behaviours/:slug` that names a view redirects to its nested URL, so no existing link breaks.
- `/behaviours/:slug/views/new` and `.../views/:view/edit` replace the old flat `edit` route.
- `/behaviours/new?filters=…` (the Sessions and saved-search entry point) asks which behavior to narrow first. That step is skipped whenever the parent is known or the project has only one behavior, so it appears only once a project actually has a choice to make.

## Reserved slug

`lat-` is now rejected on custom behavior slugs, not only facet slugs. Behavior slugs are the routing namespace, so before this a behavior named "Lat topics" would have slugified to `lat-topics` and shadowed the topic route. `TOPICS_BEHAVIOR_SLUG` is that sentinel; the topic behavior has no `custom_behaviors` row, so the UI resolves it from the slug.

## Reads

`getBehaviourCatalog` feeds the grid in one round trip. It reuses `listProjectBehavioursUseCase` per behavior so a card's counts match the tree it links to, but deliberately not `getProjectBehaviours` — that read fans out a ClickHouse intelligence aggregate per node and a PCA over every centroid, neither of which a teaser has any use for.

## Billing

Audited every AI call in behaviors against the metering system from #3915. Two were running free:

- **Facet cluster naming.** `nameTaxonomyClusterActivity` metered its topic and custom-behavior branches but not its facet branch, which routes to the same `nameClusterCore` and makes the same two generations per cluster. Fixed by giving that branch the same scope and repositories as its siblings.
- **Facet projection extraction.** One generation per sampled session, up to the proposal sample cap, on a 6-hourly sweep — the taxonomy's heaviest AI spend, and metered nowhere. Now runs under a `taxonomy-facet-extract` activity scope.

Extraction runs `FACET_EXTRACTION_CONCURRENCY` calls at a time while metering keys come from a counter, so a retry can map a key to a different call than the first attempt did. That direction is safe: a reused key dedupes to "already charged", so a retry undercharges the unflushed tail rather than double-charging it, and anything already cached in `taxonomy_facet_projections` is skipped instead of re-extracted. Documented in `dev-docs/billing.md`.

**Retry policy.** `defaultActivityRetryPolicy` marks `BadRequestError` non-retryable, which is what the credit gate raises when an organization is out of credits. The garden workflow was *replacing* that list with `["TaxonomyGardeningPlanMissingError"]` instead of extending it, so a credit-blocked run would have burned eight attempts over roughly an hour and a half against a balance that cannot move in that window. It now concatenates both. The workflow's existing catch already marks the run and behavior failed before rethrowing, so a blocked run finalizes and the next sweep retries it.

Embeddings are unaffected: every embed in this path is `inputType: "document"`, which the metering layer never charges — they ride on the generation that produced them.

## Flag and the legacy screen

Everything stays behind the existing `customBehaviors` flag; no new flag. With it off, `/behaviours` renders `LegacyBehavioursPage`, the pre-facets screen, unchanged. The flag is resolved in the index route's loader rather than in the component, so the legacy screen never paints first. The behavior and view routes carry no flag guard — reaching them directly is fine.

## Where to focus review

- `behaviour-scope.ts` — the one place `/behaviours/:slug[/views/:view]` is resolved against the project's behaviors, including the `lat-topics` sentinel and the flat-slug redirect.
- The two billing changes in `taxonomy-naming-activities.ts` and `taxonomy-gardening-activities.ts`, plus the retry-policy line in `taxonomy-gardening-workflow.ts`.
- `getBehaviourCatalog` — whether reusing the tree use-case per behavior at concurrency 4 is the right trade against a purpose-built read.

## Known gaps

- A credit-blocked behavior shows "Waiting for the next run" rather than explaining that the organization is out of credits. The waiting state deliberately never shows a red failure, so surfacing this needs a distinct state.
- `regenerate-showcase-workflow.ts` and `seed-demo-project-workflow.ts` replace `nonRetryableErrorTypes` the same way the garden workflow did. Neither is metered today, so nothing burns retries there yet; left alone as out of scope.
- Facet economics on the free plan are worth a decision before the flag widens: at the 10-behavior cap, a cold-start round is roughly 15k of a free organization's 20k monthly credits, and steady state costs about 1 credit per new session per behavior on top of the 1 credit for the trace.

## Verification

- `pnpm typecheck` green on `@app/web`, `@app/workflows`, `@domain/taxonomy`.
- Tests: `@app/web` 714 pass, `@domain/taxonomy` 253 pass, `@app/workflows` 81 pass.
- Biome clean on every changed file.
- ClickHouse repo changes are type-checked only; the chdb integration suite was not run in this worktree.


andresgutgon:
## Reshape (second commit) — one concept: **Lens**

After design review, the UX was reshaped around a single new concept. Reviewers should look at the branch as-of this commit, not the first one.

**Model.** A **Lens** is the only new user-facing noun (Global topic is the built-in default lens). Everything else is a **view = (lens × filter)**, gardened. Concretely:
- **Garden a lens** by creating a custom behavior with an *empty filter* + that facet — its whole-project view, which auto-gardens and is the tree you review. This reuses the entire existing create→generate→sweep→status pipeline, so there's **no** standalone facet-gardening path, **no** nullable-`customBehaviorId`, and **no** facet-status migration.
- A **filtered view** is just a custom behavior on the same lens. Filters drive gardening (a filter re-clusters the subset — that's the discovery, not a live rollup).
- "Behavior" is retired as a user-facing noun (the section keeps its name).

**What changed vs. the first commit:**
- Added `createLens` (facet find-or-create + deduped whole-project view).
- **Removed** the n=10 extraction preview entirely — the honest preview is the gardened whole-project tree, so it was dropped (use-case, test, constants, web fn, hook, and the 2-step grid modal all deleted).
- Filter-edit correctness: `updateCustomBehaviorFn` now purges the view's ClickHouse assignment slice + re-gardens when the filter changes, so stale edges from the old cohort don't leak (projections reused — a re-cluster, not a re-extraction).
- Web: **New lens** modal (preset gallery + fork/scratch), contextual **"Filter this view"** (replaces "New behavior"; lens implied by the current scope), and a **lens-grouped scope selector** (Global topic + a group per lens, filtered views nested).

**Kept from the first commit (still the foundation):** the preset catalog, the `lens` discriminated union + `resolveFacetLens`, the `createFacet` `lat-`/cap cleanup, and the `facetId` read-path threading through both ClickHouse repos + the PG cluster scope (what makes a facet view render a real tree with the standard drawer rollups).

**Feature flag:** Behaviours is unchanged without `customBehaviors` — all the new UI lives inside the scope header, which still returns `null` when the flag is off.

**Verification:** full `pnpm typecheck` green (88/88), `@domain/taxonomy` 251 tests pass, `knip`/Biome clean. No new migrations.

apps/web/src/domains/taxonomy/custom-behaviors.functions.ts
```diff
@@ -121,15 +140,42 @@ export const updateCustomBehaviorFn = createServerFn({ method: "POST" })
...
+        const result = yield* updateCustomBehavior({
+          id: CustomBehaviorId(data.id),
+          ...(data.name !== undefined ? { name: data.name } : {}),
+          ...(data.filterSet !== undefined ? { filterSet: data.filterSet } : {}),
+        })
+        // Changing which sessions a view scopes leaves the old cohort's edges in
+        // the assignment slice (ReplacingMergeTree never deletes), so purge the
+        // slice and re-garden from scratch over the new cohort. Projections are
+        // reused (same facet), so this is a re-cluster, not a re-extraction.
+        const filterChanged = before !== null && stableStringify(before.filterSet) !== stableStringify(result.filterSet)
```
andresgutgon: This has to be in the domain should be in the domain. The web is just one user of this

apps/web/src/routes/_authenticated/projects/$projectSlug/behaviours/-components/behavior-authoring.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+// Models the quality of a built-in preset so users aim for a real instruction,
+// not a keyword.
+export const BEHAVIOR_INSTRUCTIONS_PLACEHOLDER =
+  "e.g. Identify the main reason the user reached out: the underlying goal, not the surface request. State it as a single short phrase in the user's own terms, and treat it as unclear if the conversation doesn't make it obvious."
+
+export type BehaviorDraftErrors = Partial<Record<keyof BehaviorDraft, string>>
+
+/** Validate a draft into a `newFacet` input, mapping Zod issues back to per-field errors. */
+export function parseBehaviorDraft(
```
andresgutgon: Is this the right approach? Is not zod schema validated in the function? Use existing `useForm` and related patterns to let validation flow erros flow from backend

apps/web/src/routes/_authenticated/projects/$projectSlug/behaviours/-components/behaviour-form-modal.tsx
```diff
@@ -74,41 +147,76 @@ export function BehaviourFormModal({
   return (
     <Modal.Root open onOpenChange={(next) => (next || isSaving ? undefined : onClose())}>
       <Modal.Content size="full" height="screen" dismissible>
```
andresgutgon: Use compact modal version not Modal.Content and Modal.Root

apps/web/src/routes/_authenticated/projects/$projectSlug/behaviours/-components/behaviour-route-states.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Loader2Icon } from "lucide-react"
+import { ListingLayout as Layout } from "../../../../../../layouts/ListingLayout/index.tsx"
+
+/** Waiting on the behaviors list to resolve which behavior (or view) the URL names. */
+export function BehaviourRouteLoading() {
+  return (
+    <Layout>
+      <Layout.Content>
+        <div className="flex flex-1 items-center justify-center p-8">
+          <Loader2Icon className="h-6 w-6 animate-spin text-muted-foreground" />
```
andresgutgon: Use `<Icon icon={Loader2Icon}...`

apps/web/src/domains/taxonomy/behaviour-catalog.functions.ts
```diff
@@ -0,0 +1,168 @@
...
+        )
+        const scopes = [
+          {
+            slug: TOPICS_BEHAVIOR_SLUG,
+            name: TOPICS_BEHAVIOR_NAME,
+            description: TOPICS_BEHAVIOR_DESCRIPTION,
+            status: "ready" as CustomBehaviorStatus,
+            customBehaviorId: null,
+            facetId: null,
+            viewCount: viewCountFor(null),
```
andresgutgon: What happen with the views that are created from global topics

apps/web/src/routes/_authenticated/projects/$projectSlug/users/$userId/-components/user-behaviours-section.tsx
```diff
@@ -37,8 +38,8 @@ export function UserBehavioursSection({
       {behaviours.map((behaviour) => (
         <Link
           key={behaviour.clusterId}
-          to="/projects/$projectSlug/behaviours"
-          params={{ projectSlug }}
+          to="/projects/$projectSlug/behaviours/$behaviourSlug"
+          params={{ projectSlug, behaviourSlug: TOPICS_BEHAVIOR_SLUG }}
```
andresgutgon: Why is this `TOPICS_BEHAVIOR_SLUG` harcoded here?

andresgutgon: Maybe link to the home as before

andresgutgon:
@claude review

PR #4282, opened 2026-07-28
feat(onboarding): ask how users heard about us

### LINKED ISSUE #4277: Add "Where did you hear from us" question on the onborading
Image

Above the "phone" field in onboarding, add a "How did you hear about us?" selector (non optional) with these options (in this order pls):

- Recommended by someone
- Google or search
- ChatGPT or other AI
- Reddit
- GitHub
- X or LinkedIn
- YouTube or podcast
- Other

When "Other" is selected, show a free-text field so the user can type the source. Keep it optional so it never blocks completing onboarding.

paulacavero:
Closes #4277

## What

Adds a required **"How did you hear about us?"** select directly above the phone field on the onboarding step, with the options in the order requested:

Recommended by someone · Google or search · ChatGPT or other AI · Reddit · GitHub · X or LinkedIn · YouTube or podcast · Other

Picking **Other** reveals a free-text input, which is **also required** — so the answer is never just "Other" with no detail.

## One column, not two

The typed source is stored in `heard_about_us` **in place of** the `other` sentinel, so `other` itself is never persisted. There is no companion column.

The trade-off, worth knowing before writing queries: `heard_about_us` is no longer a fixed set of values, so counting the "Other" bucket means excluding the seven known slugs rather than matching one. Free text that happens to read `reddit` is also indistinguishable from the Reddit option.

## Notes for review

- **Slugs, not labels.** The seven real channels store as `recommendation`, `search`, `ai`, … so marketing can re-word the options without breaking historical segments. Labels and slugs live together in `packages/domain/users/src/constants.ts`.
- **Shown in the invite-claim flow too.** `/claim/$token` renders the same `RoleStep.Left` but built its own duplicate `useForm`. The field contract — defaults, the conditional required-field list, and the select-plus-text merge — lives in `role-step-form.ts` and both routes import it, so they can't drift apart.
- **The free text is only required while "Other" is selected.** The step submits via a "Next" button rather than a native form, so required fields are validated by name; that list is computed from the current choice, otherwise picking a normal channel would be blocked by an empty text field.
- **Loops.** Syncs as a single `heardAboutUs` property alongside `jobTitle` / `phoneNumber`. The `heardAboutUsOther` property created earlier in Loops is now unused and can be deleted there.
- **Two migrations.** The first adds `heard_about_us`; the second drops `heard_about_us_other`. The drop is separate rather than an edit to the first because the first had already been applied to a local dev database, and Drizzle tracks applied migrations by hash — editing it in place would leave existing environments inconsistent. Both columns are nullable, so no backfill and nothing to reverse. Applied automatically by the deploy pipeline's migrate step.
- `.migration-lock` is rewritten by the generator on purpose (conflict tripwire), so expect a conflict if another migration PR is open.
- **Reverts part of de10b78d1.** That commit fixed a real bug — `sanitizeString` turned the repository's explicit `null` into `undefined`, which `omitUndefined` then stripped, leaving a stale free-text source in Loops. The second column is gone, so the fix and its `sanitizeClearableString` helper go with it; that whole class of inconsistency is now structurally impossible.

## Verification

- `pnpm typecheck` — 90/90 tasks pass
- Tests pass for the affected packages (81 tests), including cases covering a typed source reaching Loops
- `biome check` and `knip` clean

**Not visually verified** — the machine this was written on has no local database, so the step has never been rendered in a browser. Worth a click-through on staging. One thing to eyeball: the shared `Select` defaults to `side="top"`, so the menu opens upward; Radix should flip it when there's no room, but with eight options mid-form it may read better as `side="bottom"`.

## Summary by CodeRabbit

* **New Features**
  * Added a “How did you hear about us?” question to onboarding (including claim flow).
  * Users can choose a predefined source or enter an “Other” response.
  * Attribution is saved to user profiles and propagated to marketing updates.
* **Bug Fixes**
  * Prevented advancing when required attribution fields are invalid or missing.
  * Cleared/avoided persisting empty “Other” responses.
* **Improvements**
  * Sent attribution through the marketing contact update pipeline (including Loops synchronization).


paulacavero:
pls check how this looks because i'm a vibecoder and I have no idea hehe

apps/web/src/routes/claim.$token.tsx
```diff
@@ -201,11 +204,21 @@ function ClaimOnboarding({
...
 
   const form = useForm({
-    defaultValues: { jobTitle: "", phoneNumber: "" } satisfies OnboardingFormValues,
+    defaultValues: EMPTY_ONBOARDING_FORM_VALUES,
     onSubmit: createFormSubmitHandler(
-      async ({ jobTitle, phoneNumber }) => {
+      async ({ jobTitle, phoneNumber, heardAboutUs, heardAboutUsOther }) => {
+        // `handleAdvanceFromRole` gates on this, so an empty channel here would
+        // mean a programming error rather than user input.
+        if (heardAboutUs === "") return
```
andresgutgon, line 215: Wait. Are we doing this mandatory?

paulacavero: yes

apps/web/src/routes/_authenticated/projects/$projectSlug/-components/onboarding/steps/role-step-form.ts
```diff
@@ -0,0 +1,31 @@
+import type { HeardAboutUs } from "@domain/users"
+
+/**
+ * Field contract for the "Tell us about yourself" onboarding step.
+ *
+ * Lives in its own module because two routes build this form independently —
+ * the project onboarding flow (`onboarding-flow.tsx`) and the invite-claim flow
```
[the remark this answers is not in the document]
paulacavero: Done in d00d97a67 — module JSDoc removed. The type is unexported now too, since nothing outside the file used it (both callers import the constants, not the type), which also clears the knip failure.

packages/platform/loops/src/client.ts
```diff
@@ -111,6 +111,8 @@ export const createLoopsContactsSender = (config: LoopsConfig | undefined): Mark
               firstName: sanitizeString(parsed.firstName),
               jobTitle: sanitizeString(parsed.jobTitle),
               phoneNumber: sanitizeString(parsed.phoneNumber),
+              heardAboutUs: sanitizeString(parsed.heardAboutUs),
+              heardAboutUsOther: sanitizeString(parsed.heardAboutUsOther),
```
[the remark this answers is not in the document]
paulacavero: Good catch — fixed in de10b78d1. Added `sanitizeClearableString`, which only omits the property when the caller passes nothing; an explicit null (or blank text) now sends `null` so Loops clears the old free-text source. Covered by a new test in `client.test.ts`.

apps/web/src/routes/claim.$token.tsx
andresgutgon: But if they choose "other" what happens?

andresgutgon (dismissed):

packages/domain/users/src/constants.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Options in the exact order they are presented in the onboarding select.
+ * Shared so the form and any backoffice reporting label a slug the same way.
+ */
+export const HEARD_ABOUT_US_OPTIONS: readonly { readonly value: HeardAboutUs; readonly label: string }[] = [
+  { value: "recommendation", label: "Recommended by someone" },
+  { value: "search", label: "Google or search" },
+  { value: "ai", label: "ChatGPT or other AI" },
+  { value: "reddit", label: "Reddit" },
+  { value: "github", label: "GitHub" },
+  { value: "social", label: "X or LinkedIn" },
```
andresgutgon: Don't you want to know which one?

andresgutgon (approved):

PR #4322, opened 2026-07-31
feat(cost): six coherent seeded project archetypes for the cost section

andresgutgon:
Closes LAT-809.

Every `tokens_cache_read` in `fixed-traces.ts`, `span-builders.ts` and `custom-behavior-qa.ts` was `0`, so the six-state cache classifier from #4317 could not be exercised at all — every row landed in `Correctly off` or `Not enough data`, and the panel looked like it worked while testing nothing.

Two commits: the archetypes, then a deduplication the archetypes exposed.

## 1. Six coherent archetypes, not a bag of edge cases

Five with a project to themselves. A project whose numbers tell a consistent story is what lets us judge whether the dashboard *reads* well; a project stuffed with every failure mode only proves the code paths execute.

| | Project | What it proves |
| -- | -- | -- |
| **A** | `cost-healthy-qa` | 3 models, 90 days, 100% priced, multi-turn, every cache row `Optimal` so **no card renders**. Proves the page does not cry wolf. |
| **B** | *default seed project* | One model per cache state, both zero-cost buckets, an 8-model long tail, mixed sessioned/sessionless. |
| **C** | `cost-single-turn-qa` | `turns_per_session = 1` everywhere **with an excellent ceiling**. See below. |
| **D** | `cost-regression-qa` | Two agents, one attributable cause each, two full 28-day periods. |
| **E** | `cost-tiny-qa` | 9 calls over 3 days, reaching the anchor so the last day is always partial. |
| **F** | `cost-free-qa` | Registry-zero-priced models: $0 total, 100% priced, zero warnings. |

### Two decisions to review

**`cache-economics-qa.ts` is folded into archetype B, not kept alongside it.** B *is* the demo project where findings fire, and that fixture already lived there — two fixtures on the same project competing for the same model rows was the accident to avoid. B keeps its six cohorts (so the assertions #4317 shipped still hold) and gains `correctlyOff`, the `stopCaching` shape, `unpriced` + pre-migration `unknown`, and the tail.

**⚠️ If you already seeded #4317, you have 466 stale `cache-qa` rows that will blend into B's models.** I hit this locally. `pnpm ch:seed --reset` fixes it.

**The 7-day freshness sentinel is dropped.** It existed so cohorts stayed inside "the cost page's default range" — but `resolveAnalyticsListRange` defaults to **All time**, so there is no drift to guard against. A plain presence sentinel also stops repeat seeds from stacking duplicate cohorts, which the freshness check permitted once a week.

### The bug archetype C exists to catch

LAT-798 defines the achievable ceiling over gaps across **the agent's whole traffic**, not within a session. C is the only fixture that can tell a correct implementation from one measuring within-session gaps. Run through LAT-798's own `lagInFrame` formula against the seeded rows, both partitions:

| agent | ceiling, partitioned by agent | ceiling, partitioned by session |
| -- | -- | -- |
| `intent-classifier` | **0.9833** | **0** |
| `rag-answerer` | **0.9583** | **0** |

Zero, because every session's only call is a first call and therefore an automatic miss. Without this fixture that bug ships silently and systematically tells our highest-volume customers their cache is unfixable.

### `Stop caching` reads `Investigate` today, and that is correct

That state and the `underusing` branch of `Investigate` are unreachable until 2b computes ceilings — the classifier only returns verdicts that hold for *every* possible ceiling. The data shapes are seeded now (`b-stop-caching` has `callsPerCluster: 1`, so its ceiling is exactly 0; `b-underusing` sits at 6% against a 0.8 ceiling) so 2b lights them up without new fixtures. Both are pinned in the test with that reasoning written down.

`b-underusing` and `b-investigate-overpaying` deliberately share **the same 6% measured rate** on models with and without a write premium — the clearest demonstration that a rate means nothing without the model's own price list.

### Calibration input for LAT-798

Archetype A caches by warm prefix, so its measured rate is `share x (n-1)/n` against a ceiling of `(n-1)/n` — the gap is the fresh suffix every real call carries:

| agent | actual | ceiling | gap |
| -- | -- | -- | -- |
| `support-agent` | 0.8708 | 0.9167 | 4.6pt |
| `ticket-triage` | 0.8280 | 0.9000 | 7.2pt |
| `thread-summarizer` | 0.8138 | 0.8750 | 6.1pt |

Under 2b's *strict* `actual < ceiling` rule all three become findings and `Optimal` becomes unreachable — which is exactly the tolerance LAT-798 has to settle. A is the fixture to calibrate against.

## 2. Seeded spans are now costed from the registry

`ModelConfig` carried its own per-1M-token prices, hand-written beside the `@domain/models` registry the dashboard actually derives break-even from. Two statements of the same fact — and they had already diverged: **the table was authored in cents per million tokens while `computeCost` read its argument as dollars, so every ambient-generated span recorded 100x its true cost.** Seed spend on the default project read ~$3k where the same tokens are worth ~$31.

The layering that motivated the duplication does not require it. `@domain/shared` sits below the registry and cannot import it — but the cost is not computed there. It is computed in `span-builders.ts`, one layer up, where `@domain/spans` (and through it `@domain/models`) is already a dependency. So `ModelConfig` keeps identity and behaviour, and rates are looked up where they are spent, via a `modelRegistryPricing` helper beside the `modelCacheBreakEvenRate` that already wrapped `getCostSpec`. **No new workspace dependency edge**, so the lockfile is not re-resolved and the 7-day `minimumReleaseAge` gate is not in play.

Beyond deleting the second price list:

- The unit bug becomes inexpressible — `MICROCENTS_PER_USD` is named at the one place a rate is read, mirroring the ingest path's own constant.
- **`unpriced` is now derived from a failed lookup**, the way `estimateCostFromTokens` derives it for real traffic, instead of being asserted by a fixture. Archetype B's gateway cohort states no `costSource` at all.
- Reasoning tokens bill at their own rate falling back to output, matching `computeTokenCost`, instead of being folded into the output rate.
- **`claude-3-5-haiku` has aged out of models.dev**, so registry-costing it turned a seventh of ambient traffic silently unpriced — which would have polluted archetype B's coverage story with accidental noise. Repointed to `claude-sonnet-5`, and the fixtures now assert that **exactly one** cohort resolves to unpriced, so the next retirement fails loudly instead of quietly.
- `cost-archetypes/models.ts` loses its hand-copied price block, and the test asserting "our prices agree with the registry" deletes itself — agreement is structural now.

`fixed-traces.ts` is untouched: it hand-builds `SpanRow` literals with explicit cost columns and never used `MODELS`. Its own hardcoded rates are a separate, pre-existing inconsistency.

## Verification

Beyond the 56 unit tests, seeded into the real dev ClickHouse and read back through the dashboard's own SQL.

**Archetype spend is byte-identical across commit 2** — the hand-copied rates were correct — which is what makes it a deduplication rather than a repricing. Ambient spend on the default project moved $3k → $31.34, and no ambient model resolves to unpriced.

- **`cost_source` coverage** across the set: `provider_reported` (880 calls, A), `estimated`, `unpriced` (30, B's gateway — derived), `unknown` (26 from `cost_source = ''`, confirming the SQL mirror classifies it), `no_tokens`.
- **A**: $39.21, 100% priced, 21.6% verified share. **F**: $0.00, 100% priced, all `estimated`.
- **B's cache table**: `claude-opus-4-5` 0.837 `Optimal` · `gpt-5.6` 0.05 · `claude-haiku-4-5` 0.06 `Investigate` · `gpt-5.4-mini` 0.06 `Optimal` · `gpt-5-mini` 0 @ 26k prompt `Cache it` · `gpt-5-nano` 0 @ 640 prompt `Correctly off` · `gemini-2.5-flash-lite` 8 calls `Not enough data`.
- **D**: router spend 2.12 → 5.78 with prompt held at 7000 and premium share 9.1% → 63.6%; grader 1.53 → 4.29 with model and turn count flat and prompt 4000 → 16000. Attribution is separable, not merely non-zero.
- **Break-even variety**: Anthropic-style 21.7% beside OpenAI-style 0%, including two OpenAI models on opposite sides — so a blended project figure is visibly meaningless.

Green: archetype + span-builder suites 56/56, full ClickHouse suite 40 files / 570 tests, `@domain/spans` 1357, `@domain/shared` 168, typecheck + format + knip clean.

## Summary by CodeRabbit

* **New Features**
  * Added realistic cost-analysis sample projects covering healthy, single-turn, regression, tiny, and free usage patterns.
  * Added scenarios for cache behavior, missing costs, model mix, spend changes, and long-tail traffic.
  * Added free-tier and unpriced usage examples for zero-cost analysis.
* **Improvements**
  * Improved cost and cache pricing consistency across sample data.
  * Updated seeded assistant model configuration.
* **Tests**
  * Added comprehensive validation for generated cost and usage scenarios.


packages/domain/shared/src/seed-content/agents.ts
```diff
@@ -82,8 +82,8 @@ function requiredAt<T>(items: readonly T[], index: number, name: string): T {
 const openAiMiniModel = requiredModel(openaiModels, (model) => model.model === "gpt-4o-mini", "gpt-4o-mini")
 const knowledgeAssistantModel = requiredModel(
   anthropicModels,
-  (model) => model.model === "claude-3-5-haiku",
-  "claude-3-5-haiku",
```
andresgutgon, line 86: This model is no longer present in models.dev

PR #4348, opened 2026-08-04
feat(cost): price the usage modifiers providers already report

andresgutgon:
## Why

Our Anthropic cost estimate undercounts, only ever in one direction, because we price from tokens while the bill carries multipliers tokens cannot express. Every one of them is reported by the API in the response `usage` object, and our Claude Code telemetry discards them.

Three land here:

| Modifier | Reported as | Effect |
|---|---|---|
| Prompt-cache TTL split | `usage.cache_creation.ephemeral_5m_input_tokens` / `ephemeral_1h_input_tokens` | 1h writes bill at 2x base input; we always priced the 5m 1.25x rate. The largest confirmed gap — up to ~30% of a project's bill. |
| Speed | `usage.speed` | Opus 5 / Opus 4.8 fast mode bills 2x base on input *and* output. Zero production spans capture it today. |
| Inference geo | `usage.inference_geo` | `us` is 1.1x on every token category. Orgs that opted out of global routing were auto-migrated to `us`, so nothing in the *request* says so. |

They compose. Fast mode replaces the base rate and the cache factors apply on top of it, so a cache read on fast Opus 5 is 0.1 × \$10, not 0.1 × \$5.

## What

Four layers, all additive.

**Cost — `@domain/models`.** New `cost-multipliers.ts`: hand-sourced rule tables (models.dev carries no speed/tier/TTL key at all) plus `estimateModifiedCost`, a chain where **every factor defaults to 1.0 when its modifier is absent**. That is what makes this backward compatible structurally rather than by special-casing — existing spans and every exporter that never emits these fields price exactly as before, with no backfill.

The TTL premium is expressed against the **catalog** cache-write rate, not base input. models.dev's `cache_write` is 1.25× input on every Anthropic row, i.e. the published 5m price, so 1h is 2 / 1.25 = **1.6×** the catalog rate — which lands on the published per-model dollar figure exactly (\$6.25→\$10, \$2.50→\$4, \$12.50→\$20, \$1.25→\$2). Expressing it against base input would re-derive a number the catalog already carries and would silently reprice the 5m majority.

**Ingest — `@domain/spans`.** New `resolvers/usage/modifiers.ts` reading `latitude.usage.*` candidate keys (no OTEL GenAI attribute exists for any of these — verified against the semconv registry — so `latitude.*` is the cross-exporter contract rather than squatting `gen_ai.*`), plus the two OpenAI service-tier attributes that already exist. `speed` and `service_tier` normalize onto one tier field.

**Store — `@platform/db-clickhouse`.** Migration `00056`: `tokens_cache_create_by_ttl Map(UInt32, UInt32)` keyed by seconds, plus `service_tier` / `inference_geo` scalars. Testkit `schema.sql` regenerated from the applied migration.

**Client — `packages/telemetry/claude-code`.** `Usage` widened to keep the nested `cache_creation` split, `speed`, and `inference_geo`; emitted as one flat int per observed TTL (`latitude.usage.cache_creation.ttl.300` = 148) so the values land in `attr_int` and production is queryable before the migration reaches it.

## Where review should focus

- **`tokens_cache_create` stays the authoritative total.** `tokens_total` is a MATERIALIZED column summing it, so the split is a *subset* stored alongside — a sibling column that also summed into the total would double-count and desync `traces_mv` / `sessions_mv`. Neither view is touched. There's a test asserting `tokens_total` = 348 on a span whose split is `{300: 148, 3600: 100}` of a 248 total.
- **TTL keyed by seconds, never by tier name.** A `cache_1h` column would bake one provider's one tier into the schema; OpenAI extended retention runs to 24h and Gemini takes an arbitrary TTL.
- **The 1.6× derivation** and the decision to key multipliers on the *catalog* provider/model pair rather than the reported one, so a multiplier always matches the rate it scales.
- **What is deliberately absent, and whether you agree.** Anthropic's priority tier and OpenAI's `flex`/`priority` are real premiums with no factor that holds across models, so they price at the catalog rate. That understates them — but a guessed factor would put a confident wrong number on every span carrying the tier with no way to tell afterwards. Likewise `inferenceGeoMultiplier` has **no model-version gate** despite the option only existing on Claude 4.6+: a model that cannot route this way never reports the field, so absence already gates it, while a hand-maintained version list would omit each new family on the day it ships.
- `promptCacheTtlSeconds` is untouched (it answers the *default* lifetime and feeds `judge-cache-economics` + a CH query). `purchasablePromptCacheTtlSeconds` is the new sibling.
- Anthropic has **no** long-context premium and none was added.

## Tests

- `cost-multipliers.test.ts` (18 cases): byte-identical output for the all-absent case *and* for explicitly-standard modifiers, mixed TTL, all-1h against the published rate, unsplit remainder, split-disagrees-with-scalar, fast × cache composition, geo × fast × mixed-TTL.
- `modifiers.test.ts` (15 cases): attribute resolution and end-to-end `resolveUsage` pricing, including that a provider-reported cost is left alone while the modifiers are still recorded.
- `otlp.test.ts`: emitter keys, omitted empty buckets, and that a span with no modifiers emits no `latitude.usage.*` at all.
- `span-repository.test.ts`: CH round-trip of the Map + scalars, `tokens_total` untouched, and empty reads on a row that never set them.

Full workspace typecheck 90/90. `@domain/spans` 1967, `@domain/models` 196, `@domain/destinations` 143, claude-code telemetry 81, full ClickHouse suite 590 — all pass. knip and Biome clean.

Not run: no live end-to-end against real fast-mode / 1h-TTL traffic, since nothing in production emits these attributes yet — the change is what makes them capturable.

## Summary by CodeRabbit

- **New Features**
  - Improved usage-based cost estimates using service tier, inference geography, and cache-write duration.
  - Usage records now capture cache creation tokens by retention period and pricing metadata.
  - Claude Code telemetry reports cache, speed-tier, and inference-location details when available.
  - Added storage and retrieval support for the new usage fields.

- **Tests**
  - Expanded coverage for pricing adjustments, metadata resolution, cache-duration tracking, and persistence behavior.


andresgutgon:
@geclos some client complained about claude code not counting correctly the cost. This PR add 3 modifiers that could influence the cost

1. TTL
2. Speed mode 
3. Service Tier

Honestly not sure if it's worth it to fix this. I'll leave it open until you come back.

geclos:
I don't think this is worth fixing given the complexity and edge case

PR #4364, opened 2026-08-05
feat(cost): wasted spend — what this window spent on traces that errored

andresgutgon:
Closes LAT-804 (flavour 1 of three; see **Scope** below).

The one figure on the cost page that names money which bought **nothing**. Everything else on the page names money that bought something you might have bought cheaper.

## The decision, stated

**A trace that errored is wasted in whole, not per failed span.**

Per-span sounds more honest but is the *less* accurate measure of waste: a failed call usually records no usage at all — the provider rejected it, or it never returned tokens. The money in a failed trace was spent on the steps that **succeeded** and whose output was then thrown away. Per-span would report ~\$0 for exactly the traces that wasted the most.

This is in the headline tooltip, and **no per-span framing of the same claim appears anywhere on the page** — the two would not reconcile, and this section has a hard rule against numbers that do not add up.

Errored means at least one span with `status_code = 2` — the same definition `if(error_count > 0, 'error', 'success')` gives the traces list's Status filter, so the panel's drill-down returns the very traces it counted. The alternative (root span only) would be a tighter definition that disagrees with the list it links to.

## Shape

One dollars-wasted headline with a path to the evidence, not a wide table (the LAT-795 lesson).

- **\$X wasted** via the shared `rollupCostDisplay`, so a zero reads as `-` with its unpriced note rather than as free.
- **N of M traces with usage errored** — same denominator as every other per-trace figure on the page.
- **Y% of spend**, gated behind `WASTED_SPEND_MIN_SAMPLE_TRACES = 20`. Dollars are a sum and shown at any volume; the *ratio* is withheld below the floor with a tooltip saying why. This is the LAT-795 `278x avg` lesson: a ratio needs a sample floor, not just a correct denominator.
- **Failure reasons ranked by cost**, each a bar of its share of the wasted total.
- **View errored traces** → the traces list, pre-filtered to Status = error over the same window.

**Arithmetic closes.** Each errored trace is attributed to its *first* failed span's `error_type` (`argMinIf`), so the reasons partition the errored traces: their costs sum to the headline exactly, and the truncation remainder is `total − listed` rather than a decorative "and others". A trace counted under every type it hit could not do that.

Placed directly under the spend chart, above Session — it is the strongest message on the page.

## Should this be a signal instead?

Raised before building, as asked. My answer is **split by flavour**:

- **Errored-trace spend (this PR) is a panel.** It is an accounting fact about a window, not a finding. There is nothing to validate or resolve — you do not "resolve" that 12% of last week's spend errored, you fix the errors, and the errors already have their own pipeline. As a signal it would never resolve and would re-fire every window.
- **Retry loops are a signal, not a dashboard row.** A specific agent looping on a specific tool has a beginning, an end, and something to annotate and resolve — exactly the argument LAT-822 is making for cache findings. **So I did not build it here**, and recommend it go to the signal/issue pipeline rather than becoming a row nobody revisits.

## Scope

- ✅ **Errored traces that still cost money** — this PR.
- ⏭️ **Retry loops** — assessed, recommended as a signal (above), not built.
- ⏭️ **Abandoned sessions** — out of scope, needs a definition that overlaps LAT-806.
- No public API operations (LAT-811 stays delayed).

## Seeds

Reuses the LAT-809 archetypes rather than parallel fixtures. `CostCohort` gains optional `callsPerTrace` + `failure` — both default off, so every existing cohort builds byte-identically.

Two failing cohorts on **archetype B**:
- `b-errors-rejected` — 3-call traces, 1 in 4 fails on its last call, which **reports no usage**. This is the fixture the decision turns on: per-span sees \$0 while two paid calls were discarded. A seed test asserts the two readings visibly disagree.
- `b-errors-timeout` — 2-call traces, 1 in 5 fails having consumed tokens, so both readings charge something.

Both sit at `b-optimal`'s cache margin so they add no cache finding to a table that deliberately carries one model per state. **Archetype A asserts zero errored spans** — if a well-run project showed a wasted-spend figure, the definition would be wrong.

## Where to focus review

1. **The decision itself** — whole-trace, and the "at least one span errored" definition that comes with it. A trace that errored on one step and recovered still counts. I chose consistency with the traces list over a tighter root-span rule so the drill-down reconciles; that trade is the main thing to push back on.
2. **`WASTED_TRACE_SOURCE`** (`cost-analytics-repository.ts`) — the one query that is deliberately *not* gated to billable operations, because a trace fails on whichever span errored (usually a tool span) while cost is gated inside the aggregate. Tool-span dollars stay out; tool-span failures count.
3. **The sample floor at 20** — matches `COST_PER_CALL_MIN_SAMPLE_CALLS`, but it is a guess.
4. **Panel placement** above Session.

## Verification

- Full `@platform/db-clickhouse` suite — 42 files / 628 tests, green (includes the 8 new `getWastedSpend` cases and the reused seed archetypes).
- `@domain/spans` — 2154/2154. `@app/web` — 96 files / 899 tests.
- `typecheck` clean on all three packages; `biome` and `knip` clean.

> An earlier revision of this description flagged the full ClickHouse suite as un-runnable here (OOM, exit 137). That was vitest's default parallelism against chdb on a shared machine, not this change — `--maxWorkers=2` runs it clean.

## Rebased onto the cost dashboard UI rework (#4389)

One conflict, in `cost/index.tsx`. Upstream's reworked body was taken wholesale and the Waste section re-expressed in the new idiom; the panel itself was rewritten to the new design language rather than merely made to compile:

- `SectionHeading` → the new `Section` wrapper (`heading="Waste"`).
- `projectSlug` now from `project.slug`, not `Route.useParams()` — dropped my duplicate.
- Shared `EmptyState` (with an icon) instead of a hand-rolled centered message.
- `SplitValue` on the headline figures, so the `$` and `%` mute like every other value in the section.
- `Text.H3M` headline over a `Text.H6M` label, matching the per-session cards; `px-4 py-3` padding.
- Theme-aware bar colour via `useChartCssTheme()`, alongside `otherSeriesColor` for the remainder — the fixed `hsl()` literals are gone.
- `ChartHeader` with `showWindow={false}` + `titleColor="foregroundMuted"`, as its siblings now do.
- Drill-down restyled to the section's `ArrowUpRightIcon` + `group-hover:text-primary` affordance. Kept as a `<Link>` rather than adopting `useGoToModelSessions`' button+`useNavigate`, since it targets a different page and the web-frontend skill wants a real anchor there.

The share and its sample floor, the whole-trace decision, and the reason-rows-sum-to-headline invariant are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added a Waste section to project cost analytics.
  * Shows total wasted spend, its share of overall spend, and breakdowns by error reason.
  * Includes trace counts, proportional cost indicators, “Other reasons,” and links to filtered errored traces.
  * Supports loading and empty states across selected time ranges.

* **Bug Fixes**
  * Improved attribution to the earliest error in each trace.
  * Excludes non-billable tool-span costs from wasted-spend totals.

* **Tests**
  * Added coverage for calculations, thresholds, pricing availability, and zero-failure scenarios.

### DIFF
--- a/apps/web/src/domains/cost/cost.collection.ts
+++ b/apps/web/src/domains/cost/cost.collection.ts
@@ -8,6 +8,7 @@ import {
   getCostPerSessionDecomposition,
   getCostSeries,
   getModelUsageSeries,
+  getWastedSpend,
 } from "./cost.functions.ts"
 
 /** Time window shared by every cost query — lower bound inclusive, upper bound exclusive. */
@@ -98,6 +99,25 @@ export function useCacheEconomics({
   return useQuery({
     queryKey: [...projectScopeKey(scope), "cache-economics", projectId, range],
     queryFn: () => getCacheEconomics({ data: { ...projectScopeData(scope), projectId, ...range } }),
+    staleTime: COST_STALE_TIME_MS,
+    placeholderData: keepPreviousData,
+    enabled: enabled && projectId.length > 0,
+  })
+}
+
+export function useWastedSpend({
+  projectId,
+  range,
+  enabled = true,
+}: {
+  readonly projectId: string
+  readonly range: CostTimeRange
+  readonly enabled?: boolean
+}) {
+  const scope = useProjectScope()
+  return useQuery({
+    queryKey: [...projectScopeKey(scope), "wasted-spend", projectId, range],
+    queryFn: () => getWastedSpend({ data: { ...projectScopeData(scope), projectId, ...range } }),
     staleTime: COST_STALE_TIME_MS,
     placeholderData: keepPreviousData,
     enabled: enabled && projectId.length > 0,
--- a/apps/web/src/domains/cost/cost.functions.ts
+++ b/apps/web/src/domains/cost/cost.functions.ts
@@ -8,6 +8,7 @@ import type {
   JudgedCacheModel,
   ModelUsageMeasures,
   ModelUsageSlice,
+  WastedSpend,
 } from "@domain/spans"
 import {
   COST_BREAKDOWN_DIMENSIONS,
@@ -18,6 +19,7 @@ import {
   judgeCacheEconomics,
   sessionCostTokens,
   summarizeUnpricedUsage,
+  wastedSpendShare,
 } from "@domain/spans"
 import { CostAnalyticsRepositoryLive } from "@platform/db-clickhouse"
 import { withTracing } from "@repo/observability"
@@ -116,6 +118,19 @@ export interface CostPerSessionRecord extends CostPerSessionDecomposition {
   readonly comparedToIso: string
   /** Sparkline points for the two headline blocks, spanning both windows, oldest first. */
   readonly buckets: readonly SessionCostSparkPoint[]
+}
+
+/** Every listed reason's spend, and the one row that holds whatever the list left off. */
+export interface WastedSpendOtherReasons {
+  readonly typeCount: number
+  readonly traces: number
+  readonly costMicrocents: number
+}
+
+export interface WastedSpendRecord extends WastedSpend {
+  /** Null where the window cannot support a share — the dollar figure is shown regardless. */
+  readonly wastedShare: number | null
+  readonly otherReasons: WastedSpendOtherReasons | null
 }
 
 /** Cost per session is derived here rather than in the panel: an empty bucket has none. */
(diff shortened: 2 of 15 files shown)


andresgutgon:
Not sure this is worth it. Closing for now

PR #4372, opened 2026-08-06
feat(taxonomy): offline blind static-vs-adaptive quality review generator

andresgutgon:
## Why

The shadow experiment gives us shape telemetry, and shape telemetry cannot answer the question we actually need answered. On a project where static and adaptive produce the same number of behaviours but a low `@taxonomy.shadow.diff.partitionAri`, the metric says *the two trees disagree* — not *which one groups the conversations better*. That call needs a human.

This is the generator for the artifact that human reads: both trees, side by side, labels hidden, real conversations underneath.

## What this contains

The script and nothing else. Embeddings, transcripts and the rendered report all stay in a local dump directory outside the repo (default `~/Desktop/taxonomy-blind-review`), and the file header documents the ClickHouse pull that produces them. **No customer data is in this diff.**

Three steps, each resumable:

```
pnpm --filter @app/workers exec tsx scripts/taxonomy/blind-review.ts plan   [dumpDir]
pnpm --filter @app/workers exec tsx scripts/taxonomy/blind-review.ts name   [dumpDir]
pnpm --filter @app/workers exec tsx scripts/taxonomy/blind-review.ts render [dumpDir] [templatePath]
```

## Where review should focus

**1. `buildBoth` replays `runTaxonomyClusterBuild`, not the calibration harness.** The harness's `adaptiveBuild` omits the `escalation` block, so it has not been exercising the shipped adaptive path since the near-gate fix. Worth confirming the constant list here matches `build-hierarchical-taxonomy.ts:166-197` exactly — that is the whole fidelity claim.

**2. It compares LEAF partitions, not root children.** This surprised me and it is the main judgement call in the diff. On every project I sampled, the two builders agreed on the root children (`rootChildARI` 0.92–1.00) and diverged only on which branches to split further. A top-level comparison would have shown the reviewer two identical columns. Leaves are also what `partitionAri` actually scores, so this granularity matches the telemetry the decision is based on.

**3. Both trees are named fresh, in the same run.** If one arm kept production's existing names and the other got new ones, a reviewer could de-blind on writing style without reading a single conversation. Naming reproduces `name-taxonomy.ts` — same two-call map/reduce, same verbatim policy strings, same FPS sampling, same model, deepest-first — so the only asymmetry left between the columns is the clustering itself.

**4. Two deviations from production, both applied identically to both arms**, called out in comments where they occur:
- Real agent transcripts can contain literal tool-call tags, which pull the namer into emitting a tool call instead of JSON. `defang` escapes the brackets.
- A retried naming call escalates temperature. At a fixed 0.2 a transcript that derails the namer derails it identically every time, so a same-temperature retry is not a retry.

**5. Example selection prefers the farthest-point sample over pure nearest-to-centroid.** Nearest-to-centroid picks five mutual neighbours, and in projects that run the same scheduled job repeatedly those are near-identical — the reviewer would see one quote five times. FPS is seeded at the closest-to-centroid member, so the most typical example still comes first, and it is exactly the sample the namer saw.

## Verification

- `pnpm --filter @app/workers typecheck` clean; `biome check` clean. Note `scripts/` is outside every `tsconfig` include, same as the existing scripts in this folder.
- Ran end-to-end against a real 5-project dump pulled from production ClickHouse and confirmed the script reproduces a report built by hand from the same data, byte for byte.

## Findings from the first run (report is local, not attached)

- Two of the five projects that Datadog reported as low-ARI rebuilt **identically** on today's window (`leafARI` 1.000). The low-ARI condition is a property of a window, not of a project, so any sample for this review has to be re-checked at build time rather than trusted from the dashboard.
- One project could not be named at all: its transcripts are agent sessions that defeat the namer even with the markup escaped. That is a production naming-robustness gap, not a harness problem, and deserves its own issue.

## Summary by CodeRabbit

* **New Features**
  * Added a resumable workflow for comparing taxonomy clustering approaches using production samples.
  * Added statistical comparison metrics, including partition agreement and silhouette analysis.
  * Added blinded taxonomy naming with validation and retry handling.
  * Added HTML report generation with cleaned transcript examples and prominent warning labels.
  * Added command-line steps for planning, naming, and rendering, with persisted progress for resuming interrupted runs.

### DIFF
--- a/scripts/taxonomy/blind-review.ts
+++ b/scripts/taxonomy/blind-review.ts
@@ -1,0 +1,605 @@
+#!/usr/bin/env tsx
+/**
+ * Build a local, blind, human-reviewable static-vs-adaptive taxonomy comparison.
+ *
+ * This is the "(a)" quality check behind the shadow-experiment decision. Shape
+ * telemetry cannot rank two clusterings that carve the same conversations
+ * differently — `@taxonomy.shadow.diff.partitionAri` says they disagree, not
+ * which one is right. So run both SHIPPED builders over one project's real
+ * production sample, name both trees with the identical production naming
+ * procedure, and put them side by side with the labels hidden.
+ *
+ * Granularity is the LEAF partition, because that is what `partitionAri` scores
+ * (shadow-comparison.ts) and it is where the two builders actually disagree: on
+ * the projects sampled so far they agree on the root children and differ only
+ * about which branches deserve to be split further.
+ *
+ * BLINDING INTEGRITY: both trees are named fresh in the same run, same prompts,
+ * same model, same sampling. Reusing production's existing names for one arm
+ * would let a reviewer de-blind on writing style alone.
+ *
+ * The dump and the report contain REAL CUSTOMER CONVERSATIONS. Keep both outside
+ * the repo (default `~/Desktop/taxonomy-blind-review`); never commit or upload
+ * them. This script holds no data.
+ *
+ * Four steps, each resumable:
+ *
+ *   plan    cluster both arms, score them, choose samples   -> plan.json
+ *   name    name every node of both trees via Bedrock       -> names.json
+ *   render  fill the blind template's DATA array            -> report.html
+ *
+ *   pnpm --filter @app/workers exec tsx scripts/taxonomy/blind-review.ts <step> [dumpDir]
+ *
+ * `plan` expects, per project `<short>` listed in `<dumpDir>/projects.json`:
+ *   <short>.f32        row-major float32, L2-normalized, 2048 dims
+ *   <short>.meta.json  { rows, dims, observations: [{ observationId, startTime }] }
+ *   texts.json         { [observationId]: transcript }
+ *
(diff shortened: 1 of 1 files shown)


andresgutgon:
Closing — parked, not rejected. This is the generator for the blind static-vs-adaptive review artifact (the "(a)" quality check).

The check itself came back **inconclusive**: on the sampled projects, static and adaptive agree at the top level and diverge only at the leaves, and both projects that made it into the report happened to be non-English, so the reviewer couldn't judge grouping coherence from the conversations. Given that — and that the enforce-or-not decision rests on the collapse/flicker rescues (root-level), not this leaf-level tie-breaker — we decided not to pursue a bigger blind study, so this tooling has no live use right now.

It's data-free (embeddings, transcripts and the rendered report all stay in a local dir outside the repo) and it stays on this branch, so if a real reuse comes up later — e.g. eyeballing before/after groupings to validate the dimensionality-reduction work (LAT-830) or the K-selection stability fix (LAT-842) — we can resurrect it in minutes. Not merging it into `development` on spec.

PR #4411, opened 2026-08-11
fix(taxonomy): clip a facet lens's time range to the coverage it has (LAT-862)

andresgutgon:
Closes part 1 of [LAT-862](https://linear.app/latitude/issue/LAT-862/facet-lens-session-counts-do-not-grow-with-the-selected-time-range).

image

## The bug

A customer selected a four-month range on a lens view and the session counts did not move. They were right, and the picker was misleading rather than merely limited.

A facet lens's membership lives in `taxonomy_view_assignments` and only ever covers the windows gardening actually wrote. Facet plans are forced to `mode: "off"`, so they never take the full-window reassignment that keeps the global tree — and a cohort view's topic slice — covering whole project history. Widening the range filters a set that never contained the older sessions.

The worse consequence is the trend chart. On the reporting project, real volume across Aug 3 → Aug 5 goes 155 → 217 (1.4x) while the chart showed 35 → 217 (6x): the chart was plotting a coverage ramp as behavioural change. `First seen` read "August 3" on every row, which is when the lens was created.

This is **Part 1** — make the UI honest about what the data covers. Part 2 (projection-space reassignment, which grows coverage) is deliberately left for a separate PR; sequencing it after the `persistAdaptive` decoupling is a data-correctness requirement, not a preference.

## What this does

**Derives the covered band from the data, not from the assignment span.** Per UTC day, rows pointing at a currently-active cluster (`getAssignedCountsByDay`) are compared against the observations a pass could have clustered (`getClusterableCountsByDay`), then `resolveLensCoverage` walks back from the newest day with membership while that share holds at the lens's current rate.

Three deliberate choices there, all tested:
- **The rate test is relative, not absolute.** The gardening sample is capped at 1,500 per 7 days, so a project busier than ~214 sessions/day plateaus well below full coverage. A "must be 100%" test would clip every such lens to nothing.
- **Days with no traffic are neither covered nor missing**, so a quiet day does not break the band. Aug 8 and Aug 9 on the reporting project are 39 and 48 sessions at 100% coverage — a count-based cut would have severed the band there.
- **The newest day is always included.** It is partial by nature, like the live edge of any chart.

**The clip happens in the read, not only in the picker.** `listProjectBehavioursUseCase` intersects the requested window with the covered band before computing counts, and the intelligence rollup uses the same clipped window. So no range a client sends can label a short window's counts with a long range. The trend baseline is shortened to the covered part and divided by that, otherwise a baseline reaching past coverage averages real counts against days with no membership and reads as a spike on every row.

**`First seen` stops reporting the coverage floor as a sighting.** A cluster whose earliest member sits at the floor gets `firstSeenLabel: "unknown"`, renders as "Before <date>" with an explanation, and no longer counts as new — which also stops a fresh lens filling the *New this week* segment.

**The band is stated and enforced in the UI.** "Coverage: Aug 5 – Aug 11" sits next to the picker, the calendar disables days outside it (new `minDate`/`maxDate` on `DateRangePicker`), and presets reaching past it are dropped. Coverage is served on its own query key so it can bound the range before one is applied.

It degrades gracefully: when Part 2 lands, the band widens on its own and the clip stops binding, with no UI change.

## Not implemented on purpose

Both were raised in the issue and retracted there:
- **Clamping to `min(start_time)` of the assignment table** — advertises coverage that has already been orphaned by a rebuild. This is why `getAssignedCountsByDay` filters on the active cluster ids: an orphaned row is physically present but invisible to every other read, so counting it would claim coverage the tree cannot show.
- **Removing the temporal UI from lens views** — too aggressive; trends are valid inside the fully-covered band, which is exactly what this makes selectable.

## Where to focus review

- `resolveLensCoverage` in `packages/domain/taxonomy/src/use-cases/lens-coverage.ts` — the rate reference is the pooled assigned share over the trailing gardening window excluding the newest day. Pooling (rather than max or median) is what keeps a thin day from setting the bar; excluding the newest day is what keeps a big, barely-gardened "today" from collapsing the reference and falsely widening the band.
- **The denominator ignores a view's filter set on purpose.** A filtered view's eligible population is smaller, but the test is relative, so a uniform filter rate cancels out of the comparison. The trade-off: a filter whose match rate shifts mid-window (a model switched halfway through) skews the judgement. That buys a cheap query and no session-filter compilation on a page-load path — worth challenging if you disagree.
- **Cost**: two grouped ClickHouse counts per facet-lens page load, plus the same two inside the tree read. Both are partition-pruned day aggregates over the assignment TTL horizon (60 days). No LLM calls.
- `clipRangeToCoverage` in `use-analytics-time-window.ts` is a new generic capability on a shared hook; only the behaviours screen passes coverage today, and every other caller is unaffected (bounds absent ⇒ range untouched).
- The tree read now waits for coverage to resolve (`enabled`), and that wait counts as loading so the empty state cannot flash.

## Verification

- `pnpm typecheck` (workspace), `npx knip`, `biome check` clean.
- `@domain/taxonomy`: 266 tests pass, including new `lens-coverage.test.ts` — the reporting project's measured day-by-day numbers are a fixture, and it asserts the band comes out Aug 5 – Aug 11.
- `@platform/db-clickhouse`: the two touched repository test files pass, covering per-day grouping, orphaned-row exclusion, topic/facet slice isolation and the embedding-eligibility filter.
- `apps/web` (511 tests over `src/domains` + project routes) and `apps/workflows` (100) pass.
- Not exercised against live data — the clip's behaviour on a real lens is worth a look in staging.

## Summary by CodeRabbit

* **New Features**
  * Added coverage-aware date filtering for behavior analytics.
  * Date pickers now restrict selections, presets, and navigation to available data.
  * Added coverage details and explanatory guidance for facet-scoped behavior views.
  * Added clearer handling for behaviors with unknown first-seen dates.

* **Bug Fixes**
  * Prevented analytics ranges from extending beyond available data.
  * Improved date-range handling for missing, incomplete, or unbounded coverage.
  * Adjusted first-seen and novelty displays to avoid unsupported date claims.

### DIFF
--- a/apps/web/src/domains/projects/use-analytics-time-window.ts
+++ b/apps/web/src/domains/projects/use-analytics-time-window.ts
@@ -73,7 +73,13 @@ export function clipRangeToCoverage(input: {
   if (!Number.isFinite(coverageFromMs) || !Number.isFinite(coverageToMs)) return range
   const requestedFromMs = range.fromIso ? Date.parse(range.fromIso) : Number.NaN
   const requestedToMs = Date.parse(range.toIso)
-  const fromMs = Number.isFinite(requestedFromMs) ? Math.max(requestedFromMs, coverageFromMs) : coverageFromMs
+  // Clamped at both ends: a coverage window that ends before now (data stopped) lets a
+  // recent selection start past it, and a start outside the band would report a window
+  // the picker says is unselectable.
+  const requestedOrFromMs = Number.isFinite(requestedFromMs)
+    ? Math.max(requestedFromMs, coverageFromMs)
+    : coverageFromMs
+  const fromMs = Math.min(requestedOrFromMs, coverageToMs)
   const toMs = Math.max(Math.min(Number.isFinite(requestedToMs) ? requestedToMs : coverageToMs, coverageToMs), fromMs)
   return { fromIso: new Date(fromMs).toISOString(), toIso: new Date(toMs).toISOString() }
 }
--- a/apps/web/src/domains/projects/use-analytics-time-window.ts
+++ b/apps/web/src/domains/projects/use-analytics-time-window.ts
@@ -57,6 +57,28 @@ export function resolveAnalyticsListRange(input: {
 }
 
 /**
+ * Clips a list range into the range the screen's data actually covers. Absent coverage bounds are
+ * open, so a screen without a coverage constraint keeps its range untouched, and an "All time" read
+ * on a screen with one comes back as exactly the covered band rather than an unbounded scan whose
+ * counts would be labelled with a wider range than the data spans.
+ */
+export function clipRangeToCoverage(input: {
+  readonly range: AnalyticsListRange
+  readonly coverageFromIso?: string | null | undefined
+  readonly coverageToIso?: string | null | undefined
+}): AnalyticsListRange {
+  const { range, coverageFromIso, coverageToIso } = input
+  const coverageFromMs = coverageFromIso ? Date.parse(coverageFromIso) : Number.NaN
+  const coverageToMs = coverageToIso ? Date.parse(coverageToIso) : Number.NaN
+  if (!Number.isFinite(coverageFromMs) || !Number.isFinite(coverageToMs)) return range
+  const requestedFromMs = range.fromIso ? Date.parse(range.fromIso) : Number.NaN
+  const requestedToMs = Date.parse(range.toIso)
+  const fromMs = Number.isFinite(requestedFromMs) ? Math.max(requestedFromMs, coverageFromMs) : coverageFromMs
+  const toMs = Math.max(Math.min(Number.isFinite(requestedToMs) ? requestedToMs : coverageToMs, coverageToMs), fromMs)
+  return { fromIso: new Date(fromMs).toISOString(), toIso: new Date(toMs).toISOString() }
+}
+
+/**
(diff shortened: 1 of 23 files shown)


apps/web/src/domains/projects/use-analytics-time-window.ts
```diff
@@ -56,6 +56,28 @@ export function resolveAnalyticsListRange(input: {
   return allTimeLowerBoundIso ? { fromIso: allTimeLowerBoundIso, toIso } : { toIso }
 }
 
+/**
+ * Clips a list range into the range the screen's data actually covers. Absent coverage bounds are
+ * open, so a screen without a coverage constraint keeps its range untouched, and an "All time" read
+ * on a screen with one comes back as exactly the covered band rather than an unbounded scan whose
+ * counts would be labelled with a wider range than the data spans.
```
andresgutgon: Simplify this explanation

andresgutgon:
@claude review

apps/web/src/domains/projects/use-analytics-time-window.ts
andresgutgon: Done in cb3f14c4c — collapsed to one line:

```ts
/** Clips a list range into the band the data covers. Without bounds it passes through; "All time" becomes the band. */
```

Kept only the two things the signature does not show: an absent bound passes the range through, and "All time" resolves to the covered band instead of staying unbounded. The rest was restating the code.

PR #4414, opened 2026-08-12
Custom behavior UI update [LAT-831]

ntizish:
## Behaviors home: card grid → full-width panels

- Replace the 3-column card grid with a stacked list of full-width panels (one per behavior group), matching the updated Figma design.
- Group previews now show every top-level group (no truncation), each with name, description, `% of total · N sessions`, and a trend badge — nested tree breakdowns stay behind the click into the tree.
- Add color coding for group predominance: a segmented share bar in the panel header (hover for a full legend) and a matching full-height color bar per group row, using the same categorical ramp as the cost dashboard.
- Add scope transparency: a globe/clock-arrow icon next to each panel's name with a tooltip explaining that Topics covers the full project history while every other behavior group only covers sessions since it was created.
- Add a plain "Capturing data from `<date>`" indicator next to the date picker on the facet tree page (backed by a new `createdAt` field on `FacetRecord`).
- Sort populated behaviors above empty/analyzing ones on the catalog home (stable sort).
- Route `BehaviourBadge` through the shared `Badge` component instead of a bespoke pill, so it stays consistent with the rest of the app; let `Tooltip` accept an optional `className` to widen its content when needed.
- Various layout polish: spacing between panels vs. the header, description max-width, blank-slate padding, secondary background on group rows.

https://github.com/user-attachments/assets/2dcff06e-6edc-41a1-9796-34be7716a641

## Summary by CodeRabbit

* **New Features**
  * Redesigned the Behaviors catalog from cards to a detailed panel and list view.
  * Added behavior metadata, trend indicators, session counts, percentages, and grouped previews.
  * Added scope explanations for Topics and user-created behavior groups.
  * Added “New behavior group” labeling and expanded theme background colors.
* **Improvements**
  * Added consistent trend badges, icons, and tooltip styling.
  * Behaviors are now ordered with existing groups shown first.


andresgutgon:
@claude review

apps/web/src/domains/taxonomy/facet-scope-copy.ts
```diff
@@ -0,0 +1,15 @@
...
+ * embeddings every trace already gets at ingest, so it can read the whole
+ * history for free. A facet instead needs an AI pass to extract its answer per
+ * session, which only ever runs on sessions from the point the facet existed
+ * onward — there is nothing to extract for sessions that predate it. Used by
+ * the catalog's scope icon (the facet tree still shows the real date picker).
+ * User-facing copy says "behavior group", never "facet" — that word only
+ * appears here and in the domain layer (`TaxonomyFacet`, `facetId`).
+ */
+export const FACET_SCOPE_GLOBAL_TOOLTIP =
+  "Topics tracks sessions across your project's entire history, from the very beginning."
```
andresgutgon: One thing I'm not sure is clear is that people can understand that topics (the first block in the list) is different than the other behaviors. 

So when we talk about TOPICS to the user not sure they will understand

apps/web/src/domains/taxonomy/group-colors.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+const GROUP_COLOR_CLASSES = [
+  "bg-blue-400 dark:bg-blue-600",
+  "bg-violet-400 dark:bg-violet-600",
+  "bg-fuchsia-400 dark:bg-fuchsia-600",
+  "bg-teal-400 dark:bg-teal-600",
+  "bg-amber-400 dark:bg-amber-600",
+  "bg-red-400 dark:bg-red-600",
+  "bg-sky-400 dark:bg-sky-600",
+  "bg-lime-400 dark:bg-lime-600",
```
andresgutgon: These colors could be part of our design system and be encoded in the same way `bg-background` and `text-foreground`. So then dark mode is automatic and we don't need 2 classes together. Also any part of the UI can use these same colors 

Maybe we could called with some kind of prefix `bg-background-blue` so we could have `text-foreground-blue` if needed to play in the same way we play with background and foreground

ntizish: Good call — moved these into globals.css as proper design tokens (`bg-background-blue`, `-violet`, `-fuchsia`, `-teal`, `-amber`, `-red`, `-sky`, `-lime`), following the `bg-background-blue`/`bg-background-code`/`bg-background-gray` naming pattern you suggested. Dark mode is automatic now (each token flips in `.dark`), so the `bg-X dark:bg-Y` pairing is gone, and `group-colors.ts` is deleted — the color-cycling helper is now a small local const in `behaviour-catalog-panel.tsx`, the only file that needs it. Pushed in 8bf47f2a9.

apps/web/src/domains/taxonomy/facet-scope-copy.ts
ntizish: Reworded both tooltips so each one names the *other* side of the contrast, since a user could hover either icon first:
- Topics: "Unlike the behavior groups you create, Topics is built in by default, covering your project's entire history from day one."
- Others: "Unlike Topics, this behavior group only tracks sessions received since it was created."

Pushed in 8bf47f2a9.

andresgutgon (approved):

PR #4416, opened 2026-08-12
refactor(taxonomy): split "has staging leaves" from "is an adaptive tree" (LAT-862 Part 2a)

andresgutgon:
Part 2a of LAT-862. **This PR changes no behaviour.** Facet-lens coverage is unchanged until Part 2b.

## The problem

`planPersistsAdaptive(plan)` was `(plan.leafClusters ?? []).length > 0`, and four call sites read it for two different questions:

| call site | what it actually asks |
| --- | --- |
| reassign branch | does this plan defer its assignment writes to a full-window routing pass? |
| `deprecateGardenTaxonomyClustersActivity` | same — is there a catch-up pass to run? |
| `publishClusterIds` | did this stage a fresh adaptive tree (legacy fallback for plans with no `stagedClusterIds`)? |
| `supersededByPublish` | did this stage a fresh adaptive tree, so the swap retires the whole prior tree? |

Today those coincide, because `leafClusters` is populated exactly when `persistAdaptive` is true. Part 2b needs facet plans to carry staging leaves without becoming adaptive trees, and under one boolean that flip lands on publication: `supersededByPublish` would return `supersededClusterIds ?? []` — empty on a static persist — so a facet rebuild would deprecate **nothing** and leave its old clusters active alongside the new ones.

## The change

`HierarchicalTaxonomyPlan` now carries `persistsAdaptiveTree` (set from the existing `persistAdaptive`, i.e. "the adaptive builder produced an accepted build" — not the org flag), and the plan artifact persists it. The predicate splits in two:

- `planHasStagingLeaves` — shape-keyed, as before. Reassign routing and the catch-up branch read it.
- `planPersistsAdaptiveTree` — `plan.persistsAdaptiveTree ?? planHasStagingLeaves(plan)`. Publication reads it.

The fallback keeps Temporal deploy skew safe: a plan staged by the current deploy and published by the next carries no flag, and there `leafClusters` answered the question correctly.

## Why it is byte-identical

The two predicates can only disagree when `persistAdaptive` is true and `leafClusters` is empty. `bornLeaves` gets one entry per descriptor with `children.length === 0`, and `collectNodes` emits the root itself as a descriptor — so a persisted tree always has at least one leaf, including a bare root. The divergence is unreachable by construction, not merely unlikely.

Per plan shape, `persistsAdaptiveTree` equals the old predicate's value everywhere:

| plan | old predicate | new `persistsAdaptiveTree` |
| --- | --- | --- |
| global topic, `off` | false | false |
| global topic, `enforced` accepted | true | true |
| global topic, `enforced` fallen back | false | false |
| cohort view, `off` | false | false |
| cohort view, `enforced` accepted | true | true |
| facet lens (forced `off`) | false | false |
| sample below the gardening minimum | false | false (explicit) |

## Tests

Publication is the risk here, not reassignment — #4305 published a tree before it was named and blanked Behaviours on every rebuild, #4347 let a degenerate bare-root rebuild wipe the live tree. So the case with no production instance gets a fixture rather than trust:

- A plan with staging leaves but **not** an adaptive tree still runs the full-window catch-up (shape), and its swap retires exactly `deprecatedClusterIds` while the prior root stays active (flag). Reverting `supersededByPublish` to the shape predicate fails this test, so it discriminates.
- The same plan in the pre-`stagedClusterIds` shape activates nothing, instead of falling back to the whole cluster set.
- The existing pre-change adaptive plan test now drops `persistsAdaptiveTree` entirely, pinning the shape fallback.
- Domain-side, `persistsAdaptiveTree` is asserted on off/global, off/scoped, adaptive-accepted and adaptive-fallen-back plans.

The fixture uses a cohort scope deliberately: a facet plan with leaves would also exercise `reassignFullWindowScoped`'s hard-coded `facetId: null`, which is Part 2b's to fix and should not be pinned as correct here.

## Explicitly not in this PR (all Part 2b)

`mode: "off"` in `plan-facet-garden.ts`, the hard-coded `facetId: null`, projection-space routing, and letting any facet plan gain leaves. No facet plan may gain leaves before the projection-space path and the routing guard land in the same change — `planPersistsAdaptive` flipping true for a whole-project facet routes it to `reassignFullWindowGlobal`, which overwrites `taxonomy_observations.assigned_cluster_id` and clobbers the global topic tree.

## Answered for Part 2b: projection embeddings are normalised

Verified in `extract-facet-projections.ts`: the extracted answer's embedding goes through `normalizeTaxonomyEmbedding` before it is persisted — the same helper the planner applies to observation embeddings. `routeOne` also re-normalises defensively before `cosineSimilarityNormalized`, so routing confidence is on the same footing in projection space as in observation space. Unclear projections carry an empty embedding and are dropped by `routeObservationsToLeaves`, which skips empty vectors.

## Verification

- `@domain/taxonomy` full suite: 296 passed
- `@app/workflows` taxonomy tests: 38 passed (11 in the publish file, up from 9)
- `@app/workers` taxonomy tests: 14 passed
- `typecheck` clean on `@domain/taxonomy` and `@app/workflows` (`@app/workers` has a pre-existing failure in `packages/platform/latitude-api/src/client.ts`, reproduced on the base commit)

### DIFF
--- a/apps/workflows/src/activities/taxonomy-gardening-activities.ts
+++ b/apps/workflows/src/activities/taxonomy-gardening-activities.ts
@@ -195,12 +195,18 @@ interface StoredGardenTaxonomyPlan {
   /**
    * Non-null ⇒ enforced planning fell back to static (structural/non-finite
    * adaptive output), so this plan persists the static tree even though its mode
-   * is `enforced`. Drives `planPersistsAdaptive` so downstream never runs the
-   * staging path on a fallen-back plan.
+   * is `enforced`. Clears both `persistsAdaptiveTree` and `leafClusters` so
+   * downstream never runs the staging path on a fallen-back plan.
    */
   readonly fallbackReason?: TaxonomyAdaptiveFallbackReason | null
   /** Adaptive-only: leaf id + centroid for full-window routing. Empty/absent on off. */
   readonly leafClusters?: readonly StagingLeafCluster[]
+  /**
+   * Whether the plan staged a fresh adaptive tree (every id new, prior tree
+   * retired wholesale), as opposed to upserting continuations in place. Absent on
+   * plans staged by pre-change code, where `leafClusters` stood in for it.
+   */
+  readonly persistsAdaptiveTree?: boolean
   /** Adaptive-only: the full old active tree the atomic swap deprecates. Empty/absent on off. */
   readonly supersededClusterIds?: readonly string[]
   /**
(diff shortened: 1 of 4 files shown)


andresgutgon:
@claude review

PR #4419, opened 2026-08-12
feat(taxonomy): route a facet lens in projection space (LAT-862 Part 2b)

andresgutgon:
## What

LAT-862 Part 2b. A facet lens's membership only ever covered the gardening window, so widening the date range added no sessions. [Part 1 (#4411)](https://github.com/latitude-dev/latitude-llm/pull/4411) made the UI honest about that by clipping the range to the coverage that exists. This removes the shortfall so coverage actually accumulates — Part 1's clip then widens on its own, with no UI change.

Stacked on [#4416](https://github.com/latitude-dev/latitude-llm/pull/4416) (Part 2a), which split "has staging leaves" from "is an adaptive tree". This fills that seam.

## The three changes

**1. Facet plans stage leaf centroids.** `leafClusters` is populated for a facet pass as well as an adaptive one, so the pass defers its assignment writes to a full-window routing pass instead of writing the sample's assignments. `persistsAdaptiveTree` stays `false`, so publication keeps upserting the view's continuations in place rather than retiring the whole prior tree — that is exactly the split Part 2a made.

The centroids need no conversion: `planFacetGardenUseCase` hands the planner facet projections as its observations, so `bornLeaves[].centroid` is already a projection-space centroid.

**2. Route the projection window, not the observation window.** New `FacetProjectionRepository.listWindowForReassignment` — every clear projection cached for the facet, newest first, narrowed to the cohort's sessions by the same filter compiler the sample and the observation-space window use. `routeObservationsToLeaves` is reused **unchanged**; it takes `{observationId, embedding}` plus leaf centroids and is indifferent to the space.

This is where the coverage win lands: extraction is cached per `(facet, session)` and never invalidated, so the routed window is the union of every pass, not the latest 7-day sample.

**3. `facetId` is no longer hard-coded to `null`, at both call sites.** `reassignFullWindowScoped` writes the plan's facet and picks its window source from it. The reassign target is now resolved once by `reassignmentTarget(plan)` and read by both the main reassign branch and the catch-up inside `swapAndCatchUp`. Missing the second would have left coverage correct at reassign time and stale the moment the swap landed.

## The corruption case this ordering exists to prevent

`reassignmentTarget` asks **"is this a view?" before "does it have staging leaves?"**, deliberately, rather than relying on the existing branch conditions:

| plan shape | before | now |
| --- | --- | --- |
| whole-project facet (`facetId` set, `customBehaviorId` null) | would reach `reassignFullWindowGlobal` → `observations.reassignManyById` overwrites `taxonomy_observations.assigned_cluster_id`, clobbering the **global topic tree** for every project | `sampleScoped`; never touches the inline column |
| cohort × facet | would write `facetId: null`, filing facet memberships as cohort memberships | `fullWindowScoped` with the facet stamped |

The whole-project-facet shape is unreachable from the workflow today (a facet only arrives via its behaviour, and the builder already fails fast without one), but it is pinned by a test rather than by trust.

`plan-facet-garden.ts` stays hard-coded to `mode: "off"`. Making it flag-driven is the separate follow-up that only becomes safe now that this has landed.

## Verified, not assumed

**Normalisation — confirmed.** `extract-facet-projections.ts` wraps the embedding in `normalizeTaxonomyEmbedding` before persisting (`use-cases/extract-facet-projections.ts`, the `answer.length === 0 ? [] : normalizeTaxonomyEmbedding(...)` branch), exactly as the observation path does. Both sides of `cosineSimilarityNormalized` are unit vectors, and `routeOne` re-normalises the source anyway, which is a no-op on an already-normalised vector.

**No confidence threshold — confirmed, and deliberately left alone.** `routeObservationsToLeaves` is pure nearest-centroid with no floor, and neither writer filters on confidence. `TAXONOMY_ASSIGN_ABSOLUTE_THRESHOLD = 0.65` is wired only into the online path and the builder, never into reassignment. So every clear projection lands in whichever leaf is nearest, however dissimilar — the same behaviour the observation-space path already has. **I did not add a gate.** I do think one is worth discussing: a facet's clusters answer a narrow question, so a projection that matches none of them is more likely than an off-topic observation is in the global tree. But adding it here would be a new product decision buried in a routing change, and it would add a third term to Part 1's coverage ratio. Separate issue if we want it.

**Cost — zero LLM calls per pass.** Extraction is cached per observation and already paid for; routing is dot products against ~10 centroids. Reaching sessions that were never sampled still costs one extraction each, but nothing on this path triggers that.

## Tests

- `taxonomy-gardening-publish.test.ts` — a facet pass routes the projection cache and not the observation window (distinct row counts and orthogonal embeddings make reading the wrong table visible); the facet is stamped on every edge; the catch-up after the swap routes in projection space too; a facet plan with no behaviour never reassigns the global column.
- `facet-projection-repository.test.ts` (chdb) — the window is the whole cache newest-first, skipping unclear rows and other facets.
- `build-hierarchical-taxonomy.test.ts` / `plan-facet-garden.test.ts` — a facet plan now stages leaves and empties both sample-assignment arrays, with `persistsAdaptiveTree` false.

`pnpm typecheck` (whole workspace), `pnpm knip`, `@domain/taxonomy` (304), `@app/workflows` (109), and the single chdb file all pass.

## Where to look

`reassignmentTarget` in `taxonomy-gardening-activities.ts` — it is the guard for the data-corruption case above, and both call sites read it.

Closes LAT-862.

## Summary by CodeRabbit

- **New Features**
  - Facet-based taxonomy gardening now supports projection-space reassignment and catch-up processing.
  - Reassignment windows can apply optional filters and prioritize the most recent eligible projections.
  - Facet runs now stage leaf clusters for deferred full-window routing.

- **Bug Fixes**
  - Corrected facet and behavior identifiers during reassignment.
  - Prevented unnecessary global observation reassignment when no behavior slice exists.
  - Improved handling of unclear observations and projection eligibility.

---

## QA: measured on a live stack

Two real `gardenTaxonomyWorkflow` executions on Temporal — full pipeline (sample → Bedrock/minimax extraction → cluster → reassign → publish → naming → quality gate → lineage) against a local ClickHouse and Postgres. No activity was called directly.

The workflow reads `now` from the wall clock, so a two-pass scenario three weeks apart is not reachable by faking the clock. Elapsed time was simulated by moving the **fixture** instead: seed a batch inside the 7-day sample window → garden it → shift those rows 20 days into the past → seed a second batch → garden again. Pass 2 can then only *sample* the recent batch, and the question is whether the older, already-extracted sessions are still in the lens.

Pass 1 ran once. Its derived state was snapshotted and **pass 2 was replayed twice from that identical state** — once with a worker built from `bfc194d58` (this PR's parent), once with this branch. Same fixture, same databases, only the worker's code differs.

| | parent (`bfc194d58`) | this branch |
| --- | --- | --- |
| cached projections | 98 | 98 |
| **visible assignments** | **42** | **98** |
| **coverage span** | **Aug 7–12 (5 days)** | **Jul 18 – Aug 12 (25 days)** |
| older-batch sessions visible | 0 of 56 | 56 of 56 |
| orphaned rows left in `taxonomy_view_assignments` | 56 | 0 |
| rows with an empty `facet_id` | 0 | 0 |
| observations moved off the global topic cluster | 0 | 0 |

The parent reproduces the reported defect exactly: 98 sessions extracted and cached, 42 visible, and the other 56 still physically present in the slice but pointing at clusters the rebuild deprecated — the erosion term from LAT-862, measured.

**Routing sanity.** Per topic, the re-routed older sessions score the same as their in-window twins: 0.781 vs 0.783, 0.869 vs 0.871, 0.917 vs 0.917. They are pulled in because they match, not because nearest-centroid will accept anything.

**Coverage accumulates up to a ceiling, and it is not far away.** All three taxonomy tables are written with a hardcoded `TAXONOMY_OBSERVATION_RETENTION_DAYS = 30`, and the ClickHouse TTL is `start_time + retention_days + 30`, so embeddings and assignments survive roughly **60 days**. This PR moves a lens from ~7 days to *up to* ~60; it does not make coverage unbounded. Worth saying plainly because the original report asked about a four-month range, which is unreachable regardless of this change. Part 1 is already consistent with the ceiling — `TAXONOMY_LENS_COVERAGE_HORIZON_DAYS` is pinned to exactly that TTL horizon, so the picker can never advertise a band into deleted data.

**Part 1 composes as claimed.** With no UI change, the lens header moved from a 5-day band to `Coverage: 18 Jul – 13 Aug` — the clip widened on its own as coverage grew, which is the sequencing argument in LAT-862 confirmed end to end.

**Cost.** Pass 2's extraction was a 98/98 cache hit, so routing added no LLM calls. Naming still calls the model, exactly as before.

### The missing confidence floor, now measured rather than asserted

The fixture deliberately includes one topic present only in the older batch, so pass 2's tree has no counterpart for it. All 14 of those sessions were filed into **"Export Financial Data as CSV" at 0.567 confidence**, against 0.78–0.92 for every well-matched session.

This is not new behaviour — the observation-space path has no floor either — but it sharpens the trade-off. **Before this PR those sessions were silently invisible; after it they are visible and mislabelled.** Silent omission becomes visible misclassification, precisely for the topics that fade out of a tree as it rebuilds.

The gate is still deliberately left out of this PR, as scoped. But this is a stronger argument for adding one than the reasoning offered above, and it should get its own issue rather than being folded in here.

*The QA harness is not committed — it is fixture-specific, with hardcoded QA ids. Happy to add it under a scripts path if reviewers want the scenario reproducible in CI.*

### Production prevalence, measured before scoping the follow-up

Read-only aggregates, no customer text.

**Scale.** 743 projects have taxonomy observations. Exactly **one** has ever taken the full-window reassignment path, because on the topic tree it is gated on the adaptive flag; 612 are on sample-only assignment. This PR makes the facet path shape-driven rather than flag-driven, so on merge every existing lens (4 projects, 7 lenses) and every future one starts routing without a floor. Lenses become the dominant consumer of full-window reassignment immediately, which inverts where the risk sits.

**Prevalence.** On the one project running it, 113 of 3,725 re-routed observations (**3.03%**) sit below the online path's existing `TAXONOMY_ASSIGN_ABSOLUTE_THRESHOLD` of 0.65 — p01 0.537, p05 0.694, median 0.849. Across today's facet lenses (1,776 rows, all sample assignments) the below-0.65 share is **2.59%**, median 0.892. The lens figure is a floor, not a forecast: those rows are sample members matched against clusters they helped build, which is the best case by construction. The population this PR newly routes has no production analogue, because nothing has ever routed it.

**Mechanism.** Bucketed by session week, the below-gate share does not rise with age — it is a bump in two middle weeks (4.6% and 6.7%) with the oldest weeks among the cleanest, and a median that barely moves (0.843–0.857). So the exposure is not "old sessions" but "sessions from a period whose content the current tree no longer represents". My earlier age-based reasoning was wrong.

**A fixed threshold would not transfer.** Mean cosine between two *randomly chosen* sessions, sampled per project, ranges from 0.59 to 0.78 across the seven largest projects, and one project's median random pair is 0.985. So 0.65 sits far below the noise floor in some projects — where it could never fire — and at the median random pair in others. Any gate the follow-up adds should be relative to a per-project or per-cluster baseline rather than an absolute cosine. (Centroid similarity is systematically higher than pairwise, so these numbers are not directly comparable to assignment confidence; the point is the spread, not the level.)

Caveat throughout: confidence is not a correctness label. These are assignments the online path would have refused, not confirmed misfilings.

### DIFF
--- a/apps/workflows/src/activities/taxonomy-gardening-activities.ts
+++ b/apps/workflows/src/activities/taxonomy-gardening-activities.ts
@@ -5,6 +5,7 @@ import {
   type FilterSet,
   OrganizationId,
   ProjectId,
+  type SessionId,
   TaxonomyClusterId,
   TaxonomyRunId,
 } from "@domain/shared"
@@ -14,6 +15,7 @@ import {
   CustomBehaviorStatus,
   customBehaviorFilterSetHasConditions,
   emitLineageUseCase,
+  FacetProjectionRepository,
   type HierarchicalTaxonomyPlan,
   isDisplayableTaxonomyName,
   measureTaxonomyNameQualityUseCase,
@@ -426,13 +428,18 @@ const withCustomBehaviorClickHouse = <A, E, R>(effect: Effect.Effect<A, E, R>, o
     withClickHouse(TaxonomyViewAssignmentRepositoryLive, getClickhouseClient(), OrganizationId(organizationId)),
   )
 
-// Scoped full-window reassignment reads the global taxonomy window (filtered by
-// the behavior's sessions) and writes the scoped taxonomy_view_assignments
-// slice, so it needs both ClickHouse repositories.
+// Scoped full-window reassignment reads its window (the global taxonomy window
+// filtered by the behavior's sessions, or the facet's projection cache) and
+// writes the scoped taxonomy_view_assignments slice, so it needs all three
+// ClickHouse repositories.
 const withScopedReassignClickHouse = <A, E, R>(effect: Effect.Effect<A, E, R>, organizationId: string) =>
   effect.pipe(
     withClickHouse(
-      Layer.mergeAll(TaxonomyObservationRepositoryLive, TaxonomyViewAssignmentRepositoryLive),
+      Layer.mergeAll(
+        TaxonomyObservationRepositoryLive,
+        TaxonomyViewAssignmentRepositoryLive,
+        FacetProjectionRepositoryLive,
+      ),
       getClickhouseClient(),
       OrganizationId(organizationId),
     ),
@@ -821,6 +828,15 @@ const planLeaves = (plan: StoredGardenTaxonomyPlan): ReassignmentLeaf[] =>
 // those up after the swap). More than this means the bulk write did not land.
 const STAGING_SNAPSHOT_STRAGGLER_FRACTION = 0.05
 
+// What the write needs from a window row once routing is done, in either
+// embedding space — so the observation window and the projection window are
+// interchangeable to everything downstream of the router.
+interface ReassignmentWindowRow {
+  readonly observationId: string
+  readonly sessionId: SessionId
+  readonly startTime: Date
+}
+
 // Shared kernel of both full-window reassignment targets: read the bounded live
 // window (optionally scoped to the behavior's sessions) and route every row to
 // its nearest staging leaf. The caller maps the routed assignments onto its own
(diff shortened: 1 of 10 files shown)


andresgutgon:
@claude review

PR #4435, opened 2026-08-13
docs: add local agent simulation guide

geclos:
## What does this PR do?

Adds a public Mintlify guide for running agent simulations locally, without a hosted Latitude runner.

The flow is: pull a dataset with `@latitude-data/sdk` / `latitude-sdk`, drive each row with [Scenario](https://github.com/langwatch/scenario) (TypeScript and Python), send the new sessions back through `@latitude-data/telemetry` / `latitude-telemetry`, then compare versions with Experiments using a `simulation` tag and `agentVersion` metadata.

This is the docs-only answer to "can Latitude cover multi-turn simulation?" Latitude keeps the dataset, traces, signals, and comparison. The agent still runs in the user's process.

N/A

## How was this tested?

Docs-only. Cross-checked the snippets against:

- `client.datasets.listRows` / `list_rows` pagination (`items`, `nextCursor`/`next_cursor`, `hasMore`/`has_more`)
- telemetry `capture()` (including Python async)
- Scenario's public `run()` / `userSimulatorAgent` / `judgeAgent` / `UserSimulatorAgent` / `JudgeAgent` APIs

`docs.json` parses. The new page is linked from Refine nav plus Datasets, Experiments, Regression testing, and Getting Started.

## Checklist

- [x] Lint, type-checking, and tests pass locally (docs-only)
- [ ] I have signed the CLA

[Slack Thread](https://latitudedata.slack.com/archives/C0456PX26N7/p1786633090746529?thread_ts=1786633090.746529&cid=C0456PX26N7)

<div><a href="https://cursor.com/agents/bc-e28d57f7-14d3-50df-a310-36d3fda0555a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e28d57f7-14d3-50df-a310-36d3fda0555a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive guide for running local agent simulations against datasets.
  * Documented single-turn and multi-turn simulations, telemetry, judging, result reporting, experiment comparisons, and CI gating.
  * Updated dataset, regression testing, Refine, and getting-started workflows to include local simulations.
  * Added navigation links and next-step references to the new simulations documentation.


docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+    expected = as_text(row.expected_output)
+    session_id = f"sim-{RUN_ID}-{row.row_id}"
+    agents = [
+        agent_for_row(row),
+        scenario.UserSimulatorAgent(),
+    ]
+    if expected:
+        agents.append(scenario.JudgeAgent(criteria=[expected]))
+
+    result = await scenario.run(
```
learningbizz: The judge verdict lives in a console log and a process exit code. Latitude has a Scores API, the page even links Scores from the CI section, but never posts anything to it. One scores.create per row with the judge verdict and the session id would turn this from "a pile of traces from a run" into a test result that appears in score analytics, feeds signals, and can be compared across runs. That is the difference between Latitude storing the traces and Latitude owning the result

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+    maxTurns: 8,
+    script: [
+      scenario.user(input),
+      scenario.agent(),
+      scenario.user(),
+      scenario.agent(),
+      expected ? scenario.judge() : scenario.succeed(),
+    ],
+  })
+
```
learningbizz, line 292: The judge verdict lives in a console log and a process exit code. Latitude has a Scores API, the page even links Scores from the CI section, but never posts anything to it. One scores.create per row with the judge verdict and the session id would turn this from "a pile of traces from a run" into a test result that appears in score analytics, feeds signals, and can be compared across runs. That is the difference between Latitude storing the traces and Latitude owning the result

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+3. Run it again (`AGENT_VERSION=v2`).
+4. Open [Experiments](../experiments/overview) and create two variants:
+   - Baseline: tag `simulation` and metadata `agentVersion = v1`
+   - Comparison: tag `simulation` and metadata `agentVersion = v2`
+
+The experiment will compare session counts, cost, errors, tools, signals, and behaviors on those two slices. If a [signal](../signals/overview) you care about still fires on the v2 slice, the fix did not cover that case.
+
+You can also search for `metadata.rowId` to inspect one case across versions.
+
+## Gate it in CI
```
learningbizz: should read like optional, not mandatory

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+| --- | --- | --- |
+| Tag `simulation` | `simulation` | Separates these sessions from production. |
+| Tag for the agent version | `agent-v2` | Lets an experiment compare v1 vs v2. |
+| Metadata `dataset` | `refund-regressions` | Ties the run back to the dataset slug. |
+| Metadata `rowId` | the SDK `rowId` | Ties the session to one dataset row. |
+| Metadata `agentVersion` | `v2` or a git SHA | The slice key for Experiments. |
+
+Use a fresh `sessionId` per dataset row per run, for example `sim-${runId}-${rowId}`. All turns of that conversation share it, so they group as one [session](../observability/sessions).
+
+If you already have [evaluations](../evaluations/overview) on the project, they will score simulation traces the same way they score production, unless their [triggers](../evaluations/triggers) filter them out. Keep the `simulation` tag so you can exclude this traffic from production monitors if you want to.
```
learningbizz: No warning about blast radius. Line 57 mentions that existing evaluations will score simulation traces, framed as a convenience. The consequence is that a CI run creates scores, which cluster into signals, which monitors watch, which notify Slack. Someone will page their team from a test run. The page needs an explicit "before your first run, exclude the simulation tag from your production evaluation triggers and monitors," or a recommendation to use a separate project. There's also no note that this traffic is metered.

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+        scenario.UserSimulatorAgent(),
+    ]
+    if expected:
+        agents.append(scenario.JudgeAgent(criteria=[expected]))
+
+    result = await scenario.run(
+        name=f"dataset row {row.row_id}",
+        description=input_text,
+        agents=agents,
+        thread_id=session_id,
```
learningbizz: **Blocking: `thread_id` is not a parameter of Python `scenario.run()`.** I unpacked the langwatch-scenario 1.1.1 wheel to check — the signature at `scenario/scenario_executor.py:2008` is `name, description, agents, max_turns, verbose, cache_key, debug, script, set_id, metadata, on_audio_chunk, on_voice_event, audio_playback`. No `thread_id`, and no `**kwargs` to swallow it, so this raises `TypeError` on the very first row. The TypeScript `ScenarioConfig` *does* accept `threadId`, which is why line 164 is fine and this isn't — the two snippets aren't equivalent.

Scenario mints its own thread id internally (`ScenarioState.thread_id`, a `PKSUID("scenariothread")`), so there are two ways out:

- drop the argument and pass the session id through `metadata=`, or
- invert the direction: read Scenario's `input.thread_id` inside the adapter and derive `session_id` from it.

The second is worth preferring, because it's also what line 202 already tells the reader to do ("key that state off Scenario's `input.threadId` and use the same id as `sessionId`") — right now that advice is only achievable in TypeScript.

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+   - Baseline: tag `simulation` and metadata `agentVersion = v1`
+   - Comparison: tag `simulation` and metadata `agentVersion = v2`
+
+The experiment will compare session counts, cost, errors, tools, signals, and behaviors on those two slices. If a [signal](../signals/overview) you care about still fires on the v2 slice, the fix did not cover that case.
+
+You can also search for `metadata.rowId` to inspect one case across versions.
+
+## Gate it in CI
+
+The scripts above already exit non-zero when Scenario's judge fails a row. Wire that into CI:
```
learningbizz: **Blocking for this section: a dataset with no expected outputs greenlights the build.** No expected output means no judge is appended (lines 156/301) and the script terminates in `succeed()`, so the row passes unconditionally. Wire that into CI as written and a dataset where nobody filled in expected outputs exits 0 with "0 failed" — the gate reports success while measuring nothing, which is worse than having no gate.

Documenting the semantics isn't sufficient on a page that tells people to block merges on the exit code. The script needs to either skip unjudgeable rows and surface them as skipped, or treat "no criteria" as a run-level failure so an empty dataset can't pass. My preference is the latter for CI and the former for local runs, since that matches how people actually use the two.

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+- From a [signal](../signals/overview), so the cases are the failures you care about. See [Add traces](../datasets/add-traces).
+- From [Search](../search/overview) or a saved search, for a behavior you want to keep passing.
+- By hand or CSV, for cases that never showed up in production.
+
+Each row should have at least an **input**. That is the first user message, or a short description of what the user wants.
+
+**Expected output** is optional and useful in two ways:
+
+- As the known-good answer for a single-turn check.
+- As judge criteria for a multi-turn run, for example "Issue the refund without asking for the order number twice."
```
learningbizz: Expected output is being asked to do two incompatible jobs here, and the code takes the second one literally — `criteria: [expected]` at lines 157 and 302 feeds a stored answer string straight to the judge.

A known-good answer captured from one conversation path is not a criterion. Judging a fresh multi-turn run against a stored string reintroduces exactly the brittleness simulations exist to avoid: the agent resolves the case correctly by a different route, phrases it differently, and the judge fails it. The criterion you actually want is the one in the example on line 39 ("Issue the refund without asking for the order number twice") — a behavioral assertion, which is a different kind of value than a golden answer.

These should be separate columns: a `judgeCriteria` custom column for the multi-turn judge (custom columns are already recommended two lines down at line 41, so the machinery is there), with expected output reserved for the single-turn check described at line 374.

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+  const failed = results.filter((result) => result && !result.success).length
+  console.log(`ran ${rows.length} rows, ${failed} failed`)
+  if (failed > 0) process.exitCode = 1
+} finally {
+  await latitude.flush()
+  await latitude.shutdown()
+}
+```
+
+`capture()` has to wrap the actual agent call, not a value you already computed. Scenario's user simulator and judge typically use their own model client, so they will not show up in Latitude unless you instrument that client too.
```
learningbizz: "typically use their own model client" is carrying a lot of weight for a hedge, and in Python I think it's actually wrong. `UserSimulatorAgent` and `JudgeAgent` both call `litellm.completion` (e.g. `scenario/judge_agent.py:778`), and litellm's OpenAI path goes through the same `openai` package this page hands to `instrumentations={"openai": openai}` on line 232. If that's right, those calls *are* captured — and since they fire outside `capture()`, they land as untagged, session-less traces in the live project, where production evaluations will score them and where they slip past the `simulation` tag filter that the rest of the page depends on for isolation.

So the hedge points the wrong way: the reader is told simulator traffic probably won't show up, when the likely outcome is that it shows up in the worst possible form. Worth running it once and then stating it definitively in whichever direction it lands, plus the remedy if they are captured (point Scenario at a separate, uninstrumented client, or filter its spans out).

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+RUN_ID = os.environ.get("SIMULATION_RUN_ID", str(uuid.uuid4()))
+
+latitude = Latitude(
+    api_key=os.environ["LATITUDE_API_KEY"],
+    project=PROJECT,
+    instrumentations={"openai": openai},
+)
+client = AsyncLatitudeClient(api_key=os.environ["LATITUDE_API_KEY"])
+
+scenario.configure(default_model="openai/gpt-4.1-mini")
```
learningbizz, line 356: Both libraries reach for the global OpenTelemetry tracer provider, and nothing on the page says the order matters. Scenario's tracing setup attaches its processors to an already-registered concrete provider, and calls `trace.set_tracer_provider()` itself when there isn't one (`scenario/_tracing/setup.py:121-214`); the Latitude SDK does the same attach-or-create dance. The order in this snippet is the safe one — `Latitude(...)` on line 229 before `scenario.configure()` on line 236 — but that's incidental, and a reader who moves the Latitude constructor into `main()` or lets Scenario configure tracing first ends up with Scenario's provider and silently loses every trace, with no error to explain it.

One sentence, here and in the TypeScript section, saying construct `Latitude` before Scenario configures its tracing, and that this is why the lines are in this order, turns luck into instruction.

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+  return rows
+}
+
+function agentForRow(row: { rowId: string }): AgentAdapter {
+  const sessionId = `sim-${RUN_ID}-${row.rowId}`
+  return {
+    role: AgentRole.AGENT,
+    call: async (input) => {
+      return capture(
+        "simulated-agent-turn",
```
learningbizz: Every session's root span is named `simulated-agent-turn` (same on the Python side, line 273), and a session's name falls back to its root span name — so the Sessions list that line 346 sends people to will be N identically-named rows. That's precisely the moment you're trying to spot the one row that regressed.

Folding the row id into the capture name (`\`simulated-agent-turn-${row.rowId}\``, or just the row id) makes the list navigable at a glance. Right now `metadata.rowId` search on line 359 is the only way to tell two sessions apart, which is a lot of ceremony for "which case failed".

docs/test-and-fix/simulations.mdx
```diff
@@ -0,0 +1,381 @@
...
+
+## Prepare a dataset
+
+Build the dataset the way you already do for [regression testing](./regression-testing):
+
+- From a [signal](../signals/overview), so the cases are the failures you care about. See [Add traces](../datasets/add-traces).
+- From [Search](../search/overview) or a saved search, for a behavior you want to keep passing.
+- By hand or CSV, for cases that never showed up in production.
+
+Each row should have at least an **input**. That is the first user message, or a short description of what the user wants.
```
learningbizz: Worth stating on this line that the input column should hold plain text. `asText` / `as_text` stringifies a structured cell, and that string becomes the simulated user's opening message at lines 166 and 311 — so a row whose input is a message array or a `{ role, content }` object (which is what you get from [Add traces](../datasets/add-traces), the very path recommended on line 30) produces a simulated user whose first utterance is raw JSON. The run completes, the traces look fine, and the results are meaningless.

Either constrain the column here, or handle the message-array shape explicitly by seeding `script` with the stored turns instead of a single `scenario.user(input)`.

learningbizz (commented):
Nice framing on this one. "Latitude holds the dataset, traces, signals and comparison, the agent runs in your process" is the right way to answer the multi-turn question without overclaiming, and the Experiments v1/v2 slice comparison is the strongest part of the page.

I pulled the actual packages and checked the snippets against them. The TypeScript SDK usage is accurate: `client.datasets.listRows(projectSlug, datasetSlug, { cursor, limit })` returning `{ items, nextCursor, hasMore }` matches `@latitude-data/sdk` 9.9.0, `DatasetRow` has `rowId` / `input` / `expectedOutput` / `custom` as used, `AsyncLatitudeClient.list_rows` matches on the Python side, Python `capture()` does accept an async callable so `await capture(...)` is correct, and 200 is the documented max row limit.

The Python example does not run, though, and there are a few behavioral issues that I think matter more than the typos. Comments inline; the `thread_id` one is blocking, the no-criteria one is blocking for the CI section specifically. I've skipped the points already raised by the bots and in my earlier comments (Scores API, blast radius, the fixed two-turn script vs `max_turns`, skipped-row counting, `str()` vs `json.dumps`).

**On the Scenario dependency**, separate from the code: recommending LangWatch's framework by name in Latitude's docs is defensible and developers will respect the honesty. The risk is that a page titled "Simulate agents locally" invites a direct feature comparison against their Simulations product, where they have a purpose-built visualizer and we have a tag-filtered sessions list. Promptfoo and Inspect only get a passing mention at line 12. Consider leading with the plain loop (currently buried at line 372) as the default path, presenting the three runners as equal options, and keeping the Experiments comparison as the section that is genuinely specific to Latitude.

PR #4445, opened 2026-08-15
fix(spans): skip Wrapped report generation when project or org was deleted

geclos:
## What does this PR do?

Fixes an uncaught `NotFoundError` in the weekly Wrapped report worker (`runWrappedUseCase`, `packages/domain/spans/src/wrapped/use-cases/run-wrapped.ts`).

**Datadog issues addressed:**
- [`ET-375` (`NotFoundError`)](https://app.datadoghq.eu/error-tracking/issue/4dea99da-6bbd-11f1-9869-da7ad0900005) — `workers`, regressed 2026-08-14 (previously resolved 2026-07-20)
- [`ET-376` (`NotFoundError$1`)](https://app.datadoghq.eu/error-tracking/issue/4deecbfe-6bbd-11f1-bca4-da7ad0900005) — same underlying event (Datadog's minifier-fingerprint split of the same stack), regressed 2026-08-14

Both are low-volume so far (1 occurrence each in the 7d window) but flagged as regressions of an issue that was already fixed once, in a different call site of the exact same bug class — worth catching before it becomes a recurring pile-up.

**Root cause:** `fanOutWeeklyRunUseCase` enumerates projects with Claude Code activity from ClickHouse, then publishes one `runForProject` BullMQ job per project on the `wrapped` topic. If a project (or its organization) is deleted between that enumeration and the job actually running, `ProjectRepository.findById` / `OrganizationRepository.findById` throw `NotFoundError`, which was previously uncaught — failing the whole BullMQ job and surfacing as an unhandled exception in Error Tracking.

This is the same race condition class as [PR #4099](https://github.com/latitude-dev/latitude-llm/pull/4099), which fixed the equivalent issue for `organizations.findById` / `incidents.findById` in `requestAgentDispatchUseCase` by wrapping the lookup in `Effect.catchTag("NotFoundError", ...)` and returning a skip result instead of propagating the failure. This PR applies the identical pattern to `runWrappedUseCase`, which wasn't covered by that earlier fix.

**Fix:** wrap both `findById` calls with `Effect.catchTag("NotFoundError", () => Effect.succeed(null))` and return `{ status: "skipped", reason: "project-not-found" | "organization-not-found" }` instead of letting the error propagate.

Closes #

## How was this tested?

- Added two regression tests to `run-wrapped.test.ts`: one where the project has been deleted, one where the organization has been deleted. Both assert the use case returns a `skipped` result instead of throwing, and that no report is persisted.
- Confirmed both new tests **fail** against the pre-fix code (uncaught `NotFoundError` propagates and the job errors) and **pass** with the fix applied.
- Full `@domain/spans` test suite passes except for 28 pre-existing failures in an unrelated file (`ingest-spans.test.ts`, a `Uint8Array.prototype.toBase64` Node-version mismatch in this sandbox) — confirmed identical failure count/content on `development` before this change, so unrelated to this fix.
- `pnpm exec biome check` clean on both changed files.

**Post-deploy verification:** occurrences of Datadog issues `ET-375` / `ET-376` should stop appearing for this failure mode (the underlying race can still occur, but it will now resolve as a silent skip rather than an error).

## Checklist

- [ ] I have signed the CLA

---
_Generated by [Claude Code](https://claude.ai/code/session_01Tnt3JSaoeDDrLEW278AVK9)_

### DIFF
--- a/packages/domain/spans/src/wrapped/use-cases/run-wrapped.test.ts
+++ b/packages/domain/spans/src/wrapped/use-cases/run-wrapped.test.ts
@@ -126,13 +126,13 @@ const makeOrganization = (): Organization => ({
   updatedAt: new Date("2026-01-01T00:00:00.000Z"),
 })
 
-const makeOrganizationRepository = (organization: Organization): (typeof OrganizationRepository)["Service"] => ({
+const makeOrganizationRepository = (organization: Organization | null): (typeof OrganizationRepository)["Service"] => ({
   findById: (id) =>
-    id === organization.id
+    organization && id === organization.id
       ? Effect.succeed(organization)
       : Effect.fail(new NotFoundError({ entity: "Organization", id })),
   findByIdForUpdate: (id) =>
-    id === organization.id
+    organization && id === organization.id
       ? Effect.succeed(organization)
       : Effect.fail(new NotFoundError({ entity: "Organization", id })),
   listByUserId: () => Effect.die("listByUserId not used"),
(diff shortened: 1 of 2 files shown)


geclos:
Closing this in favor of #4600, which reimplements the identical fix fresh against current `development`.

This branch (`claude/trusting-allen-xrx4hd`) has no common git ancestor with the current `development` history any more (`git merge-base` fails between them), so it can't be merged or rebased as-is — it's been sitting as an unmergeable draft since 2026-08-15 while the underlying bug (Datadog `ET-375`/`ET-376`) kept recurring in production (last occurrence 2026-09-04). #4600 has the same diff, same tests, verified against the current codebase.

---

https://claude.ai/code/session_01DWqgcpLQWH9gEpNafyWzBC

---

PR #4538, opened 2026-09-01
fix(api-keys): reject revoked keys on token-hash lookup

neoxelox:
## What does this PR do?

Revoking an API key did not actually stop it from authenticating. `ApiKeyRepositoryLive.findByTokenHash` matched on `token_hash` alone, ignoring the `deleted_at` soft-delete column that revocation sets. `validateApiKey` trusted that row, returned a valid auth result, and cached it in Redis under the 5-minute valid-key TTL — so a revoked key kept working, and each use re-armed the positive cache entry for another window. The revoke path's `DEL apikey:<hash>` was working correctly; the next request just repopulated the cache from a DB lookup that never checked revocation.

Both `apps/api` and `apps/ingest` authenticate through this one validator, so both were affected. `apps/api` additionally accepted the revoked key for `access: "destructive"` operations.

The fix has two parts:

`findByTokenHash` now filters on `deleted_at IS NULL`, matching what `list()` already did. This is the root cause and the only change needed to close the hole. The port's method now documents that contract, and the in-memory fake in `@domain/api-keys/testing` applies the same filter so fakes and the live adapter can't drift.

`validateApiKey` additionally drops a key that fails the domain's `isActive` predicate before building an auth result. The port returns the full `ApiKey` entity including `deletedAt`, so this is a plain read of data the validator already has — no extra query, no leak of persistence concerns into the auth layer. It exists so that no repository implementation, present or future, can hand a revoked row to the auth boundary and have it accepted. A revoked key now takes the same negative path as an unknown one: `null`, cached for the 5-second invalid-key TTL, no `onKeyValidated` touch.

### Follow-up hardening in this PR

Auditing every place a key is checked found no other unpatched revocation path: both bearer entry points funnel into the two shared validators, and MCP tool calls re-enter the API middleware. It did surface adjacent gaps, fixed in the two later commits:

- **API keys outlived their organization.** No FK links `api_keys` to `organizations`, and none of the four org-delete paths touched them. `validateApiKey` accepted such a key with `isSandbox` forced to `false`; the API then failed at the organization-context middleware with a 404 rather than a 401, and ingest accepted the key outright. The validator now resolves the key and its organization together and takes the same negative path for a missing, revoked, or orphaned key.
- **Organization deletion revokes every API key and OAuth key.** New `teardownOrganizationUseCase` (`@domain/organizations`) revokes all active API keys, revokes all OAuth keys (deleting tokens, disabling applications), busts both Redis caches, and purges projects in one org-scoped transaction. It is wired into the settings danger zone, sandbox deletion, the expired temporary-org reaper, and account deletion. It composes new `revokeAllApiKeysUseCase` and `revokeAllOAuthKeysUseCase`, which reuse the existing single-key revoke logic.
- **OAuth cache hits re-verify the application.** The cache-hit re-check in `validateOAuthAccessToken` now joins `oauth_applications` and rejects a disabled or unbound application, not only a removed member, so disabling an application takes effect on the next request inside the cache window. OAuth revocation itself never had the original bug: token rows are hard-deleted, so there was no soft-delete flag to ignore.
- **Better Auth's `GET /api/auth/mcp/get-session` is blocked** at the web auth catch-all. It echoes the raw token row, refresh token included, to any holder of a valid access token and performs none of our checks; nothing of ours calls it.

`@domain/sandboxes` lists `@domain/oauth-keys` only so the exported `deleteSandboxUseCase` type stays portable under declaration emit; knip ignores it for that workspace. `apps/workers` gains `@platform/api-key-auth` and `@platform/oauth-token-auth` for the cache-invalidator layers. Docs: `dev-docs/organizations.md` gets a deletion section; `authentication.md` and `mcp.md` get cross references.

### Scope note on ingest

`apps/ingest` has no vitest setup or HTTP test harness, and its auth middleware is `validateApiKey(...) → if (!result) 401`. Rather than add a test toolchain to that app, ingest is covered at the shared validator with its exact wiring, alongside the pre-existing `matches ingest wiring` test. The API surface is covered end-to-end through its real middleware and routes.

None.

## How was this tested?

Tests were written first and observed failing against the vulnerable code, at all three layers:

- `packages/platform/db-postgres/src/repositories/api-key-repository.test.ts` (new, PGlite): a revoked row must not resolve — 2 failed, 1 passed before the fix.
- `packages/platform/api-key-auth/src/validate-api-key.test.ts`: cold-cache validation of a revoked key, and the warm-cache sequence (authenticate → revoke through the real `revokeApiKeyUseCase` + `ApiKeyCacheInvalidatorLive` → next call), plus the ingest wiring — 3 failed, 4 passed before the fix.
- `apps/api/src/middleware/auth.test.ts`: authenticate with a key (200), `DELETE /v1/api-keys/{id}` (204), re-request with the same bearer — expected 401, **got 200** before the fix.

After the fix all three suites pass. Each guard was then confirmed load-bearing by temporarily restoring the old code:

| restored | db-postgres | api-key-auth | api |
| --- | --- | --- | --- |
| repository filter only | 2 failed | pass (validator guard catches it) | pass |
| both guards | 2 failed | 3 failed | 1 failed |

The follow-up commits add these tests, each observed failing before its fix:

- `validate-api-key.test.ts`: a key whose organization row was deleted is rejected.
- `apps/api/src/middleware/auth.test.ts`: the same scenario end to end expects 401 (got 404 before).
- `validate-oauth-token.test.ts`: a cached token is rejected once its application is disabled.
- `revoke-all-api-keys.test.ts`, `revoke-all-oauth-keys.test.ts`, `teardown-organization.test.ts`: unit tests on the in-memory fakes covering revocation, cache busting, application disabling, project purge, and the `ProjectDeleted` event.

Follow-up gates: `typecheck` clean for `@domain/api-keys`, `@domain/oauth-keys`, `@domain/organizations`, `@domain/sandboxes`, `@app/web`, `@app/workers`; tests passing for those domain packages plus `@platform/api-key-auth`, `@platform/oauth-token-auth`, and the API auth middleware; `pnpm knip` clean.

Gates run locally on Node 25 (the repo's required version):

- `pnpm turbo check` (biome) — 65 tasks, clean
- `pnpm turbo typecheck` (tsgo) — 67 tasks, clean
- `pnpm knip` — clean
- `pnpm turbo test` for `@platform/db-postgres` (405), `@app/api` (28 files), `@platform/api-key-auth` (7), `@domain/api-keys`, `@domain/organizations`, `@domain/sandboxes`, `@repo/operations`, `@app/workers`, `@tools/live-seeds` — all passing

All 22 CI checks were green on the first commit; the follow-up commits re-run the same checks.

`@app/workers` (`signals.ts`) and `@tools/live-seeds` also call `findByTokenHash` to resolve the telemetry key; both want active keys only, and a revoked telemetry key now failing the job matches the behavior `dev-docs/flaggers.md` already documents.

## Checklist

- [x] I have signed the CLA


neoxelox:
Confirmed the `list()` observation — the fake returns `[...apiKeys.values()]` unfiltered while `ApiKeyRepositoryLive.list()` filters `deletedAt IS NULL`, so the drift is real.

Leaving it out of this PR deliberately. It has no security consequence: `list()` is org-scoped and only feeds the key-management UI, so a revoked key surfacing there is a display bug, not an auth bypass, and nothing authenticates off it. The port also doesn't document an active-only contract for `list()` the way it now does for `findByTokenHash`, so tightening the fake alone would leave the same question open one level up. Fixing it here would mean widening a security fix with an unrelated change to a shared test double that three domain suites build on.

Worth a follow-up that settles the `list()` contract at the port and aligns the fake with it.

---

PR #4589, opened 2026-09-07
fix(models): unblock the models.dev refresh and bundle the current catalog

neoxelox:
## What does this PR do?

The daily **Update Models.dev Data** workflow has been failing since 2026-09-04 ([run 173](https://github.com/latitude-dev/latitude-llm/actions/runs/34123227510)). It downloads the catalog fine, then its verification gate fails, so the bundled snapshot has been stuck and no refresh PR has opened for four days.

### Root cause

Two fixtures named catalog rows that models.dev has since retired. Neither is a code regression — the registry and the span builder both behave correctly — but both are hard failures on a refresh:

1. **`registry.test.ts`** pinned `vercel` / `xai/grok-4.5` in four places to exercise the Vercel AI Gateway path. models.dev dropped that one listing; `xai/grok-4.5` still exists under the `xai` provider, only Vercel stopped reselling it.
2. **`cost-archetypes/models.ts`** named `openrouter` / `nvidia/nemotron-nano-9b-v2:free` as one of archetype F's two free tiers. That listing is gone too, so the cohort built from it seeded *unpriced* rows into the archetype that exists precisely to show free ≠ unpriced — failing three of its assertions.

The second only surfaced once the first was fixed, since the workflow runs the four suites in sequence.

### Solution

Both fixtures now read the catalog instead of naming rows in it, so a retired listing degrades to a different fixture rather than a red build.

**Gateway fixture.** `findGatewaySlugAlsoSoldByItsVendor()` picks a priced `<vendor>/<model>` gateway listing whose vendor is itself a provider pricing the bare model. That second condition is what keeps the coverage meaningful: precedence is only observable when the lookup has a real alternative to fall through to, so resolving to `vercel` is a choice and not the only entry available. The refreshed catalog offers **98 such slugs across 15 vendors**, sorted so a failure names the same one every run. This follows the pattern `findAmbiguousBareId()` already established in this file for exactly this reason.

The precedence test got *stronger*, not weaker — it now also asserts `pricedProvider`/`pricedModel`, which the deleted `it.each` row covered separately.

**Archetype F.** `FREE_MODELS` is now the subset of a candidate list that the registry still prices at zero. Two hardcoded `it.each` rows naming `xai/grok-4.5` were dropped from the historical production-pair tables; the two other `gateway /` rows there still resolve and stay.

**Catalog refresh.** `models.dev.json` regenerated by the update workflow itself (192 → 213 providers), bringing in `openai/gpt-6-astra` at $10/$50 per 1M in/out with cache rates, and retiring the two rows above.

N/A — scheduled workflow failure, no filed issue.

## How was this tested?

Because this session's egress policy blocks `models.dev`, the refresh was produced by **dispatching the repository's own `update-models-dev-data.yml` against this branch** rather than fetching locally. That means the workflow's own gate is the primary evidence:

- ❌ [Run 174](https://github.com/latitude-dev/latitude-llm/actions/runs/34124495105) (gateway fix only) — registry/metering/evaluation-execution passed against the fresh catalog; archetypes failed with the 3 free-tier assertions above. This is what exposed the second fixture.
- ✅ [Run 175](https://github.com/latitude-dev/latitude-llm/actions/runs/34125367875) — **"Verify catalog-coupled tests" green**, all four commands against the freshly downloaded catalog. It then opened #4588 with the snapshot, which is the JSON committed here verbatim.

Re-run locally against that committed snapshot (all four workflow commands):

| Command | Result |
| --- | --- |
| `@domain/models` `src/registry.test.ts` | 76 passed |
| `@platform/ai` `src/metering.test.ts` | 14 passed |
| `@domain/evaluations` `src/runtime/evaluation-execution.test.ts` | 9 passed |
| `@platform/testkit chdb:build` + `@platform/db-clickhouse` archetypes | 51 passed |

Both fixes were confirmed load-bearing rather than assumed:

- The archetype fix was validated by deleting `nvidia/nemotron-nano-9b-v2:free` from the local snapshot to simulate the churn: **3 failed / 48 passed** without the change (the same three CI reported), **51 passed** with it.
- The gateway fix is demonstrated by run 174, where the registry suite passed against a catalog that no longer lists `xai/grok-4.5`.

Also run:

- `turbo typecheck --filter @platform/db-clickhouse --filter @domain/models` — 55/55 tasks clean.
- `pnpm --filter @domain/models check` — the 2 pre-existing `noExcessiveCognitiveComplexity` warnings only, unchanged from `origin/development` (verified by stashing). No new diagnostics.

Two things I could **not** verify locally, both environmental:

- `@domain/spans` `ingest-spans.test.ts` fails 29 tests here with `bytes.toBase64 is not a function`. This sandbox runs Node 22; the repo pins Node 25, where `Uint8Array.prototype.toBase64` exists. **Identical 29 failures on unmodified `origin/development`**, so it is not this change. CI will settle it.
- `models.dev` itself is unreachable from this session, hence the dispatch-based refresh described above.

## Notes for the reviewer

- **`gpt-6-astra` coverage asserts the entry, not the rate.** It pins `pricedProvider`/`pricedModel` and requires both rates positive, rather than `$10`/`$50`. A prefix fallback onto a neighbouring `gpt-6` would report `costImplemented: true` just as readily at the wrong price, so identity is the invariant with discriminating power — and pinning the dollar figures would recreate exactly the brittleness this PR removes, breaking the refresh on the next price change. The parser reads the base `cost` block and never consults `tiers`, so long-context tier selection is not a live risk worth a test.
- **The retired free-tier candidate is deliberately left in the list.** Pruning it would leave a single candidate and no slack the next time one is retired; the filter exists to tolerate exactly that.
- **This does not close the class of problem, only the two instances of it.** Other suites still name catalog rows directly (`archetypes.test.ts` asserts break-even rates for specific models; the historical production-pair tables still list literal pairs). Those are all live today and de-brittling them is out of scope here, but they are the next things to fail on a future refresh.

## Checklist

- [ ] I have signed the CLA

---

https://claude.ai/code/session_01M5GvGXpHomWaFq12bVDbea

---
_Generated by [Claude Code](https://claude.ai/code/session_01M5GvGXpHomWaFq12bVDbea)_

---
<a href="https://app.blacksmith.sh/latitude-dev/codesmith/latitude-llm/pr/4589"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791378858&installation_model_id=435055&pr_number=4589&repository=latitude-dev%2Flatitude-llm&return_to=https%3A%2F%2Fgithub.com%2Flatitude-dev%2Flatitude-llm%2Fpull%2F4589&signature=c9aae5030ef97af568d7c547d3b1488e20deaf0421f6708b738ae3f7bb2604d9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


neoxelox:
Thanks — taking these separately.

**Comment length: fixed** in 105b1eb. Both multi-line blocks I added are now one line each, keeping only the trap in each (the gateway fixture needs its vendor to price the bare model or the precedence assertions have nothing to rule out; archetype F's candidates are filtered on live pricing). Rationale stays in the PR description. Re-ran after the trim: registry 76 passed, archetypes 51 passed, lint unchanged at the 2 pre-existing `noExcessiveCognitiveComplexity` warnings.

**Long-context tier: not changing it here.** The mechanism is described correctly — `parseModelsDevData` reads `raw.cost.{input,output,cache_read,cache_write,reasoning}` and never consults `cost.tiers` — but this is neither new nor accidental:

- It predates this PR by a wide margin. The **previous** bundled snapshot already carried **377** models with a `tiers` / `context_over_*` block (the refreshed one has 434). `gpt-6-astra` is one more instance of long-standing behaviour, not the case that introduces it.
- It is a recorded decision, not an oversight. `packages/domain/spans/src/helpers/model-registry-pricing.ts` documents it: *"Tiered pricing collapses to the base tier: long-context tiers scale every rate together, so they barely move a ratio between them, and the dashboard already prices everything else at the base tier."* Pricing `gpt-6-astra` at its context tier while 433 other models stay on their base tier would make the dashboard internally inconsistent, which is the thing that note is guarding.

So implementing this would change pricing semantics for the whole catalog — a deliberate product decision, well outside a PR whose job is to unblock the daily refresh. Worth its own issue if the team wants to revisit the base-tier policy; I'd rather not fold it in behind a test fix.

On the test not catching it: that is accurate, and intentional. The assertion targets `pricedProvider`/`pricedModel` because the failure mode this PR guards against is a prefix fallback onto a neighbouring `gpt-6` — which would report `costImplemented: true` at another model's price. Pinning `$10`/`$50` would re-create exactly the brittleness this PR removes: the refresh would break on the next price change, which is the bug being fixed.

---

PR #4676, opened 2026-09-17
fix(ai): use the json tool for Anthropic schemas with unsupported keywords

### LINKED ISSUE #4675: taxonomy: cluster naming fails with Anthropic provider — structured-output schema uses maxItems/minLength/maxLength the Anthropic API rejects (gardenTaxonomyWorkflow never succeeds)
## Summary

On a self-hosted stack with `LAT_AI_GENERATION_PROVIDER=anthropic` (model `claude-sonnet-4-6`), every `gardenTaxonomyWorkflow` run fails in `nameTaxonomyClusterActivity` with:

```
AIError: AI generation failed (anthropic/claude-sonnet-4-6):
  output_config.format.schema: For 'array' type, property 'maxItems' is not supported
```

8 attempts with backoff over ~23 minutes, then the workflow fails; 4 runs/day on the `0 */6 * * *` gardening cron. 13 of 13 runs in our Temporal visibility window failed, zero successes ever, so the project's Topics taxonomy never gets named. Reproduced on `v0.3.102`; `name-taxonomy.ts` and `ai-vercel/src/ai.ts` are byte-identical through `v0.3.104` and `development` HEAD (`3f6ca7f50`), so it is still present.

## Root cause

1. **The naming schemas use JSON Schema keywords Anthropic's structured-output subset rejects.** `packages/domain/taxonomy/src/use-cases/name-taxonomy.ts`:
   - `candidateThemesSchema` (~L80-85): `z.array(...).min(1).max(5)` → `minItems: 1, maxItems: 5` (the 400 above)
   - `finalNameSchema` (~L86): `z.string().min(3).max(80)` / `.min(20).max(280)` → `minLength` / `maxLength`
   - `contrastiveThemesSchema` (~L88-97): `.min(1).max(4)` plus outer `.min(2)` → `maxItems: 4, minItems: 2`
   - `contrastiveNamesSchema` (~L98-108): outer `.min(2)` plus string `min/max`
   
   Per Anthropic's structured-outputs docs, `maxItems`, `minLength`, `maxLength`, `minimum`/`maximum` are unsupported and `minItems` is only allowed for 0 and 1. All four schemas are invalid on that path, so a fix that only drops `maxItems` moves the 400 to the next call (`proposeCandidateThemes` → `nameCluster`).

2. **The generate path never sanitizes or reroutes the schema per provider.** `packages/platform/ai-vercel/src/ai.ts` (~L386) passes the caller's zod schema straight into `Output.object({ schema })`. The repo already has `BEDROCK_UNSUPPORTED_SCHEMA_KEYWORDS` / `loosenSchemaForBedrock` (~L487-528, `maxItems` is on that list), but it is only applied to agent-loop tool schemas when `provider === "amazon-bedrock"`, not to structured outputs, and not for `anthropic`.

3. **Why the default configuration never hits it.** `TAXONOMY_DEFAULT_NAMING_MODEL` is `amazon-bedrock` / `minimax.minimax-m2.5`; `@ai-sdk/amazon-bedrock@5.0.0-beta.41` only uses native structured output for Anthropic models on Bedrock and routes everything else through a non-strict `json` tool, whose `input_schema` accepts these keywords. With `@ai-sdk/anthropic@4.0.0-beta.37`, `structuredOutputMode` defaults to `auto`, which picks the native `output_config.format` path for every model in its capability table (sonnet-4-6/4-5, opus-4-7/4-6/4-5/4-1, haiku-4-5). `structuredOutputMode` is only reachable through `providerOptions`, which neither `resolveGenerationConfig` nor `name-taxonomy.ts` sets, so there is no env-level escape hatch. Operators who follow https://docs.latitude.so/deployment/configuration.md and set `LAT_AI_GENERATION_PROVIDER=anthropic` inherit that provider into `TAXONOMY_NAMING` with no warning.

## Suggested fixes (any one of these)

- When the resolved provider is `anthropic`, pass `providerOptions: { anthropic: { structuredOutputMode: "jsonTool" } }` on the structured-output path (schema then rides in a non-strict tool `input_schema`, which accepts all four schemas; verified against the API directly).
- Or extend the existing keyword stripper to the `generate` structured-output path, keyed on provider (`anthropic` needs `maxItems`, `minItems > 1`, `minLength`, `maxLength` removed).
- Or at minimum a note in the configuration docs that Anthropic-native structured outputs reject these schemas, and that `TAXONOMY_NAMING` should be overridden to a non-Anthropic route.

## Related observation: Claude 5 models on this SDK pin

`@ai-sdk/anthropic@4.0.0-beta.37` predates the Claude 5 family, so `claude-sonnet-5` / `claude-opus-5` are unknown ids: they would take the `jsonTool` route (which avoids the schema problem) but the SDK still forwards `temperature`, which those models reject with `` `temperature` is deprecated for this model ``. Since most generation features hard-code a temperature default (taxonomy naming 0.2, facet extraction 0.1, moment classifier 0, flagger classifier/extractor 0, flagger annotator 0.2) and `resolveGenerationConfig` only lets env change the number, not omit it, Claude 5 cannot currently serve Latitude's own generation features. A bump of `@ai-sdk/anthropic` would address that side.

## Workaround attempted (does not work, and exposes a second gap)

We tried the per-feature override away from the Anthropic-native path: `LAT_AI_TAXONOMY_NAMING_PROVIDER=custom` + `LAT_AI_TAXONOMY_NAMING_MODEL=MiniMaxAI/MiniMax-M3` on an OpenAI-compatible endpoint. The Anthropic 400 goes away, but the call now fails client-side:

```
AIError: AI generation failed (custom/MiniMaxAI/MiniMax-M3):
  No object generated: response did not match schema. (finishReason=stop)
  text: {"candidateThemes":["...","...","...","..."]}
  cause: [{"expected":"array","code":"invalid_type","path":["candidates"]}]
```

`@ai-sdk/openai-compatible` defaults `supportsStructuredOutputs` to false and `createCustomProvider` does not override it, so the request degrades to `response_format: {type:"json_object"}` with no schema, and the AI SDK build pinned here does not inject the schema into the prompt. The system prompt in `proposeCandidateThemes` (~L274) says only "Return only schema-valid JSON" with no shape example, so the model cannot know the required keys (`candidates[].theme`, `candidates[].examples`). Any provider without server-side schema support will fail the same way. A short shape description in the prompt (or `supportsStructuredOutputs: true` for endpoints that accept `json_schema`) would close that second gap.

So on `v0.3.102` an operator on `LAT_AI_GENERATION_PROVIDER=anthropic` has no env-only way to make taxonomy naming succeed; we reverted to waiting for a code fix. We are happy to open a PR for the `jsonTool` routing if that direction is acceptable.

## Environment

Self-hosted Latitude `v0.3.102` (images `latitudedata/*:0.3.102`) with the stock stack: Postgres 16 + pgvector, ClickHouse 26.2, Temporal 1.27.2. `LAT_AI_GENERATION_PROVIDER=anthropic`, `LAT_AI_GENERATION_MODEL=claude-sonnet-4-6`. Happy to test a patch.

jkbdco:
## What does this PR do?

Anthropic's structured-output subset rejects several JSON-schema keywords that Zod emits from `.min()`, `.max()` and `.multipleOf()`: `maxItems`, `minItems` above 1, `minLength`/`maxLength`, the numeric bounds, and `additionalProperties` set to anything other than `false`. A generation whose schema carries any of them fails with a 400 from `output_config.format.schema`, so a feature with such a schema cannot run on the `anthropic` provider at all. Taxonomy cluster naming is one such feature: all four of its schemas are affected, so `gardenTaxonomyWorkflow` fails on every attempt for operators who set `LAT_AI_GENERATION_PROVIDER=anthropic`.

The Anthropic provider already offers a `jsonTool` structured-output mode that carries the schema in a non-strict tool `input_schema`, which that subset does not constrain. This change detects the unsupported keywords on the schema and selects that mode for the affected calls only:

- `ANTHROPIC_UNSUPPORTED_SCHEMA_KEYWORDS` plus a recursive walk (`hasAnthropicUnsupportedSchemaKeyword`) with value checks for `minItems` (only 0 and 1 are supported) and `additionalProperties` (only `false` is supported; an absent keyword is not flagged).
- `schemaNeedsAnthropicJsonTool(schema)` converts the Zod schema with `z.toJSONSchema` and returns `false` if conversion throws, leaving the native path in place.
- `withAnthropicStructuredOutputMode` merges `anthropic.structuredOutputMode = "jsonTool"` into the provider options only when the serving provider is `anthropic` and the schema is non-compliant. A caller-supplied `structuredOutputMode` always wins. Compliant schemas keep the native `output_config.format` path.

Detection is keyed on the provider that actually serves the call (`servedBy`), so a fallback attempt is judged on its own terms. The walk matches by key name, the same trade-off the existing Bedrock keyword stripper makes; a false positive only routes a call through the json tool, which accepts the whole schema.

Out of scope: Anthropic's structural exclusions (recursive schemas, complex enum members, external `$ref`), which no keyword check can detect, and the prompt-shape gap on providers that never receive the schema, which #4675 describes separately.

Closes #4675

## How was this tested?

- 12 new tests in `packages/platform/ai-vercel/src/ai.test.ts`: 8 for `schemaNeedsAnthropicJsonTool` (each keyword, the `minItems` 0/1 boundary, `z.looseObject`, a realistic compliant object, an unconvertible schema) and 4 for the generate path (non-compliant schema on `anthropic` gets `jsonTool`; compliant schema stays native; another provider is untouched; a caller-supplied mode is preserved). The first generate-path test fails with the call-site change reverted.
- `pnpm --filter @platform/ai-vercel typecheck`, `check` and `test`: clean, 35/35 passing (the one biome complexity warning on `generateWithModel` is pre-existing and unchanged by this diff).
- Verified against the Anthropic API directly that a `json` tool with `minItems`/`maxItems`/`minLength`/`maxLength` in its `input_schema` is accepted and returns a well-formed `tool_use`.
- Verified end to end on a self-hosted deployment of `v0.3.102` plus this commit (`LAT_AI_GENERATION_PROVIDER=anthropic`, `claude-sonnet-4-6`): a taxonomy gardening run for a project whose previous 13 runs had all failed on the `maxItems` 400 completed in 85 s, with 5 of 5 leaf clusters named and no schema errors.

## Checklist

- [x] I have signed the CLA

---
<a href="https://app.blacksmith.sh/latitude-dev/codesmith/latitude-llm/pr/4676"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792241836&installation_model_id=435055&pr_number=4676&repository=latitude-dev%2Flatitude-llm&return_to=https%3A%2F%2Fgithub.com%2Flatitude-dev%2Flatitude-llm%2Fpull%2F4676&signature=d6b41c090444dfe837a36f61131e4c3bb7ec2a2a35078d5953fa1d7129fb2808"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


jkbdco:

